PR #48, opened 2018-01-09
Call dismiss in main thread

jcesarmobile:
App crashes on browser close if dismiss is not run in main thread


riderx:

PR #49, opened 2018-01-09
Move export js to the bridge

jcesarmobile:
Moving the exportAvocadoJS to Bridge class again, I moved it to the VC for the native-bridge.js injection, but it's better here.


riderx:

PR #57, opened 2018-01-11
Inject cordova.js and cordova plugins js

jcesarmobile:
Will inject cordova.js from the public root and any file from public/plugins and subfolders if there is at least one class with CDVPlugin super class


mlynch:
⚡️

riderx:

PR #74, opened 2018-01-19
Handle Cordova plugins native code

jcesarmobile:
Search for cordova plugins in node_modules and add them to the app as local CocoaPods in the project.

For now it assumes all the native code will be in `src/ios` folder. Might need to change that to cover other plugins where the code is in a different folder by reading the plugin.xml content and create the .podspec based on the `source-file` and `header-file` tags instead of doing a `*.{swift,h,m}`. 
The PR already read the xml, but doesn't use it for now.

### DIFF
--- a/cli/src/ios/common.ts
+++ b/cli/src/ios/common.ts
@@ -41,16 +41,18 @@ export async function getIOSPlugins(allPlugins: Plugin[]): Promise<Plugin[]> {
 }
 
 export async function resolvePlugin(plugin: Plugin): Promise<Plugin|null> {
-  const iosManifest = plugin.manifest.ios;
-  if (!iosManifest) {
+  let iosPath = '';
+  if (plugin.manifest && plugin.manifest.ios) {
+    if (!plugin.manifest.ios.src) {
+      throw 'capacitor.ios.src is missing';
+    }
+    iosPath = join(plugin.rootPath, plugin.manifest.ios.src);
+  } else if (plugin.xml) {
+    iosPath = join(plugin.rootPath, "src/ios");
+  } else {
     return null;
   }
   try {
-    if (!iosManifest.src) {
-      throw 'capacitor.ios.path is missing';
-    }
-
-    const iosPath = join(plugin.rootPath, iosManifest.src);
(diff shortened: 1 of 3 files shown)


cli/src/ios/common.ts
```diff
@@ -41,16 +41,19 @@ export async function getIOSPlugins(allPlugins: Plugin[]): Promise<Plugin[]> {
...
 
 export async function resolvePlugin(plugin: Plugin): Promise<Plugin|null> {
-  const iosManifest = plugin.manifest.ios;
-  if (!iosManifest) {
+  let iosPath = '';
+  if (plugin.manifest && plugin.manifest.ios) {
+    if (!plugin.manifest.ios.src) {
+      throw 'capacitor.ios.src is missing';
+    }
+    let iosPath = join(plugin.rootPath, plugin.manifest.ios.src);
```
mlynch: Looks like you redefined `iosPath` here?

cli/src/plugin.ts
```diff
@@ -42,15 +44,30 @@ export async function resolvePlugin(name: string): Promise<Plugin | null> {
     const rootPath = resolve('node_modules', name);
     const packagePath = join(rootPath, 'package.json');
     const meta = await readJSON(packagePath);
-    if (!meta || !meta.capacitor) {
+    if (!meta) {
       return null;
+    } else {
```
mlynch: No need for this else, if you remove it it'll make the code cleaner and less idented

cli/src/plugin.ts
```diff
@@ -42,15 +44,30 @@ export async function resolvePlugin(name: string): Promise<Plugin | null> {
...
+        const pluginXMLPath = join(rootPath, 'plugin.xml');
+        const xmlMeta = await readXML(pluginXMLPath);
+        return {
+          id: name,
+          name: fixName(name),
+          rootPath: rootPath,
+          xml: xmlMeta.plugin.platform
+        };
+      } else {
+        return null;
```
mlynch: Same this condition being unnecessary, it would fall to the return null anyways

PR #75, opened 2018-01-19
Fix implicit declaration errors on some plugin classes

jcesarmobile:
Plugins that use methods from `CDVAvailability` such as `IsAtLeastiOSVersion` fail to build. Importing it in CDVPlugin.h fixes it.


riderx:

PR #77, opened 2018-01-19
Fix crash on iPad modals

jcesarmobile:
Show Actions and Show Sharing were crashing on iPad because they are action sheets and need some extra configuration. Used the centered way as the other plugins.

BTW, "Native Prompt" button also makes the app crash, but that's the browser one, right? so I don't think we can fix it.

Ah, WKWebView actually overwrites the browser alert, prompt and confirm, the problem was the prompt didn't have type `.alert`, fixed now.


riderx:

PR #86, opened 2018-01-23
Fix IonicRunner build

jcesarmobile:
IonicRunner was not building for this reasons:
Capacitor.podspec was deleted from ios/Capacitor, so pod install didn't work.
Changed the path to use the root podspec
Also changed the path for CapacitorCordova to use the root podspec and deleted the old one
Moved CAPAppDelegate.swift class to Capacitor folder so CocoaPods can find it.


mlynch:
Awesome, stupid xcode putting my files in the wrong folders

PR #90, opened 2018-01-24
Serve all content and not just public folder

jcesarmobile:
Change the web server to serve all the app content and not just the public folder.
If we don't serve all app content, Cordova camera plugin images can't be set to an img tag because the image url returned by the plugin is not accessible from the web server.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -54,7 +54,7 @@ class CAPBridgeViewController: UIViewController, WKScriptMessageHandler, WKUIDel
   }
   
   func loadWebView() {
-    if Bundle.main.path(forResource: "public/index", ofType: "html") == nil {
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/CAPBridgeViewController.swift
```diff
@@ -67,7 +67,7 @@ class CAPBridgeViewController: UIViewController, WKScriptMessageHandler, WKUIDel
     startWebServer(port: port!)
 
     print("⚡️  Loading index.html...")
-    let request = URLRequest(url: URL(string: "http://localhost:\(port!)/")!)
+    let request = URLRequest(url: URL(string: "http://localhost:\(port!)\(indexPath)")!)
```
mlynch: We can't request index.html here, we need the root slash path for proper SPA routing. Many routers rely on `/` to be the index path

jcesarmobile:
Closed in favour of #91

PR #102, opened 2018-02-01
Change cordova.js to also work on Android

jcesarmobile:
Forgot to commit the cordova.js changes for Android compatibility.

### DIFF
--- a/core/cordova.js
+++ b/core/cordova.js
@@ -947,7 +947,11 @@
                 action,
                 actionArgs
             };
-            window.webkit.messageHandlers.bridge.postMessage(command);
+            if (window.androidBridge) {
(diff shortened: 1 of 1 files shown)


core/cordova.js
```diff
@@ -947,7 +947,11 @@
                 action,
                 actionArgs
             };
-            window.webkit.messageHandlers.bridge.postMessage(command);
+            if (window.androidBridge) {
```
mlynch: I wonder if we could just export `window.webkit.messageHandlers` on Android as well? Then we would have to modify this? I haven't checked if that's possible yet though

mlynch:
Merging for now. Made note to follow up on the modifying cordova.js #104

PR #103, opened 2018-02-02
Cordova android tooling

jcesarmobile:
Create cordova_plugins.js and copy Cordova plugin files to Android public folder
Write config.xml with the needed plugin entries.


mlynch:
🤘

Issue #153, opened 2018-02-19
Race condition in Android web server


JStumpp:
We try to build a capacitor app with stencil and vanilla javascript. In an android emulator it works (with Android Webview v58). But on a real device with latest Version 64 it does not work because of the following error in chrome remote debugger:
```Failed` to load module script: The server responded with a non-JavaScript MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.```

This issue is already discussed here: https://stackoverflow.com/questions/48113359/es6-modules-imports-in-apache-cordova-hybrid-app-throw-mime-type-error

Any idea how to solve this?

**Comment by jcesarmobile** [comment]
Could you provide a sample project?

**Comment by JStumpp** [comment]
Sure, I just used the Stencil App Starter and added Capacitor. Here you find the example:
https://github.com/JStumpp/stencil-capacitor
Just run the android app. Don't be fooled by the prerendering, the app starts, but the JS is not loaded.
Android logcat also shows: `02-19 12:26:48.596 25765-25765/com.getcapacitor.app I/chromium: [INFO:CONSOLE(0)] "Failed to load module script: The server responded with a non-JavaScript MIME type of "". Strict MIME type checking is enforced for module scripts per HTML spec.", source: https://58b69576-5bc8-41b3-a368-b6fb2885dd4ccapacitorapp.net/build/app/cufq1558.js (0)
`

**Comment by JStumpp** [comment]
Seems to be a stencil issue: https://github.com/ionic-team/stencil/issues/517

**Comment by jcesarmobile** [comment]
That issue looks like some webview incompatibility.
But since Capacitor serves from http and not file, we can probably fix it anyway, so reopening.

**Comment by JStumpp** [comment]
Oh, true. It should be handeled here: https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java#L365
Any idea why it is not "application/javascript"?

**Comment by mlynch** [comment]
I'm noticing this even with the patch, I wonder if it's a race condition? Occasionally the mimetype will come back for those .js files as an icon

**Comment by janpio** [comment]
I could recreate the `mimetype=icon` problem by building the Android project while I have a `native-bridge.js` and `cordova.js` file (copied over from `/core`) but no `cordova_plugins.js` and `public/plugins` folder.

Logcat:
```
[...]

PR #175, opened 2018-02-24
MimeType was not set correctly

JStumpp:
The MimeType was never set. Additionally the URLConnection.guessContentTypeFromName does not recognize *.js types. But this is required to enable ES6 Modules like used in stenciljs.

This fixes https://github.com/ionic-team/capacitor/issues/153


mlynch:
Thanks!

PR #182, opened 2018-02-27
Fix and improve tests

jcesarmobile:
All tests were failing because it was using cap from bin and we only have capacitor there
Also gave more timeout for iOS update test as pod install takes time
Improved the plugin.xml to be more real and wrote the CordovaPlugins podspec and not just the Resources one.


mlynch:
Me doing my happy dance: 🕺

Issue #142, opened 2018-02-13
Support adding custom pods


mlynch:
We auto-generate the Podfile, but we should support adding your own stuff to this, even if through our command line

jcesarmobile:
At the moment it's possible with Cordova plugins by adding them to the plugin.xml like this:

```
<platform name="ios">
    <framework src="FirebaseMessaging" type="podspec" spec="~> 2.0.0"/>
</platform>
```

Maybe we can add something similar in the Capacitor plugin package.json?

Or if we do it from the CLI, then where will we store the pods added to keep track of them? in capacitor.config.json?

jcesarmobile:
Users are free to write on the Podfile
Plugins will have a .podspec where they can add dependiencies

So we have both cases covered, closing

PR #195, opened 2018-02-28
Instead of generating the Podfile, just fill it with dependencies

jcesarmobile:
Followed a similar approach than for filling the android gradle dependencies.
Instead of generating the whole Podfile, read the existing one and replace the dependencies between the comments

This could be a good way of implementing #142, so we don't have to do anything else, users can just write in the Podfile and add the dependencies they want to use on their projects without worrying about them being removed by the CLI.


mlynch:
Good idea, wanted to avoid the parsing but seems minimal

PR #196, opened 2018-02-28
link "success-ees"

janpio:
link the "competing" product mentioned

(also some line breaks removed to make style consistent)


mlynch:
👍

PR #197, opened 2018-02-28
multiple tiny changes in "with Ionic" docs

janpio:
- Fix Ionic version
- clarify what build requires `cap copy`
- Link Ionic Native
- fix "Cordova" link (right now links to https://capacitor.ionicframework.com/docs/getting-started/basics/cordova which is an endless load loop)


mlynch:
🍻

janpio:
good point about the default 🥇

PR #203, opened 2018-02-28
more "building your app"

janpio:
- add equivalent sentence how to build for Android as already exists for iOS

(you are too fast in merging the PRs so I couldn't add it to the other one :p)


mlynch:
⚡️

PR #221, opened 2018-03-01
Fix for potential console logging issues.

IT-MikeS:
This PR fixes potential issues with console logging in `native-bridge.js`.

The issues are caused when the WebView or Chrome version is too low.  Causing `Object.keys(win.console)` to return no keys and if low enough, to not have the `Proxy` object defined.

Fix includes:
- Added Polyfill for `Proxy`.
- Added logic to check for keys from `win.console` and use fallback logging if none are available.

These changes have been tested on stock google provided android emulator images in the forms of:
- Android 5.1 on API 22 with WebView/Chrome version 39
- Android 7.0 on API 24 with WebView/Chrome version 51


mlynch:
Thanks Mike. As far as I know the Proxy polyfill doesn't support dynamic getters/setters which is what I'm using them for, unfortunately. Will play with this but not crazy about adding it or having another param to JS Export. Maybe there's another way...

PR #258, opened 2018-03-05
Add more extensions to plugin files

jcesarmobile:
Some plugins might include c files (.c), c++ files (.cc) or Objective-C++ (.mm), add those to the podspecs.


mlynch:
We should have written capacitor in C++

PR #259, opened 2018-03-05
Added fallback logging to native-bridge.js

IT-MikeS:
Fallback logging to old console.log/error as no keys are exposed by Object.keys(console) until more recent versions of webview/chrome


mlynch:
Awesome, thanks for resubmitting! Makes it easy to have one PR for one small fix so we can review them in isolation. Looking forward to more PRs from you 😄

Issue #271, opened 2018-03-06
Update capacitor version in gradle


jcesarmobile:
We have `implementation 'ionic-team:capacitor-android:0.0.+'` in the build.gradle, but with the Alpha we bumped to 1.0.0, so we should update it to `implementation 'ionic-team:capacitor-android:1.0.+'` or 
 `implementation 'ionic-team:capacitor-android:1.0.0-alpha.7'`

jcesarmobile:
Fixed in https://github.com/ionic-team/capacitor/pull/275

PR #275, opened 2018-03-06
Update build.gradle dependency for Capacitor

IT-MikeS:
Updated to `ionic-team:capacitor-android:1.0.+` .

Using `+` to account for future release.

Closes issue #271


mlynch:
Nice catch!

Issue #253, opened 2018-03-04
"Plugin does not have web implementation." could mention which plugin


janpio:
```
capacitor.js:2
 Uncaught (in promise) 
Plugin does not have web implementation.
```

could mention which plugin triggered this console log.

IT-MikeS:
Realistically this can be handled by the user very simply.
```
    const { Keyboard } = Plugins;
    Keyboard.show().then(() => {
      console.log('Keyboard shown');
    }).catch((e) => {
      console.error('Keyboard plugin:', e);
    });
```

that will output `Keyboard plugin: Plugin does not have web implementation.` when running in npx cap serve

janpio:
Why should the user have to do any extra implementation when the library can just handle this and help the user to actually find out which of the several plugins he just added is generating this output?

(In my case this is the default state after creating a project - that is using the Splashscreen APIs)

IT-MikeS:
I can look into doing this but the reason is not so simple at the moment is how it uses a Noop Plugin for plugins that dont have an impl. for web/android/ios

jcesarmobile:
done in #276

PR #276, opened 2018-03-06
Use plugin name in no impl. error message

IT-MikeS:
"Plugin does not have web implementation." error message now uses the plugin name instead of just "Plugin"

ex: `Keyboard does not have web implementation.`

Addresses issue #253

### DIFF
--- a/core/src/web-runtime.ts
+++ b/core/src/web-runtime.ts
@@ -3,39 +3,21 @@ export class CapacitorWeb {
   platform = 'web';
   isNative = false;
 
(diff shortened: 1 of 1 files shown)


mlynch:
Awesome, thanks for the PR! A few small things to watch for: you can remove the commented out stuff, no need to keep it in (it'll be in git history anyways). Watch the formatting, we use 2 spaces. Cheers

IT-MikeS:
Will do, maybe we should a .editorconfig file so people get auto formatting compliance?

Issue #268, opened 2018-03-05
Different error message on `npx cap init`


janpio:
```
λ npx cap init
C:\Projects\Capacitor\npmInitTest\node_modules\@capacitor\cli\bin\capacitor
Usage: npx cap init appName appId

Example: npx cap init "My App" "com.example.myapp"

[error] Must provide an app name
Invalid App ID "undefined". Must be in domain form (ex: com.example.app)
```

The error messages for the missing `appName` and `appId` are quite different in style and formatting:
![image](https://user-images.githubusercontent.com/183673/37003405-393cf59e-20cd-11e8-9493-ef22bf50c0f8.png)

PR #277, opened 2018-03-06
Making app name and app id error messages more similar

SonnyCampbell:
[Original Issue](https://github.com/ionic-team/capacitor/issues/268)


janpio:
Doesn't make it partially worse by always implying the string "undefined" was used?

mlynch:
Thanks, going to tweak this a little bit more but wanted to get you on the contributor sheet 🚀

SonnyCampbell:
Thanks @mlynch.

Yeah you're right @janpio. I wasn't really sure what sort of id requirements would be needed going forward, but for now it looks like the only check is to make sure the name exists.

PR #279, opened 2018-03-06
Improve example project and its documentation

janpio:
- update` example/README` and `.github/CONTRIBUTING.md`
- copy over `native-bridge.js` from `/core` to `/www` in `copy` script

also:
- rename example project from `avocado` to `capacitor-example`
- update package-lock.json files
- remove `@ionic/cli-plugin-cordova`
- update `@ionic/app-scripts` to a version that likes recent NPM on Windows

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,40 +1,49 @@
-Build Example
-=====================
+# Capacitor Example App
 
-## Build Capacitor Core Module
+This Example app can be used to develop Capacitor. It uses Capacitor to create native iOS and Android apps, the web app itself is an Ionic app, but the important bit is that the native apps are set up in a way that enables development of the native code by loading it from the parent folder instead of as an external dependency.
 
-```
-cd to core
+Note: This project does _not_ use the Capacitor CLI. Please follow the steps below to build, prepare and run the apps.
 
+## 1. Build Capacitor Core Module
+
+Start by building the Capacitor Core Module in `/core`:
+
+```bash
+cd ../core
+
+npm install
 npm run build
 npm link
 ```
 
+## 2. Build Example App
 
-## Build Example App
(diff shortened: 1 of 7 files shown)


example/README.md
```diff
@@ -1,40 +1,49 @@
...
 
+## 3. Build the native Capacitor Apps
 
-## Build iOS App
+Now that everything is in place you can build the native Capacitor Apps:
 
-```
+### a) Build iOS App
+
+```bash
```
jcesarmobile: This is not bash, nor commands, so not sure if should be formated like this.
Also, should open IonicRunner.xcworkspace, not the .xcodeproj

janpio: - correct @ bash
- `xcodeproj` is what was already there. will fix.

janpio:
Note:

I also hat to do the following to get the iOS project building:

```
cd ios/IonicRunner

pod repo update
pod update CapacitorCordova
pod update Capacitor
pod install
```

I am not 100% sure this was _all_ needed or just a local fluke, so I am not adding it to the document yet but just noting it down here.

jcesarmobile:
pod install is needed, the others shouldn't, but if you needed them, maybe should be added.

janpio:
Will do a new checkout tomorrow on my super slow Mac and check again. I think I got error messages on `pod install` because the lockfile was out of date I guess.

janpio:
As planned I tested with a new checkout. Running `pod update` in `ios/IonicRunner` was actually enough to update the `Podfile.lock` and install the Pods into `/Pods` - so I added those as the additional iOS instructions.

After that I could successfully build and run the iOS project, so I think this should be good now.

I also improved the introduction paragraph a bit and explained more what this project is and how it works.

mlynch:
Is this ready for merge review?

janpio:
Yes, this contains all I know about the `example` project now @mlynch ;)

mlynch:
🚀

PR #285, opened 2018-03-06
[Work in Progress] Show warning when SplashScreen is hidden by default timeout

janpio:
One of the most "popular" problems with the Cordova splash screen plugin was that many users didn't manually hide the splash screen, and so had to a) either wait longer than necessary for the splashscreen to go away while their app was waiting ready in the background or b) look at a white screen after the timeout while their app was still loading.

Quite often a complaint about a "slow app" (directed at Ionic) was actually a badly optimized use of the Cordova splash screen plugin. So one would have to instruct them to user it correctly and in an optimized way.

This shouldn't happen for Capacitor.

This PR adds a warning log to the console and device log when the SplashScreen is hidden by the default timeout, and not a manual call. It outputs instructions what to do, including a link to the SplashScreen docs where this could be explained in more detail.

Limitation:
For now it just prints to the device log, as I couldn't figure out how to log to the actual browser console that the user is more probable to look at via remote debugging (and where there is a lot less noise and the warning more probable to be noticed). See #286 for issue asking about that.


mlynch:
Decided this looked ready enough to merge, going to tweak a few things but I just made that timeout configurable so it's a good time to add this. Cheers!

janpio:
Nice, thanks. Will have a look at your changes later and maybe create a follow up if anything I was planning is still missing.

PR #294, opened 2018-03-07
iOS - Allow CAPBridge to be used outside of CAPBridgeViewController

joshdholtz:
## Reason
I would like to be able to use `CAPBridge` outside of `CAPBridgeViewController` for custom iOS projects where I'm not able to use `CAPBridgeViewController` but do have access to the `WKWebView` (https://github.com/turbolinks/turbolinks-ios)

## Solution
- Created `CAPBridgeDelegate` protocol
  - Moved `CAPBridge` logic into `CAPBridgeDelegate` protocol extension
- `CAPBridgeViewController` now adheres to `CAPBridgeDelegate`
  - Still has same functionality but more flexible

## Example Usage Somewhere Else
```swift
class AppDelegate: UIResponder, UIApplicationDelegate {
    var bridge: CAPBridge?

    func initBridge() {
        configuration.userContentController.add(self, name: "bridge")
        bridge = CAPBridge(self, configuration.userContentController)
    }
}

extension AppDelegate: WKScriptMessageHandler {
    func userContentController(_ userContentController: WKUserContentController, didReceive message: WKScriptMessage) {
        guard let bridge = bridge else {
            return
        }
        self.userContentController(userContentController, didReceive: message, bridge: bridge)
    }
}

extension AppDelegate: CAPBridgeDelegate {
    var bridgedWebView: WKWebView? {
        return session.webView
    }
    
    var bridgedViewController: UIViewController? {
        return window?.rootViewController
    }
}
```

### DIFF
--- a/Stuff.swift
+++ b/Stuff.swift
@@ -1,0 +1,8 @@
+//
+//  Stuff.swift
+//  Capacitor
+//
+//  Created by Josh Holtz on 3/7/18.
+//
+
+import Foundation
--- a/Stuff.swift
+++ b/Stuff.swift
@@ -1,8 +1,0 @@
-//
-//  Stuff.swift
-//  Capacitor
-//
-//  Created by Josh Holtz on 3/7/18.
-//
-
-import Foundation
--- a/example/ios/IonicRunner/IonicRunner.xcodeproj/project.pbxproj
+++ b/example/ios/IonicRunner/IonicRunner.xcodeproj/project.pbxproj
@@ -294,7 +294,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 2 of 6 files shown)


jcesarmobile:
Please, use CocoaPods 1.4.0

joshdholtz:
My bad, didn't mean to commit those

janpio:
(As I fought with that earlier: Isn't it actually correct to commit the new `SPEC CHECKSUMS` in `Podfile.lock`?)

mlynch:
I got an infinite loopdey loop when testing this:

screenshot 2018-03-08 09 31 59

mlynch:
Okay I see the issue. The convenience constructor, when passed a real instance of `CAPBridgeViewController`, keeps calling itself rather than the true init function. Maybe the argument labels should be defined to remove ambiguity?

mlynch:
Actually, why is the `convenience` method even needed? It works without it and it seems redundant

mlynch:
@joshdholtz can you give me commit privileges on your repo/branch? I have some tweaks I want to push to this PR

joshdholtz:
@mlynch I _THINK_ you should have access now 😊 That infinite loop was a test for you. You passed 💪

mlynch:
Is this like saw 🤡

joshdholtz:
I am not mentally old enough or mature enough to watch Saw yet

mlynch:
Hmm still getting permission denied:
screenshot 2018-03-08 10 44 57

PR #301, opened 2018-03-09
[WIP] Electron plugin support!

IT-MikeS:
Currently this will work and load the `Filesystem` plugin on electron.

Need to add:
```
<script>
    try {
      require('@capacitor/electron/plugins');
      console.log('loaded electron plugins');
    } catch(e) {
      console.warn('Not in Electron.');
    }
  </script>
```

in your `index.html` right after the `<body>`.

this will load in all core plugins for electron or throw a warning if you arent in electron.

Then just use `@capacitor/core` in your app such as:

```typescript
import { Component } from '@angular/core';
import { NavController } from 'ionic-angular';
import {Plugins, FilesystemEncoding, FilesystemDirectory} from "@capacitor/core";


@Component({
  selector: 'page-home',
  templateUrl: 'home.html'
})
export class HomePage {

  constructor(public navCtrl: NavController) {
    console.log('Loaded at: ', (new Date()).toString());

    const { Device, Filesystem } = Plugins;

    Device.getInfo().then((deviceInfo) => {
      console.log('Device Info: ');
      console.log(JSON.stringify(deviceInfo));
    }).catch((e) => {
      console.error(e);
    });

    Filesystem.readFile({
      path: 'testfile.txt',
      encoding: FilesystemEncoding.UTF8,
      directory: FilesystemDirectory.Documents
    }).then((data) => {
      console.log(data);
    }).catch((err) => {
      console.error('FS Error:', err);
    });
  }

}
```

and all _*SHOULD*_ work.


mlynch:
Nice! That seems right on. Can you rename `electron/inject.js` to `electron-bridge.js`?

IT-MikeS:
Yeah for sure

PR #318, opened 2018-03-12
Update camera site doc

peterpeterparker:
Update the camera documentation:

- Fix doc 'base64' 
- Modify angular example to display an example where the camera is specified as source


mlynch:
🤙

PR #321, opened 2018-03-13
Use local Capacitor repos from NPM

### LINKED ISSUE #317: Use local Capacitor and CapacitorCordova from NPM
Use local Capacitor and CapacitorCordova from NPM changing the Podfile to point to the local copy instead of downloading from CocoaPods.

jcesarmobile:
Copy Capacitor and CapacitorCordova to assets folder and make the Podfile point to them

Closes #317


mlynch:
Thoughts on having those be separate npm packages instead of CLI assets?

For example, `capacitor-ios` for the iOS pod and `capacitor-android` for the Android one

jcesarmobile:
Main problem of that are the podspec files as they are in the root, outside the the ios folder.
If we move them inside then we won't be able to publish.

mlynch:
We can keep them at the root, just copy them to a temporary location before doing the deploy, much like the assets script

jcesarmobile:
Closing in favor of #329

PR #322, opened 2018-03-13
[WIP] Electron plugin support!

IT-MikeS:
Got core working, need to test user custom built plugins now.

Before Merge: 
- Test custom user plugins (should work the same as PWA) **[UPDATE: added a simple custom plugin to the test app and it just works by importing it after installing from npm. see: https://github.com/IT-MikeS/Capacitor_Test ]**
- Add `@capacitor/electron` as dependency to `package.json` of `electron-template`
- Publish `@capacitor/electron`

### DIFF
--- a/electron/src/plugins.ts
+++ b/electron/src/plugins.ts
@@ -1,0 +1,18 @@
+import { Plugins } from './global';
+import { mergeWebPlugins, mergeWebPlugin, WebPlugin } from '../../core/src/web/index';
+
+export * from '../../core/src/web/browser';
+export * from '../../core/src/web/camera';
+export * from '../../core/src/web/clipboard';
+export * from '../../core/src/web/geolocation';
+export * from '../../core/src/web/device';
+export * from '../../core/src/web/share';
+export * from '../../core/src/web/modals';
+export * from '../../core/src/web/storage';
+export * from './electron/filesystem';
+
+mergeWebPlugins(Plugins);
+
+export const registerElectronPlugin = (plugin: WebPlugin) => {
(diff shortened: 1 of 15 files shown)


electron/src/plugins.ts
```diff
@@ -0,0 +1,18 @@
+import { Plugins } from './global';
+import { mergeWebPlugins, mergeWebPlugin, WebPlugin } from '../../core/src/web/index';
```
mlynch: These should be importing from `@capacitor/core`

electron/src/plugins.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export * from '../../core/src/web/browser';
```
mlynch: What are you trying to do here? Should we change something in Core so you don't have to do this?

electron/src/plugins.ts
IT-MikeS: using `@capacitor/core` is giving build grief

electron/src/plugins.ts
IT-MikeS: This is so if bundles it all together in an `electron-bridge.js` in the dist of `@capacitor/electron` this is then injected into the apps `index.html`. You can reference `@capacitor/core` in the app and it will all play nice together, see the test app I posted, and give building it a shot, Im quite horrible at explaining this.

IT-MikeS:
Okay I address the imports and have the imports coming from `@capacitor/core` !

mlynch:
Consider this one...merged

csi

PR #327, opened 2018-03-14
Update landing-page.scss

sagarkaurav:
change email address textbox font color and weight for better readability.


mlynch:
Thanks!

sagarkaurav:
@mlynch i think that the css is not applied properly. because the css on input inside cta-form overwritten twice.
capacitor
I dont think thats big issue and you have time for this. i can solve the css problem if you can just point how to setup and run site code on local dev machine because i can't find out any instruction

PR #347, opened 2018-03-16
Make android a package

jcesarmobile:


mlynch:
Assume this is step 1 then?

jcesarmobile:
Yeah, this is only to publish. As all files are there we don't need extra scripts. Can you do a release?

PR #351, opened 2018-03-19
Electron Network plugin support

IT-MikeS:
Adds the core plugin `Network` to electron.

**Limitation:**
- At this time I can not find a way to detect type of connection, be it WiFi, Ethernet, ectra. So for now, the `connectionType` will either be set to `wifi` or `none` depending on the status of `connected`.

### DIFF
--- a/electron/src/electron/network.ts
+++ b/electron/src/electron/network.ts
@@ -24,13 +24,13 @@ export class NetworkPluginElectron extends WebPlugin implements NetworkPlugin {
       if(!window.navigator)
         reject('Network info not available');
       let connected = window.navigator.onLine;
-      resolve({connected: connected, connectionType: connected ? 'wifi' : 'none'});
+      resolve({connected: connected, connectionType: connected ? 'unknown' : 'none'});
     });
   }
 
   addListener(eventName: 'networkStatusChange', listenerFunc: (status: NetworkStatus) => void): PluginListenerHandle {
     let thisRef = this;
-    let onlineBindFunc = listenerFunc.bind(thisRef,{connected: true, connectionType: 'wifi'});
+    let onlineBindFunc = listenerFunc.bind(thisRef,{connected: true, connectionType: 'unknown'});
     let offlineBindFunc = listenerFunc.bind(thisRef,{connected: false, connectionType: 'none'});
     if(eventName.localeCompare('networkStatusChange') === 0) {
       window.addEventListener('online', onlineBindFunc);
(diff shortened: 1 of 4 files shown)


janpio:
FYI: cordova-browser has an `unkown` connection type for a similar reason: https://github.com/apache/cordova-plugin-network-information/blob/master/src/browser/network.js

IT-MikeS:
Yeah I am trying to conform with the current spec that the others use, but I'm playing with changing the the spec a bit

electron/src/electron/network.ts
```diff
@@ -0,0 +1,51 @@
...
+  constructor() {
+    super({
+      name: 'Network',
+      platforms: ['electron']
+    });
+  }
+
+  getStatus(): Promise<NetworkStatus> {
+    return new Promise((resolve, reject) => {
+      if(!window.navigator)
```
mlynch: Try to avoid `if` and `else` without curlies. Can you change this to `if (!window.navigator) {`

electron/src/electron/network.ts
```diff
@@ -0,0 +1,51 @@
...
+        reject('Network info not available');
+      let connected = window.navigator.onLine;
+      resolve({connected: connected, connectionType: connected ? 'unknown' : 'none'});
```
mlynch: Can we use the connection type here if available? https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API

electron/src/electron/network.ts
```diff
@@ -0,0 +1,51 @@
...
+    });
+  }
+
+  addListener(eventName: 'networkStatusChange', listenerFunc: (status: NetworkStatus) => void): PluginListenerHandle {
+    let thisRef = this;
+    let onlineBindFunc = listenerFunc.bind(thisRef,{connected: true, connectionType: 'unknown'});
```
mlynch: Same comment: can we use connection type? https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API

electron/src/electron/network.ts
IT-MikeS: Okay, will fix

electron/src/electron/network.ts
IT-MikeS: The current stable version of electron does not have `navigator.connection` implemented as it runs on `Chrome v59` and `navigator.connection` is only in `Chrome v61 +`. The master branch in electron is in beta for `2.0.0` and this will bring `Chrome v61` so I will update all of this when `2.0.0` hits stable

mlynch:
Nice work!

PR #352, opened 2018-03-19
Use NPM for Android dependency

### LINKED ISSUE #344: Use NPM for Android dependency
To make it easier to manage dependencies for Android, add our Android library to NPM and use that instead for new projects.

We can continue to deploy to bintray for those that are using their own dependency management for android, but we can keep it simple for those wanting us to manage that through npm

jcesarmobile:
Closes #344


mlynch:
awwww yisss!!! 🤙

Issue #307, opened 2018-03-11
[Camera] On Android devices width and height inverted


peterpeterparker:
[X] I would to submit a bug

**Current behavior**

When I take a picture with the Capacitor camera plugin on my Android device (Samsung Edge) the result is kind of inverted (see pictures behind)

**Expected behavior**

A not inverted picture ;)

Note: This doesn't happen on iOS, everything alright there

**Code**

ts:

    async takePicture() {
        const { Camera } = Plugins;
    
        const imageUrl = await Camera.getPhoto({
          allowEditing: false,
          correctOrientation: false,
          height: 1080,
          width: 1080,
          quality: 90,
          resultType: CameraResultType.Base64,
          saveToGallery: false
        });
    
        this.image = this.sanitizer.bypassSecurityTrustResourceUrl(imageUrl && (imageUrl.base64_data));
      }

html:

        <img [src]="image" />
       <button (click)="takePicture()" ion-button color="primary">Take Picture</button>

scss:

        img {
          max-width: 400px;
          max-height: 400px;
        }

**Side notes**

A side effect of https://github.com/ionic-team/capacitor/issues/217 maybe?

**Screenshots**

![screenshot_20180311-102641](https://user-images.githubusercontent.com/16886711/37251839-a023596c-2517-11e8-8d57-b6c50cee812b.png)

![screenshot_20180311-102712](https://user-images.githubusercontent.com/16886711/37251840-a4b7bc84-2517-11e8-9bf3-695c2c4b39be.png)

![screenshot_20180311-102754](https://user-images.githubusercontent.com/16886711/37251842-aab47dca-2517-11e8-8ea4-2d409284a757.png)

mlynch:
Okay can't reproduce in the simulator. Can you try something: set `saveToGallery` to `true` and share the image that was saved in the gallery when you take the photo?

Will test on a real device tomorrow

mlynch:
I may have resolved this if you can test `1.0.0-alpha.12`

peterpeterparker:
@mlynch sorry for my late answer, it was a loooong day

anyway unfortunately even with `1.0.0-alpha.13` I still face the same problem on the same phone

but, I don't face it on another phone and I don't face it when I import a photo from the gallery (congrats for the new feature btw.)

maybe something related to the Android version?

Samsung Edge / Android 7.0 / take photo => result image inverted

Samsung Edge / Android 7.0 / from gallery => no problem
[...]

PR #359, opened 2018-03-21
Android camera orientation

peterpeterparker:
Hi,

Here's a PR to fix the camera orientation related to the issue https://github.com/ionic-team/capacitor/issues/307

Furthermore to the proposal of fix itself, I did a couple of other changes.

Summarized this PR contains:

- Fix for the camera orientation using ExifInterface (only supported on Android API >= 24, that's why I tried to use it only if the Build version match this)
- Refactoring of inner class/enums in separate class/enums
- Added the missing close to InputStreams
- Added a new logError method
- Set the default quality to 90 instead of 100 (100 could generate big image and should be use when aware of the "risk")
- Move ImageUtils to new package camera since it's only use by the camera

These are proposals, of course I would not mind if you reject it ;)

Best regards
David


mlynch:
Awesome! 🚀

PR #361, opened 2018-03-22
Fix windows Android Studio path

jcesarmobile:
I was looking into a windows issue and npx cap open android wasn't working despite https://github.com/ionic-team/capacitor/commit/49e94b3c8d33962ab15ca2d3464eb358ad36bba0

As windowsAndroidStudioPath is set, it's using it's value, so changing the default path it there too.
Also changed the path in some docs


mlynch:
Cool, and I want to move this to a .rc or a local config instead since it's machine-specific

PR #362, opened 2018-03-22
fix(template): gitignore

manucorporat:


mlynch:
🚀 looking forward to more 😊

PR #365, opened 2018-03-22
Electron SplashScreen

IT-MikeS:
Experimental support for splash screen in Electron using Capacitor API


IT-MikeS:
I see where you're coming from, personally I like this order as it doesn't lead to weird taskbar flickering of the icon on windows and the main window will cover the splash and let the splash hide behind the scenes.

IT-MikeS:
Just added config options for `capacitor.config.json`. These are the defaults that can be changed:

```
"plugins": {
  "SplashScreen": {
    "imageFileName": "splash.png",
    "windowWidth": 400,
    "windowHeight": 400,
    "textColor": "#43A8FF",
    "loadingText": "Loading",
    "textPercentageFromTop": "",
  }
}
```

`imageFileName` can be something along the lines of `splash.gif` if you wanted as well.

IT-MikeS:
New config option:

- transparentWindow: boolean
  - will make the window transparent, so if you have a custom shaped Gif or static image, it will give more custom appearance

IT-MikeS:
Added another config option:

- autoHideLaunchSplash: boolean (defaults to true)
  - if true the splashscreen will hide when the main windows dom is finished loading
  - if false the splashscreen will stay up with the main window hidden until SplashScreen.hide() is called, this is good if you have some http calls or other setup type work to do after the dom is rendered.

mlynch:
w00t! Nice work 🚀

PR #368, opened 2018-03-23
Fixed order of watchPosition parameters

JStumpp:
Did not match actual code: https://github.com/ionic-team/capacitor/blob/19ed80532d68856c89101c8e08bd4a6c421ec1c6/core/src/web/geolocation.ts#L39


mlynch:
Nice catch

PR #370, opened 2018-03-23
[WIP] Toast/Notification Plugin for Electron

IT-MikeS:
Electron exposes native notifications through the HTML5 Notification API, simple plugin for that.

Needs expanding.


mlynch:
I love it, only concern is the expectations that this is a Toast UI component, rather than a notification. Might make more sense to keep this in a different Plugin?

Also, if we go the Toast UI route, we can use an Ionic web component (like we do the Camera), which would mean little UI work

IT-MikeS:
Very true, maybe this is more of the Electron equivalent of `Push Notifications` as a local option and a wire up for a service too.

mlynch:
Cool, how about this: re-implement this using the Ionic Toast component in Ionic 4: https://github.com/ionic-team/ionic/blob/master/core/src/components/toast/test/basic/index.html

This is the same approach we're taking for the Camera. For now, you can test by adding this script to the index file:

`<script src="https://unpkg.com/@ionic/core@0.1.5-2/dist/ionic.js"></script>`

IT-MikeS:
Yeah I'll work on this over the next little bit

Issue #371, opened 2018-03-26
Android keydown events are not captured


JStumpp:
The following code works in the browser and under iOS to prevent entering only numbers into an input field:
this.numberEl = this.el.querySelector("input");
        this.numberEl.addEventListener('keydown', (event: KeyboardEvent) => {
            if ((event.metaKey || event.code == 'ArrowLeft') || (event.code == 'ArrowRight') ||
                (event.code == 'ArrowUp') || (event.code == 'ArrowDown') ||
                (event.code == 'Delete') || (event.code == 'Backspace')) {
                return;
            } else if (event.key.search(/\d/) == -1) {
                event.preventDefault();
            }
        });

Under Android this does not work / does not get called.

The problem and workaround is described here:
https://stackoverflow.com/questions/45312553/android-webview-html-input-keypress-doesnt-fire

Pull request is in the works...

PR #372, opened 2018-03-26
Capturing keypresses in Android

JStumpp:
Closes https://github.com/ionic-team/capacitor/issues/371

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/CapacitorWebview.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapacitorWebview.java
@@ -1,0 +1,19 @@
+package com.getcapacitor;
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.inputmethod.BaseInputConnection;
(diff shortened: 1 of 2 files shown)


mlynch:
Thanks, can you explain a bit about what this does? Also `Webview` should be `WebView`

mlynch:
Okay cool did some research. I'll merge after the rename to `CapacitorWebView`

JStumpp:
CamelCase is fixed.

mlynch:
Sweet, thanks!

jcesarmobile:
This was causing some problems and had a few limitations, so we have changed it to use the default keyboard as the default option, but it's still possible to use the `BaseInputConnection` keyboard by using a preference in `capacitor.config.json`. You'll have to set it if you want to continue capturing keypresses.

See https://github.com/ionic-team/capacitor/pull/718 and https://github.com/ionic-team/capacitor/pull/717 for more information

ArmorDarks:
Note that this might be related to newly appeared #757

PR #378, opened 2018-03-26
Using official github android gitignore template

JStumpp:
Capacitor was missing important ignores:
https://github.com/github/gitignore/blob/master/Android.gitignore


janpio:
Maybe add that source URL as a comment as well?

JStumpp:
Good point, added!

mlynch:
I assume with the recent fix to `gitignore` on platform add, this won't cause any breaking? We ship a `.iml` to maintain the project for fast import to Android Studio

JStumpp:
Oh, you are shipping .iml, I did not realize that. Doesn't this create more problems than it solves?

mlynch:
I don't think so. The Android mantra is to be IDE agnostic, but we're very much focusing on Android Studio, so I think that's the only sticking point. It makes the first import into Android studio seamless, and the `.iml` references are all local. Basically when you first run `npx cap open android` it Just Works, no import prompts, no errors, etc.

JStumpp:
Ok, i had an issue with the gradle import that I had to click gradle sync twice. So maybe the .iml is ok for a seamless first import. But when working in a team checkin in .iml is usually not a good idea.

mlynch:
yea makes sense, I don't think this would break that but needs testing

PR #391, opened 2018-03-27
Fix a trivial doc typo

peterpeterparker:
In https://capacitor.ionicframework.com/docs/apis/filesystem

Replace `encoding: 'utf8'` with `encoding: FilesystemEncoding.UTF8`

### DIFF
--- a/site/docs-md/apis/filesystem/index.md
+++ b/site/docs-md/apis/filesystem/index.md
@@ -16,7 +16,7 @@ leave out the `directory` param to use a full file path.
 ## Example
 
 ```typescript
-import { Plugins, FilesystemDirectory } from '@capacitor/core';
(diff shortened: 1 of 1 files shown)


mlynch:
Thanks!

peterpeterparker:
@mlynch no worries

but you know what, if you allow me to share a thought, I ask my self if it would not be better to even don't display UTF-8 in the examples of the File plugin. Still displaying it in the doc but not in the examples.

I just copy/pasted the code to handle the result of the Camera plugin but it seems that reading the file result as utf8 of a photo isn't supported on iOS (at least in the simulator). Afterwards I just removed utf8 and everything worked just fine. 

So I'm just thinking that some other users might do the same "dumb" copy/paste as I just did and may actually report issue where actually there aren't

Issue #394, opened 2018-03-27
Set Action Buttons language (Modal, Camera)


douglaszaltron:
An initial solution to not leave the action buttons with the English language set, to get the language of the "localization native development region" @mlynch 

https://gist.github.com/douglaszaltron/e2a03cbb5b51e8579e71f3e62b7789f4

https://gist.github.com/douglaszaltron/c64e29c4728ebdba6ba859aeaed2c938

https://gist.github.com/douglaszaltron/c0b579e2fff9f47e14a7f135c667f124

mlynch:
Cool, would love that as a PR if you're up for it!

jcesarmobile:
closing as we also have https://github.com/ionic-team/capacitor/issues/872 for the same problem

PR #402, opened 2018-03-28
Set actions buttons language / Snackbar

douglaszaltron:
Feat(android): Snackbar

About #394

I'm working on the Snackbar for IOS, soon to pull the version.


mlynch:
Thanks, will wait on the iOS implementation. If you'd like to separate out the localization and snackbar into two PRs that could be easier, but I'm fine with one for now

douglaszaltron:
@mlynch I do not have much mastery in swift, I'm having a bit of work, I've come up[ here](https://gist.github.com/douglaszaltron/42464a5a2e9d2ace0cfa0a09be4b67cc), I'm trying to auto adjust the height, depending on the text, but I'm not getting it. need help! =/

PR #422, opened 2018-04-02
doc(getting-started): move dependencies before installation

kensodemann:
I believe when starting with a project, people tend to read and follow along with the "Getting Started" section in the order it is there. In that light, it makes sense to put the checking of required dependencies before the installation section.


jcesarmobile:
The Installation docs have a "Before you start" section where it links the 
 dependencies section (https://capacitor.ionicframework.com/docs/getting-started/#before-you-start)

But I'm ok with this.

PR #423, opened 2018-04-02
Handle allowEditing for Android

### LINKED ISSUE #384: allowEditing not implemented for Android
Either implement it or document that it's not supported for Android

In Cordova, camera plugins launches an intent to an edit app that does the crop. It doesn't really work very well if you don't use photos app.

jcesarmobile:
Closes #384


mlynch:
w00t!

PR #435, opened 2018-04-05
Handle no permissions set when requested

jcesarmobile:
Make handleRequestPermissionsResult return an error when some of the requested permission is not present in the AndroidManifest.xml

Added some functions that can be useful for plugins when checking permissions

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
@@ -3,6 +3,7 @@ package com.getcapacitor;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
+import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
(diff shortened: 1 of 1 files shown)


mlynch:
Are there any cases, (I read something with camera), where you can request permissions without defining them?

jcesarmobile:
This is for when we have code for requesting the permissions, but the permissions are not in the AndroidManifest.xml.
At the moment, the code doesn't crash, but returns an error like the user denied the permission, while the real problem is he wasn't even prompted, with this PR, the error message will indicate that the permissions are not set in the AndroidManifest.xml

Camera works without permission because we use an Intent to Camera app, and that doesn't really need the Camera permission, but if the user adds the Camera permission in the AndroidManifest.xml, the Intent won't work unless the app has asked for the permissions (strange Android things), so better be safe than sorry and keep the code asking for Camera permission despite it's not really needed

mlynch:
Makes sense thanks

PR #438, opened 2018-04-06
[Android] Allow custom webviews to load capacitor

joshdholtz:
This is the Android equivalent PR for https://github.com/ionic-team/capacitor/pull/294

## What got changed
- Added a new boolean to the `Bridge` constructor to setup a `WebClient` to load from a custom URL
- The new custom URL `WebClient` will inject all the `JSInjector` javascript at `pageFinished`
  - This also saves off the previous `WebClient` and will call those methods as well just incase the custom implementation depended on that `WebClient`

## Proof
This is using a Turbolinks Android project showing GPS and Modal plugins working 

screen shot 2018-04-06 at 1 36 43 pm

```java
public class MainActivity extends AppCompatActivity implements TurbolinksAdapter, TabLayout.OnTabSelectedListener {

    @BindView(R.id.turbolinks_view) TurbolinksView turbolinksView;
    @BindView(R.id.tab_view) TabLayout tabView;
    private JsonArray tabs;

    protected Bridge bridge;
    public CordovaInterfaceImpl cordovaInterface;
    private ArrayList<PluginEntry> pluginEntries;
    PluginManager pluginManager;
    private CordovaPreferences preferences;

    private int activityDepth = 0;

    private String lastActivityPlugin;

    private List<Class<? extends Plugin>> initialPlugins = new ArrayList<>();

    @Override
        setContentView(R.layout.activity_main);
        ButterKnife.bind(this);

        tabView.addOnTabSelectedListener(this);

        TurbolinksSession.getDefault(this)
                .activity(this)
                .adapter(this)
                .view(turbolinksView)
                .visit("https://myurl.com");

        loadConfig(this.getApplicationContext(),this);
        this.load(savedInstanceState);
    }

    /**
     * Load the WebView and create the Bridge
     */
    protected void load(Bundle savedInstanceState) {
        Log.d(Bridge.TAG, "Starting BridgeActivity");

        WebView webView = TurbolinksSession.getDefault(this).getWebView();

        WebView webview = new WebView(this);
        webView.setWebContentsDebuggingEnabled(true);

        cordovaInterface = new CordovaInterfaceImpl(this);
        if (savedInstanceState != null) {
            cordovaInterface.restoreInstanceState(savedInstanceState);
        }

        MockCordovaWebViewImpl mockWebView = new MockCordovaWebViewImpl(this.getApplicationContext());
        mockWebView.init(cordovaInterface, pluginEntries, preferences, webView);

        this.pluginManager = mockWebView.getPluginManager();
        cordovaInterface.onCordovaInit(this.pluginManager);
        bridge = new Bridge(this, webView, initialPlugins, cordovaInterface, this.pluginManager, true);

        if (savedInstanceState != null) {
            bridge.restoreInstanceState(savedInstanceState);
        }
    }
    public Bridge getBridge() {
        return this.bridge;
    }

    /**
     * Notify the App plugin that the current state changed
     * @param isActive
     */
    private void fireAppStateChanged(boolean isActive) {
        PluginHandle handle = bridge.getPlugin("App");
        if (handle == null) {
            return;
        }

        App appState = (App) handle.getInstance();
        if (appState != null) {
            appState.fireChange(isActive);
        }
    }

    @Override
    public void onSaveInstanceState(Bundle outState) {
        super.onSaveInstanceState(outState);
        bridge.saveInstanceState(outState);
    }

    @Override
    public void onStart() {
        super.onStart();

        activityDepth++;

        this.bridge.onStart();

        Log.d(Bridge.TAG, "App started");
    }

    @Override
    public void onRestart() {
        super.onRestart();
        this.bridge.onRestart();
        Log.d(Bridge.TAG, "App restarted");
    }

    @Override
    public void onResume() {
        super.onResume();

        fireAppStateChanged(true);

        this.bridge.onResume();

        Log.d(Bridge.TAG, "App resumed");
    }

    @Override
    public void onPause() {
        super.onPause();

        this.bridge.onPause();

        Log.d(Bridge.TAG, "App paused");
    }

    @Override
    public void onStop() {
        super.onStop();

        activityDepth = Math.max(0, activityDepth - 1);
        if (activityDepth == 0) {
            fireAppStateChanged(false);
        }

        this.bridge.onStop();

        Log.d(Bridge.TAG, "App stopped");
    }

    @Override
    public void onDestroy() {
        super.onDestroy();
        Log.d(Bridge.TAG, "App destroyed");
    }

    @Override
    public void onRequestPermissionsResult(int requestCode, String permissions[], int[] grantResults) {
        if (this.bridge == null) {
            return;
        }

        this.bridge.onRequestPermissionsResult(requestCode, permissions, grantResults);
    }

    @Override
    protected void onActivityResult(int requestCode, int resultCode, Intent data) {
        if (this.bridge == null) {
            return;
        }
        this.bridge.onActivityResult(requestCode, resultCode, data);
    }

    @Override
    protected void onNewIntent(Intent intent) {
        if (this.bridge == null) {
            return;
        }

        this.bridge.onNewIntent(intent);
    }

    public void loadConfig(Context context, Activity activity) {
        ConfigXmlParser parser = new ConfigXmlParser();
        parser.parse(context);
        preferences = parser.getPreferences();
        preferences.setPreferencesBundle(activity.getIntent().getExtras());
        pluginEntries = parser.getPluginEntries();
    }

    @Override
    public void onReceivedError(int errorCode) {}

    @Override
    public void pageInvalidated() {}

    @Override
    public void requestFailedWithStatusCode(int statusCode) {}

    @Override
    public void visitCompleted() {}

    @Override
    public void visitProposedToLocationWithAction(String location, String action) {}

    @Override
    public void onTabUnselected(TabLayout.Tab tab) {}

    @Override
    public void onTabReselected(TabLayout.Tab tab) {}
}
```

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -240,14 +240,13 @@ public class Bridge {
 
       private void injectScriptFile(WebView view, String script) {
 
-        // String-ify the script byte-array using BASE64 encoding !!!
+        // Base64 encode string before injecting as innerHTML
         String encoded = Base64.encodeToString(script.getBytes(), Base64.NO_WRAP);
-        Log.d(TAG, "Trying to inject - " + encoded);
         view.loadUrl("javascript:(function() {" +
                 "var parent = document.getElementsByTagName('head').item(0);" +
                 "var script = document.createElement('script');" +
                 "script.type = 'text/javascript';" +
-                // Tell the browser to BASE64-decode the string into your script !!!
+                // Base64 decode injected javascript
                 "script.innerHTML = window.atob('" + encoded + "');" +
                 "parent.appendChild(script)" +
                 "})()");
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -1,14 +1,18 @@
 package com.getcapacitor;
 
+import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
+import android.graphics.Bitmap;
 import android.net.Uri;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.text.TextUtils;
+import android.util.Base64;
 import android.util.Log;
 import android.webkit.ValueCallback;
 import android.webkit.WebResourceRequest;
@@ -114,7 +118,6 @@ public class Bridge {
   // Any URI that was passed to the app on start
   private Uri intentUri;
 
-
   /**
    * Create the Bridge with a reference to the main {@link Activity} for the
    * app, and a reference to the {@link WebView} our app will use.
@@ -122,6 +125,16 @@ public class Bridge {
    * @param webView
    */
   public Bridge(Activity context, WebView webView, List<Class<? extends Plugin>> initialPlugins, CordovaInterfaceImpl cordovaInterface, PluginManager pluginManager) {
+    this(context, webView, initialPlugins, cordovaInterface, pluginManager, false);
+  }
+
+  /**
+   * Create the Bridge with a reference to the main {@link Activity} for the
+   * app, and a reference to the {@link WebView} our app will use.
+   * @param context
+   * @param webView
+   */
+  public Bridge(Activity context, WebView webView, List<Class<? extends Plugin>> initialPlugins, CordovaInterfaceImpl cordovaInterface, PluginManager pluginManager, boolean usingCustomURL) {
     this.context = context;
     this.webView = webView;
     this.initialPlugins = initialPlugins;
@@ -145,10 +158,14 @@ public class Bridge {
     // Register our core plugins
     this.registerAllPlugins();
 
-    this.loadWebView();
+    if (!usingCustomURL) {
+      this.loadLocalWebView();
+    } else {
+      this.loadCustomWebView();
+    }
   }
 
-  private void loadWebView() {
+  private void loadLocalWebView() {
     final String appUrlConfig = Config.getString("server.url");
 
     String authority = null;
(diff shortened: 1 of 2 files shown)


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -194,10 +211,70 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
-    // Get to work
-    webView.loadUrl(appUrl);
+    // Saving off old client so no functionality is lost if
+    // custom webview implements an existing client
+    final WebViewClient oldClient = webView.getWebViewClient();
+    webView.setWebViewClient(new WebViewClient() {
+
+      @Override
+      public void onPageStarted(WebView view, String url, Bitmap favicon) {
+        if (oldClient != null) {
```
mlynch: What's the purpose of this override?

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -194,10 +211,70 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+      }
+
+      @Override
+      public void onPageFinished(WebView view, final String location) {
+        if (oldClient != null) {
+          oldClient.onPageFinished(view, location);
+        }
+
+        JSInjector injector = getJSInjector();
+        injectScriptFile(view, injector.getScriptString());
```
mlynch: Think we could make this a feature of the webview client instead of needing overrides?

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -194,10 +211,70 @@ public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request
...
+      }
+
+      private void injectScriptFile(WebView view, String script) {
+
+        // Base64 encode string before injecting as innerHTML
```
mlynch: What's the encoding process for?

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
joshdholtz: TurboLinks Android has its own `WebClient` that it depends on so overriding it with completely makes it 💣 This allows both clients to be called. The `oldClient` which is the one TurboLinks created and then this one which injects the Capacitor javascript and whatnot

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
joshdholtz: I'm not sure what you mean by "feature of the webview client" 🙃

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
joshdholtz: Since I'm using the `innerHTML` here, and writing javascript from java, I want to make sure that nothing with quotes or double quotes gets messed up when writing/injects of this javascript. So I base64 encode the string in java and then base64 decode the string using `window.atob` for safety

joshdholtz:
@mlynch Heyyy 👋 Would we still be able to get this guy looked at / merged? 😊

jcesarmobile:
A few things:

1. You removed
```
// Get to work
webView.loadUrl(appUrl);
```
From the `loadLocalWebView`, that will kill the local implementation as it won't load the initial url.

2. I think the `usingCustomURL` name might be confusing as we allow to configure the server.url to load an external url and people might think that's the custom url, so maybe `usingCustomWebView` makes more sense and this will look better:

```
if (!usingCustomWebView) {
  this.loadLocalWebView();
} else {
  this.loadCustomWebView();
}
```

3. You might want to also override this 2 methods:

```
public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceRequest request)
public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request)
```

Maybe TurboLinks doesn't need them, but other custom WebViews might, or even TurboLinks might add them in the future.

4. Also note that the `WebViewLocalServer` has some logic so if you have a `file://` url, you can replace `file://` scheme with `https://localhost:port/_capacitor_` and the `WebViewLocalServer` will return the actual file. This is needed if you use a Cordova plugin that returns `file://` paths (camera, file picker, file chooser, etc), and that won't work when using a custom WebView.

joshdholtz:
@jcesarmobile Thanks for the review! I will look into these 😊

aminland:
What's the status of this?

joshdholtz:
@aminland I need to clean this back up next week! I would still like this merged in if you all still want it 😊

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
jcesarmobile: `getWebViewClient()` requires API 26, but capacitor supports 21, so this won't work in a great part of devices.

Do you plan to continue with this or I can close the PR?

jcesarmobile:
closing since it has issues and conflicts and the requested changes have not been addressed, feel free to create a new PR or reopen in the future if you find time to finish.

Issue #227, opened 2018-03-01
Use gradle variables for setting the library versions


jcesarmobile:
Right now we are replacing the library versions  on the build.gradle with javascript, but gradle actually supports the variables, so maybe we can create the variables from plugin.xml but let users overwrite them in the build.gradle instead of using plugin.xml or package.json Cordova variables

https://medium.com/@ivan_m/use-variables-to-manage-gradle-android-dependencies-library-version-numbers-92f0e362a2fc

Related PR where we replace the variables: https://github.com/ionic-team/capacitor/pull/222

jcesarmobile:
Related https://github.com/ionic-team/capacitor/issues/206

PR #447, opened 2018-04-10
Added common dependencies variables

jcesarmobile:
Related to https://github.com/ionic-team/capacitor/issues/227

I've been having a lot of problems with Cordova plugins setting 27+ for android support libraries. There is a conflict as we set it to 26.1.0.

So this is my idea, I've created a dependencies.gradle file in the template with the common cordova variable names  set to a value that works for us. I apply it in the project build.gradle so it's accesible to all modules. This file won't be overwritten by us, it's for the user to manage, he can also add his own variables there.

This is just the initial step, next steps should be
1. Remove the CLI code that changes the plugin variables with a default version, it should leave the variable names and then gradle will take care of it.
2. Document the dependencies.gradle so users know what it is and how to use it.
3. Edit the app template, plugin template, capacitor android, etc, to use those variables instead of fixed values.

@mlynch WDYT?


nerdic-coder:
This would be great!

PR #451, opened 2018-04-11
Write edit-config and config-file tag content into AndroidManifest.xml

jcesarmobile:
Write edit-config and config-file tag content into the capacitor-andorid-plugins module's AndroidManifest.xml so manifest merger does the job for us


mlynch:
Looks fine, but that 4 promise nesting looks over complicated. Any way to simplify that?

Issue #450, opened 2018-04-11
Tune scroll decel speed on iOS


mlynch:
Make it native

wtrocki:
Resolved

jcesarmobile:
going to close since the issue is unclear

if it's about the webView scrollView decelerationRate it can be accomplished in different ways now, like providing your own webView class that is configured with the value you want, or also, any plugin can access the webView and change the value at will.

PR #454, opened 2018-04-11
Tune in scrollView deceleration rate

wtrocki:
Attempt to resolve: https://github.com/ionic-team/capacitor/issues/450


mlynch:
Awesome! I need to play with this a bit. The only thing that I'm unsure about is that this isn't the default value for `UIScrollView`, but I know in the past it has seemed like the webview's scroll is different.

jcesarmobile:
Default value is UIScrollViewDecelerationRateNormal. Cordova allows to set it with a config.xml preference. The docs say:

> This property controls the deceleration speed of momentum scrolling. normal is the default speed for most native apps, and fast is the default for Mobile Safari.

wtrocki:
Closing as code is no longer relevant

PR #461, opened 2018-04-13
Adding electron splash screen config object as optional parameter

hamzatrq:
The splash screen was not customizable because the splashOptions object was hardcoded made it an optional parameter on the constructor so that others can easily customize the splash screen.


mlynch:
Thanks!

PR #475, opened 2018-04-17
Android improvements

danielpassos:
Some improvements to the Android example app.


mlynch:
Thanks for the PR, we will review this over time since it's not critical and see if it makes sense to adopt. Cheers, hope to see more from you in the future!

PR #494, opened 2018-04-23
Fix for Electron splashscreen causing launch hang due to missing optional param

IT-MikeS:


mlynch:
Those `typeof`s have been bothering me from the last PR. Any chance you could tweak these to be simpler:

```
textColor: splashOptions.textColor || '#43A8FF'
```

IT-MikeS:
Yeah for sure

mlynch:
Awesome! 🚀

PR #504, opened 2018-04-30
Electron clipboard

techiediaries:
The electron API provides support for natively accessing the clipboard so I added an initial support for the clipboard plugin. Also added the ability to test electron in the example project.


electron/src/electron/clipboard.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  async read(_options: ClipboardRead): Promise<ClipboardReadResult> {
+
+    return new Promise<ClipboardReadResult>((resolve, reject)=>{
+        const availableFormats = clipboard.availableFormats();
+
+        for(const format of availableFormats){
+            if(format === "text/plain"){
+                return Promise.resolve(clipboard.readText());
```
mlynch: Can you change these to use the provided resolve and reject functions from the new Promise call?

techiediaries: Ok done.

mlynch:
Sweet, thanks for the PR!

PR #506, opened 2018-05-01
Electron Clipboard Improvements  

techiediaries:
Fixed some TS errors when building the @capacitor/electron package and solved getting an undefined value when pasting from the clipboard in the example.

### DIFF
--- a/electron/src/plugins.ts
+++ b/electron/src/plugins.ts
@@ -3,7 +3,7 @@ import { mergeWebPlugins, mergeWebPlugin, WebPlugin } from '@capacitor/core';
 
 export * from '@capacitor/core/dist/esm/web/browser';
 export * from '@capacitor/core/dist/esm/web/camera';
-export * from '@capacitor/core/dist/esm/web/clipboard';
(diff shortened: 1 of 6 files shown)


mlynch:
Sweet! did you mean to leave in that commented out `export`?

techiediaries:
TS complains about having two exports with the same name Clipboard. I wanted to remove it but thought about maybe rename it? I guess there is no point from keeping it?

electron/src/plugins.ts
```diff
@@ -3,7 +3,7 @@ import { mergeWebPlugins, mergeWebPlugin, WebPlugin } from '@capacitor/core';
 
 export * from '@capacitor/core/dist/esm/web/browser';
 export * from '@capacitor/core/dist/esm/web/camera';
-export * from '@capacitor/core/dist/esm/web/clipboard';
```
IT-MikeS: Since you are defining a native Electron plugin, there is no need to export the Web version any longer. This line can be removed. Nice work on the plugin :+1:

PR #547, opened 2018-05-17
Cleanup properties

wtrocki:
## Motivation

When setting up project I always needed to remove local properties. 
They are not recommended to be in the repository. I have removed them for both IOS and Android.

After removing this project can be automatically recognized as gradle project which can help potential contributors to get started faster.


mlynch:
👌

PR #551, opened 2018-05-21
Remove CORS workaround

jcesarmobile:
The CORS workaround causes problems with POST, PUT, DELETE and ORIGIN, and for GET will create a new connection that might not have the webview cookies, so removing it from the code.

People needing this because they can't configure the server to enable CORS will have to use a plugin that use native code for the connections.


janpio:
Maybe re-add it as an configurable option or plugin in the future?
From what I see at the Ionic forum, CORS problems are really widespread.
(With beta it should get obvious if this will be the same case here)

jcesarmobile:
It was only working for GET, so I think a full native networking plugin would be better than this patch.

Also, most APIs don't face the CORS problem (tested only a few). If it's a custom service, you probably own the server and can configure it too.

Issue #535, opened 2018-05-11
Local notifications core plugin


meltuhamy:
Although there is a [cordova plugin](https://github.com/katzer/cordova-plugin-local-notifications), I believe local notifications should be a **core** plugin - it would be great to have a unified API for displaying, scheduling and interacting with notifications on all platforms including PWAs. Also, notification APIs are constantly improving on each platform, and although the cordova plugin tries to keep up, it often _breaks_ between platform releases. Having it a core plugin would hopefully ensure proper maintenance. 

I was wondering if such a plugin would appear on the roadmap any time soon?

jcesarmobile:
It's already implemented for iOS, but still missing for web, Android and Electron

chuckgreenman:
I've got some open time coming up this weekend.  Will take a crack at notifications on Android.

PR #553, opened 2018-05-22
Local Push notifications for Android

wtrocki:
## Motivation

Implement LocalNotifications in Android core #535

## Progress

1) [x] Implement basic scheduling
2) [x] Implement cancel
3) [x] Wrap other required plugin methods
4) [x] Implement ActionHandler
5) [x] Timers - notifications in future
6) [x] Testing and simple verification
7) [x] On (cron jobs) support
8) [x] Fix cancel bugs (on notification dismissed)
9) [x] Verify listener support
10) [x] Fix on activity result bug (remove workaround) 
11) [x] Fix bug with "Reply to action not dismissing the notification"


android/capacitor/build.gradle
```diff
@@ -41,6 +42,8 @@ repositories {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
     implementation 'com.android.support:appcompat-v7:27.1.1'
+    implementation 'com.android.support:support-compat:27.+'
```
jcesarmobile: I prefer using `27.1.1` as it's the version used in the other support libraries we have.
Using `27+` will require less updates, but can also be less predictable.

In case we stick to `27+` then we should use that for all the libraries we have.

jcesarmobile:
1. I don't think the activity is needed
1. b. Not sure if that's what you mean, but plugins have to overwrite handleOnActivityResult to get and handle intent results
2. I think the timer is the most important part of local notifications (also the harder), I think people interested in them what to schedule and show in the future, not at the moment. But maybe for a first version can be enough, better wait for confirmation from @mlynch.

About the problem, what are you doing to get that?

wtrocki:
> b. Not sure if that's what you mean, but plugins have to overwrite handleOnActivityResult to get and handle intent results

This answers my question. I was looking for a way to get activity launch intent and `handleOnActivityResult` could be the way to get that when application is launched from notification.

> I think the timer is the most important part of local notifications. 

I mean could we just implement it without some corner cases - for example when user restarts the phone. This requires additional permissions and can make core plugins require permissions that are very specific (launchOnStartup etc.)

jcesarmobile:
Ah, then yeah, people can report those corner cases and maybe be implemented in the future.

No, I haven't seen that error

android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotification.java
```diff
@@ -0,0 +1,199 @@
...
+        call.error("id field missing in Notification cancel method ");
+        return;
+      }
+    }
+    call.success();
+  }
+
+  @PluginMethod()
+  public void getPending(PluginCall call) {
+    // IOS specific
```
jcesarmobile: We can probably keep track of the scheduled notifications and return that instead of asking the system.
Anyway it's ok if we don't have this yet, but I would rather return an error with non implemented message than returning success

android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotification.java
```diff
@@ -0,0 +1,199 @@
...
+    // Requires special permissions that may not be desired for a plugin.
+    // There is no need to implement this method for Android
+    // https://developer.android.com/reference/android/service/notification/NotificationListenerService#getActiveNotifications()
+    call.success();
+  }
+
+  @PluginMethod()
+  public void registerActionTypes(PluginCall call) {
+    // IOS specific call
```
jcesarmobile: Same here, if not implemented probably better to call error with not implemented

android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotification.java
```diff
@@ -0,0 +1,199 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+
+/**
+ * Plugin for creating local notification
+ */
+@NativePlugin()
+public class LocalNotification extends Plugin {
```
jcesarmobile: Class should be LocalNotifications, as that's the iOS name, otherwise it won't work with current API as the plugin execution is based on the class name.

android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotification.java
```diff
@@ -0,0 +1,199 @@
...
+      JSONObject extra = notification.getExtra();
+      if (extra != null) {
+        Bundle extras = new Bundle();
+        extras.putString("data", extra.toString());
+        mBuilder.addExtras(extras);
+        intent.putExtras(extras);
+      }
+
+      // notificationId is a unique int for each notification that you must define
+      notificationManager.notify(Integer.valueOf(notification.getId()), mBuilder.build());
```
jcesarmobile: Interesting, will change the iOS implementation and example code to use ints too

wtrocki:
Thanks for review. Going to address comments and put scheduler code.

android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotification.java
wtrocki: Going to do that in this PR.

jcesarmobile: Too late, is already changed 😜

jcesarmobile:
Ah, another thing I noticed is the phone was vibrating, but not showing the notification. I think that’s because it doesn’t have an icon. So if no icon it’s provided it should show the app icon.

wtrocki:
Added scheduler support. There is still plenty TODO elements that I'm going to resolve bit by bit.

wtrocki:
@jcesarmobile Ready for review. I would love to get what would be missing/essential to implement and what could be done in separate PR's.

jcesarmobile:
Code looks good.
A few things:
1. `SimpleDateFormat` complains about the final `X`, I don't think it's available on Android, I've tried replacing it with `Z` or `'Z'`. It works with both, but the notification fires right away, not in 10 seconds. Not sure if the problem is because of that, as the Z is for the time zone or if it's another thing.
2. JsonParserUtils.java looks similar to the JSObject class, maybe you can move the code there (`getInt` and `getBoolean` as `getString` is already there), and use JSObject class instead of JSONObject class.
3. Notification scheduled for each 1 minute fires right away and then every minute, I think it shouldn't fire right away, only every minute.
4. Cancel doesn't really cancel, it just remove an already fired notification, cancel should cancel the trigger, right now I had to delete the app to stop receiving the notifications 😂 , so it's more like a delete than a cancel.
5. Schedule doesn't support `on`, only `at`, should support both as iOS.

wtrocki:
@jcesarmobile World class review! Making changes now!

jcesarmobile:
Sorry, missed your edit.
What I've read is that Android's SimpleDateFormat is different from Java 7 one and doesn't accept the X format (or at least didn't accept it on Android 6 and older, not sure if they have fixed it later)

I think dates are easier for users than timestamps, so I would keep it unless it's extremely complex to handle them. I've taken a look into the Cordova plugin and it also uses dates. But doesn't use SimpleDateFormat to handle them.
In case it's changed to use timestamps, it should also be changed for iOS, as it's already working fine with Date.

About the remove, for iOS I've added it to the push plugin, but it can remove both, I assume it's the same for Android, so not sure what to tell you. Maybe we should add it in both push and local.

wtrocki:
Couple minor bugs found during the extensive tests. 
Going to address them. End to end functionality is already implemented.

wtrocki:
Remaining bugs:
- Repeating cron based notifications id's are removed when user interacts with them. We need to save them twice as repeating id's and as actual notification instance.
- Interaction with notification custom actions is not dismissing notification. 
This is probably related with lack if icons. Need to figure out how to provide icons (we can associate them with id's or add additional field to js or in last resort setup fixed transparent icon for now)

jcesarmobile:
Latest changes look awesome!

There are 2 code problems.
1. Looks like you did a project wide replace that changed "save" with "appendNotificationIds" in multiple places.
2. The iOS fix you did is reintroduced in this PR, probably needs a rebase with master to remove it.

I've also seen 2 functionality problems
1. Sometimes the get pending doesn't return all of them, not sure if that's what you mean in your first point of remaining bugs about ids getting removed.
2. On the "on 1 minute" example if I tap the notification it fires again and again (if I continue tapping it) for the next minute, should only fire once, and when tapped disappear.
3. When an "every" is already displayed, the next one won't fire, not really sure if this is really a bug.  As the id is the same, it's possible that android doesn't show the new one as the id didn't change and it's already displayed.
4. localNotificationActionPerformed doesn't return title, body and probably other fields, only the id and extra (when available)

If you fix the 2 code problems I think this is good to be merged.
The pending issues can be fixed in future PRs.

wtrocki:
> There are 2 code problems.
Resolved!

> Sometimes the get pending doesn't return all of them,

It depends how we would see this working. Pending notification will be removed when user dismiss or action on the notification. If notification with the same id is already active in the bar it will be replaced.

Number of the notifications visible in the notification bar should be the same as number of pending notifications.

However in the demo application I decided to randomly assign notification id and they are overlapping. This is probably causing that issue. Going to fix this in demo application to assign more randomized notification numbers.

ios/Capacitor/Capacitor/Plugins/LocalNotifications.swift
```diff
@@ -271,6 +271,9 @@ public class CAPLocalNotificationsPlugin : CAPPlugin {
     case "day":
       let newDate = cal.date(byAdding: .day, value: 1, to: now)!
       return DateInterval(start: now, end: newDate)
+    case "day":
```
jcesarmobile: This is what I mean by the iOS fix you did and you are adding back

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -39,7 +39,7 @@
   // Reference to the PluginHandle wrapper for this Plugin
   protected PluginHandle handle;
 
-  // A way for plugins to quickly save a call that they will
+  // A way for plugins to quickly appendNotificationIds a call that they will
```
jcesarmobile: you missed this one

jcesarmobile:
> > Sometimes the get pending doesn't return all of them,
> 
> It depends how we would see this working. Pending notification will be removed when user dismiss or action on the notification. If notification with the same id is already active in the bar it will be replaced.
> 
> Number of the notifications visible in the notification bar should be the same as number of pending notifications.
> 
> However in the demo application I decided to randomly assign notification id and they are overlapping. This is probably causing that issue. Going to fix this in demo application to assign more randomized notification numbers.
> 
> **Is that going to resolve this issue?**

I don't think using random ids in the example will fix it.

Let's give you an example:
I click the  "Schedule every minute"
Click the "Get pending" and get the one with id = 3
When it fires I tap it
Click the "Get pending" and get none
But the notification keeps firing every minute.
I have to manually cancel it with `window.Capacitor.Plugins.LocalNotifications.cancel({notifications: [{id:3, options:""}]})` as the cancel button first queries for the pending ones.

> When an "every" is already displayed

> We can cancel previous ones with that id. Does this should be only for timed notifications or general rule = Cancel existing notification with that id and create new.  We can detect that basing on PendingIntent. Android is not going to 
> push notification if it's not different. However if any of the value in PendingIntent will be different then it will be replaced - according to flags we setup. So for case of every - we always have the same content of the notification.
> 
> **Should we cancel existing notification every time or for timer notifications?**

I think when you schedule one and there is already another one with the same id, it should cancel the previous one and then schedule the new one.

> > localNotificationActionPerformed doesn't return title
> 
> I actually wanted to ask if we really need to return it. On IOS api already has that data. On Android we need to save this and retrieve it (it's tricky as pending intent that is triggering notification is part of the notification). Also tried to  make LocalNotification object Parcelable and save it to the bundle but this did not worked great. Is this information useful for js user? If this is useful I can think of couple hacks to get that data into the intent or fetch it from NotificationManager
> 
> **WDYT?**

Yeah, probably doesn't make much sense for Local Notifications, but I think we should do it for parity with iOS. As the id is present, can't you just check on the stored notifications and get the information form there?
Anyway, this is not very important, can be added in the future (or never if it's too hard)

> > On the "on 1 minute" example if I tap the notification it fires again and again (if I continue tapping it)
> 
> To be sure. It fires many times when notification is pressed or scheduled? 
> Cannot reproduce that 

Example:
I click the schedule when clock matches next minute button
The notification appears after 1 minute
I tap or swipe the notification and it fires again
I tap or swipe the notification and it fires again
I tap or swipe the notification and it fires again
If I wait for ~ 1 minute and tap  or swipe the notification, it doesn't fire again

It's like as it fires "on x minute", it continues firing as long as it's the same minute where it was supposed to be fired.

wtrocki:
> I think when you schedule one and there is already another one with the same id, it should cancel the previous one and then schedule the new one.

💯 👍 Fully agree! On my way to resolve it.

> Let's give you an example:

💯 👍  This is the issue: "Repeating cron based notifications id's are removed "
On my way to resolve it.

> Example:

Good catch 👍 
I will need to save time unit that triggered change in the intent and ignore this unit later due to fact if next trigger is going to still match that minute. 
This explains why this works this way: endless loop until next minute will happen.
This code deserves some unit tests so going to write them to cover all the cases.

PR #568, opened 2018-05-31
Animate iOS statusbar when changing style

sbannigan:
This will fade the status bar from white to black (and black to white) when changing the style instead of it just changing immediately. Just wanted the transition to look a little cleaner.


mlynch:
Looks sweet, thanks for the PR!

Issue #567, opened 2018-05-31
Found 0 Capacitor plugins twice


jcesarmobile:
When you start an Android project without plugins, it shows
Found 0 Capacitor plugins for android:
Found 0 Capacitor plugins for android:

Second one should be 0 Cordova plugins

PR #575, opened 2018-06-08
capacitor-567: Always print cordova plugins

wtrocki:
See #567

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@capacitor/cli",
-  "version": "1.0.0-alpha.25",
+  "version": "1.0.0-alpha.39",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/cli/src/android/update.ts
+++ b/cli/src/android/update.ts
@@ -25,8 +25,6 @@ export async function updateAndroid(config: Config) {
     }
   }
 
-  printPlugins(capacitorPlugins, 'android');
-
   removePluginsNativeFiles(config);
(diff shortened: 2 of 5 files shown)


jcesarmobile:
@wtrocki thanks for the PR, but the problem really is that the "Found 0 Capacitor plugins for android:" appears twice.
I though the problem was that it was showing "Capacitor" instead of "Cordova" on the second one, but with your change now I get

> Found 0 Capacitor plugins for android:
> Found 0 Cordova plugins for android
> Found 0 Capacitor plugins for android:

So it still appears twice.

jcesarmobile:
I've fixed it in https://github.com/ionic-team/capacitor/pull/576

But maybe it's a good thing to always print the Cordova plugins too, @mlynch what do you think?

I see you've made some changes since I reviewed, so you'll have to revert the changes back to just print Cordova plugins. But wait from Max confirmation if he likes to always print Cordova plugins or not

wtrocki:
Let's close that then

wtrocki:
Even after mentioned PR we are still calling print method twice. 
In my PR we just do that once so less input is printed. 
Let me know if you will want any further improvements. 

Bonus is that I learned cli architecture (it is impressive) so feel free to ping me on any future improvements :)

jcesarmobile:
print method is called multiple times, but it's one for Capacitor plugins, other for Cordova plugins and another for incompatible Cordova plugins.
The problem was on the incompatible plugins print, as it had a length === 0 (no incompatible plugins found), it was going to the remaining else and printed as 0 Capacitor plugins, so by checking the type on the else we avoid that problem

PR #633, opened 2018-07-03
Push notifications ios

jcesarmobile:
Moved part of the LocalNotifications code to a common class as both Push and Local are handled by UNUserNotificationCenter

### DIFF
--- a/ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
+++ b/ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
@@ -9,7 +9,6 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
 
   public override init(){
     super.init()
-    print("init notification center delegate")
     let center = UNUserNotificationCenter.current()
     center.delegate = self
   }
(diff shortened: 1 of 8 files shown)


ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
```diff
@@ -0,0 +1,145 @@
...
+  }
+  /**
+   * Request permissions to send notifications
+   */
+  public func requestPermissions() {
+    print("native request permissions")
+    // Override point for customization after application launch.
+    let center = UNUserNotificationCenter.current()
+    center.requestAuthorization(options:[.badge, .alert, .sound]) { (granted, error) in
+      // Enable or disable features based on authorization.
```
mlynch: Something supposed to be here?

ios/Capacitor/Capacitor/Plugins/PushNotifications.swift
```diff
@@ -1 +1,83 @@
...
+
+  }
+
+  @objc public func didFailToRegisterForRemoteNotificationsWithError(notification: NSNotification){
+    guard let error = notification.object as? Error else {
+      return
+    }
+    notifyListeners("didFailToRegisterForRemoteNotificationsWithError", data:[
+      "error": error.localizedDescription
+      ])
```
mlynch: A few of these dangling `])` at the wrong indentation level

jcesarmobile:
Fixed that and removed some print lines I left behind.

pushNotificationReceived is fired when the notification is received and the app is in foreground, or when you open the app by tapping a notification from notification center.

mlynch:
Saweeeet!

PR #642, opened 2018-07-09
Gradle scoped packages fix

### LINKED ISSUE #641: Android build failing with scoped packages
Plugins using scoped packages are failing to build on android

triniwiz:
Fixes #641


jcesarmobile:
Thanks for the PR, but `getGradlePackageName` is just returning the id as the id is `@teammaestro/capacitor-single-signon`, so the switch goes to the default every time.

If what you want to do is to replace the `@` and the `/` then I think it can be changed to just:
```
return id.replace('@','').replace('/', '-');
```

Another option would be to remove the `@teammaestro/` part on the names from the gradle files as that's only used for naming the modules, so as long as the names match in both `capacitor.settings.gradle` and on `capacitor.build.gradle`, it doesn't need to be the actual package name. We are open for your opinion here.

Let me know if you want to update your PR. If not, I will proceed with the first option.

triniwiz:
Ok will do

jcesarmobile:
Great, thanks! Merged.

PR #643, opened 2018-07-10
iOS - Adds presentationStyle to BrowserOpenOptions

### LINKED ISSUE #616: Browser is opening in popover instead of fullScreen
By Default the SafariViewController should open in fullScreen mode based on more standards out there. Currently it is being hard set to popover. I would like to create a PR setting this to fullScreen mode instead.

Please review my fork / PR (Cannot figure out how to PR back to this repo): https://github.com/jpinkster/capacitor/pull/1

Lines:
https://github.com/ionic-team/capacitor/blob/99455d2a2e419ca12e3ee991b70c6f7216448924/ios/Capacitor/Capacitor/Plugins/Browser.swift#L20

sbannigan:
Fixes #616. Also changes default to fullscreen. See: [https://developer.apple.com/documentation/uikit/uimodalpresentationstyle](UIViewController API) for other possible presentation styles. I just stuck with fullScreen and popover for now.

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -237,6 +237,11 @@ export interface BrowserOpenOptions {
    * A hex color to set the toolbar color to.
    */
   toolbarColor?: string;
+
+   /**
+   * iOS only: The presentation style of the browser. Defaults to fullscreen.
+   */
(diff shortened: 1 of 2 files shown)


mlynch:
Thanks for the PR!

PR #648, opened 2018-07-11
Change authority

### LINKED ISSUE #636: On app reboot `localStorage` is lost
First of all, thanks for all your efforts in making PWA great again!

I'm not sure is it a bug or my misunderstanding.

Our app API writing some data to `localStorage`, and relies on it during auth. After building the app with Capacitor for Android, everything worked fine. However, if the app was forcefully closed (as shut down process), the next time app turned out to launch with clear `localStorage`.

That's happens everytime the app is shutted down.

My understanding was that `localStorage` should persist until Android or user _somtimes_ decides to clear it. However, in my case, it happens every time.

I assume that might happen because `localStorage` is destroyed with Chrome's webview when the app is closed, or maybe each time app starts it gets a different domain name, which will also mean a new, clear `localStorage`.

Is it correct behavior?

I know that easy solution is Capacitor's `Storage`, however, we can't use it for 3rd-party tools, like AWS Amplify, which using directly `localStorage` methods. I could override `localStorage` methods with Capacitor's `Storage`, but that won't work because Capacitor's one are Promise-based.

Anyway, it's still solvable by watching changes in `localStorage` and gradually pushing them to the `Storage`, and on hard shut down restore `localStorage` state from `Storage`. Though, I wanted to ensure that it is correct approach and that wiped `localStorage` on every shut down is expected behavior.

jcesarmobile:
Removed the random UUID generation on the authority and use localhost instead.

The port will be picked from server.port configuration in capacitor.config.json.
If the configuration is not present, then it will generate a random port number and store the port in SharedPreferences for reuse.

Closes #636

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -16,6 +16,7 @@ import android.webkit.WebResourceResponse;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
 import android.webkit.WebViewClient;
+import android.content.SharedPreferences;
 
 import com.getcapacitor.plugin.Accessibility;
 import com.getcapacitor.plugin.App;
@@ -52,6 +53,7 @@ import java.net.URL;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ThreadLocalRandom;
 
 
 /**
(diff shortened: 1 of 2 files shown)


ArmorDarks:
@jcesarmobile Thanks a lot for the fix! Is there any ETA for release?

mlynch:
1.0.0-beta.3 just landed, should be in that!

PR #650, opened 2018-07-11
Make it possible to configure iOS port with server.port

jcesarmobile:
The port will be picked from server.port configuration in capacitor.config.json.
If the configuration is not present, keep the old behaviour and use a random port stored in UserDefaults


imsingh:
it looks like port value from config is not used. The reason seems to be `self.config` is empty when you try to fetch ```server.port``` value.

jcesarmobile:
Fixed in https://github.com/ionic-team/capacitor/pull/742.

@imsingh if you find some problem in the future, it's better if you create a new issue, as comments in old Pull Requests might be missed.

imsingh:
@jcesarmobile Thanks. I will make sure next time.

Issue #590, opened 2018-06-19
Uncaught (in promise): SplashScreen does not have web implementation (no Promise to chatch)


mburger81:
Yes we know, SplashScreen is for mobile devices and probably not supported also on PWA, at least it is not documented.

But if we run SplashScreen on `ionic serve`, we run into this error. So having a look from PWA documentation https://capacitor.ionicframework.com/docs/basics/progressive-web-app there is

_**If your app makes native plugin calls that don't have a web substitute, such as SplashScreen.show(), the app will allow those calls without crashing. Calls that return a promise will return a rejected promise, which you should be handling in your app anyways.**_

But neither `show()` nor `hide()` has promise to catch. We can only check for `Capacitor.isPluginAvailable` and omit the calls.

Same thing on `Network` plugin. Shouldn't work the plugins in ionic serve as for PWA's?

mlynch:
`show()` should return a promise you can catch

bhandaribhumin:
@mlynch same issue with Plugins.StatusBar.setBackgroundColor({ color: `#000000` }); 
ERROR Error: Uncaught (in promise): StatusBar does not have web implementation.

josh-m-sharpe:
@mlynch philosophically, isn't it capacitor's job to "catch" this exception?  Or better yet, shouldn't capacitor provide a web implementation for things like SplashScreen and simply be a noop?

Sampath-Lokuge:
This works for me on `StatusBar`

`import { Plugins, PluginRegistry, Capacitor } from '@capacitor/core';`

`const { SplashScreen, StatusBar }: PluginRegistry = Plugins;`

`if (Capacitor.isPluginAvailable('StatusBar')) { StatusBar.show() };`

PR #652, opened 2018-07-12
fix(SplashScreenPlugin): use correct return type

pwespi:
`show` and `hide` actually return a `Promise`. Related to #590.

Also should the `callback?: Function` be there?


mlynch:
🙌

PR #672, opened 2018-07-23
Destroy webview on app exit

### LINKED ISSUE #669: [Android] App.exitApp() doesn't close WebView
I have this code in my app:
```
import { Plugins } from '@capacitor/core';
const { App } = Plugins;

App.addListener('backButton', () => {
  App.exitApp();
});
```
When I press the hardware back button the app closes as expected. When I open it again there is a second WebView instance in Chrome DevTools and I get this error from firebase:

```
Error enabling offline storage. Falling back to storage disabled: FirebaseError: [code=failed-precondition]: There is another tab open with offline persistence enabled. Only one such tab is allowed at a time. The other tab must be closed or persistence must be disabled.
```

When I keep closing and reopening the app, more and more WebView instances are added.

So I guess `App.exitApp()` should either close the WebView or the same WebView instance should be used when reopening the app.

Capacitor v1.0.0-beta.1

jcesarmobile:
Call removeAllViews and destroy on webview object when the app is closed so it doesn't create a new webview on next launch

Closes #669

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 public class BridgeActivity extends AppCompatActivity {
   protected Bridge bridge;
+  private WebView webView;
   public CordovaInterfaceImpl cordovaInterface;
   private ArrayList<PluginEntry> pluginEntries;
   PluginManager pluginManager;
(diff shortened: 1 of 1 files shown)


mlynch:
If we remove it on detach do we need to add it on attach?

jcesarmobile:
No, it's onDetachedFromWindow because you can't call destroy while it's still attached, so we need to wait for it.

If you mean that as we destroy it we need to recreate it, onDetachedFromWindow is called when the activity is destroyed, so onCreate and our init methods will be called when the app is opened next time, and we create the webview on init method, so no need to do it on onAttach

PR #673, opened 2018-07-24
Append special chars to active element

jcesarmobile:
super.dispatchKeyEvent(event); doesn't dispatch special chars properly, so they are not passed to the webview.

This appends them to the active element, but won't fire Keyboard Events for them.


mlynch:
Worth a try I suppose!

Issue #534, opened 2018-05-10
Network does not have web implementation.


ix-xerri:
I am using the Network API but getting a console error with:

`ERROR Error: Uncaught (in promise): Network does not have web implementation.`

This is on v1.0.0-alpha.37

ix-xerri:
I thought it was working since there is paw written on Network docs. Any idea when it will be available on the web?

mlynch:
Re-opening to track since we need to add this

IT-MikeS:
At the moment this may not be possible with out doing a fetch to test connectivity and connectionType being `unknown` a lot of the time due to the acceptance of the `Network Information API` in current browsers.

https://caniuse.com/#feat=netinfo

ix-xerri:
Seems like all was fixed in `alpha.38`

ix-xerri:
Any plans of having an implementation in the browser soon? It hit beta and the error still persists.
Thanks

mburger81:
Same for me, not sure why there is a discrimination between `PWA` and `ionic serve` or `browser apps`. We are always running in a browser and we have always to use web apis, so or it work or doesn't.
Right?
Ionic Native Network (cordova) is working well (not for all properties)
Thx!

ix-xerri:
Since its currently on beta, I don't think this feature will be implemented on web soon, right?

ix-xerri:
Since its currently on beta, I don't think this feature will be implemented on web soon, right?

mlynch:
Added a basic network plugin that is ported from the Electron one. It won't really do anything on browsers that don't support it but at least it won't throw a bunch of errors.

PR #675, opened 2018-07-25
[WIP]: Network web implementation

John-Luke:
Fixes: #534

As a lot of browser have not this feature or haven't the `type` property in `navigator.connection` (like Chrome for desktop, for android already have), I return `unknow` as the type of connection.

### Links:
 - [NetworkInformation API](https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation)
 - [NetInfo caniuse](https://caniuse.com/#feat=netinfo)
 - [Tutorial with an old spec of this feature](https://www.davidbcalhoun.com/2010/using-navigator-connection-android)

_P.S.: I put WIP mark because I don't know if this implementation works completely or if I should do any work on it_

### DIFF
--- a/core/src/web/net.ts
+++ b/core/src/web/net.ts
@@ -1,0 +1,37 @@
+import { WebPlugin } from './index';
+
+import {
+  NetworkPlugin,
+  NetworkStatus
+} from '../core-plugin-definitions';
+
+declare var navigator: any;
+
+export class NetworkPluginWeb extends WebPlugin implements NetworkPlugin {
+  constructor() {
+    super({
+      name: 'Network',
+      platforms: ['web']
+    });
+  }
+
+  getStatus(): Promise<NetworkStatus> {
+    return new Promise<NetworkStatus>((resolve, reject) => {
+      const connection = navigator.connection || navigator.mozConnection || navigator.webkitConnection;
+      if (!connection) {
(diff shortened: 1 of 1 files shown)


mlynch:
Looks right to me! Should see about supporting the events that it emits. I am blanking on how that's done in the web impl right this second...

TwitchBronBron:
For supporting the events that it emits, can you use [the electron implementation](https://github.com/ionic-team/capacitor/blob/6a170d34a810ce09d15d7817c6ee102eee6d84e8/electron/src/electron/network.ts#L44) for reference?

nethzyr:
You can use `connection.effectiveType` for Google Chrome.

kerosin:
Like TwitchBronBron suggested, listening to connection change events would be great.

*online* and *offline* events on ```window``` are supported on all major browsers (see https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/Online_and_offline_events).

On a few browsers there is also additional a *change* event fired on ```navigator.connection``` (mainly Android, see https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/onchange)

mlynch:
Closing as we ported over the network plugin from Electron which was very similar but also handled the status events. Thanks for the PR!

PR #679, opened 2018-07-26
fix usage of == operator in string comparison

### LINKED ISSUE #678: Android: Do not use the == operator for String comparison in Java
In Java you should **never** compare Strings with the == operator because unlike Javascript it compares the object references and not the actual content.

Comparing with == might work because of sth called [String interning](https://en.wikipedia.org/wiki/String_interning) which is a VM optimization but it's not secure to rely on that and there are cases in which it definitely does not work.

I will create a PR as soon as I did some further testing.

BR, Michael

moberwasserlechner:
There are only a few places where == is actually used. Most others are fine.

Fixes #678.


mlynch:
Awesome, thanks for the PR!

PR #681, opened 2018-07-26
Small inconsistencies in the Device plugin's Core and Android implemenation

moberwasserlechner:
Android: 
1) The platform value should be lower case to match the docs in [DeviceInfo](https://github.com/ionic-team/capacitor/blob/78dfcc419b1e3417086569ca9b3d5411531a4391/core/src/core-plugin-definitions.ts#L380)
2) Added a few null checks to prevent NPEs

Core: 
There are 2 platform values in this class: "web" and "pwa". I changed it to "web" because of the doc in DeviceInfo.


mlynch:
Looks good, thanks!

PR #684, opened 2018-07-27
iOS: Fix BackgroundTask beforeExit method name

### LINKED ISSUE #685: Background Task plugin not working for iOS
Method has incorrect name in [https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/DefaultPlugins.m#L18](DefaultPlugins.m). Should be `beforeExit`.

sbannigan:
Renamed from 'exec' to 'beforeExit'. Would throw an error when trying to schedule a background task. Fixes #685


jcesarmobile:
Nice catch, thanks!

Issue #655, opened 2018-07-13
Links are opened inside the WebView (iOS)


danielsimon1:
All links inside iOS Capacitor apps are opened inside the WebView, also external ones. After a user clicks an external link, the site just opens in place and there is no way back to the app. Closing and re-opening the app is required.

So my suggestion is that when a user clicks a link, the standard browser gets opened (just like in Android).

PR #699, opened 2018-08-01
Fixes iFrame behavior from opening in a new window

sean-perkins:
Reverts: https://github.com/ionic-team/capacitor/pull/664

Currently all framed content is opening in a new window. This is a bad workflow since certain types of content (Microsoft docs, E-learning docs) have custom viewers that have to be embedded to render correctly. 

Let the app decide how it consumes link taps instead of forcing a behavior for iOS.

Also, Cordova's behavior does not check for http/https as well: https://github.com/apache/cordova-ios/blob/d44cec6266f5890802268e996722a380a0e51927/CordovaLib/Classes/Private/Plugins/CDVUIWebViewEngine/CDVUIWebViewDelegate.m#L199

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -175,7 +175,7 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScr
     }
     
     if let scheme = navUrl.scheme {
-      let validSchemes = ["tel", "mailto", "facetime", "sms", "maps", "itms-services", "http", "https"]
+      let validSchemes = ["tel", "mailto", "facetime", "sms", "maps", "itms-services"]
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Will review this tomorrow, but it looks like this will bring https://github.com/ionic-team/capacitor/issues/655 back, so the proper fix should be to find a way of opening links outside, but iframes inside. 

I think that should be possible by checking the `WKNavigationType` in the `WKNavigationAction` and see if it's `WKNavigationType.linkActivated`. But still, links in the iframes will open outside.

Cordova handles this with the whitelist plugin, but we don't have that in Capacitor

jcesarmobile:
I've created [this PR](https://github.com/ionic-team/capacitor/pull/704) to allow internal iframe navigation while keeping current link behaviour (to be consistent with Android), going to close your PR for now.

But you made good points about letting the user decide, so, can you create a new issue about it so we can discuss it there? you can ping max and me there.

PR #700, opened 2018-08-02
Android: Unify logs

moberwasserlechner:
To make Capacitor's log stream better searchable by Logcat all Capacitor Android runtime and plugin log tags should have the same form.

I introduced a new class `com.getcapacitor.LogUtils` with a **core** and **plugin** tag method, which should be used for the obvious purposes.

```
Log.i(LogUtils.getCoreTag(), "Your message");
Log.i(LogUtils.getPluginTag(), "Your plugin message");
```

These methods might have subTags: String[] param, joining the Strings together separated by a slash.

For plugins extending `com.getcapacitor.Plugin` the super method `com.getcapacitor.Plugin#getLogTag()` should be used, which builds the log tag using a prefix and the class name of the sub class.

Refactorings during the cleanup:
* Replaced all `System.out.println()` and `printStackTrace()` with log statements
* Changed some log levels, where appropriate
* Removed unused or not needed logging helpers

After changing this I was able to use the filter string `/Capacitor` in LogCat and get only log statements belonging to Capacitor or search by `Capacitor/Plugin` to get everything related to plugins.

BR,
Michael

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/AndroidProtocolHandler.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/AndroidProtocolHandler.java
@@ -18,7 +18,6 @@ import java.util.List;
 
 
 public class AndroidProtocolHandler {
-  private static final String TAG = "AndroidProtocolHandler";
 
   private Context context;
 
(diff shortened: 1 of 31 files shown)


mlynch:
Looks awesome, thanks so much for doing this 🚀

PR #714, opened 2018-08-08
Add input file functionality to Android webview

### LINKED ISSUE #631: File Inputs not working inside Capacitor apps on Android
I'm trying to implement a file input inside my Capacitor app. On iOS everything is working as expected, but on all tested Android devices just nothing happens when the user clicks on the file input element. I tried a simulator running a Google Pixel 2 and a Huawai P9 Plus. If the app is built with Cordova, the file input works again.

**Steps to reproduce:**
1. Follow the steps to create a project for a Capacitor app (I created an Ionic project with the sidemenu template)
2. Add the following snippet somewhere in your HTML code: `<input type="file">`
3. Build the app for Android

**Expected behavior:**
When the user clicks on a file input element inside a Capacitor app, some kind of dialog should pop up to select a file on the device.

jcesarmobile:
This code is needed to show a file chooser when an input with type file is clicked.

Used a fake CordovaPlugin and `cordovaInterface.startActivityForResult` to do the job because Capacitor's startActivityForResult will require a PluginCall and a Plugin to properly work and that's harder to fake.

Closes #631

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
@@ -1,5 +1,7 @@
 package com.getcapacitor;
 
+import android.content.ActivityNotFoundException;
+import android.content.Intent;
 import android.net.Uri;
 import android.util.Log;
 import android.webkit.ConsoleMessage;
@@ -11,12 +13,15 @@ import android.webkit.ValueCallback;
 import android.webkit.WebChromeClient;
 import android.webkit.WebView;
 
+import org.apache.cordova.CordovaPlugin;
+
 /**
  * Custom WebChromeClient handler, required for showing dialogs, confirms, etc. in our
  * WebView instance.
  */
 public class BridgeWebChromeClient extends WebChromeClient {
   private Bridge bridge;
+  static final int FILE_CHOOSER = PluginRequestCodes.FILE_CHOOSER;
 
   public BridgeWebChromeClient(Bridge bridge) {
     this.bridge = bridge;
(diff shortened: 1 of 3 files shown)


android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
```diff
@@ -134,8 +139,33 @@ public void onGeolocationPermissionsShowPrompt(String origin, GeolocationPermiss
...
   @Override
-  public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, FileChooserParams fileChooserParams) {
-    return super.onShowFileChooser(webView, filePathCallback, fileChooserParams);
+  public boolean onShowFileChooser(WebView webView, final ValueCallback<Uri[]> filePathCallback, FileChooserParams fileChooserParams) {
+    Intent intent = fileChooserParams.createIntent();
+    if (fileChooserParams.getMode() == FileChooserParams.MODE_OPEN_MULTIPLE) {
+      intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true);
+    }
+    try {
+      bridge.cordovaInterface.startActivityForResult(new CordovaPlugin() {
```
mlynch: Why using the Cordova stuff for this? We should have some of our own utils

jcesarmobile: I explained it on the PR comment.

The capacitor's Bridge startActivityForPluginWithResult needs a PluginCall and a Plugin to work, but this will be called from the input type file button click, so there won't be a real Plugin or PluginCall invoking it.
I could have tried faking a Plugin and PluginCall, but doing it with the fake CordovaPlugin was easier.

mlynch: Sounds good

PR #717, opened 2018-08-09
Add an Android preference 'captureInput' to use special keyboard

### LINKED ISSUE #676: Some more Keyboard-issues
Hi.

1. Keyboard not respecting input type "number".

2. Keyboard not giving me word completion (suggested words).

Both issues on Android. Haven't tried it on iOS.

### LINKED ISSUE #701: Special characters bugg with keyboard
This pull almost fixed the previous issue where special characters didnt appear at all.
[https://github.com/ionic-team/capacitor/pull/673](url)

Now the bugg is that if the special charachter is the last one in a field and you leave focus it will be removed. If I type the Swedish city Malmö in a field and then leave focus I will only have Malm left.

jcesarmobile:
Added a new `captureInput` preference for Android that allows the keyboard key capture, otherwise it will use the default keyboard as the `BaseInputConnection` has a few issues (no autocompletion, no emoji, problems with special chars, not respecting type number, etc)

Closes #676
Closes #701

Also reuse the same `BaseInputConnection` instead of creating a new one every time.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
@@ -12,13 +12,22 @@ import android.webkit.WebView;
 import java.util.Date;
 
 public class CapacitorWebView extends WebView {
+  private BaseInputConnection capInputConnection;
+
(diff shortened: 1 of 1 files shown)


ArmorDarks:
Note that this might be related to newly appeared #757

PR #736, opened 2018-08-14
Document cordovaSwiftVersion and minVersion for iOS

### LINKED ISSUE #734: cordova-plugin-zeroconf - 'watch' not working [ios]
```javascript
if (window.cordova) {
    const zeroconf = window.cordova.plugins.zeroconf;
    zeroconf.watch('_http._tcp.', 'local.', result => {
      var action = result.action;
      var host = result.service;
      if (action == 'added') {
        console.log('service added', host);
      } else if (action == 'resolved') {
        console.log('service resolved', host);
      } else {
        console.log('host removed', host);
      }
    });
  }
```

Xcode reports as follows.
> To Native Cordova ->  ZeroConf watch ZeroConf336797565 ["options": [_http._tcp., local., any]]
Error: Plugin Optional("ZeroConf") does not respond to method call watch:.
Ensure plugin method exists and uses @objc in its declaration

When I put a debugger in I see the method exits. 
```
watch: {
   function(type, domain, success, failure)
   arguments: TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context.
   caller: TypeError: 'arguments', 'callee', and 'caller' cannot be accessed in this context.
   length: 4
   name: "watch"
   prototype: watch {}
   Function Prototype
}
```

jcesarmobile:
We never documented this two preferences

Closes #734

### DIFF
--- a/site/docs-md/basics/configuring-your-app.md
+++ b/site/docs-md/basics/configuring-your-app.md
@@ -52,6 +52,14 @@ The current ones you might configure are:
     // You can use a simpler keyboard enabling this preference
     // Be aware that this keyboard has some problems and limitations
     "captureInput": true
+  },
+  "ios": {
+    // Configure the Swift version to be used for Cordova plugins.
+    // Default is 4.0
+    "cordovaSwiftVersion": "3.2",
+    // Minimum iOS version supported by the project.
+    // Default is 10.0
+    "minVersion": "10.3"
   }
 }
 ```


Nudelsieb:
That's great work, thanks. However, I believe there should also be some details why this could be necessary. The relevance is quite imortant, IMHO. For example, the documentation of using Cordova plugins should be more detailed and contain this information or link to it.

PR #737, opened 2018-08-15
Make native-bridge.js ES5

jcesarmobile:


mlynch:
I think this is good for now. Down the road we should build this in typescript and then transpile down so we don't have to lose some of the niceties (like multi-line strings)

PR #740, opened 2018-08-16
[WIP] fix(layout): fix offset bottom on iOS devices (closes #337)

### LINKED ISSUE #337: Offset at the bottom after app launch on iOS devices
There is a small offset at the bottom of the webview directly after the splash-screen disappeared. On reals devices the offset quickly vanishes (see gif), in the simulator it stays. I've tried with the example app and with one of our apps (both are Ionic projects). It happens on the iPhoneX (where the offset is biggest) and other devices too (see screenshot).

```
@capacitor/cli: 1.0.0-alpha.13
@capacitor/core: 1.0.0-alpha.13
```

offset-iphonex

offset-sim-iphone8_x

paulelie:
Ref: https://github.com/ionic-team/capacitor/issues/337

Fixed with adding `modalPresentationStyle` to `overFullScreen` property on storyboard

### DIFF
--- a/e2e/ios/App/App/Base.lproj/Main.storyboard
+++ b/e2e/ios/App/App/Base.lproj/Main.storyboard
@@ -1,17 +1,17 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="13771" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="BYZ-38-t0r">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="14113" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="BYZ-38-t0r">
     <device id="retina4_7" orientation="portrait">
         <adaptation id="fullscreen"/>
     </device>
     <dependencies>
         <deployment identifier="iOS"/>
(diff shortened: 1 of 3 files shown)


mlynch:
@paulelie awesome, thanks so much for the PR!

A-Industry:
Also running into this issue now
screen shot 2018-08-28 at 15 17 44

AVatch:
Just to point out this white spacing issue disappears if orientation is changed or the status bar visibility is toggled.

That being said perhaps there is some constraint being set to the bottom safe area region on iPhones which does not initialize properly?

paulelie:
Still working on it (when i have time); BTW could you try this snippet on your app.component:

1. After your imports:
```
declare const document:any;
```
2. In constructor :
```
constructor(platform: Platform){
platform.ready().then(()=>{
 document.querySelector('meta[name=viewport]').setAttribute('content', 'width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=1.0, user-scalable=no');
})
}
```

A-Industry:
Yes that helps :) But now the application header is not shown under the the statusbar. So it is has a grey/white background
screen shot 2018-08-28 at 15 48 03

A-Industry:
I found an easy fix: I removed the app.component.ts fix and just added
`html[mode="ios"] {
  height: 100vh;
}`

to global.scss. Now it works as expected.

Devleaf-IT:
I'm still having this issue, even with the above mentioned fixes. Does anyone know the status of this problem? Thank you!

paulelie:
> I'm still having this issue, even with the above mentioned fixes. Does anyone know the status of this problem? Thank you!

Hello,
Sorry for late answer.. It partially fix it, because if you use an native plugin, it will break your layout again... I've plan to check out/investigate more deeply on my extra-time.. will post on this feed if i found something
BTW thanks @A-Industry for your trick ;)

jcesarmobile:
should be fixed by https://github.com/ionic-team/capacitor/pull/856, so going to close

PR #753, opened 2018-08-20
Android: make splash resource name configurable

moberwasserlechner:
Creating custom icons and splashscreens following Josh Morony's great post works like charm. https://www.joshmorony.com/adding-icons-splash-screens-launch-images-to-capacitor-projects/

To generate my splashscreens I used one of the tools listed https://apetools.webprofusion.com/app/#/tools/imagegorilla. This tool generates the image files needed but names it `screen` instead of `splash`.

Instead of renaming all files it would be nice to let users configure the resource name for the splashscreen image.
```
  "android": {
    "splashResourceName": "screen",
  }
```
I searched but didn't find anything according on ios. If there is I'm happy to add a config param there as well.

BR, Michael

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Splash.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Splash.java
@@ -12,10 +12,6 @@ import android.view.WindowManager;
 import android.view.animation.LinearInterpolator;
 import android.widget.ImageView;
 
-import com.getcapacitor.plugin.SplashScreen;
-
-import org.json.JSONObject;
-
 /**
  * A Splash Screen service for showing and hiding a splash screen in the app.
  */
@@ -25,10 +21,9 @@ public class Splash {
     void error();
   }
 
-  private static final String SPLASH_DRAWABLE = "splash";
+  public static final String CONFIG_KEY_PREFIX = "plugins.SplashScreen.";
 
-  public static final int LAUNCH_SHOW_DURATION = 3000;
-
+  public static final int DEFAULT_LAUNCH_SHOW_DURATION = 3000;
   public static final int DEFAULT_FADE_IN_DURATION = 200;
   public static final int DEFAULT_FADE_OUT_DURATION = 200;
   public static final int DEFAULT_SHOW_DURATION = 3000;
@@ -41,8 +36,10 @@ public class Splash {
 
   private static void buildViews(Context c) {
 
-    int splashId = c.getResources().getIdentifier("splash", "drawable", c.getPackageName());
-    Drawable splash = c.getResources().getDrawable(splashId);
+    String splashResourceName = Config.getString(CONFIG_KEY_PREFIX + "androidSplashResourceName", "splash");
+
+    int splashId = c.getResources().getIdentifier(splashResourceName, "drawable", c.getPackageName());
+    Drawable splash = c.getResources().getDrawable(splashId, c.getTheme());
 
     splashImage = new ImageView(c);
 
@@ -59,25 +56,7 @@ public class Splash {
    * @param a
    */
   public static void showOnLaunch(final BridgeActivity a) {
-    Bridge b = a.getBridge();
-    if (b == null) {
-      return;
-    }
-
-    PluginHandle splashPluginHandle = b.getPlugin("SplashScreen");
-    if (splashPluginHandle == null) {
-      return;
-    }
-
-    SplashScreen splashPlugin = (SplashScreen) splashPluginHandle.getInstance();
-
-    Integer durationValue = (Integer) splashPlugin.getConfigValue("launchShowDuration");
-
-    int duration = LAUNCH_SHOW_DURATION;
-    if (durationValue != null) {
-      duration = durationValue.intValue();
-    }
-
+    Integer duration = Config.getInt(CONFIG_KEY_PREFIX + "launchShowDuration", DEFAULT_LAUNCH_SHOW_DURATION);
     show(a, duration, 0, DEFAULT_FADE_OUT_DURATION, true, null, true);
   }
 
@@ -86,7 +65,7 @@ public class Splash {
    * @param a
    */
   public static void show(final Activity a) {
-    show(a, LAUNCH_SHOW_DURATION, DEFAULT_FADE_IN_DURATION, DEFAULT_FADE_OUT_DURATION, DEFAULT_AUTO_HIDE, null);
+    show(a, DEFAULT_LAUNCH_SHOW_DURATION, DEFAULT_FADE_IN_DURATION, DEFAULT_FADE_OUT_DURATION, DEFAULT_AUTO_HIDE, null);
   }
 
   /**
--- a/android/capacitor/src/main/java/com/getcapacitor/Config.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Config.java
@@ -76,7 +76,7 @@ public class Config {
     // Search until the second to last part of the key
     for (int i = 0; i < parts.length-1; i++) {
       String k = parts[i];
-      o = this.config.getJSONObject(k);
+      o = o.getJSONObject(k);
     }
     return o;
   }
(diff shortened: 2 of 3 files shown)


android/capacitor/src/main/java/com/getcapacitor/Splash.java
```diff
@@ -41,8 +36,11 @@
 
   private static void buildViews(Context c) {
 
-    int splashId = c.getResources().getIdentifier("splash", "drawable", c.getPackageName());
-    Drawable splash = c.getResources().getDrawable(splashId);
+    // there are splashscreen creators naming the file sth other than "splash"
+    String splashResourceName = Config.getString("android.splashResourceName", "splash");
+
+    int splashId = c.getResources().getIdentifier(splashResourceName, "drawable", c.getPackageName());
+    Drawable splash = c.getResources().getDrawable(splashId, c.getTheme());
```
jcesarmobile: is the `c.getTheme()` needed? the original code didn't have it.

moberwasserlechner: The previous used method `getDrawable(id)` is deprecated.

jcesarmobile: Ah, cool. Looks good to me then

moberwasserlechner:
Btw. I found a hardcoded splash screen resource name in the iOS part of Capacitor as well. 

see https://github.com/ionic-team/capacitor/blob/76830d9d24f4f9ce0b8478fc37b3c5fe74a38362/ios/Capacitor/Capacitor/Plugins/SplashScreen.swift#L53

If we can get this configurable as well the developers can always change the storyboard resource name in XCode. 

This is one of the best things about Capacitor. A great framework but the ability to use the platform specfic tools if they want to.

jcesarmobile:
Yeah, but that's what I explained yesterday, first the launch storyboard is launched and it will load the image named "Splash", then a Capacitor view will appear with the image called "Splash".
Changing that you can make the second image configurable, but not the first one from the Launch storyboard. The image on the Launch Storyboard is set on Interface Builder. Sadly Launch Storyboards don't accept any code, so it can't be configured on runtime.

moberwasserlechner:
Ok I missed that. I thought if I change the Launch storyboard in my XCode capacitor project it would do the trick. But good to know. thx

moberwasserlechner:
I think we have to differentiate between the **Capacitor framework** and the **template files**, which are generated once a developer **adds a platform** to his or her project.

I try to explain what a mean step by step by a example project "cap-splashscreen-poc" and enable Capacitor on it. I follow the documentation and do:

```
mkdir cap-splashscreen-poc
cd cap-splashscreen-poc
npm init
npm install --save @capacitor/core @capacitor/cli
npx cap init
*   Your Capacitor project is ready to go!  *

Add platforms using "npx cap add":

  npx cap add android
  npx cap add ios
  npx cap add electron
```

After that I run `npx cap add android`. Now the CLI creates the dir `cap-splashscreen-poc/android` with files from `CAP_REPO/android-template`. 

grafik

These files are now part of the "cap-splashscreen-poc" project and most of them will not be touched by the CLI whenever I run `npx cap update`. The "cap-splashscreen-poc" developer is now responsible for them.

I now follow the tutorial to replace the default icons and splashscreen generated by the CLI with my own. I have to change the splashscreen images and 2 xml configs. 

grafik

The 2 xml files are:

*android/app/src/main/res/drawable/launch_splash.xml*
grafik

*android/app/src/main/res/values/styles.xml*
grafik

The only thing I need to do now is to configure the Capacitor Android runtime in `capacitor.config.json`
```
 "android": {
    "splashResourceName": "screen",
  }
```
then run `npx cap sync` and build my app in Android Studio to test it on a device or in emulator. 

**Conclusion:**
I have this PR running in a real project (the screenshots are from there). I tested it on Android 5-8 and it works great.

In my opinion it's just a **documentation issue** and not a technical one. Right now the Capacitor platform templates include good defaults and a developer only needs to know what to change if they want to use differently named splashscreens files.

I'll add the documentation if you could point me to the right markdown files.

jcesarmobile:
Probably best place would be Splash docs https://github.com/ionic-team/capacitor/blob/master/site/docs-md/apis/splash-screen/index.md

But when I looked at them I remembered we have some configurations in place for the plugin like:

```
{
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 5000
    }
  }
}
```

So probably makes sense to move the `splashResourceName` there instead of having it inside android object.
But that makes me wonder how to deal with it on iOS, should we have different ios and android objects inside SplashScreen object and each one with a `splashResourceName`? or have different keys like `AndroidSplashResourceName` and `isSplashResourceName`? Or use same name for both?

moberwasserlechner:
I think the most flexible way is to use 2 properties because

* Generators like imagegorilla will name it differently. For Android it's "screen". For iOS it's "LaunchImage".
* There might be different naming conventions for iOS and Android. In Android / Java most resource files start with a lowercase letter. In iOS I don't exaclty know but it seems that the first letter is upper-case. I observed this from  Capacitor's defaults. Android: "splash" vs iOS: "Splash". Developers should feel "at home" on their platform.

I have no preference where to put the property in `capacitor.config.json`:

```
{
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 5000
      "androidSplashResourceName": "screen"
      "iosSplashResourceName": "LaunchImage"
    }
  }
}
```
or
```
{
  "android": {
    "splashResourceName": "screen",
  }
  "ios": {
    "splashResourceName": "LaunchImage",
  }
}
```
or
```
{
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 5000
      "android": {
        "splashResourceName": "screen",
      }
      "ios": {
        "splashResourceName": "LaunchImage",
      }
    }
  }
}
```
Please choose one. :)

moberwasserlechner:
I tested the following config
```
{
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 5000,
      "android": {
        "splashResourceName": "screen"
      },
      "ios": {
        "splashResourceName": "LaunchImage"
      }
    }
  }
}
```
and found a bug in `com.getcapacitor.Config`, which didnot read values if there are on nested level 2+.

I cleanup a few things as well and will add some unit tests tomorrow.

moberwasserlechner:
I hope I didn't messed up the branch by doing `git pull --rebase` form the upstream.

jcesarmobile:
Sorry, we talked about it and decided to go with the first one to avoid nesting
```
{
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 5000
      "androidSplashResourceName": "screen"
      "iosSplashResourceName": "LaunchImage"
    }
  }
}
```
BTW, iOS is less likely to be changed as the name is not of the image, but for the asset. The asset has 3 images and you can drag any image there, doesn’t matter the image names. Unless there is a service that generates the whole asset with the 3 images (not sure if you can just drag an asset file to Xcode).

About the problem with the nested config values I would prefer a separate PR for it.

And yeah, something went wrong with your rebase, all those commits shouldn’t be there, can you try to squash them?

moberwasserlechner:
Np.  `androidSplashResourceName` it is.

I need to fix the Config in this PR. I cannot access the `androidSplashResourceName` in `com.getcapacitor.Splash#buildViews` because I'm not able to access `com.getcapacitor.Plugin#getConfigValue` from there. Moreover using the the `Config` for `launchShowDuration` as well makes the code simpler.

I will change only the parts really required to have a small PR.

> BTW, iOS is less likely to be changed a ...

This is fine for me. I could not have tested it anyway because I need to get Android running for my app first and have not setup my iOS dev environment properly.

moberwasserlechner:
I've reverted everything to get one single clean commit and added a basic documentation for changing splashscreen images with some instruction what to change in the user's Android app.

Let's do any further improvements to the docs in a separate PR.

jcesarmobile:
Didn't need to go with a single commit, I was going to squash on merge anyway, but it's appreciated.

Merged, thanks!

PR #779, opened 2018-08-29
Do Not Merge: Attempt to add moveFile and copyFile for iOS and Android

sean-perkins:
I'm not as familiar with the Swift (or Obj. C)/Android side of things, need an extra set of eyes here. 

The `Filesystem` API is missing the utility functions for both moving and copying files.

iOS spec: https://developer.apple.com/documentation/foundation/filemanager/1413529-moveitem
Android spec: (File API): https://developer.android.com/reference/java/io/File#renameTo(java.io.File)


jcesarmobile:
Closing as copy (#1758) and rename (#1624) were merged and add the same functionality.

PR #800, opened 2018-09-12
Catch ActivityNotFoundException to prevent app crash.

cthos:
If you link out to an activity (in my case this was `slack://` in the webview, but the user does not have an Activity installed which can handle slack - the app will crash entirely.

This PR will prevent the crash, but I do not know what I should be doing to inform the end user that this has occurred, probably send some sort of Capicator event to the browser?


ArmorDarks:
Sounds like a good addition, because the default behavior is unexpectedly destructive.

PR #803, opened 2018-09-16
[Android] Remove correct call

michaeltintiuc:
I won't pretend to be a Java guru, but while trying to fiddle with the backButton events, I noticed that you could not remove the created event listener, after hours of debugging I was able to fix it. Here's the JS example:

```js
...
  created() {
    this.backEvent = App.addListener('backButton', this.handleHardwareBackButton)
  },
  destroyed() {
    this.backEvent.remove()
  },
```

Given the above code, the removeListener method was trying to remove a non-existent `call` object even though the `savedCall` condition was passing.

In fact, looks like this is how it's done for the [iOS counterpart](https://github.com/ionic-team/capacitor/blob/aaa301a5874b37bdf18d305f1e9e26cc5092dd7c/ios/Capacitor/Capacitor/CAPPlugin.m#L101)


mlynch:
Ah, good catch, surprised that was overlooked. Either way, appreciate the PR!

PR #804, opened 2018-09-17
Make window.open to launch safari

### LINKED ISSUE #798: iOs - window.open target=_blank in a iframe is silently ignored
## Version
1.0.0-beta.7

## Platfrom
iOS built application

## Question

It seems that the window.open target=_blank events are being ignored with the current WKWebView implementation. The following hook gets called in the WKUIDelegate implementations when the window.open is called:
```swift
    public func webView(_ webView: WKWebView, createWebViewWith configuration: WKWebViewConfiguration, for navigationAction: WKNavigationAction, windowFeatures: WKWindowFeatures) -> WKWebView? {
        ...
    }
```
The default implementation seems to `return nil` as [stated here](https://developer.apple.com/documentation/webkit/webuidelegate/1387785-webview) (_By default, this method returns nil_).
Is there a way to implement this WKWebView hook without modifying the `node_modules/@capacitor/ios/ios/Capacitor/Capacitor/CAPBridgeViewController.swift` file?

Following code would be implemented:
```swift
    public func webView(_ webView: WKWebView, createWebViewWith configuration: WKWebViewConfiguration, for navigationAction: WKNavigationAction, windowFeatures: WKWindowFeatures) -> WKWebView? {
        if (navigationAction.request.url != nil) {
            UIApplication.shared.open(navigationAction.request.url!, options: [:], completionHandler: nil)
        }
        return nil
    }
```

jcesarmobile:
This method is needed to make window.open work.

The code will make window.open to open Safari (or other apps based on the url scheme)

Closes #798

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -365,6 +365,13 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScr
     self.present(alertController, animated: true, completion: nil)
   }
 
+  public func webView(_ webView: WKWebView, createWebViewWith configuration: WKWebViewConfiguration, for navigationAction: WKNavigationAction, windowFeatures: WKWindowFeatures) -> WKWebView? {
+    if (navigationAction.request.url != nil) {
+      UIApplication.shared.open(navigationAction.request.url!, options: [:], completionHandler: nil)
+    }
+    return nil
+  }
+
   public func getWebView() -> WKWebView {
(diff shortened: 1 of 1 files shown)


goodwillsandy:
Facing this issue, find problem description below:
"I am opening my app(using capacitor), then go on login page, there on clicking on login option, it opens a iframe & in that iframe, on clicking on button it should be opening a page in new webview, but url (`navigationAction.request.url`) is coming as nil"

However when i try the same in project created using Cordova then , `<preference name="AllowNewWindows" value="true" /> ` in config.xml make it works.

PR #790, opened 2018-09-04
feat(splash): splash can be animated on android

ptitjes:
Fixes #789.

PR #806, opened 2018-09-18
[Android] Add missing import statement to splash

vially:
This should fix a compile error introduced in #790.


jcesarmobile:
Thanks!

Issue #540, opened 2018-05-14
Input element with capture on android


sandstrom:
It would be nice if `<input type="file" accept="image/*" capture>` would work on Android.

- Currently works just nice on builds for iOS running in the capacitor webview.
- Also works just fine running on Chrome in Android.
- But doesn't work when running a Capacitor build for Android (in the webview).

For easy testing, this page contains a few html snippets with file input + accept + capture:
http://anssiko.github.io/html-media-capture/

mlynch:
Should be fixed in the next release

sandstrom:
@mlynch Awesome! 🥇

PR #809, opened 2018-09-19
Add support for media capture using input fields on Android

vially:
By default, the Android webview does **not** open the camera for inputs like:
```html
<input type="file" accept="image/*" capture>
```
This pull-request fixes that. The motivation for this functionality is that it might replace the need for using the *native* [Camera](https://capacitor.ionicframework.com/docs/apis/camera) plugin with purely *browser* APIs. Obviously, the native camera plugin will still be needed for more advanced use-cases, but if all you need is to take a picture with the camera you will now be able to do so without any native plugin (and the same code will also work in browsers).

This functionality was previously requested in #540 and it was [supposedly](https://github.com/ionic-team/capacitor/issues/540#issuecomment-388920583) already [fixed](https://github.com/ionic-team/capacitor/commit/733d7971f8cb64a9e4c9f85196c9ea59bbf24d2b). But the [current code](https://github.com/ionic-team/capacitor/blob/46cc2083dd106f217fe2ab154a2b64d15f51d452/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java#L142-L170) doesn't seem to contain the logic needed for handling it.

# Limitations

This initial implementation only covers some very basic cases and it has some limitations that could be improved in the future:

- an [Action Chooser](https://developer.android.com/reference/android/content/Intent.html#ACTION_CHOOSER) should be shown when both `image/*` and `video/*` are present in the `accept` array. The current implementations defaults to showing the camcorder in this case.
- video capture is currently disabled on Android M and older because it looks like its [Intent](https://developer.android.com/reference/android/provider/MediaStore#ACTION_VIDEO_CAPTURE) is returning a `file://` URI on those devices (instead of the expected `content://` URI)
- if the app contains `<uses-permission android:name="android.permission.CAMERA" />` in its merged manifest file and the permission is not granted, we don't show the camera/camcorder and fallback to the default file chooser (more info [here](https://developer.android.com/reference/android/provider/MediaStore#ACTION_IMAGE_CAPTURE) and [here](https://stackoverflow.com/q/32789027))
 
***Note:*** This is already working as expected on iOS because it's handled automatically by `WKWebView`.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
@@ -1,9 +1,11 @@
 package com.getcapacitor;
 
+import android.Manifest;
 import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.Intent;
 import android.net.Uri;
+import android.provider.MediaStore;
 import android.util.Log;
 import android.webkit.ConsoleMessage;
 import android.webkit.GeolocationPermissions;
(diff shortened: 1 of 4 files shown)


jcesarmobile:
Yeah, you are right, I've downgraded to 1.0.0-alpha.38 as that's the version where that change was included and didn't work, and when I added the input file support in a more recent version I didn't handle camera option.

I'll review your PR, thanks!

PR #810, opened 2018-09-20
Make input file capture ask for permission if not granted

jcesarmobile:
At the moment, an input file with capture (`<input type="file" accept="image/*" capture>`) won't show the camera if the permission is not granted beforehand, which is most likely. 
This PR will make the input file request the permission if it wasn't granted and continue the flow if granted. 

@vially as you sent the previous PR that added capture support, can you give it a try?

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -680,7 +680,7 @@ public class Bridge {
   public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
     PluginHandle plugin = getPluginWithRequestCode(requestCode);
 
-    if (plugin == null) {
+    if (plugin == null || requestCode == 0) {
       Log.d(LOG_TAG, "Unable to find a Capacitor plugin to handle requestCode, try with Cordova plugins " + requestCode);
       try {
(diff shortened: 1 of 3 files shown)


vially:
Great work! I'll give it a try tomorrow and report back if I find any issues.

There is something that I would like to mention though. While I was researching some things for #809 I found out that [you don't need the camera permission](https://developer.android.com/training/permissions/usage-notes#perms_vs_intents) if you're going to use the camera using it's intent.

However, this only works if `<uses-permission android:name="android.permission.CAMERA" />` hasn't been added to the manifest file (by the user or a plugin).

Maybe this should be mentioned somewhere in the documentation because I think some very basic uses of the camera (through the input field or even through the camera plugin) might get away without needing this permission at all.

vially:
I just tested it and it works great (tested on actual device running Android M).

jcesarmobile:
@vially thanks for testing!

I'm aware of not needing Camera permission for Camera plugin or input file, but as you say, if the user (or a plugin) adds it in the future it will cause those to crash if not handled properly, so better handle it and keep the permission on the base project to help us catch this kind of problems while in development.
Also, the camera permission is needed if you want to use the camera from `navigator.mediaDevices.getUserMedia`

But yeah, we should document that it's not needed for Camera plugin nor input file and they can safely remove it if not using `getUserMedia`

PR #812, opened 2018-09-21
Request native permissions needed for getUserMedia

### LINKED ISSUE #802: WebRTC audio, "could not start audio source"
My app uses WebRTC audio fine in the browser, but when I open the app on my Android phone, I get the console error "DOMException: Could not start audio source".

Here's how I'm trying to read audio:
`const stream = await navigator.mediaDevices.getUserMedia({audio: true, video: false});`

I have given my app the Android microphone permission (`<uses-permission android:name="android.permission.RECORD_AUDIO" />`).

jcesarmobile:
On Android 6+ you have to manually ask for the permissions, so this will ask for them when using getUserMedia

Closes #802

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
@@ -22,6 +22,7 @@ import com.getcapacitor.plugin.camera.CameraUtils;
 import org.apache.cordova.CordovaPlugin;
 import org.json.JSONException;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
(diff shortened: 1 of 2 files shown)


jdnichollsc:
@mlynch @jcesarmobile  can we use `getUserMedia` from iOS?

jcesarmobile:
No, Apple doesn’t provide it for WKWebView

jdnichollsc:
thanks, ok so native plugins is the only way, thanks for confirming my suspicions! 👍

petermercy:
@jcesarmobile To be clear: There is no way to access getUserMedia in an IOS Ionic app. Is that correct?

I see the Ionic/Capacitor Camera plugin https://capacitor.ionicframework.com/docs/apis/camera/, but this appears to only support photos. Is there any way to access a video stream?
Thanks!

jcesarmobile:
No, it's not possible to use getUserMedia.

You can record video with an html input with type file and capture attribute, but shows a native UI, doesn't provide the stream.

petermercy:
And there's no plugin that you're aware of? (I've been searching everywhere for one, including in Cordova plugins that might be compatible and haven't found anything).
Thanks for your quick reply! I really appreciate it!

PR #820, opened 2018-09-26
Config param to disable web content debugging

moberwasserlechner:
I hope to get better performance, if this debugging could be disabled by a config parameter.

According to the docs `https://developer.android.com/reference/android/webkit/WebView#setWebContentsDebuggingEnabled(boolean)` it defaults to false.

BR


jcesarmobile:
Do you really get better performance disabling debugging?

Anyway, I think this is a good feature to have, so merged it.

moberwasserlechner:
I don't really know yet but I thought it might not hurt to disable it for production releases.

jcesarmobile:
That's bad, it should be disabled by default on release. I think it used to, but now on the docs it says it doesn't. Will change that, but keep the preference in case people want to disable it for debug builds or enable it for release.

PR #821, opened 2018-09-26
Disable SplashScreen plugin using config param

moberwasserlechner:
My pure Angular app uses a splashscreen `div` in `index.html`, which is removed after the initial routing is done in my AppComponent.

This is not better than using the SplashScreen plugin, its only different. I only would need a way to disable the SplashScreen plugin, because I don't need it.

```
// remove the splashscreen
    const splash = this.document.body.querySelector('#splash');
    this.router.events.subscribe((event) => {
        if (event instanceof NavigationEnd) {
          if (splash) {
            splash.remove();
          }
        }
      }
    );
```

A new config parameter `plugins.SplashScreen.enabled`, which defaults to `true` would allow me to disable the plugin.

BR

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
@@ -76,7 +76,9 @@ public class BridgeActivity extends AppCompatActivity {
     cordovaInterface.onCordovaInit(pluginManager);
     bridge = new Bridge(this, webView, initialPlugins, cordovaInterface, pluginManager);
 
-    Splash.showOnLaunch(this);
+    if (Config.getBoolean("plugins.SplashScreen.enabled", true)) {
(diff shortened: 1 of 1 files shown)


jcesarmobile:
To accept this, it should also be implemented for iOS and documented.

BTW, on iOS it's not possible to disable the launch image (or launch storyboard) that appears right before the splash screen. You should add a white or black image or something that matches your html splash

moberwasserlechner:
@janpio @jcesarmobile thx for your feedback and time.

I tested it on a few more devices. I thought that the gap between the background image from
```
    <style name="AppTheme.NoActionBarLaunch" parent="AppTheme.NoActionBar">
        <item name="android:background">@color/background_color_splash</item>
    </style>
```
and the my splashscreen from the webview is shorter. 

But I get a white screen before the splashscreen div from the webview kicks in.

So I withdraw/delete this PR.

PR #824, opened 2018-09-26
Make camera full screen and allow presentationStyle

### LINKED ISSUE #819: iOS Camera is not fullscreen
In the camera plugin for iOS it does not account for iPads and always displays the camera as a popover. It needs to be fullscreen for iOS standards. As a temporary fix to the problem I have commented out these two lines in the getPhoto function

`//imagePicker!.popoverPresentationController?.delegate = self
  //self.setCenteredPopover(self.imagePicker!)`

And then in the showCamera function I have added this line after the source type is set to camera.

`imagePicker.modalPresentationStyle = .fullScreen`

I am unsure if this fix break anything for other platforms or use cases as I am developing solely for iPad.

jcesarmobile:
Make camera full screen as recommended, but allow presentationStyle option to show as popover if desired.
This is only for Camera source, for Photos it should still use popover despite the setting.

Closes #819

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Camera.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Camera.swift
@@ -38,8 +38,6 @@ public class CAPCameraPlugin : CAPPlugin, UIImagePickerControllerDelegate, UINav
     self.call = call
     self.settings = getSettings(call)
 
-    let sourceType = getSourceType(settings.source)
-    
     // Make sure they have all the necessary info.plist settings
     if let missingUsageDescription = checkUsageDescriptions() {
       bridge.modulePrint(self, missingUsageDescription)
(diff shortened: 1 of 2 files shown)


ios/Capacitor/Capacitor/Plugins/Camera.swift
```diff
@@ -123,41 +111,40 @@ public class CAPCameraPlugin : CAPPlugin, UIImagePickerControllerDelegate, UINav
...
     
+    self.configurePicker()
+    
     self.imagePicker!.sourceType = .photoLibrary
     self.imagePicker!.allowsEditing = settings.allowEditing
     
     self.bridge.viewController.present(self.imagePicker!, animated: true, completion: nil)
   }
-  
-  func getSourceType(_ source: CameraSource) -> UIImagePickerControllerSourceType {
```
mlynch: How are we doing source type now?

jcesarmobile: showCamera sets it to .camera 
https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/Camera.swift#L132
showPhotos to .photoLibrary
https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/Camera.swift#L144

So that function was not really doing anything

mlynch: Awesome

Issue #826, opened 2018-09-27
Android: Icons are loaded on every router action


moberwasserlechner:
**Issue**

In my `index.html` I have various icons

```
  <link rel="shortcut icon" href="favicon.ico">
  <link rel="icon" type="image/png" sizes="32x32" href="assets/img/icon/favicon-32x32.png">
  <link rel="icon" type="image/png" sizes="16x16" href="assets/img/icon/favicon-16x16.png">
```

Those icons are loaded once on startup but also on every page change handled by angular router. So I see the following multiple times.

```
D/Capacitor: Handling local request: https://localhost:9911/favicon.ico
D/Capacitor: Handling local request: https://localhost:9911/assets/img/icon/favicon-16x16.png
D/Capacitor: Handling local request: https://localhost:9911/assets/img/icon/favicon-32x32.png
```

I don't think loading those icons is needed at all.  Although its a local server, not loading not required resources should increase the performance. It's a small improvement but many small steps will have a impact.

**Solution / Workaround**
I don't know how to differ a icon request from a normal asset. The file name is not a good approach. The `link` attribute `rel="icon"` would be but I don't have that. afaik.

I will most propably remove the icons from my `index.html` for the android (future ios) build using webpack or use a `index.capacitor.html` without that parts.

**Version info:**
Angular: 6.1.8
Capacitor: 1.0.0-beta.7 it also occured
Android: 5-8

moberwasserlechner:
So I created new configurations in `angular.json` for Android and replacing the default `index.html` with a `index.capacitor.html`, which does only include the absolutely needed stuff. No icons, no manifest, no noscript.

This works great and I got rid of the png icons but the `favicon.ico` remains and is still loaded/handled on every route change.

```
D/Capacitor: Handling local request: https://localhost:9911/favicon.ico
D/Capacitor: Handling local request: https://localhost:9911/favicon.ico
D/Capacitor: Handling local request: https://localhost:9911/favicon.ico
```
The WebView is automatically looking for `/favicon.ico`, although I have not referenced it in a link element.

This could be prevented by Capacitor because the name will not change. It would be nice but so far only its only a logging issue. I'll give it a try and do some testing and eventually create a PR.

moberwasserlechner:
If I do as suggested in https://stackoverflow.com/a/41223255

```
    if ("/favicon.ico".equalsIgnoreCase(path)) {
      try {
        return new WebResourceResponse("image/png", null, null);
      } catch (Exception e) {
[...]

PR #830, opened 2018-09-27
Android: Handle WebView '/favicon.ico' requests

moberwasserlechner:
Please see #826 for details.

BR, Michael

If there is any reason why the favicon.ico should be loaded please let me know.


mlynch:
Seems reasonable, thanks!

Issue #841, opened 2018-10-01
Spawning a UI Window from Plugin Method Causes Unpredictable results


cthos:
If you spawn a ViewController from a plugin on iOS, the following warning appears in the console:
`Main Thread Checker: UI API called on a background thread`

And

```
[Animation] +[UIView setAnimationsEnabled:] being called from a background thread. Performing any operation from a background thread on UIView or a subclass is not supported and may result in unexpected and insidious behavior.
```

Which is certainly true. I've experienced weird rendering bugs, where certain things in a scene won't render in the proper order - unpredictably each launch. Sometimes it works fine, sometimes labels will be moved under other elements.

I've tested this out by removing the `dispatchQueue.async` call here: https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/CAPBridge.swift#L385

With the Async wrapper removed, the UIView runs great.

It'd be nice if there was a way to tell Capacitor that I _specifically_ want a method to run in the foreground because I'll be launching UI elements to skirt around this whole issue.

mlynch:
You need to execute any UI operations on the main thread, for example: https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/Modals.swift#L20

cthos:
Nice, thanks. 

Should that be mentioned in the `Presenting Native Screens` section of the Capacitor docs, or is that an integral iOS thing that I should have known?

jerrychong25:
> You need to execute any UI operations on the main thread, for example: https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/Modals.swift#L20
> 
> ```swift
>     DispatchQueue.main.async {
>       self.bridge.viewController.present(alert, animated: true, completion: nil)
>     }
> ```

Thanks! This solved my problem.

quanganh206:
> You need to execute any UI operations on the main thread, for example: https://github.com/ionic-team/capacitor/blob/master/ios/Capacitor/Capacitor/Plugins/Modals.swift#L20
> 
> ```swift
>     DispatchQueue.main.async {
>       self.bridge.viewController.present(alert, animated: true, completion: nil)
>     }
> ```

Thanks, everything going good. But is there anyway I can show my custom view with fullscreen size like capacitor's one?

[...]

PR #842, opened 2018-10-02
Add documentation on running screens on the main thread from Capacito…

cthos:
…r Plugins.

Just adding some documentation as discussed over in #841.

Running build locally produced way more changed files than just these two (`.html` files), but since these are the relevant ones I just included them.


mlynch:
Awesome, thanks for the PR!

PR #855, opened 2018-10-04
Add Motion implementation

jcesarmobile:
Maybe we should use original names instead of accel and orientation?

### DIFF
--- a/core/native-bridge.js
+++ b/core/native-bridge.js
@@ -14,6 +14,29 @@
   addLegacyHandlers(win);
 
   capacitor.Plugins = capacitor.Plugins || {};
+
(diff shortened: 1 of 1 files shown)


mlynch:
Why would this be in the native bridge? In theory plugins like `Motion` should be able to register themselves. We shouldn't be hard coding those into the bridge, which should be plugin-agnostic for the most part

jcesarmobile:
But there is no Motion plugin (in the native side), so where should this be instead?

mlynch:
In theory it should work like any web fallback for a plugin. So, take our `Geolocation` plugin. It has native implementations, but if you're running on Electron, for example, the web one is the only one available so that on runs instead. Motion should be the same. If you import it and use it, the web version will get called because there is no native one.

PR #856, opened 2018-10-04
Fix iOS webview bottom space (#337)

sbannigan:
Possible fix for #337. Until now I have been adding [cordova-plugin-disable-ios11-statusbar](https://github.com/jcesarmobile/cordova-plugin-disable-ios11-statusbar) to my capacitor projects to fix this issue, so this basically applies the same thing to capacitor's webview. Not sure if its the proper way to solve the issue, but it fixes the issue on the native side instead of requiring CSS changes.


mlynch:
Awesome will give it a try!

mlynch:
Weird, wasn't able to get your code to compile, had to change it to 

```swift
webView?.scrollView.contentInsetAdjustmentBehavior = .never
```

PR #860, opened 2018-10-05
Fix passing ArrayBuf arguments to Cordova plugins

mxdmedia:
Added call to massageArgsJsToNative as was in original Cordova implementation, which fixes broken passing of ArrayBuf arguments to legacy Cordova plugins.

### DIFF
--- a/core/cordova.js
+++ b/core/cordova.js
@@ -931,6 +931,8 @@
                     { success: successCallback, fail: failCallback };
             }
 
+            // Properly encode ArrayBuf action arguments
+            actionArgs = massageArgsJsToNative(actionArgs);
             actionArgs = JSON.parse(JSON.stringify(actionArgs));
             var command = {
                 type: 'cordova',


mlynch:
Awesome, thanks for the PR. Warrants some further testing!

jcesarmobile:
Did you try it in iOS, Android or both?

I've created a sample plugin that pass and return an ArrayBuffer and with your change it works fine on iOS, but on Android it still doesn't work.

I think the problem is now on Capacitor's native code, so just commenting to check if you tested on Android and got it working there.

jcesarmobile:
You probably only tested on iOS as the `massageArgsJsToNative` function is iOS only.
I've found the problem on Android, Cordova does the ArrayBuffer processing in a different way on Android platform.
Going to merge your PR and I'll send a separate PR for Android.

Thanks!

mxdmedia:
Correct, my PR was only tested on iOS.

jcesarmobile:
https://github.com/ionic-team/capacitor/pull/868

PR #867, opened 2018-10-08
fix(cli): resolve node modules

### LINKED ISSUE #792: Monorepo Support
```
> capacitor add android
✔ Installing android dependencies in 1.64s
✔ Adding native android project in: /Users/dan/apps/angular/monorepo/packages/capacitor-app/android in 52.64ms
✔ Syncing Gradle in 1.11s
✔ add in 2.80s
✔ Copying web assets from www to android/app/src/main/assets/public in 274.30ms
[error] Unable to find node_modules/@capacitor/core/native-bridge.js. Are you sure @capacitor/core is installed? This file is required for Capacitor to function
```

`@capacitor/core` should be found via node resolution algorithm

### LINKED ISSUE #865: [cli sync] node modules are not resolves correctly with workspaces/hoisting
When you use `yarn` with workspaces or `lerna` with hoisting, node modules are dispatched in different `node_modules` folders along the path to the current directory. Node.js specify the [resolution algorithm](https://nodejs.org/api/modules.html#modules_loading_from_node_modules_folders) in that case.

Capacitor's CLI does not adhere to this specification and fails to work in such an environment. This a blocking issue in our case because we have to handle our application in monorepos.

Here is the error from `cap`:
```
[error] Unable to find node_modules/@capacitor/core/native-bridge.js. Are you sure @capacitor/core is installed? This file is required for Capacitor to function
```

```sh
$ npx cap -V    
1.0.0-beta.8
```

ptitjes:
Fixes #865.
Fixes #792.

This PR is quite big but it does a few changes in a consistent manner:

- It introduces an asynchronous `resolveNode(config, ...pathSegments)` function that resolves the path formed by the `pathSegment`s  in a `node_modules` folder. It tries to go up to the filesystem root until it finds an existing corresponding element in the filesystem.
- It uses `resolveNode` everywhere `fs.resolve` was used to resolve a file in a node module.
- It fixes the creation of the references in the generated `gradle` files.
- It removes the hard-coded references in the android template and adds those to the generated file. (This is a BREAKING CHANGE!)
- It removes the use of `npm list` to check for the relative versions of the `cli` and platform modules.

With this PR, `npx cap sync` works correctly and the gradle builds work correctly.

The iOS counterpart has not been worked on, as I can't test it.


ptitjes:
@mlynch Would you mind to review ?

kalicki:
I think this PR fix a problem https://github.com/ionic-team/capacitor/issues/875

mlynch:
Thanks for the PR. This one is pretty complex so I don't think it's going to get merged with much haste. 

Do these changes impact iOS? i.e. if we merged this right now would it break iOS?

My biggest concern is the android gradle template, as you mentioned that's a breaking change. However, breaking changes are better done today than after we ship 1.0...

dwieeb:
Should also resolve https://github.com/ionic-team/capacitor/issues/792

ptitjes:
@mlynch, @jcesarmobile any chance this can make it to master for the next release ?

dwieeb:
@ptitjes If you bump the engine requirement to Node 8.9.0+ you can use the `paths` option of `require.resolve()`: https://nodejs.org/dist/latest-v8.x/docs/api/modules.html#modules_require_resolve_request_options

This would be preferred.

ptitjes:
@dwieeb Thanks! I did not know that API.

Tell me if I'm wrong but, from the docs, I get that only the second (optional) argument `paths` exists since Node 8.9.0. The function seems to exists since 0.13.

We don't need that `paths` argument here. As I understand it, it is used to specify the locations from where to run the resolution algorithm. Yet we only want the default resolution paths to be used (`./`, `~/.npm/`, `/usr/local/node`, ...).

So I will rewrite my `resolve()` function based on `require.resolve()` and `path.*()`. Thanks for the suggestion.

Also I will add a commit to fix the requirement reported in the error you mention https://github.com/ionic-team/capacitor/issues/792#issuecomment-432855490.

cli/src/android/update.ts
```diff
@@ -92,18 +102,19 @@ if (hasProperty('postBuildExtras')) {
...
   let applyArray: Array<any> = [];
   cordovaPlugins.map( p => {
+    const pluginPath = relative(p.rootPath, pluginsFolder!);
     const frameworks = getPlatformElement(p, platform, 'framework');
     frameworks.map((framework: any) => {
       if (!framework.$.type && !framework.$.custom) {
         frameworksArray.push(framework.$.src);
       } else if (framework.$.custom && framework.$.custom === 'true' && framework.$.type && framework.$.type === 'gradleReference') {
-        applyArray.push(`apply from: "../../../../${p.id}/${framework.$.src}"`);
+        applyArray.push(`apply from: "${pluginPath}/${framework.$.src}"`);
```
jcesarmobile: Here `../../../../` is not the plugins folder, but `node_modules`. Then `${p.id}` is the specific cordova plugin folder that have the gradle file. So it should be something like `${resolveNode(p.id)}`.

ptitjes: Ah, thanks ! So I have to resolve it in my patch. Thanks for the info.

dwieeb:
> Yet we only want the default resolution paths to be used

It looks like you're using the function to resolve node modules from `config.app.rootDir`. I think `require.resolve('module', { paths: [config.app.rootDir] })` should work.

ptitjes:
@dwieeb Yes that's finally what I went. However, I just tested it and that won't work: Unfortunately, `require.resolve` can only resolve packages that have a `main` property in their `package.json`. `@capacitor/android` don't have one (and that would make no sense to have one anyway).

I looked for possible dependencies that do that without luck. I have to resort to implementing the algorithm by hand. :(

ptitjes:
@mlynch, @jcesarmobile, I rebased and took the opportunity to clean more my code and split into multiple commits. I think that this, and the fact @jcesarmobile moved the cordova plugin root, makes the commits a lot more understandable.

The only two remaining explicit references to `node_modules` are [in the `config.ios.capacitorRuntimePod` and `config.ios.capacitorCordovaRuntimePod` variables](https://github.com/ionic-team/capacitor/blob/master/cli/src/config.ts#L52-L53). I don't really know how those should be handled.

Please make any comments that could help me enhance this PR. This PR is vital for me and, I believe, those that maintain large projects with dependency libraries and tools inside monorepos.

**Side note:** While passing over that code, I believe that since @jcesarmobile moved the cordova plugin root and made it configurable, [the `./capacitor-cordova-android-plugins/` path in the `android-template`](https://github.com/ionic-team/capacitor/blob/df4de0f53d6fd3bf4795b98844feaf97613a071f/android-template/settings.gradle#L3) should not be hardcoded but generated in `capacitor.settings.gradle` at update.

jcesarmobile:
For iOS we can remove those variables and replace them with the resoultion of `@capacitor/ios` path. I can do it in a separate PR if you don't have a mac to test.

The cordova plugin root is not really configurable, I just put the variable in the config file to reuse it instead of using the name to avoid errors, but can't be configured.
We can change it to the capacitor.settings.gradle, but I don't think we will change the name/path again, so as you wish.


BTW, this change causes problems on windows.
Paths like this don't work on gradle
`project(':capacitor-android').projectDir = new File('..\node_modules\@capacitor\android\capacitor')`
should be like this
`project(':capacitor-android').projectDir = new File('../node_modules/@capacitor/android/capacitor/')`

ptitjes:
I do have a Mac but I run it on Linux 😁 I am not very comfortable to develop on Mac however. So maybe it is better if you wish to handle it. Tell me and if necessary, I will make an effort.

Ah, yeah I see the config act as a central point to share paths, like the app root. So then it is good.

For the Gradle paths on Windows, just to make it clear, they must be Unix-like and not Windows-like. Do I understand correctly ? If yes, I will see how I can fix that then.

ptitjes:
@jcesarmobile I opted for the simplest way, by replacing occurrences of `\` by `/`, as there is no way to force the `path` module to use the `/` separator. I think it should work like that but I can't test on Windows.

cli/src/android/update.ts
```diff
@@ -54,9 +54,10 @@ export async function installGradlePlugins(config: Config, capacitorPlugins: Plu
 
   const settingsPath = join(config.app.rootDir, 'android');
   const dependencyPath = join(config.app.rootDir, 'android', 'app');
+  const relativeCapcitorAndroidPath = relative(settingsPath, capacitorAndroidPath).replace('\\', '/');
```
jcesarmobile: This only replaces the first occurrence, try with something like `.replace(/\\/g, '/');`

cli/src/android/update.ts
```diff
@@ -68,10 +69,11 @@ project(':${getGradlePackageName(p.id)}').projectDir = new File('${relative(sett
   let frameworksArray: Array<any> = [];
   let prefsArray: Array<any> = [];
   cordovaPlugins.map( p => {
+    const relativePluginPath = relative(dependencyPath, p.rootPath).replace('\\', '/');
```
jcesarmobile: Same as avobe, only replaces the first occurrence.

cli/src/android/update.ts
```diff
@@ -102,7 +115,7 @@ export async function handleCordovaPluginsGradle(config: Config,  cordovaPlugins
       if (!framework.$.type && !framework.$.custom) {
         frameworksArray.push(framework.$.src);
       } else if (framework.$.custom && framework.$.custom === 'true' && framework.$.type && framework.$.type === 'gradleReference') {
-        applyArray.push(`apply from: "../../../../${p.id}/${framework.$.src}"`);
+        applyArray.push(`apply from: "${relative(pluginsFolder, p.rootPath)}/${framework.$.src}"`);
```
jcesarmobile: here you also need to replace the `\` with `/` on the relative result.

ptitjes: Rha, missed it!

ptitjes:
@jcesarmobile I fixed the changes you requested. That should be good now.

As of `capacitor add android` hanging on Windows, I have no idea what the problem could be in these commits. I checked on Linux and it doesn't hang whether `@capacitor/android` is already installed or not!

cli/src/common.ts
```diff
@@ -324,15 +331,61 @@ export async function printNextSteps(config: Config, appDir: string) {
...
+
+function resolveNodeFrom(start: string, id: string): string | null {
+  let basePath = resolve(start);
+  let modulePath;
+  while (true) {
+    modulePath = join(basePath, 'node_modules', id);
+    if (existsSync(modulePath)) {
+      return modulePath;
+    }
+    if (basePath === '/') {
```
jcesarmobile: This is was makes the cli hang on windows, as `basePath` will never be `/` on windows.
I've tried getting the base path like this instead of using a hardcoded value
`parse(process.cwd()).root` (needs to add `parse` in the `path` import, and you can remove `relative` from `path` import as it's not used in this file).
I've tried this in windows and mac and seems to work fine

ptitjes: Great. I'll do that tonight.

ptitjes:
@jcesarmobile I took the liberty to add three more commits, in order to test the fixes:
- Fix the android tests by using `npm install` to setup the `node_modules` directory (otherwise, when `cap add android` executes `npm install --save @capacitor/android`, `node_modules/@capacitor` gets wiped) and copying the `cordova-android-plugins` after `cap add android` (otherwise it fails).
- Bump Jest to get `describe.each`.
- Add tests for the mono-repo layout.

cli/src/common.ts
ptitjes: Fixed as requested.

ptitjes:
Any feedback ? Should I add the tests in a different PR ?

jcesarmobile:
Sorry, didn't have time to review the test changes, I would prefer if you send them in a separate PR as this was very close to being merged and the tests changes will require a new review.

On a quick look, couldn't you use the existing `runCommand` function from cli instead of creating a new one?
And on `npm install`, wouldn't it better to install using the relative path to cli and core instead of `@capacitor/core`  and `@capacitor/cli` as that will install previous published version instead of using current code? Wouldn't affect much at the moment because we don't really do anything with them at the moment in any test, but might in the future.

jcesarmobile:
@ptitjes so, are you planning on removing the tests and adding them in a separate PR?

ptitjes:
@jcesarmobile Yes but I had no time to do it this week-end. I'll try tomorrow morning. Otherwise, if you want to merge the first commits, could you cherry-pick them ?

ptitjes:
@jcesarmobile So I forced-pushed to remove the test commits. I will make a PR as soon as this PR gets merged. Sorry for the noise...

mlynch:
w00t! Thanks so much for the PR

dwieeb:
🎉

ptitjes:
Thank you for your patience guys ! <3

PR #911, opened 2018-10-22
feat(statusBar): set default ios status bar options based on plist

sbannigan:
This will allow the app to either hide the status bar by default or use the light content style by default based on the `Info.plist`, which prevents the status bar flashing an undesired state from the bridge view controller's defaults.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -163,9 +163,7 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScr
   }
 
   public func setStatusBarDefaults() {
-    if let url = Bundle.main.url(forResource:"Info", withExtension: "plist"),
-      let plist = NSDictionary(contentsOf: url) as? [String:Any] {
-        
+    if let plist = Bundle.main.infoDictionary {
       if let statusBarHidden = plist["UIStatusBarHidden"] as? Bool {
         if (statusBarHidden) {
           self.isStatusBarVisible = false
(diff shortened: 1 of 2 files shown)


ios/Capacitor/Capacitor/CAPBridgeViewController.swift
```diff
@@ -160,6 +162,24 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScr
 
   }
 
+  public func setStatusBarDefaults() {
+    if let url = Bundle.main.url(forResource:"Info", withExtension: "plist"),
+      let plist = NSDictionary(contentsOf: url) as? [String:Any] {
```
jcesarmobile: You can get the info plist information with `let plist = Bundle.main.infoDictionary`.

Other than that, looks good to me.

Might be good to also document this in the Statusbar plugin section

sbannigan: Changes made and added a line to the status bar docs!

site/docs-md/apis/status-bar/index.md
```diff
@@ -10,6 +10,8 @@ The StatusBar API Provides methods for configuring the style of the Status Bar,
 
 This plugin requires "View controller-based status bar appearance" (`UIViewControllerBasedStatusBarAppearance`) set to `YES` in `Info.plist`. Read about [Configuring iOS](../ios/configuration) for help.
 
+The status bar visibility defaults to visible and the style defaults to `StatusBarStyle.Dark`. You can change these defaults by modifying the `UIStatusBarHidden` and `UIStatusBarStyle` in the `Info.plist`.
```
jcesarmobile: Default value is `StatusBarStyle.Light` as it sets the status bar to `.default` value, which is the one set as default on `CAPBridgeViewController.swift`.

Also, as `UIStatusBarHidden` and `UIStatusBarStyle` are not in the info.plist in the base project, I think it's better to say "by adding" than "by modifying"

sbannigan: 👍

PR #912, opened 2018-10-22
Call success on Toast and minor fixes

jcesarmobile:
We weren't calling success on toast show, so the promise was pending.
Also removed some `;`, a commented line and adjusted the max size a little as toast was going offscreen if the text was too long.


mlynch:
YEAH TOAST!

PR #918, opened 2018-10-23
Fix: added accuracy to GeolocationPosition object

bartbaas:
Using Capacitor on Android, no accuracy is recieved on the Geolocation plugin. This fix adds the property to the GeolocationPosition object.


jcesarmobile:
Nice catch, thanks!

bartbaas:
@jcesarmobile thank you for merging.

PR #927, opened 2018-10-25
Move cordova plugin files out of node_modules

jcesarmobile:
Renamed `capacitor-android-plugins` to `capacitor-cordova-android-plugins` as it's only used for cordova plugins.

Moved the place were cordova plugin files are copied, from inside `node_modules/@capacitor/cli/assets/` to `ios/capacitor-cordova-ios-plugins` and `android/capacitor-cordova-android-plugins`
Folders are gitignored by default as they get recreated on update/sync


mlynch:
Nice!

ptitjes:
@jcesarmobile I am rebasing #867, and I am having difficulties to understand your changes.

In `handleCordovaPluginsGradle(...)`, you changed the `pluginsFolder` to a new location based on a config variable `config.android.assets.pluginsFolderName`. However, a few lines below, you produce the same relative path for the plugin's directory starting with a hardcoded `../../../../`.

Shouldn't you make that path relative to the `pluginsFolder` path computed before, using `path.relative(...)` ? Or did I miss something ?

See at line:
https://github.com/jcesarmobile/capacitor/blob/b79e8f4aa9ac78d56bc2a8a74e3714e89f5c225e/cli/src/android/update.ts#L105

jcesarmobile:
we have moved the cordova plugin folders outside node_modules so people can commit them if they want and because writing in node_modues is not a good idea.

I missed that one on the changes, it should search for p.id path or copy the gradle file somewhere in `pluginsFolderName`.

PR #933, opened 2018-10-27
Custom html support electron splash screen

arsasm:
Added customHtml splash option to get a string with the html content to let users full customization.


mlynch:
Awesome!

PR #956, opened 2018-11-05
Ability to configure splash autohiding on app launch in `capacitor.co…

leshik:
…nfig.json`

### DIFF
--- a/site/docs-md/apis/splash-screen/index.md
+++ b/site/docs-md/apis/splash-screen/index.md
@@ -44,6 +44,18 @@ If your app needs longer than 3 seconds to load, configure the default duration 
 }
 ```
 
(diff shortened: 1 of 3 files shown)


site/docs-md/apis/splash-screen/index.md
```diff
@@ -44,6 +44,18 @@ If your app needs longer than 3 seconds to load, configure the default duration
 }
 ```
 
+If you want to be sure the splash never hides before the app is fully loaded, set `launchAutoHide` to `false` in your `capacitor.config.json`:
+
+```json
+{
+  "plugins": {
+    "SplashScreen": {
+      "launchAutoHide": false
```
jcesarmobile: Can you also add it to the "Configuration" section? (just a few lines below)

leshik: Sorry, missed it – now updated.

PR #958, opened 2018-11-05
Dedicated fileExists method for iOS and Android

### LINKED ISSUE #957: Need dedicated method for checking if a file exists
I have been using stat for this, but it's not semantically ideal and it is a common enough action that I think it should be part of the core API.

bdsimmons:
Fixes #957 

Adds `Filesystem.fileExists`


jcesarmobile:
closing since there was no response and has conflicts

oliveryasuna:
Can this please be readdress? We cannot rely on `stat` throwing an exception because the exception could be thrown for other reasons.

Issue #951, opened 2018-11-02
WebView on Android strips request headers


leshik:
At least when using the external server url in `capcacitor.config.json`, Android WebView strips request headers; e.g. compare this from Capacitor app:

<img width="222" alt="screenshot 2018-11-02 16 13 59" src="https://user-images.githubusercontent.com/443678/47905964-d1cdaa00-deba-11e8-9f9d-14d316a4ae9d.png">

to this when using the Chrome browser:

<img width="235" alt="screenshot 2018-11-02 16 21 07" src="https://user-images.githubusercontent.com/443678/47906215-53bdd300-debb-11e8-8ac6-75549b590772.png">

In the first case (note the yellow exclamation mark and the lack of _view source_ link) GraphQL request was broken, because in fact `Content-Type: application/json` header (and others, too) was never sent with the request to the server (checked with tcpdump). By default WebView uses `x-www-urlencode` which doesn't work.

iOS version doesn't suffer from this problem.

PR #962, opened 2018-11-07
Android: Intercept proxyied requests in handleProxyRequest for GET method only

leshik:
Related to #951 and #955.

Turned out, there is a lifelong known [issue](https://issuetracker.google.com/issues/36918490) that `POST` requests can't be intercepted with `WebViewClient.shouldInterceptRequest(WebView, WebResourceRequest)` method. In my case this resulted in all GraphQL queries became duplicated – the first one is broken, with `Content-Length: 0` and an empty body, then the second normal one comes. I didn't came up with any better solution than to execute `handleProxyRequest` for `GET` requests only.

@jcesarmobile what do you think?

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -316,52 +316,54 @@ public class WebViewLocalServer {
    */
   private WebResourceResponse handleProxyRequest(WebResourceRequest request, PathHandler handler) {
     final String method = request.getMethod();
-    if (method.equals("GET")) try {
-      String path = request.getUrl().getPath();
-      URL url = new URL(request.getUrl().toString());
-      Map<String, String> headers = request.getRequestHeaders();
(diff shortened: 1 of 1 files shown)


android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -315,8 +315,8 @@ private String readAssetStream(InputStream stream) {
    * @return
    */
   private WebResourceResponse handleProxyRequest(WebResourceRequest request, PathHandler handler) {
-    try {
-      final String method = request.getMethod();
+    final String method = request.getMethod();
+    if (method.equals("GET")) try {
```
jcesarmobile: Looks good, but add curly braces to the if and put the try in next line

Something like 
```
if (method.equals("GET")) { 
  try { 
    //code
  } catch () {
    //code
  }
}
```

leshik: @jcesarmobile oh sure, now fixed.

PR #966, opened 2018-11-12
Add `allowNavigation` option that allows Capacitor to open certain URLs in its WebView.

leshik:
This PR adds `allowNavigation` option to `capacitor.config.json` that allows to configure additional hosts that Capacitor will open in its WebView.

Ref. https://forum.getcapacitor.com/t/how-to-execute-the-code-before-main-app-is-fully-loaded/200

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -53,6 +53,8 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.Arrays;
+import java.util.ArrayList;
 
 
 /**
@@ -91,6 +93,7 @@ public class Bridge {
   private String localUrl;
   private String appUrl;
   private String appUrlConfig;
+  private String[] appAllowNavigationConfig;
   // A reference to the main WebView for the app
   private final WebView webView;
   public final CordovaInterfaceImpl cordovaInterface;
@@ -155,6 +158,7 @@ public class Bridge {
 
   private void loadWebView() {
     appUrlConfig = Config.getString("server.url");
+    appAllowNavigationConfig = Config.getArray("server.allowNavigation");
 
     String port = getPort();
     String authority = "localhost" + ":" + port;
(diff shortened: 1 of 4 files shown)


site/docs-md/basics/configuring-your-app.md
```diff
@@ -41,7 +41,14 @@ The current ones you might configure are:
     // If you don't configure it, a random port will be assigned and persisted.
     "port": "8787",
     // You can make the app to load an external url (i.e. to live reload)
-    "url": "http://192.168.1.33:8100"
+    "url": "http://192.168.1.33:8100",
+    // Normally all external URLs are opened in the browser. By setting this option, you tell
+    // Capacitor to open URLs belonging to these hosts inside its WebView.
+    "allowedHosts": [
```
mlynch: I wonder if `allowedHosts` is the right name for this config. My first impression of this is that it meant allowed URLs I can load in general, including fetch, etc.

Maybe `loadHostsInWebView`?

mlynch:
Thanks for the PR! Left one comment.

site/docs-md/basics/configuring-your-app.md
leshik: @mlynch ok, I've made the change that you've suggested, and also rebased the PR to the current master.

leshik:
@mlynch resolved, hope it will go into `1.0`! :)

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -195,26 +200,26 @@ public WebResourceResponse shouldInterceptRequest(WebView view, WebResourceReque
 
       @Override
       public void onPageFinished(WebView view, final String location) {
-        if (appUrlConfig != null) {
+        if (appUrlConfig != null || appLoadHostsInWebViewConfig != null) {
```
jcesarmobile: This will inject the Capacitor js in the external loaded url, is that desired?

jcesarmobile:
Added a comment.

Also, I think I prefer a name like "allow-navigation", or "allowNavigation" to be more like the Cordova option that does the same (don't change it yet, let's see what @mlynch thinks about it)

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
leshik: @jcesarmobile exactly, my use case perfectly aligns with that – so I don't have to bundle `Capacitor` in the website's `package.json` and has `Capacitor` object ready to use right away!

leshik:
I agree, `allowNavigation` sounds more natural.

mlynch:
`allowNavigation` sounds good to me

leshik:
Thanks, @mlynch.
@jcesarmobile I changed the name and rebased again.

PR #972, opened 2018-11-16
Remove CGDWebServer

### LINKED ISSUE #975: iOS: restricted network ports
Hi,

In the midst of development all of the sudden I was hit by this error on iOS:
```
WebView failed provisional navigation
Error: Not allowed to use restricted network port
```
The URL was `http://localhost:6000` at the time, which is ok as `getRandomPort()` can return any number between 3000 and 9000, but! It turns out some of these ports are not ok for iOS (are restricted). Not sure how to address this, maybe some blacklisting? @jcesarmobile any ideas on what port numbers aren't allowed on iOS? Deleting and then reinstalling the app solves the problem as another random port gets chosen, but any user can be hit by this at any arbitrary point in time.

For now I've put the static port number in my config, but as long as we have this `getRandomPort()` function, something should be done.

jcesarmobile:
Use WKURLSchemeTask instead
Bump min deployment target to iOS 11 in example app, app template and plugin template

Closes #975


ios/Capacitor/Capacitor/CAPAssetHandler.swift
```diff
@@ -0,0 +1,68 @@
...
+                data = try Data(contentsOf: fileUrl)
+            }
+            let mimeType = mimeTypeForExtension(pathExtension: url.pathExtension)
+            let expectedContentLength = data.count
+            let headers =  [
+            "Content-Type": mimeType
+            ]
+            let urlResponse = URLResponse(url: localUrl, mimeType: mimeType, expectedContentLength: expectedContentLength, textEncodingName: nil)
+            let httpResponse = HTTPURLResponse(url: localUrl, statusCode: 200, httpVersion: nil, headerFields: headers)
+            if url.pathExtension == "mp4" {
```
mlynch: Just a note to check for other video types like .mov

ios/Capacitor/Capacitor/CAPAssetHandler.swift
```diff
@@ -0,0 +1,68 @@
...
+
+class CAPAssetHandler: NSObject, WKURLSchemeHandler {
+    
+    func webView(_ webView: WKWebView, start urlSchemeTask: WKURLSchemeTask) {
+        var startPath = ""
+        let url = urlSchemeTask.request.url!
+        let stringToLoad = url.path
+        let scheme = url.scheme
+
+        if scheme == "capacitor" {
```
mlynch: Do we want to go with a custom scheme like this, or something more generic we can use on cordova as well? i.e. `hybrid://` or `webapp://` or `webview://`?

jcesarmobile: I'm using `capacitor` for app assets (public folder) and `capacitor-asset` for other files such us downloaded images, camera pictures, it requires to use convertFileSrc to replace the `file` with `capacitor-asset`, which reminds me I also have to change the camera plugin as it has the `webPath` and will not work with this change.

But I think we should keep this schemes here and user different ones on the webview plugin, Apple recommends to namespace schemes with company name or similar. Also, Cordova will add this in the future, so better have our own schemes.

ios/Capacitor/Capacitor/CAPAssetHandler.swift
jcesarmobile: Done, added all the audio and video extensions supported on iOS and tested that they worked.

PR #977, opened 2018-11-19
Fix casing in ios/Capacitor/Capacitor/CAPBridgedPlugin.m

gotbadger:
The Foundation header include is incorrectly cased which means builds will fail on case sensitive systems.


mlynch:
Thanks!

PR #980, opened 2018-11-19
Fix android tests

ptitjes:
This PR does:
- Fix android tests by
  - using `npm install` to install `@capacitor/core` and `@capacitor/cli`. Otherwise later use of `npm install` for `@capacitor/android` or `@capacitor/ios` will wipe the `node_modules/@capacitor/` folder,
  - copying the Android Cordova assets after 'add android', otherwise `cap add android` will fail as the `android` directory already exists.
- Bump `jest` version to allow the `describe.each` construct.
- Use the `describe.each` construct to run both the `add` and `update` tests on a standard app layout and a monorepo-like layout.

**Notes:**
1. To run only the android tests, I use the following command. I believe two `package.json` commands could be setup in the future, with two jobs in a `.travis.yml` file (with one running on the `macos` image with `Xcode`), however this is currently above my skills.
```bash
npx jest --testPathPattern='.*\.android\..*' --verbose --silent
```
2. The very first run of the above command on my not-clean repo fails. The subsequent runs succeed. I believe this is due to one of my `node_modules` to be dirty and would not happen on a clean git clone.

Happy to fix any requested changes.

### DIFF
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -4,12 +4,46 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
+      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "dev": true,
+      "requires": {
+        "@babel/highlight": "^7.0.0"
+      }
+    },
+    "@babel/highlight": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/highlight/-/highlight-7.0.0.tgz",
+      "integrity": "sha512-UFMC4ZeFC48Tpvj7C8UgLvtkaUuovQX+5xNWrsIoMG8o2z+XFKjKaN9iVmS84dPwVN00W4wPmqvYoZF3EGAsfw==",
+      "dev": true,
+      "requires": {
+        "chalk": "^2.0.0",
+        "esutils": "^2.0.2",
+        "js-tokens": "^4.0.0"
+      },
+      "dependencies": {
+        "esutils": {
+          "version": "2.0.2",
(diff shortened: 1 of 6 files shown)


jcesarmobile:
You can remove the whole `copyAndroidCordovaAssets` function and usages as we were copying the folder manually because we weren't installing the CLI, but as your change installs the CLI, the folder will be created on add command. 

About note 2, I randomly get failures too. I do `rm -rf node_modules`, then `npm install` and then the jest command you shared. Sometimes fails the first time, sometimes it works, on the second run it always work. Not sure what's going on, but not related to having a dirty node_modules as I'm deleting it

ptitjes:
@jcesarmobile Also about note 2, the random failure seems to disappear when I change the `npm install` of `core` and `cli` to use the latest versions on npm's central repository, instead of relative file specifiers. Maybe this can help us nail down the problem.

For the record, here is the exact failure that I get:
```
    Failed: "✖ Installing android dependencies: 
    ✖ add: 
    [error] npm WARN test-app@ No description
    npm WARN test-app@ No repository field.
    npm WARN test-app@ No license field.

    npm ERR! path /tmp/tmp-16547UsVnZd7ssUwQ/test-app/node_modules/.staging/@capacitor/cli-ff3138c2/node_modules/@babel/code-frame
    npm ERR! code ENOENT
    npm ERR! errno -2
    npm ERR! syscall rename
    npm ERR! enoent ENOENT: no such file or directory, rename '/tmp/tmp-16547UsVnZd7ssUwQ/test-app/node_modules/.staging/@capacitor/cli-ff3138c2/node_modules/@babel/code-frame' -> '/tmp/tmp-16547UsVnZd7ssUwQ/test-app/node_modules/.staging/@babel/code-frame-4e03bde6'
    npm ERR! enoent This is related to npm not being able to find a file.
    npm ERR! enoent 

    npm ERR! A complete log of this run can be found in:
    npm ERR!     /home/didier/.npm/_logs/2018-11-23T18_32_15_380Z-debug.log
```

ptitjes:
```bash
`npx jest --testPathPattern='.*\.android\..*' --verbose --silent --runInBand`
```

@jcesarmobile Yep, `--runInBand` works correctly all the time. Not sure how to fix that in another way (tried adding a `--cache=${tmpDir}/npmCache` but this fails too).

Also, removed `copyAndroidCordovaAssets`. Would you mind to review again please ?

ptitjes:
Thanks!

PR #983, opened 2018-11-20
feat: add support for yarn package manager

### LINKED ISSUE #979: Support yarn when installing deps for cap add
If there is a `yarn.lock` in the root directory of a capacitor project, running `cap add [platform]` should install deps with yarn instead of npm. For projects using yarn, adding a platform takes way more time then it should, an a `package-lock.json` is generated unnecessarily. I could make a PR if requested.

nklayman:
fixes #979. Uses [yarn](http://yarnpkg.com/) instead of npm to install deps if the project has a `yarn.lock` in it's root. Otherwise, npm will be used.


nklayman:
That's only used in one location, as it differs from the normal way of installing deps. Everywhere else, the `installDeps` function is used.

geekox86:
Greetings gentlemen. Great job and teamwork.
What is the status on this important PR?

nklayman:
I fixed the merge conflict, it's ready to go now.

geekox86:
Wonderful job, thank you @nklayman !

Can we proceed with merging this PR with the official library?

We really appreciate your efforts guys.

geekox86:
Greetings gentlemen, what is the status of this PR?

geekox86:
Is it possible or easy to solve these failing tests and merge for next version release (please)?

nklayman:
I found a fix, but I also want to add tests with yarn. I should have it ready later today

geekox86:
Thx man, I really appreciate your efforts!

nklayman:
Force push to rerun tests.

nklayman:
The tests should pass now, and project creation with yarn is tested.

nklayman:
Whoops, didn't actually change the default :man_facepalming:.

cli/src/tasks/create.ts
```diff
@@ -1,19 +1,22 @@
 import { Config } from '../config';
 import { copy } from './copy';
+import { exec } from 'child_process';
```
jcesarmobile: exec is not used, remove import

cli/src/common.ts
```diff
@@ -390,3 +427,22 @@ function resolveNodeFrom(start: string, id: string): string | null {
...
+  // npmClient in config
+  const npmClient = config.cli.npmClient;
+  if (npmClient === 'yarn') {
+    return true;
+  } else if (npmClient === 'npm') {
+    return false;
+  }
+  // yarn.lock
+  return existsSync(join(projectDir || process.cwd(), 'yarn.lock'));
+    // process.argv.includes('--npm-client') ? process.argv[process.argv.indexOf('--npm-client') + 1] : process.argv.includes('--yarn') || existsSync(join(process.cwd(), 'yarn.lock'));
```
jcesarmobile: don't leave comments behind

nklayman:
Good point with the init command, I removed the option there. It will only prompt for the npm client if yarn is installed. Likely, if someone has installed yarn, they would like to use it. I moved the npm-client to an argument for the create command as suggested. The commands will check if yarn or npm is used for the project as described in https://github.com/ionic-team/capacitor/issues/979#issuecomment-446407593.

jcesarmobile:
`capacitor create` is not using yarn for installing the packages, not when using `--npm-client yarn`, nor when it prompts for a npm client to use and I choose yarn.

The `--npm-client` option should be in `create` and in `init` commands, not in `add`.

I don't really like the prompt, even if it only appears when yarn is installed. I would prefer if you remove it. But if you leave the prompt, then `init` should also prompt for consistency. 

As `init` and `create` are only run once per project, if `--npm-client` is provided it should be written into the `capacitor.config.json` file for the following commands (`add`) to use the desired option instead of making those commands to prompt or have that option.
If you keep the prompt, then the choice should be persisted to `capacitor.config.json` too

nklayman:
the `getOrCreateConfig` function was removing the `npmClient` from the config object passed to it. This caused the create command to ignore it. I moved the `--npm-client` option from the add command to the init command as well.

jcesarmobile:
thanks for the PR and all the edits!

geekox86:
Greetings all, I can still see that the current docs are stating:

> Capacitor does not officially support yarn nor will we be able to fix or keep open any issues filed for it.

Any plans to update the docs to reflect the efforts done by this PR?

Thank you.

jcesarmobile:
Yeah, the full docs are being revisited/rewritten and this is one of the things being changed

geekox86:
Awesome! Thank you very much.

PR #996, opened 2018-11-23
fix(cli): rewrite Gradle path separators on Windows (#992)

### LINKED ISSUE #992: sync android produces invalid characters in gradle file on 1.0.0-beta.11
It seems after updating to beta.11 `sync android` produces mixed slashes in my gradle file.

This does not compile on Windows and the paths seem to be in the normal format returned by nodejs' `path.join`.

image

ptitjes:
Fixes #992.


ptitjes:
Thank you!

PR #1004, opened 2018-11-26
Fix ios tests

ptitjes:
As for Android:
- removes the lines that copies the cordova assets as I believe they are now copied by `cap add` as for Android
- use `describe.each` to `add.ios.spec.ts` and `update.ios.spec.ts` to also test for the monorepo-like project layout
- add `-i` (aka. `--runInBand`) to `test` script in `package.json`

Under MacOS (Pod is required for the iOS part), all tests passes with `npm run test`! 🎉
Ready to be added to Circle CI.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -19,7 +19,7 @@
     "lint": "tslint --project tsconfig.json",
     "assets": "node scripts/copy-assets.js",
     "prepublishOnly": "npm run build",
-    "test": "jest",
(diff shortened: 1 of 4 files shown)


ptitjes:
Also `typescript` can be updated to `^3.1.6`. `cli` still builds correctly and tests pass correctly (without `jest` complaining for the old `typescript` versions).

Tell me if you want that I add a commit for that.

PR #1017, opened 2018-11-28
Remove script tags from plugins code

jcesarmobile:
If a plugin have a script tag in the .js code, even if it's a comment as on cordova-plugin-buildinfo, it breaks the js injection.
Remove any script tag and it's content from the plugin's .js files


mlynch:
What if they are injecting it into the page? Would a better approach be to escape it?

jcesarmobile:
I've only found it in a comment, so removing or escaping doesn't make a difference in that case.

To inject a script they should use `document.createElement('script');` 

adding a script like `<script>whatever</script>` will break our implementation as we put all the plugins code in a script tag, so whenever the </script> tag from the .js file appear, it closes our script instead

PR #1021, opened 2018-11-29
Prompt for Camera permission before using it

### LINKED ISSUE #959: IOS Camera should check for permission
Hi. The Camera plugin should check the authorization status prior to accessing the camera. One way that worked for me was to add a listener on the plugin and be notified of the status so I can notify the user with a dialog. In all cases, the camera should not be accessed if not authorized: https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_ios
Cheers

jcesarmobile:
Closes #959

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Camera.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Camera.swift
@@ -120,26 +120,33 @@ public class CAPCameraPlugin : CAPPlugin, UIImagePickerControllerDelegate, UINav
       return
     }
 
-    let presentationStyle = call.getString("presentationStyle")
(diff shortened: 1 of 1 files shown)


mlynch:
I'm not convinced this is necessary. My understanding of the permission flow is that if you use the UIImagePickerController, it handles the prompts for you. AVCaptureDevice is a lower level API and I think would be used when doing manual capture/streaming of camera.

If you test the Camera plugin, it properly asks you for permissions as soon as you use it.

jcesarmobile:
It's not necessary, but I think it's better UX as if the user deny the permission it will just call the permission error, while current implementation will leave the camera UI with black image and let the user take the black image picture.

PR #1029, opened 2018-12-03
Use capacitor:// scheme for Android

jcesarmobile:
Changed the local server to load from `capacitor://` scheme instead of https
Also added `capacitor-asset` for handling `file://` urls and `capacitor-content` to handle `content://` urls

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -83,6 +83,9 @@ public class Bridge {
 
   // The name of the directory we use to look for index.html and the rest of our web assets
   public static final String DEFAULT_WEB_ASSET_DIR = "public";
+  public static final String CAPACITOR_SCHEME_NAME = "capacitor";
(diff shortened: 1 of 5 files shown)


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -160,9 +160,8 @@ private void loadWebView() {
     appUrlConfig = Config.getString("server.url");
     appAllowNavigationConfig = Config.getArray("server.allowNavigation");
 
-    String port = getPort();
-    String authority = "localhost" + ":" + port;
-    localUrl = "https://" + authority;
+    String authority = "app";
+    localUrl = "capacitor://" + authority + "/";
```
mlynch: Should we ditch authority completely and move to a fixed, configurable hostname? The possibility of clash seems rare considering it's local to the app, no?

jcesarmobile: I used `app` as authority to be consistent with iOS, but we can make it configurable on both

PR #1038, opened 2018-12-06
Filesystem plugin based on IndexedDb

Beutlin:


mlynch:
This is awesome! Thanks for the contribution.

One thing stood out as something I am not sure we want: the `validateParams` feature. Cordova used this kind of runtime type checking but you get that "for free" with typescript and we don't need to re-invent the wheel. 

Thoughts? cc @jcesarmobile

Beutlin:
Thanks for your feedback. I have removed the validation check.

Currently the following methods I miss in the FileApi:
- moveFile
- moveFolder
- copyFile
- copyFolder
- removeRecursive
- doesExist

What is the best practice. Do you or someone from Ionic want to define the interface? Or can I send a suggestion as PR?

There is already a merge commit #779. but it is questionable if the typical use case is to move a file from CACHE directory into the APPLICATION directory. In my opinion it is more likely, that someone wants to rename the path.

Beutlin:
Is there any reason why this PR is not merged?
It is my second attempt to extend the filesystem plugin for the web platform.
As far as I know this is one of your major goals: Provide a universal plugin api, which can be used anywhere (iOS, Android, Web, Electron, ...).

PR #1058, opened 2018-12-14
Add scrollEnabled configuration option

### LINKED ISSUE #1055: API to disable <body> scrolling
https://github.com/ionic-team/cordova-plugin-ionic-webview/blob/486d412ad365b76aae40b25c758551310368a35b/src/ios/CDVWKWebViewEngine.m#L649

Ionic works better if the `<body>` scrolling is completely disabled.

**Some possible solution**
- Provide an API that ionic can use without importing capacitor directly
- Provide some API that can be used directly using the capacitor config, so ionic starters can be preconfigured to disable scrolling
- Provide some runtime API so the value can be changed later by the developer.

jcesarmobile:
I've put inside ios configurations

```
"ios": {
      "scrollEnabled": false
}
```
Closes #1055

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -62,6 +62,9 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScr
     setKeyboardRequiresUserInteraction(false)
     
     bridge = CAPBridge(self, o)
+    if let scrollEnabled = CAPConfig.getValue("ios.scrollEnabled") as? Bool {
+        webView?.scrollView.isScrollEnabled = scrollEnabled
+    }
   }
   
   override public func viewDidLoad() {


jcesarmobile:
`isScrollEnabled` default is true, the PR only sets a value if one is set on the preferences, if none is set keep the default. I think we should keep the current default.

I couldn't find a way of doing this on Android and webview plugin only does it for iOS

mlynch:
👍

ihsanberahim:
the commit is merge in 1.3.0 but the line missing

https://github.com/ionic-team/capacitor/blob/b941e88e7915cadbb453df3a9d6bc634be172116/ios/Capacitor/Capacitor/CAPBridgeViewController.swift

Not sure why/how.

jcesarmobile:
It was moved https://github.com/ionic-team/capacitor/blob/a77d51db73bfa3002a02762fb5ab2a49b6717be1/ios/Capacitor/Capacitor/Plugins/Keyboard.m#L52

PR #1069, opened 2018-12-21
Make electron use web toast implementation

jcesarmobile:
Also changed `__dirname` to `process.cwd();` as `__dirname` is the folder where the script is, so not the folder we want when `@capacitor/electron` is linked or locally installed.


electron/src/plugins.ts
```diff
@@ -10,7 +10,7 @@ export * from '@capacitor/core/dist/esm/web/storage';
 export * from './electron/filesystem';
 export * from './electron/network';
 export * from './electron/splashscreen';
-export * from './electron/toast';
+export * from '@capacitor/core/dist/esm/web/toast';
```
mlynch: do we even need to export this then? The idea is that web implementations should just work if they are the sole implementation of a given plugin at runtime

jcesarmobile: I don't know for sure, but the other web plugins are exported here, so I did the same for toast, I can try to remove all of them and check if they continue working

jcesarmobile: Yeah, it's not needed, I've removed all of them.

PR #1073, opened 2018-12-25
Update web.html

iliyaZelenko:
maybe I don’t understand something, but why return "Promise.resolve" if the `async` function returns `Promise.resolve` by itself? (if there are no `throw`).


jcesarmobile:
Yeah, you are right, it's not really needed to return a promise. But it should really return `Promise.resolve({ value: options.value });` or `Promise.resolve(options)` as it's an echo plugin that should echo whatever it receives.

Anyway, going to close the PR, you shouldn't edit the .html files, those are generated from the .md files before we publish a new version. 
Feel free to send a new PR editing the corresponding .md file (https://github.com/ionic-team/capacitor/blob/master/site/docs-md/plugins/web.md) with the suggestion I made.

PR #1077, opened 2018-12-26
Follow pod symlinks

sergiog90:
Hi,

we have a project with [lerna](https://github.com/lerna/lerna) who uses symlinks to manage dependencies and we found that CocoaPods doesn't follow symlinks.

The problem is that CocoaPods doesn't follow symlinks when `pod install` is runned so is needed to resolve it when making the PodFile. If the dependency (custom plugin) is a symlink CocoaPods don't imports any source code when installing/adding it.

[7051](https://github.com/CocoaPods/CocoaPods/pull/7051) - CocoaPods closed pull request.

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -3,6 +3,7 @@ import { CheckFunction, checkPlatformVersions, logFatal, resolveNode, runCommand
 import { copySync, readFileAsync, readFileSync, removeSync, writeFileAsync, writeFileSync } from '../util/fs';
 import { Config } from '../config';
 import { join, relative, resolve } from 'path';
+import { realpathSync } from 'fs';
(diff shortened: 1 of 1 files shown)


jcesarmobile:
PR looks good, but I would like to test it before merging. Can you provide some sample project that uses lerna and steps to reproduce from there?

sergiog90:
[capacitor-symlink.zip](https://github.com/ionic-team/capacitor/files/2726688/capacitor-symlink.zip)

I create another project without Lerna but with same problem.

Structure:
```
- demo: Ionic app with plugin installed
- plugin (Capacitor Plugin)
- symlink
- - plugin -> ../plugin
```
Plugin is installed with `npm i ../symlink/plugin` like Lerna.

If you open de demo app in Xcode yo can see that Plugin Development Pod is empty, no source code is imported.

PR #1098, opened 2019-01-04
Fix capacitor scheme files not loading on live reload

### LINKED ISSUE #1096: Android: Show image from local file
In my app I have functionality that downloads image from url and saves it local file system. And then I show this image inside img tag. On iOS device everything works like it should be. But on android I have an error: 
`GET https://localhost:6469/_capacitor_/data/user/0/io.ionic.starter/files/images/das1dp5nnlcacs-5baa4e982ff98.jpg net::ERR_CONNECTION_REFUSED`

File exists on device - I checked it by using `Filesystem.stat()` for this file:
```
{
    mtime: 1546531603000
    size: 210927
    type: "file"
    uri: "file:///data/user/0/io.ionic.starter/files/images/das1dp5nnlcacs-5baa4e982ff98.jpg"
}
```

To convert local file url to proper img url I'm using `window.Ionic.WebView.convertFileSrc()` function. 
Again, on iOS this works great, but not on Android. 

Am I doing something wrong?

jcesarmobile:
Closes #1096

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -181,6 +181,14 @@ public class WebViewLocalServer {
       return null;
     }
 
+    if (request.getUrl().getScheme().equals(capacitorContentScheme) || request.getUrl().getScheme().equals(capacitorFileScheme)) {
+      InputStream responseStream = new LollipopLazyInputStream(handler, request);
+      String mimeType = getMimeType(request.getUrl().getPath(), responseStream);
+      int statusCode = getStatusCode(responseStream, handler.getStatusCode());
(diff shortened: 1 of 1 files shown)


GuilhermeBCC:
https://github.com/ionic-team/capacitor/blob/d06c1951ea28f6cb7ba85402a2676f3f316b2341/android/capacitor/src/main/java/com/getcapacitor/plugin/Filesystem.java#L80

I'm getting an error when I try to access u.getScheme (because is null)

I think all code needs to be refactored for something like:

"file" .equals (u.getScheme ()) 

or

if (u.getScheme ()! = null && u.getScheme (). equals ("file"))

In fact for me I had to pull out the whole line to work with live reload

jcesarmobile:
Can you create a new issue and provide an example of an url that have a null scheme or the sample code you use?

Issue #814, opened 2018-09-21
iOS 12 App Touch-Areas are offset after keyboard is hidden


ImperialDevelopment:
**Steps to reproduce:**
- Type into an input, which will scroll the Application's view
- Press a button, while the input is focued

**What happens:**
-> The UI shifts down, but touchareas of UI-Elements are still offset. Causing always touching the wrong button

Tested with iPhone X and iPhone 8 Simulator on iOS 11 and iOS 12.
Only happens on iOS 12 (both Simulators)

When you take a look at the View Hierarchy it seems, some Subviews of the WebView are still offset.
Screenshot of Hierarchy attached.

![bildschirmfoto 2018-09-21 um 18 29 11](https://user-images.githubusercontent.com/30909348/45894380-734ae180-bdce-11e8-8d02-ee2c4a8f2484.png)

jcesarmobile:
Are you using Xcode 10, right? 
I think this is a bug in the SDK 12 included in Xcode 10.

Can you try with Xcode 9 and see if you can reproduce there?

ImperialDevelopment:
You are right, does not happen with Xcode 9 / iOS 11.
But when you open an URL and load a website in a WKWebView everything works fine with iOS 12.
So it seems there is an issue in the Capacitor-Layer.

jcesarmobile:
I meant, can you test this in Xcode 9 but on iOS 12 device?

ImperialDevelopment:
Just tried it, built with Xcode 9 and run on iOS 12 device works fine.

luke-rogers:
Issue

I am also seeing this issue when using keyboard which causes input to scroll into view (may be through using keyboard arrows) then closing the keyboard, the buttons in my toolbar cannot be clicked. When you press them nothing happens. This issue is consistently reproducible and makes the app unusable.

Inline with this issue, it does not seem to affect my app when running on iOS 11.

Not super sure the cause of this issue, many issues have been raised across other projects, but it is also effecting capacitor apps.
ionic-team/cordova-plugin-ionic-webview#176
apache/cordova-ios#417

nikmartin:
We just filed the same bug against ionic 4, but was told it was probably a capacitor issue as far as ionic is concerned, spo rather than open another issue, I'll paste it here:



# Bug Report

**Ionic Info**
Run `ionic info` from a terminal/cmd prompt and paste the output below.

```
[...]

PR #1119, opened 2019-01-11
Implement view resizing on keyboard show/hide

MarkChrisLevy:
Very simple implementation of view resizing when keyboard showing/hidding. The logic is copied from cordova-plugin-ionic-keyboard. There are two config parameters related to resizing:

- **resize**, allowed value is either `true` or `false`. Default value is `true`.
- **resizeMode**, allowed value is `native`, `ionic` or `body`.  Default value is `native`.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Keyboard.m
+++ b/ios/Capacitor/Capacitor/Plugins/Keyboard.m
@@ -49,7 +49,7 @@ typedef enum : NSUInteger {
   }
     
   BOOL doesResize = YES;
-  if ([[self getConfigValue:@"resize"] isEqualToString:@"false"]) {
+  if ([[self getConfigValue:@"resize"] isEqualToString:@"none"]) {
     doesResize = NO;
     self.keyboardResizes = ResizeNone;
     NSLog(@"CAPKeyboard: no resize");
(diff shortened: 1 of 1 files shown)


mmlleevvyy:
Imho there could be only one config param **resize**, with values `native`, `ionic`, `body` or `none`. Agree?

sneko:
I can confirm it solves my issue described there: https://github.com/ionic-team/ionic/issues/16797 (cc @paulstelzer ).

I noticed a little issue with keyboard showing up slower than the content area being resized. You can see it in the video attached below: 
[ScreenRecording_01-13-2019 12-44-22.zip](https://github.com/ionic-team/capacitor/files/2753038/ScreenRecording_01-13-2019.12-44-22.zip)

Also, I'm still facing the known offset issue described here: https://github.com/ionic-team/capacitor/issues/814 . There is no specific steps to reproduce, it's random :/

Thanks for the PR :)

sneko:
While inspecting the ion-content element with my browser and my iPhone, I can see this white space is not part of the `ion-content`.

I can confirm it by setting a background color to it and it doesn't change this space color.

But I figured out it works when setting the background color on either the <html or the <body element.

mlynch:
Thanks for the PR. `resize` is a bit too generic, maybe `keyboardResize`? cc @jcesarmobile

jcesarmobile:
@mlynch `getConfig` picks the config for the current plugin, so `resize` will be inside a Keyboard object
```
"plugins": {
    "Keyboard": {
        "resize": "none"
    }
}
```

MarkChrisLevy:
Ok, I made a change in config parameters, only one parameter needed "resize" with possible values native, ionic, body or none. By default resizing is enabled and set to native mode.

ios/Capacitor/Capacitor/Plugins/Keyboard.m
```diff
@@ -42,28 +42,38 @@ @implementation CAPKeyboard
...
         self.keyboardResizes = ResizeIonic;
+        NSLog(@"CAPKeyboard: resize mode - ionic");
       } else if ([resizeMode isEqualToString:@"body"]) {
         self.keyboardResizes = ResizeBody;
+          NSLog(@"CAPKeyboard: resize mode - body");
       }
     }
-    // NSLog(@"CAPIonicKeyboard: resize mode %d", self.keyboardResizes);
+      
+    if (self.keyboardResize == ResizeNative) {
```
stripathix: I think It should be **keyboardResizes** instead of **keyboardResize**.  It throws an error that keyboardResize is not defined.

MarkChrisLevy: Indeed. The "s" was lost, when I changed config param name. Thanks!

sneko:
Any idea to fix the difference of synchronization between the keyboard showing up and the content area as I posted in a reply of this PR (https://github.com/ionic-team/capacitor/pull/1119#issuecomment-453823941)? @EricKit also described this behavior: https://github.com/ionic-team/ionic/issues/16797#issuecomment-457855566

On another side, if I focus on an input, the keyboard will show up, but if now I try to scroll the view, the view have its padding scrollable here is the result:
image

jcesarmobile:
Please, don’t comment on closed issues or merged PRs. If something is not working fine report a new issue, it’s very hard to keep track of problems if they are “hidden”. I saw it this time, but might not see it next time, and even if I saw it I might forget

PR #1120, opened 2019-01-12
 Attempt to fix splash blink, do not merge

jcesarmobile:
Adds a "fake" UIImageView with the splash image on top of WKWebView while initialization to avoid the white flash.

Might be worth to move part of the SplashScreen plugin to `CAPBridgeViewController` and make this "fake" image to be the real one and use SplashScreen plugin to hide and show it.


stewwan:
I think this can be merged. already tested and I ended up by merging this changes on this PR too #1125

PR #1125, opened 2019-01-15
Feat/photo save video to album (ios)

stewwan:


stewwan:
sample app can be found here

https://github.com/stewwan/capacitor-photos-save-video

MarkChrisLevy:
@stewwan I think, that it should not be included in Capacitor repo, as this is additional plugin, not used by all apps. BTW I made a plugin, that does the same (only images for now, but both ios and android supported) - https://github.com/co-mmons/capacitor-multimedia-library-plugin

stewwan:
@MarkChrisLevy let's see what core team think. for me it makes totally sense since capacitor already has the Photo api built in, otherwise I would publish it from my own npm.

@jcesarmobile should review soon

MarkChrisLevy:
@stewwan My bad, I didn't see Photos plugin, as it is not presented in the docs site... In that case I will make a PR with Android implementation :-)

stewwan:
I ended up by doing a separate plugin for this.

https://github.com/stewwan/capacitor-media

PR #1129, opened 2019-01-16
Remove header form data

GuilhermeBCC:
Occasionally you can send a dataURL with header. To prevent an error while creating the file on Android and save developer time by discovering this.


GuilhermeBCC:
Issue created.

https://github.com/ionic-team/capacitor/issues/1130

android/capacitor/src/main/java/com/getcapacitor/plugin/Filesystem.java
```diff
@@ -247,6 +255,10 @@ private void saveFile(PluginCall call, File file, String data) {
       call.error("Unsupported encoding provided: " + encoding);
       return;
     }
+    //remove header
+    if(data.indexOf(',') != -1){
+      data = data.split(',')[1];
```
jcesarmobile: This doesn't compile. Can you change the single quotes to double quotes on both indexOf and split? (On indexOf seems to be allowed, but change it too for consistency)

Also, can you remove all the other changes you made? Let's stick to removing the header form data in this PR, and if you want create a new PR for the other changes and explain what they solve.

GuilhermeBCC:
Sorry. I was trying to delete [Add error handling], but by the way it's impossible. I'll create another PR and close this one.

PR #1135, opened 2019-01-20
docs(android/troubleshooting): typo missing period

rtpHarry:


rtpHarry:
ah ok... I did think it was strange making the website in actual html :P

PR #1137, opened 2019-01-20
Fix android filesystem getUri

### LINKED ISSUE #1136: Filesystem getUri on Android should not throw error when file not exists
Currently on Android, getUri() of Filesystem plugin throws "File not exists" error, when file not exists at given path/directory. On iOS getUri do not throw that error, so on Android it should work the same. Why not to throw error? Because it is a method, that should return native path, not to check if file exists. With that path we can call other native methods that requires a path (e.g. downloading files from external http resources).

MarkChrisLevy:
Fixes #1136


mlynch:
Makes sense, thanks!

PR #1167, opened 2019-01-27
Adds base64String property as a return of the Camera Plugin.

javebratt:
To make the return data compatible with Firebase Cloud Storage I'm adding the property that returns the base64 data without the encoding: "data:image/jpeg;base64,"

Please do let me know if there's any more info I need to add or if this is the proper way of adding a PR to the project 😄


mlynch:
Thanks @javebratt! My only concern is memory usage for large images as we effectively are duplicating it.

What if we stripped the metadata and just returned that in a different field?

jcesarmobile:
Agree with max, should be a separate field returned when using a different CameraResultType

javebratt:
I just pushed the update, but I'm not sure if I'm missing something :-P

android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
```diff
@@ -335,6 +335,8 @@ private void returnResult(PluginCall call, Bitmap bitmap, Uri u) {
       returnBase64(call, exif, bitmapOutputStream);
     } else if (settings.getResultType() == CameraResultType.URI) {
       returnFileURI(call, exif, bitmap, u, bitmapOutputStream);
+    } else if (settings.getResultType() == CameraResultType.BASE64NOMETADATA) {
```
jcesarmobile: Looks like you made changes on `CameraResultType` class, but didn't commit them. And you also have to do the changes on the types in @capacitor/core

jcesarmobile:
Added a comment, but I think it should work this way:

Current `CameraResultType.Base64` should return a `CameraPhoto.base64String`, and it should return what this PR adds, just the base 64 string.
Then add a new `CameraResultType.DATA_URL` (or other similar name) that should return `CameraPhoto.base64Data` (what we are currently returning)

Thoughts @mlynch ?

javebratt:
I like this approach since it seems standar, but it would introduce breaking changes wouldn't it? There might be people relying in the added metadata for their apps.

jcesarmobile:
We are still on Beta, so I think we are allowed to break things, but waiting on @mlynch opinion.

javebratt:
Alright, I pushed it without breaking changes, if Max gives the all clear I can refactor it to create the new `CameraResultType.DATA_URL`, I do agree it makes more sense.

javebratt:
Hey @mlynch could we get your opinion here? To see what the best approach would be 😸

mlynch:
Yea let's do `DATA_URL` for the one w/ the metadata, and we will have to break the current behavior for `Base64`

javebratt:
Alright, I'm on bedrest until Friday, I'll take care of it then :-)

javebratt:
Alright, so, if I understood correctly, then this PR should do the trick, let me know if there's anything missing from it.

It's adding a new return type `DATA_URL` that returns the `base64string` with the encoding metadata. and it changes the `base64` return type to return the `base64string` without the metadata.

PR #1172, opened 2019-01-28
Fix Universal Links not passing data to app

jcesarmobile:
Not sure if related to the Swift 4.2 conversion, but looks like when it has `Any` it doesn't work and the `application continue` method is not called. Changing it to `UIUserActivityRestoring` makes it work again.


sneko:
Just tested, it works! Can't wait to see that released :) thank you 👍

PR #1175, opened 2019-01-28
Add getInfo method to Status Bar plugin

sbannigan:
Gets the current state of the Status Bar including style, visibility, and color (Android only).

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -1379,6 +1379,10 @@ export interface StatusBarPlugin extends Plugin {
    *  Hide the status bar
    */
   hide(): Promise<void>;
+  /**
+   *  Get info about the current state of the status bar
+   */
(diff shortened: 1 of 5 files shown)


core/src/core-plugin-definitions.ts
```diff
@@ -1379,6 +1379,10 @@ export interface StatusBarPlugin extends Plugin {
    *  Hide the status bar
    */
   hide(): Promise<void>;
+  /**
+   *  Get info about the current state of the status bar
+   */
+  getInfo(): Promise<void>;
```
jcesarmobile: Should be StatusBarInfoResult instead of void.

Will give it a try tomorrow

sbannigan: Fixed. Cool thanks 👍

sbannigan:
Import added!

PR #1179, opened 2019-01-29
Pass custom data to the client

orlyapps:
Currently all custom data are not passed to the client. Can we add this? Or is there maybe a better way to handle this?


jcesarmobile:
Thanks for the PR.
I've just merged a [similar one](https://github.com/ionic-team/capacitor/pull/1201/files) with a few Push improvements, and one of them is to add the userInfo inside a data object (to be consistent with Android), so going to close this one.

Issue #1099, opened 2019-01-05
electron-packager not working after run application


AnteaterKit:
I am try to create app for macos with electron-packager.
But i get error after run application.
at line [https://github.com/ionic-team/capacitor/blob/master/electron/index.js#L11]

<img width="582" alt="screen shot 2019-01-05 at 21 24 42" src="https://user-images.githubusercontent.com/5115677/50727789-e59ec200-1130-11e9-8987-56d022514e45.png">
node -v v10.0.0
"@capacitor/cli": "^1.0.0-beta.13",
"@capacitor/core": "^1.0.0-beta.13"

How i can fix it?

AnteaterKit:
i an try to debug

mainWindow.loadURL(await injectCapacitor(`file://${__dirname}/app/index.html`),

__dirname = /Users/sasha/ngonestarter/electron/capacitor-app-darwin-x64/capacitor-app.app/Contents/Resources/app/app/index.html

and this file in application folder.

<img width="350" alt="screen shot 2019-01-05 at 22 15 58" src="https://user-images.githubusercontent.com/5115677/50728224-88f2d580-1137-11e9-977c-397bcd9f7332.png">

AnteaterKit:
../electron/capacitor-app-darwin-x64/capacitor-app.app/Contents/Resources/app/node_modules/@capacitor/electron/index.js
let rootPath = process.cwd();
rootPath is empty.

AnteaterKit:
i fix it local by code:

const realPath = __dirname.split('/app')[0] + '/app';

    this.splashWindow.loadURL(`data:text/html;charset=UTF-8,${splashHtml}`, {baseURLForDataURL: `file://${realPath}/splash_assets/`});

    this.splashWindow.webContents.on('dom-ready', async () => {
      this.splashWindow.show();
      setTimeout(async () => {
        this.mainWindowRef.loadURL(await injectCapacitor(`file://${realPath}/app/index.html`), {baseURLForDataURL: `file://${realPath}/app/`});
      }, 4500);
    });

What is the legal solution of this problem?

lincolnthree:
@jcesarmobile Hey! So I tried to compile and use this fix, but I am still getting the same error. How can I try to verify this? I must have done something wrong, but... the problem persists.

What I did:
Added the following to my app index.ts:
```
import * as path from 'path';
(global as any).__basedir = path.resolve(__dirname);
```

Then I did a few build steps:
```
[...]

PR #1190, opened 2019-02-01
[FIX] User a global.__basedir for Electron's rootPath, process.cwd() …

birgerstoeckelmann:
This sets a global variable in the electron-template to the current path. This path is used as the `rootPath` in the electron module to fix the issue https://github.com/ionic-team/capacitor/issues/1099.

`process.cwd()` references the directory the node process was called from, not the directory of the executable.


jcesarmobile:
Thanks!

PR #1201, opened 2019-02-06
Standardize data passed to pushNotificationReceived and pushNotificationActionPerformed events

pegler:
This standardizes the data passed with the pushNotificationReceived and pushNotificationActionPerformed events to be title, body, and data. It also adds platform specific data that may be useful for some applications.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -164,7 +164,7 @@ public class PushNotifications extends Plugin {
   public void fireNotification(RemoteMessage remoteMessage) {
     JSObject remoteMessageData = new JSObject();
 
-    Map<String, Object> data = new HashMap<String, Object>();
+    JSObject data = new JSObject();
     for (String key : remoteMessage.getData().keySet()) {
       Object value = remoteMessage.getData().get(key);
       data.put(key, value);
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -4,6 +4,7 @@ import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.content.Context;
 import android.support.v4.app.NotificationCompat;
+import android.net.Uri;
 
 
 import android.util.Log;
@@ -161,13 +162,27 @@ public class PushNotifications extends Plugin {
   }
 
   public void fireNotification(RemoteMessage remoteMessage) {
+    JSObject remoteMessageData = new JSObject();
+
     Map<String, Object> data = new HashMap<String, Object>();
     for (String key : remoteMessage.getData().keySet()) {
       Object value = remoteMessage.getData().get(key);
       data.put(key, value);
     }
-    JSObject remoteMessageData = new JSObject();
     remoteMessageData.put("data", data);
+
+    RemoteMessage.Notification notification = remoteMessage.getNotification();
+    if (notification != null) {
+      Uri link = notification.getLink();
+      remoteMessageData.put("title", notification.getTitle());
+      remoteMessageData.put("body", notification.getBody());
+      remoteMessageData.put("click_action", notification.getClickAction());
+
+      if (link != null) {
+        remoteMessageData.put("link", notification.getLink().toString());
+      }
+    }
+
     notifyListeners("pushNotificationReceived", remoteMessageData, true);
   }
 
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -173,13 +173,13 @@ public class PushNotifications extends Plugin {
 
     RemoteMessage.Notification notification = remoteMessage.getNotification();
     if (notification != null) {
-      Uri link = notification.getLink();
       remoteMessageData.put("title", notification.getTitle());
       remoteMessageData.put("body", notification.getBody());
       remoteMessageData.put("click_action", notification.getClickAction());
 
+      Uri link = notification.getLink();
       if (link != null) {
-        remoteMessageData.put("link", notification.getLink().toString());
+        remoteMessageData.put("link", link.toString());
       }
     }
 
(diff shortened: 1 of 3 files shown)


pegler:
I also fixed an issue I came across related to PushNotifications on iOS. Even when the app was in the foreground, the alert was still displayed. The change to CAPUNUserNotificationCenterDelegate.swift fixes that.

We are currently using this branch on an internal build of our app and it's working great.

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
```diff
@@ -62,11 +62,9 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
...
     }
 
-    if (self.bridge?.isAppActive())! {
-      completionHandler([.badge, .sound, .alert])
-    } else {
-      completionHandler([.badge, .sound])
-    }
+    // This method is only run when the app is in the foreground, therefore
+    // always suppress the alert notification
+    completionHandler([.badge, .sound])
```
jcesarmobile: This delegate is also used for local notifications, and I think it's desired to show them even when in foreground, so you should only remove the alert for push notifications.

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
```diff
@@ -133,7 +131,9 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
       "title": content.title,
       "subtitle": content.subtitle,
       "body": content.body,
-      "badge": content.badge ?? 1
+      "badge": content.badge ?? 1,
+      "data": content.userInfo["custom_data"],
```
jcesarmobile: I think it should be better to just pass the whole userInfo as data instead of forcing to use "custom_data" key as that is not a standard key and we don't want to force people to modify their payload to match this.

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
```diff
@@ -133,7 +131,9 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
...
+      "aps": content.userInfo["aps"],
```
jcesarmobile: As this is iOS specific and will be inside data if you do the previous change, this will be duplicated, so better remove it.

android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
```diff
@@ -161,13 +162,27 @@ public static void sendRemoteMessage(RemoteMessage remoteMessage) {
...
     for (String key : remoteMessage.getData().keySet()) {
       Object value = remoteMessage.getData().get(key);
       data.put(key, value);
     }
-    JSObject remoteMessageData = new JSObject();
     remoteMessageData.put("data", data);
+
+    RemoteMessage.Notification notification = remoteMessage.getNotification();
+    if (notification != null) {
+      Uri link = notification.getLink();
```
jcesarmobile: Can this be a String instead of Uri?

core/src/core-plugin-definitions.ts
```diff
@@ -1248,6 +1248,9 @@ export interface PushNotification {
   badge?: number;
   notification?: any;
   data?: any;
+  aps?: any;
+  click_action?: any;
```
jcesarmobile: click_action in Android is `String`, so this should be `string`

core/src/core-plugin-definitions.ts
```diff
@@ -1248,6 +1248,9 @@ export interface PushNotification {
...
+  link?: any;
```
jcesarmobile: If you change the native code to be String, then change it here too to `string`

core/src/core-plugin-definitions.ts
```diff
@@ -1248,6 +1248,9 @@ export interface PushNotification {
   badge?: number;
   notification?: any;
   data?: any;
+  aps?: any;
```
jcesarmobile: Remove aps as should be included in data (see following comment)

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
pegler: I moved this to a different pull here and addressed your feedback: https://github.com/ionic-team/capacitor/pull/1223

android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
pegler: I'm a bit rusty with Java so I don't know the best practice here. getLink returns a Uri object or null. Is there a shorthand way of getting the String value of the Uri or null?

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
pegler: We're using FCM to send iOS notifications. custom_data is standard there but I can see now that it's not a standard iOS attribute.

android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
jcesarmobile: It has a .toString() that should work, but didn’t test with an actual link

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
jcesarmobile: Interesting, are you using some plugin? Or you just integrated it into your project?

android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
pegler: But getLink can return null, which would throw an exception if we called toString there

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
pegler: APN tokens can be registered with FCM so that we can have a single API for sending messages. There's no ionic/capacitor plugin. It's just the format that FCM sends the data.

android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
jcesarmobile: Sorry, I didn’t notice you are already doing the toString later, so that’s enough

PR #1220, opened 2019-02-11
retain actionPerformed events until consumed

### LINKED ISSUE #1178: LocalNotifications on app start after notification action
I try to implement LocalNotifications into my app but it won't work if the app has been stopped.

As I see it the event fires before the listener is ready. And after that it seems broken because it doesn't work anymore.

I use beta 17 with Ionic 4.0

**app.component.ts**
```
constructor(
        private platform: Platform,
        private router: Router,
        private route: ActivatedRoute,
        private asLocalNotifications: AsLocalNotificationsService,
        private ngZone: NgZone,
    ) {
        LocalNotifications.addListener('localNotificationActionPerformed', (notification) => {
            this.ngZone.run(() => {
                this.router.navigate([notification.notificationRequest.extra.link], {relativeTo: this.route});
            });
        });
        // (<any>window).skipLocalNotificationReady = true;
        this.initializeApp();
    }

    initializeApp() {
        this.platform.ready().then(() => {
            this.asLocalNotifications.checkNotifications();
            // this.localNotifications.fireQueuedEvents();
            SplashScreen.hide();
        });
    }
```
I had the listener inside `initializeApp()` before.

How can I get it working or is there a workarround?

### LINKED ISSUE #1181: pushNotificationActionPerformed not working
I listen to the event like this:

```
PushNotifications.addListener('pushNotificationActionPerformed', (notification) => {
      console.log('#### actionPerformed: ' + JSON.stringify(notification))
    })
```

When pressing on a Push Notification the following lines is shown in the logcat:

```
D/Capacitor/Plugin/LN: LocalNotification received: null
D/Capacitor/Plugin/LN: Activity started without notification attached
V/Capacitor/Plugin/LocalNotifications: Notifying listeners for event localNotificationActionPerformed
D/Capacitor/Plugin/LocalNotifications: No listeners found for event localNotificationActionPerformed
```

I don't use Local Notifications in my application.

Possibly related to this issue: https://github.com/ionic-team/capacitor/issues/1178

pegler:


solojuve1897:
@pegler: Your commit is iOS-specific so it shouldn't fix the problem I encountered on Android. This is, to be clear, only a partial fix in that case.

pegler:
Sorry about that. I took a step back and documented the different scenarios around Push Notifications and the issues I've encountered. This includes the issue you mentioned in #1181

PR #1223, opened 2019-02-13
only show alert for LocalNotifications

### LINKED ISSUE #1109: Push notifications show when app is already open
I have implemented push notifications from capacitor which are great. I am having 2 issues however that I can't see a solution for:

1. There is now permanently a 1 badge on the app no matter what I do. I have deleted and rebuilt the app, and this is still there.

2. When the app is open, the notification still shows which is silly. For example when I am on my chat page, the messages come through, along with a notification, which I want to disable.

The documentation isn't too clear as to how I can solve these 2 issues,

Any help is greatly appreciated

pegler:
Closes #1109

### DIFF
--- a/ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
+++ b/ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
@@ -42,23 +42,23 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
     let request = notification.request
     var plugin: CAPPlugin
     var action = "localNotificationReceived"
-    var presentation_options: UNNotificationPresentationOptions = [];
+    var presentationOptions: UNNotificationPresentationOptions = [];
 
     var notificationData = makeNotificationRequestJSObject(request)
     if (request.trigger?.isKind(of: UNPushNotificationTrigger.self))! {
       plugin = (self.bridge?.getOrLoadPlugin(pluginName: "PushNotifications"))!
       action = "pushNotificationReceived"
       notificationData = makePushNotificationRequestJSObject(request)
-      presentation_options = [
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
```diff
@@ -42,14 +42,24 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
     let request = notification.request
     var plugin: CAPPlugin
     var action = "localNotificationReceived"
-    print(notification)
+    var presentation_options: UNNotificationPresentationOptions = [];
```
jcesarmobile: Swift convention is to use camel case instad of snake case, so should be `presentationOptions`

ios/Capacitor/Capacitor/CAPUNUserNotificationCenterDelegate.swift
```diff
@@ -42,14 +42,24 @@ public class CAPUNUserNotificationCenterDelegate : NSObject, UNUserNotificationC
...
+
     var notificationData = makeNotificationRequestJSObject(request)
     if (request.trigger?.isKind(of: UNPushNotificationTrigger.self))! {
       plugin = (self.bridge?.getOrLoadPlugin(pluginName: "PushNotifications"))!
       action = "pushNotificationReceived"
       notificationData = makePushNotificationRequestJSObject(request)
+      presentation_options = [
+        UNNotificationPresentationOptions.badge,
```
jcesarmobile: No need to add the `UNNotificationPresentationOptions` , this can be one line `presentationOptions = [.badge, .sound]`

Same for the other with the .alert

pegler:
Thanks for the review. Updated.

PR #1232, opened 2019-02-17
Fix iOS status bar style

sergiog90:
Fix apply status bar style on iOS.

DARK was pointing to light style and LIGHT to default.
Now they are now applied correctly.


jcesarmobile:
The way it's implemented it means

light style -> dark icons
dark style -> light icons

If people find it confusing maybe we can change it, but should be changed for Android too as it has this same behaviour at the moment

sergiog90:
In iOS when we set (on native code) lightStyle the status bar text color becomes white and when we use the default style text becomes black.

jcesarmobile:
I know, that's why it looks inverted, but in Android it works the other way around. 

When doing a cross platform framework you have to make all platforms behave the same way, because most users won't know the native behaviour, but want to get light or dark text for both platforms and not having if iOS => set to light because on iOS light = light text and if Android => set to dark because on Android dark = light text. That's something we abstract for them.

Again, I'm open to changing the behaviour, but then somebody from Android might complain about the same.
Maybe it would be better to just document light = dark text and dark = light text, or change the method to setTextStyle and change light to light text and dark to dark text.

sergiog90:
It would be enough to update the documentation with this.

Thanks!

PR #1243, opened 2019-02-19
Set always accept cookies

jcesarmobile:


dminkovsky:
I don't believe this works with WKWebView/does anything. Might be better to just revert this?

> WKWebView — Each WKWebView instance has its own cookie storage. See the WKHTTPCookieStore class for more information. 

(https://developer.apple.com/documentation/foundation/httpcookiestorage)

7-plus-t:
@jcesarmobile: As correctly stated by @dminkovsky this change has no effect on the cookie handling of the WKWebView. Shouldn't this change be reverted? It is misleading developers to believe that the cookie accept policy is set while it's actually not.

Issue #1233, opened 2019-02-18
Device.getLanguageCode returns wrong language


maedewiza:
**Description of the problem:**
getLanguageCode() function returns `"{"value":"en"}"` while I have device's language set to Polish on iPhone X.

**Affected platform**
- [ ] Android
- [x] iOS
- [ ] electron
- [ ] web

**OS of the development machine**
- [ ] Windows
- [x] macOS
- [ ] linux

**Capacitor version:**
`1.0.0-beta.17`
**node version:** 
`v11.10.0`
**npm version:**
`6.8.0`
**CocoaPods version:**
`1.6.0`

**Steps to reproduce:**
Set device language (probably to every other than English), in my case to Polish, and use this code.
```
import { Plugins } from '@capacitor/core';
const { Device } = Plugins;

...

    Device.getLanguageCode().then(lang => {
      console.log(lang);
    });
```

PR #1264, opened 2019-02-25
Fixed getLanguageCode()

maedewiza:
Fixed https://github.com/ionic-team/capacitor/issues/1233

Locale.current.languageCode **doesn't return the correct language** used by device, **becouse Capacitor didn't implement localization to project.**
After changing to Locale.preferredLanguages[0] it returns phone lang properly.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Device.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Device.swift
@@ -37,7 +37,7 @@ public class CAPDevicePlugin: CAPPlugin {
   }
   
   @objc func getLanguageCode(_ call: CAPPluginCall) {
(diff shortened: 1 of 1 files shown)


maedewiza:
Ok, done.

jcesarmobile:
navigator object is provided by the browsers, not by capacitor

it's navigator.language, not navigator.locale

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/language

sharmaa-nishantt:
@maeisdev, @jcesarmobile: Then navigator.language will not work for us. We need to fetch the device locale/language e.g: en-US. Is there any other way or alternate for the same?

also,
Do you any idea when this version <4.0.0 of the plugin (still in beta)> will be released?

Issue #1262, opened 2019-02-24
Crash while launching app: Swift class extensions and categories on Swift classes are not allowed to have +load methods (lldb) 


maedewiza:
**Description of the problem:**
The application compiles successfully but does not start. On start up, throws error:
`objc [1138]: Swift class extensions and categories on Swift classes are not allowed to have + load methods(Lldb).` 

**Affected platform**
- [ ] Android
- [x] iOS
- [ ] electron
- [ ] web

**OS of the development machine**
- [ ] Windows
- [x] macOS
- [ ] linux

**Other information:**
Tested without any plugins or changes.

Capacitor version:
1.0.0-beta.17

node version:
11.9.0

npm version:
6.8.0

CocoaPods version:
1.6.0

**Steps to reproduce:**
Use latest Xcode 10.2 beta 3 (10P99q) and build any app. Use latest iOS 12.2 to test it.

jcesarmobile:
This was fixed here https://github.com/ionic-team/capacitor/pull/1242

maedewiza:
I have replaced the contents of the CAPPluginCall.swift file with this pull request, which you sent and the error still occurs.

jcesarmobile:
Have you done a product -> clean build folder before running?  Xcode 10 doesn’t rebuild cocoa pods projects

maedewiza:
Yes, I done it. Please test on XCode 10.2 beta 3

maedewiza:
Maybe it could help:
[https://github.com/pmusolino/Wormholy/issues/51](url)
[https://stackoverflow.com/questions/54853209/swift-error-while-starting-app-after-update-to-xcode-10-2-beta-3](url)

jcesarmobile:
Sorry, you are right, this is a different issue

jratiwanich:
This crash is only happening when I deploy to a device with iOS 12.2. For an older phone or iOS 12.1 or lower, it won't crash. I'm getting this error below.


`objc[1919]: Swift class extensions and categories on Swift classes are not allowed to have +load methods`
`Message from debugger: Terminated due to signal 9`

>Capacitor 1.0.0-beta.17

exelban:
The similar crash on iOS 12.2-12.3.
On 12.1 or lower all works fine.

```
objc[1669]: Swift class extensions and categories on Swift classes are not allowed to have +load methods
(lldb) 
```

Capacitor 1.0.0-beta.19

grgur:
[...]

PR #1266, opened 2019-02-25
Remove +load method from Plugins

jcesarmobile:
Removed the + (void)load method from CAPBridgedPlugin as it causes problems on Xcode 10.2 beta 3 and iOS 12.2 beta.
Also some related code that looks unused. 

Closes https://github.com/ionic-team/capacitor/issues/1262


maedewiza:
This works, thanks!

Issue #1195, opened 2019-02-05
Move capacitor/core "@types/lodash": "^4.14.120" to devDependencies


timowolf:
It seems to me that "@types/lodash": "^4.14.120", should be a devDependencies in the capacitor/core package.json

Is it correct as a direct dependency? Do I need this during runtime?

Thanks a lot.

timowolf:
Thanks. This is great.

PR #1275, opened 2019-02-26
Move @types/lodash to devDependencies

maedewiza:
@types package is not needed during runtime.
Fixed https://github.com/ionic-team/capacitor/issues/1195


jcesarmobile:
Thanks for the PR, but while reviewing I've noticed that it's not used at all, so I've removed it.

PR #1283, opened 2019-03-01
Cleanup package.json and update dependencies

maedewiza:
1. @ionic/core from core is deprecated and never used, so I removed it
2. @types/tar from cli should be in devDependencies, so I moved it
3. Updated dependencies


jcesarmobile:
Thanks for the PR, but cli doesn't build with this dependencies. Updating packages, specially when there is a major bump, it usually also requires code changes.

jcesarmobile:
Going to close as the CLI doesn't build with this changes. Future PRs will run tests, which will make it easier to spot this kind of problems.

PR #1288, opened 2019-03-04
Warns the user for providing unknown commands

### LINKED ISSUE #1287: Warn the user for unknown commands
As per now `capacitor <random_command>` does nothing. I think it is good to have some sort of warnings such that the helps shows up along with a message, `Unknown command <random_command>.`

jamesgeorge007:
Closes #1287 
Shows up help and a suitable message if the user provides an unknown command.


jamesgeorge007:
@jcesarmobile What do you think?

PR #1292, opened 2019-03-06
feat(init): add an option for setting the webdir of a project.

jthoms1:
This pull request is to add an option to the init command of Capacitor. This option will make it possible to set the `webDir` setting in the capacitor config.  By default it is set to 'www'.

The default works for Angular projects but for React and Vue will require a different directory. By opening up this option it is now possible for the Ionic CLI to set this in the creation of the project.


jcesarmobile:
If I just pass `--web-dir` without a value, then `webDir` value is set to `true`. Is that expected?

PR #1313, opened 2019-03-12
pushNotificationActionPerformed on android

jurepurgar:
@jcesarmobile Would something like this be OK for solving  the issue with pushNotificationActionPerformed event not being fired on android? Am not sure about data that should be sent to JS.


solojuve1897:
When approved this should close the following: https://github.com/ionic-team/capacitor/issues/1181

solojuve1897:
@jurepurgar: There's one thing one can mention here. The data which is returned for Android differs from the one returned for iOS.

For iOS the object returned uses the PushNotificationActionPerformed-interface.

```
export interface PushNotificationActionPerformed {
  actionId: string;
  inputValue?: string;
  notificationRequest: any;
}
```
https://github.com/ionic-team/capacitor/blob/master/core/src/core-plugin-definitions.ts

jurepurgar:
@solojuve1897 I've updated the code so returned data is more compliant with the interface. However, the data still not contain the original notification (like on iOS), since I'm not sure how to get it on Android. I'm just passing everything I get from extras bundle to notificationRequest property. It seems that extras contain stuff from notification's data property.

Even it this gets merged there is still a problem that pushNotificationActionPerformed event is not fired it the app is closed (not in background) when notification is tapped. I'm not sure how to solve this.

@jcesarmobile This PR does not solve all of the issues regarding pushNotificationActionPerformed event but, I believe it does a step closer.

solojuve1897:
@jurepurgar Cool. I was thinking about this the toString()-method on value. If value is ever null it will throw an exception. Is it possible for it to be null?

jurepurgar:
@solojuve1897 I'm not sure if it is possible it is ever null, but I've added a check anyway.

jurepurgar:
Does anybody know why CI is failing for this PR?

jcesarmobile:
you probably need to rebase with master because when you sent the PR it was not configured to run tests, and now it complains about missing configuration

jcesarmobile:
Looks like you did something wrong on the rebase, all those commits from other people shouldn't appear here

jurepurgar:
Yes, sorry. Should I just make a new PR and close this one?

jcesarmobile:
Yeah, if you are not able to remove those commits better send a new PR

jurepurgar:
@jcesarmobile  I've created a new PR #1339 . Again, sorry for my lack of git knowledge :)

PR #1334, opened 2019-03-19
[Android] Fix crash when using mailto, tel, sms style urls

nate-eisner:


jcesarmobile:
I think it's simpler to just check in `matchhots`

Issue #1221, opened 2019-02-12
Push Notification issues


pegler:
While trying to implement push notifications I have encountered quite a few bugs and inconsistencies across platforms. I want to recap the different scenarios here along with the issues I've encountered.

Maybe I am approaching some of these scenarios incorrectly and there is a different way of handling it. I'm happy to work on pull requests for anything that is deemed a bug. I am fairly new to Capacitor so I will need some guidance and feedback on the changes.

### App in foreground

The PushNotification.pushNotificationReceived event is correctly sent.

Issues:
 - The alert is always shown. https://github.com/ionic-team/capacitor/pull/1201/commits/72ad0c52f7b38da1f39d7f2d5fe704b88026a6a5
 - The data from the notification is not passed to the listener via the PushNotification object: https://github.com/ionic-team/capacitor/pull/1201

### App in background and notification tapped

The PushNotification.pushNotificationActionPerformed event is correctly sent.

Issues:
 - The data from the notification is not passed to the listener via the PushNotification object: https://github.com/ionic-team/capacitor/pull/1201

### App closed and notification tapped

Issues:
 - The PushNotification.pushNotificationActionPerformed event is sent before the listeners can be added: https://github.com/ionic-team/capacitor/pull/1220

### App in foreground

The PushNotification.pushNotificationReceived event is correctly sent.

Issues:
 - Notification title and body are not passed via the PushNotification object: https://github.com/ionic-team/capacitor/pull/1201

### App in background and notification tapped

The app is opened but I do not see a way of detecting this scenario.

Issues:
 - PushNotifications.pushNotificationActionPerformed event is not sent. While debugging, it looks like the LocalNotifications.localNotificationActionPerformed event is sent when the push notification is tapped.

### App closed and notification tapped

The app is opened but I do not see a way of detecting this scenario.

Issues:
 - PushNotifications.pushNotificationActionPerformed event is not sent. Unlike scenario above, LocalNotifications.localNotificationActionPerformed does not appear to be sent.

---

Affected platform
- [x] Android
- [x] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [ ] Windows
- [x] macOS
- [ ] linux

Capacitor version:
1.0.0. beta 17

solojuve1897:
Good report. All Android-specific I also have encountered.

denisroldan:
Waiting for this to be fixed too. Thanks for reporting!

pegler:
[...]

Issue #971, opened 2018-11-15
Android: Deeplinks -  App opens twice sometimes


Nasicus:
I'm not sure if this is capacitors fault, but I had quite the trouble with this.

Basically what happens is this:

If you have the app already open and click on a deep link in another app which supports inline-browsers (like Gmail) the app will be opened a second time.
You can see this when you look at your recent app list.
That in itself might not even be that bad. However for us our plugins respectively the callbacks of the plugins don't work correctly anymore.  Both app instances will just have a blank screen and be unusable. You have to close them again. But even then: The deeplink will never work correctly.

The solution for this is quite simple:

You have to set the `launchMode` on the `MainActivity` to `singleTask`.

```
android:launchMode="singleTask"
```

See this [stack ](https://stackoverflow.com/a/25997627/1712056) answer.

It may be a good idea to at least update the "sample app" (which you can create by using capacitor) to include this `launchMode`. 

As I said: I'm not sure if there is anything really wrong in capacitor or if this is the expected behavior.

jcesarmobile:
Closing as this PR https://github.com/ionic-team/capacitor/pull/1477 set the template launchMode to singleTask

KovshefulCoder:
Thanks, helped even with Jetpack Compose and Navigaton (navigation:navigation-compose)

PR #1349, opened 2019-03-22
Android pushNotificationActionPerformed 

jurepurgar:
This PR does not solve all of the issues regarding pushNotificationActionPerformed event as described in #1221  and #1181 but, I believe it is a step closer, and at least gets people a little unblocked.

App launchMode must be set to singleTask or handleOnNewIntent wont be called as described in #971 

The returned data does not contain the original notification (like on iOS), since I'm not sure how to get it on Android. I'm just passing everything I get from extras bundle to notificationRequest property. It seems that extras contain stuff from notification's data property.

Even it this gets merged there is still a problem that pushNotificationActionPerformed event is not fired if the app is closed (not in background) when notification is tapped. I'm not sure how to solve this.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -3,6 +3,8 @@ package com.getcapacitor.plugin;
 import android.app.NotificationChannel;
 import android.app.NotificationManager;
 import android.content.Context;
+import android.content.Intent;
+import android.os.Bundle;
 import android.support.v4.app.NotificationCompat;
 import android.net.Uri;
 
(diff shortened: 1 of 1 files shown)


jurepurgar:
@jcesarmobile Sample payload is:

`
{
  "notification": {
    "title": "Test",
    "body": "Test"
  },
  "data": {
    "customProp1": "test",
    "customProp2": "test2"
  }
}
`

jcesarmobile:
BTW, it's working fine even without setting the launch mode to singleTask.
Also works fine even if the app was fully closed, probably related to this changes I merged on Friday https://github.com/ionic-team/capacitor/pull/1347

The part about not being able to get the notification, that's an Android limitation, you can only get the notification part when the app is open, when in background or closed you can only get the the data part.

jurepurgar:
Thanks. Great to hear it is working even if the app is initially closed.

PR #1359, opened 2019-03-26
Add timestamp and altitudeAccuracy to native geolocation plugins

sbannigan:
This adds the `timestamp` and `altitudeAccuracy` properties to the built in Geolocation plugin to match the [current web API for geolocation](https://developer.mozilla.org/en-US/docs/Web/API/Position), as well as the equivalent [Cordova plugin](https://github.com/apache/cordova-plugin-geolocation).

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Geolocation.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Geolocation.swift
@@ -63,7 +63,7 @@ class GetLocationHandler: NSObject, CLLocationManagerDelegate {
     coords["altitudeAccuracy"] = location.verticalAccuracy
     coords["speed"] = location.speed
     coords["heading"] = location.course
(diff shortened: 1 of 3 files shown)


ios/Capacitor/Capacitor/Plugins/Geolocation.swift
```diff
@@ -60,8 +60,10 @@ class GetLocationHandler: NSObject, CLLocationManagerDelegate {
     coords["longitude"] = location.coordinate.longitude
     coords["accuracy"] = location.horizontalAccuracy
     coords["altitude"] = location.altitude
+    coords["altitudeAccuracy"] = location.verticalAccuracy
     coords["speed"] = location.speed
     coords["heading"] = location.course
+    ret["timestamp"] = NSNumber(value: Double((location.timestamp.timeIntervalSince1970 * 1000)))
```
jcesarmobile: This is returning 2 decimals, better use Int instead of Double, or round it

PR #1361, opened 2019-03-26
Update plugins.md

stewwan:
add capacitor-fcm


stewwan:
Screen Shot 2019-03-27 at 14 58 08

does this count as analytic?

jcesarmobile:
that's not very helpful as there is only one native view and users will be almost always there, it should contain real analytics methods to be added to the analytics section

Issue #1294, opened 2019-03-07
[iOS] Vibration from push notification comes through even though app is active


solojuve1897:
Description of the problem:
Using Beta. 18. When app is in active state (i.e. in the foreground) push notifications doesn't show anymore - as expected - but it still vibrates when the notification was supposed to have arrived which is strange and unexpected. This is not the current behaviour on Android where if the app is in the inactive state the push notification will be visibile and carry with it a sound, and if it is in the active state nothing is either heard or seen.

If relevant here is my iOS-payload:
```
{
 "aps": {
   "sound": "default",
   "alert": {
     "title": "test title",
     "body": "test body"
   }
 }
}
```
If I don't include the "sound"-property there will be no vibration, either the app is active or not.

And here's my Android-payload:
`{"notification":{"title":"test title", "body":"test body"}}`

Not sure if the issue is mine and related to the payload or with the Push Notification plugin of Capacitor.

Affected platform
- [ ] Android
- [x] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [ ] Windows
- [x] macOS
- [ ] linux

Other information:

Capacitor version:
Beta. 18
node version:

npm version:

CocoaPods version:

Steps to reproduce:

Link to sample project:

PR #1372, opened 2019-03-29
[iOS]  Don't vibrate when pushnotification is recieved if app in foreground

solojuve1897:
Closes:
https://github.com/ionic-team/capacitor/issues/1294


jcesarmobile:
We should make this configurable, you don't want the vibration, but other users might. Same thing for the badge?

solojuve1897:
@jcesarmobile Yes, you may be correct. What's seems strange for me maybe isn't strange for others. But if I'm not mistaken on Android it doesn't vibrate at the moment when app is in foreground. With this PR we will at least be consistent over the platforms and later on can introduce the ability to configure these options.

jcesarmobile:
I've sent a PR that makes it configurable, by default only the badge will be set when the app is in foreground, will close this one once it's merged.

PR #1403, opened 2019-04-09
Notifications crash fix and handled extra argument value

msepena:
Issue: 

1. After receiving Push/Local Notifications app crashing immediately.
2. In iOS also handled the extra parameter which is not receiving back when notification fired.


jcesarmobile:
Can you provide a sample app or sample javascript of the LocalNotification you schedule that makes the app crash? I've not seen that on any of my tests.

Also, the extra is not sent on purpose

PR #1417, opened 2019-04-18
1381 - PR1128 broke my use case AKA star mask in `server.allowNavigation` 

### LINKED ISSUE #1381: PR#1128 broke my use case
Description of the problem: `Capacitor` object is undefined when loading the external URL with `allowNavigation`.

Affected platform
- [x] Android

OS of the development machine
- [x] macOS

Capacitor version: `1.0.0-beta.19`

I tried to upgrade my application from `1.0.0-beta11` to `1.0.0-beta19` and things get broken on Android. In Chrome console, I can see the `Capacitor` object and its methods, but it's undefined when I try to invoke any (such as `isAndroid`).

While I understand #1128 is a great improvement, I'm not sure what should I do on my end to fix the issue. Reverting #1128 resolves it, but there should be a better way.

My config is as follows: `server.allowNavigation` contains a list of allowed URLs, then in `www/index.html` the only code is `location.href = 'https://one-of-allowed-urls-here'`.

@jcesarmobile could you please advice on what can I try?

### LINKED ISSUE #1389: Capacitor code not injected to external websites when using * in allowNavigation
Description of the problem:
When using * in allowNavigation, the Capacitor javascript is not injected to the external page
Create an app that just redirects to https://www.google.es (or your country one) and add
```
"server": {
    "allowNavigation": [
      "*.google.es"
    ]
  }
```
Check Capacitor object and it won't be there.

Change to

```
"server": {
    "allowNavigation": [
      "www.google.es"
    ]
  }
```
and Capacitor object should be there

Affected platform
- [x] Android
- [ ] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [x] Windows
- [x] macOS
- [x] linux

Other information:

Capacitor version: beta 19

node version:

npm version:

CocoaPods version:

Steps to reproduce:

Link to sample project:

pokebadgerswithspoon:
Makes it possible use star masks in `server.allowNavigation` option of `capacitor.config.json`.
Examples:
- `*.example.com` would match `app.example.com`, `www.example.com` but not `example.com`
- `192.168.*.*` is valid

What the changes are:
- `HostMask` is a matcher, it parses text into masks and check strings against it `boolean matches(String host)`, i.e. `HostMask.Parser.parse("192.168.*.*).matches("192.168.1.1")` would return `true`
- `HostMaskTest` is a test suite for the above, more mask examples are inside
- `UriMatcher` was tought to use `HostMask` for the host part
- `Bridge` contained some matching of its own, that was replaced with use of `HostMask`

Closes #1381 
Closes #1389

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -9,6 +9,7 @@ import android.net.Uri;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
+import android.util.Base64;
 import android.util.Log;
 import android.webkit.ValueCallback;
 import android.webkit.WebResourceRequest;
@@ -41,6 +42,7 @@ import com.getcapacitor.plugin.StatusBar;
 import com.getcapacitor.plugin.Storage;
 import com.getcapacitor.plugin.background.BackgroundTask;
 import com.getcapacitor.ui.Toast;
+import com.getcapacitor.util.HostMask;
 
 import org.apache.cordova.CordovaInterfaceImpl;
 import org.apache.cordova.PluginManager;
@@ -96,7 +98,7 @@ public class Bridge {
   private String localUrl;
   private String appUrl;
   private String appUrlConfig;
-  private String[] appAllowNavigationConfig;
+  private HostMask appAllowNavigationMask;
   // A reference to the main WebView for the app
   private final WebView webView;
   public final CordovaInterfaceImpl cordovaInterface;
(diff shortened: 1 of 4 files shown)


leshik:
@mlynch @jcesarmobile any objections? we need this badly.. :)

Venryx:
For anyone else who hits the issue I did and finds this page, just wanted to mention: the URI matcher does _not_ consider the port segment to be a separate segment.

Example: For the URI `192.168.0.31:4001`, it counts `31:4001` as one segment. (ie. it splits only on `.`, not `:`)

So to match it, you must use `192.168.*.*`, rather than `192.168.*.*:*`.

PR #1427, opened 2019-04-24
Docs: Added new Guides section & Ionic Framework app guide

dotNetkow:
"Bonus": includes a few small HTML page updates that were likely missed last time Master branch was updated.


site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+npm install -g ionic
+```
+
+Next, create a new Ionic app based on the "tabs" starter project:
+
+```bash
+ionic start capApp tabs
+```
+
+## Add Capacitor
```
jcesarmobile: Ionic CLI has Capacitor integrations, so as you are starting with Ionic CLI, might be better to continue using it to add Capacitor, unless you want to focus on Capacitor CLI usage instead of Ionic CLI.
Example of starting the app with Capacitor already installed `ionic start capApp tabs --capacitor`
Example of adding Capacitor to a regular Ionic app you already created `ionic integrations enable capacitor`

Both of them do the npm install for core and cli and the init.

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+
+Next, implement the `takePicture()` method:
+
+```typescript
+export class Tab2Page {
+  async takePicture() {
+    const image = await Plugins.Camera.getPhoto({
+      quality: 100,
+      allowEditing: false,
+      resultType: CameraResultType.Base64,
```
jcesarmobile: This is changing on next release, base64 will be just the base64 string, you'll want `CameraResultType.DataUrl` instead.

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+      source: CameraSource.Camera
+    });
+
+    this.photo = this.sanitizer.bypassSecurityTrustResourceUrl(image && (image.base64Data));
```
jcesarmobile: After changing the resultType  to `CameraResultType.DataUrl`, you'll get it with `image.dataUrl` instead of base64Data.

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+export class Tab2Page {
+  photo: SafeResourceUrl;
+
+  constructor(private sanitizer: DomSanitizer) {  }
+
+  async takePicture() {
+    const image = await Plugins.Camera.getPhoto({
+      quality: 100,
+      allowEditing: false,
+      resultType: CameraResultType.Base64,
```
jcesarmobile: Same as before, use `CameraResultType.DataUrl`.

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+      source: CameraSource.Camera
+    });
+
+    this.photo = this.sanitizer.bypassSecurityTrustResourceUrl(image && (image.base64Data));
```
jcesarmobile: Same as before, use `image.dataUrl`

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+
+![Camera API on the web](/assets/img/docs/guides/ionic-framework/camera-web.png)
+*"I'm pretty handsome." - The Author*
+
+Next up, let's run this app on iOS and Android.
+
+## iOS
+Capacitor iOS apps are configured and managed through Xcode, with dependencies managed by CocoaPods. Before running this app on an iOS device, there's a couple of steps to complete.
+
+From the Terminal, run the Capacitor `sync` command, which updates all native code dependencies as well as copies all web assets (the Ionic Angular app in this case) into the native iOS project:
```
jcesarmobile: Sync is only needed when you install new plugins, so use copy instead to just copy the app assets (www folder), it's way faster.

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+
+```shell
+npx cap open ios
+```
+
+Within Xcode, click on `App` in the Project Navigator on the left-hand side, then within the `Signing` section, select your Development Team. 
+
+![Xcode - Selecting Development Team](/assets/img/docs/guides/ionic-framework/xcode-signing.png)
+
+Next, for the Camera plugin to work, we must configure the "Privacy - Camera Usage" permission. iOS displays a modal dialog automatically after the first time that `Camera.getPhoto()` is called, prompting the user to allow the app to use the Camera. To set this, the `Info.plist` file must be modified ([more details here](../ios/configuration/)). To access it, click "Info," then expand "Custom iOS Target Properties."
```
jcesarmobile: We already add a default text, so it's not really needed to add it, just change it in case you don't like it

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+![Xcode build button](/assets/img/docs/guides/ionic-framework/xcode-build-button.png)
+
+Upon tapping the Camera button on Tab Two, the permission prompt will display. Tap OK, then take a picture with the Camera. Afterward, the photo shows in the app:
+
+![iOS Camera permissions](/assets/img/docs/guides/ionic-framework/ios-permissions-photo.png)
+
+## Android
+Capacitor Android apps are configured and managed through Android Studio. Before running this app on an Android device, there's a couple of steps to complete.
+
+From the Terminal, run the Capacitor `sync` command, which updates all native code dependencies as well as copies all web assets (the Ionic Angular app in this case) to the native Android project:
```
jcesarmobile: Same again, use copy instead of sync as you didn't install any plugin and it's faster (sync = update + copy)

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+npx cap sync
+```
+
+Next, run the Capacitor `open` command, which opens the native Android project in Android Studio:
+
+```shell
+npx cap open android
+```
+
+Similar to iOS, we must enable the correct permissions to use the Camera. Configure these in the `AndroidManifest.xml` file. Android Studio will likely open this file automatically, but in case it doesn't, locate it under `android/app/src/main/`.
```
jcesarmobile: Same as on iOS, Capacitor already includes all the required permissions, and in this case user don't even have to configure a text, so not sure if it's worth commenting.

site/docs-md/guides/ionic-framework-app.md
```diff
@@ -0,0 +1,250 @@
...
+
+![Android Manifest location](/assets/img/docs/guides/ionic-framework/android-manifest.png)
+
+Scroll to the `Permissions` section and ensure these entries are included:
+
+```xml
+<uses-permission android:name="android.permission.CAMERA" />
```
jcesarmobile: Despite it sounds illogical, CAMERA permission is not needed for camera plugin

site/docs-md/guides/ionic-framework-app.md
dotNetkow: excellent! didn't know that. will tweak

site/docs-md/guides/ionic-framework-app.md
dotNetkow: yep, noted. Will update later

site/docs-md/guides/ionic-framework-app.md
dotNetkow: yep.  the idea is to get the developer following this to be more comfortable working in Xcode/making changes.

site/docs-md/guides/ionic-framework-app.md
dotNetkow: haha. I took that from here, so should we change? https://capacitor.ionicframework.com/docs/apis/camera

dotNetkow: Removed that line from this PR

PR #1442, opened 2019-04-29
Added notification haptics (iOS)

maedewiza:
According to the iOS Feedback's specification (https://developer.apple.com/design/human-interface-guidelines/ios/user-interaction/feedback/), I've added a notification type feedback that has not been implemented yet.

@jcesarmobile

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Haptics.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Haptics.swift
@@ -23,6 +23,25 @@ public class CAPHapticsPlugin: CAPPlugin {
       }
     }
   }
+
+  @objc public func notification(_ call: CAPPluginCall) {
+      DispatchQueue.main.async {
+          if let type = call.options["type"] as? String {
+              var notificationType = UINotificationFeedbackGenerator.FeedbackType.success
+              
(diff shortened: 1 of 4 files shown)


ios/Capacitor/Capacitor/Plugins/Haptics.swift
```diff
@@ -23,6 +23,25 @@ public class CAPHapticsPlugin: CAPPlugin {
...
+      DispatchQueue.main.async {
+          if let type = call.options["type"] as? String {
+              var notificationType = UINotificationFeedbackGenerator.FeedbackType.success
+              
+              if type == "WARNING" {
+                  notificationType = UINotificationFeedbackGenerator.FeedbackType.warning
+              } else if type == "ERROR" {
+                  notificationType = UINotificationFeedbackGenerator.FeedbackType.error
+              }
+              let generator = UINotificationFeedbackGenerator()
```
jcesarmobile: Instead of having a duplicated generator, you can put it before the first if and use the same on both cases

maedewiza: done

maedewiza:
done

PR #1461, opened 2019-05-02
Android 4.1 support via Crosswalk

LeviticusMB:
So I don't *really* expect you to merge this, even though it would be fantastic, but ... This PR makes Capacitor run on Android 4.1+ (API v16) via the *Crosswalk* `WebView` (whether to use the system or Crosswalk `WebView` can be selected at run-time, for instance by comparing the version numbers).

We would like to move away from Cordova, but are not yet ready to leave our customers using legacy devices behind. I can imagine that quite a lot of others are in the same situation and this could at least be a stop-gap solution.

The patch is not *that* invasive, although it does add a compile-time dependency on the shared Crosswalk AAR from the `download.01.org` Maven repo.


jcesarmobile:
Thanks for the PR, but considering crosswalk was abandoned 2 years ago we are not going to merge this.
Would be cool if you could turn this into a plugin somehow, or we are open to make modifications to make it possible to use your own webview as in this PR or if you need any change to make it work as a plugin.

https://github.com/ionic-team/capacitor/pull/438

PR #1470, opened 2019-05-05
[Docs] Camera API docs refresh

dotNetkow:
- Update Angular code example - remove unnecessary code
- Add link to Ionic Angular camera guide
- Mention that for Android, CAMERA permission is actually not required (via Julio), so as to avoid confusion.
- Small formatting changes

### DIFF
--- a/site/docs-md/apis/camera/index.md
+++ b/site/docs-md/apis/camera/index.md
@@ -25,7 +25,7 @@ This API requires the following permissions be added to your `AndroidManifest.xm
 <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 ```
 
-The storage permissions are for reading/saving photo files. Note that `android.permission.CAMERA` is not required.
+The storage permissions are for reading/saving photo files.
 
 Read about [Setting Android Permissions](../../android/configuration/) in the [Android Guide](../../android/) for more information on setting Android permissions.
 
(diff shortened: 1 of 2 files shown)


site/docs-md/apis/camera/index.md
```diff
@@ -25,7 +25,7 @@ This API requires the following permissions be added to your `AndroidManifest.xm
 <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
 ```
 
-The first permission is for Camera access, and the storage permissions are for reading/saving photo files.
+The storage permissions are for reading/saving photo files. Note that `android.permission.CAMERA` is not required.
```
jcesarmobile: I removed the CAMERA permission the other day, so not sure if this note makes sense

site/docs-md/apis/camera/index.md
```diff
@@ -88,8 +89,8 @@ export class CameraPage {
       source: CameraSource.Camera
     });
     
-    // Example of using the Base64 return type. It's recommended to use CameraResultType.Uri
-    // instead for performance reasons when showing large, or a large amount of images.
+    // Recommended to use CameraResultType.DataUrl instead of Base64 for
```
jcesarmobile: DataUrl is as bad as Base64, recommended is Uri

site/docs-md/apis/camera/index.md
dotNetkow: added it on the idea that people will be surprised that Camera permission isn't needed. I will remove

dotNetkow:
> Added two comments.
> But does it make sense to keep the Angular example if we have the guide that uses Angular as it's more complete?

Good point, updated!

site/docs-md/apis/camera/index.md
dotNetkow: Updated

site/docs-md/apis/camera/index.md
dotNetkow: Maybe it would confuse them after all :)

jcesarmobile: Yeah, I forgot to remove that part, so that part of your change was OK, just not the note

Issue #1475, opened 2019-05-07
[Android] tapping push notification does nothing when app not running or in background


CFT-Chris:
Description of the problem:

Affected platform
- [x] Android
- [ ] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [x] Windows
- [ ] macOS
- [ ] linux

Other information: This is the same as issue #1192, which was closed under the impression it was fixed with the latest beta (22). It still reproduces for me, but I have investigated it and am putting in a PR very soon.

Capacitor version: beta.22

node version: 10.13

npm version: 6.4

CocoaPods version: n/a

Steps to reproduce:

Send Firebase (FCM) message to Capacitor app not running in foreground. Tap on notification in Android's notification panel.

Either nothing happens (app is not started or background app not brought to foreground), or the app is started up without sending the notification data to any of its listeners `pushNotificationReceived` or `pushNotificationActionPerformed`.

Link to sample project: n/a

jcesarmobile:
Can you show your notification payload? It works fine for me without all the changes in your PR

CFT-Chris:
Thanks for the quick reply. 

I am testing on Android 5.0 (minimum required by Capacitor).

I tried all payload combinations.  Without "notification", then without "data", with both "notification" and "data", with "click_event" and without, etc.

Just an example of what I am sending (via Node.js backend POST request to Firebase cloud) which works with the changes I proposed in the PR:
```
request({
  url: 'https://fcm.googleapis.com/fcm/send',
  method: 'POST',
  headers: {
    'Content-Type' :' application/json',
    'Authorization': 'key=' + /* FCM key */
  },
  body: JSON.stringify({
    to: / *** PushReg Id ***/,
    data: { tm: Date.now() },
    notification: {
      sound: 'default',
      click_action: "CAPACITOR_FIREBASE_MESSAGING_ACTIVITY"
    }
  }
})
```

Without "notification" the notification doesn't even arrive at all if the app is not running (background or killed).

Just looking at the code itself, I can't figure out how Android ends up calling [handleOnNewIntent in PushNotifications.java](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java#L59) as there is nothing in the AndroidManifests, apart from `CapacitorFirebaseMessagingService`, that might deal with incoming Firebase messages when the app is not running in the foreground.

[...]

Issue #1192, opened 2019-02-04
I am not able to differ background push notification from foreground push notification


mateusduraes:
Hello.

I have the plugin implementation just like the following.

```typescript
PushNotifications.register();

PushNotifications.addListener('registration', (token: PushNotificationToken) => {
  // registering token on my backend
});

PushNotifications.addListener('registrationError', (error: any) => {
  console.log('error on register ', JSON.stringify(error));
});

PushNotifications.addListener('pushNotificationReceived', (notification: PushNotification) => {
  // receive push notification (opened application)
  console.log('notification', notification.data);
});

PushNotifications.addListener('pushNotificationActionPerformed', (notification: PushNotificationActionPerformed) => {
  // never called
  console.log('notification action performed', notification);
});
```

When i receive the push notification with the app opened (foreground), my console.log is executed as expected.
But, if the app is closed, when i tap the notification, the app become open, but none of ther listeners are fired.

Am i doing something wrong?

LucasPMM:
I am having the same problem.

mateusduraes:
Oh yes, i am sorry. Im using the exact same versions.

mateusduraes:
I believe it is related to https://github.com/ionic-team/capacitor/issues/1178

jcesarmobile:
should be fixed in beta 22

LucasPMM:
Was solved? I don't think so..

FinnBossen:
@oliverandersencox @gusmarra 
Have you solved the Problem I have the same Issue. Would be nice if you have a Solution

oliverandersencox:
@FinnBossen yes. If you add this to your capacitor config file:

`"plugins": {
  "PushNotifications": {
    "presentationOptions": ["badge", "sound", "alert"]
  }
}`

then remove the options you wish to have disabled for foreground notifications. 

https://capacitor.ionicframework.com/docs/apis/push-notifications

gusmarra:
@FinnBossen the only thing I had to do was to update the capacitor version. I was not using the last version and it did not support this.

PR #1476, opened 2019-05-07
fix(android): Push notifications activity

CFT-Chris:
#### Short description of what this resolves:
For Android, handle intent of tapped push notification when app is in background or not running.

#### Changes proposed in this pull request:
- Come up with a click_action, I named it `CAPACITOR_FIREBASE_MESSAGING_ACTIVITY`, so that Android knows what to tell the app if a user taps on an unhandled Firebase notification in the notification panel.
- Via [androidManifest.xml](https://github.com/ionic-team/capacitor/compare/master...CFT-Chris:push-notifications-activity?expand=1#diff-6128f7864c486ab3743bfd0eeb0f59ce), tell Android to let `CapacitorFirebaseMessagingActivity` handle the new click_action.
- Try to send the Intent to the app running in the background, or otherwise save the Intent (similar to how a RemoteMessage is saved if the app is running in the foreground) within the [PushNotifications](https://github.com/ionic-team/capacitor/compare/master...CFT-Chris:push-notifications-activity?expand=1#diff-4dac91fa926404ed754bb98f291de6ee) module. Once a `pushNotificationActionPerformed` listener is added, the most recent saved Intent can be sent to it.
- Make sure to bring the app from the background to the foreground (if running), or start the app if it's not running at all. Because of the call to [startActivity](https://github.com/ionic-team/capacitor/compare/master...CFT-Chris:push-notifications-activity?expand=1#diff-fb95a009fc46e610d856022923a35dbc) with the app's launch intent, make sure the [launchMode](https://github.com/ionic-team/capacitor/compare/master...CFT-Chris:push-notifications-activity?expand=1#diff-2f26da12130dc2e0f9a6e8ee6f4e13b6) for `MainActivity` is set to `singleInstance` so it doesn't end up calling `BridgeActivity.init()` more than once if the app was running in the background when the notification was tapped. Without `singleInstance`, the Ionic webview will be recreated and multiple webviews will be opened within the app.

**Capacitor Version**: 1.0.0-beta.22

**Fixes**: #1475 (originally #1192)

### DIFF
--- a/android-template/app/src/main/AndroidManifest.xml
+++ b/android-template/app/src/main/AndroidManifest.xml
@@ -14,7 +14,8 @@
             android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale"
             android:name="com.getcapacitor.myapp.MainActivity"
             android:label="@string/title_activity_main"
-            android:theme="@style/AppTheme.NoActionBarLaunch">
+            android:theme="@style/AppTheme.NoActionBarLaunch"
+            android:launchMode="singleInstance">
 
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
--- a/android-template/app/src/main/AndroidManifest.xml
+++ b/android-template/app/src/main/AndroidManifest.xml
@@ -15,7 +15,7 @@
             android:name="com.getcapacitor.myapp.MainActivity"
             android:label="@string/title_activity_main"
             android:theme="@style/AppTheme.NoActionBarLaunch"
-            android:launchMode="singleInstance">
+            android:launchMode="singleTask">
 
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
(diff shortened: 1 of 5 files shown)


solojuve1897:
Nice to see this issue about the be solved once and for all. Thanx! One question: Why singleInstance instead of singleTask?

CFT-Chris:
> Nice to see this issue about to be solved once and for all. Thanx! One question: Why singleInstance instead of singleTask?

Based on what it says [here](https://developer.android.com/guide/topics/manifest/activity-element.html#lmode) I didn't want the launch Intent to be handled more than once, as I thought that was what was creating multiple webviews. But then I just contradicted what I wrote in my initial PR comment that BridgeActivity init() (and load()) were spawning the new views. 

I believe `singleTask` would work fine, though I don't know of any other reasons why MainActivity would need to be sent (and handle) another Intent again which makes `singleInstance` fine as well, I would imagine? I could also just be misinterpreting the spec or misunderstanding something.

CFT-Chris:
Closing this one as changes are redundant. As long as a click_action is **not** defined in the FCM notification part of the payload, and a notification part does exist, then there is no need for any of these changes. The Intent is routed through the MainActivity to the PushNotification plugin instead.  

Per the discussion in #1475, the bugs I still see can be resolved simply by setting the launchMode to `singleTop` (anything but `standard`), which I can submit in a new PR.

PR #1477, opened 2019-05-08
fix(android): change MainActivity launchMode

CFT-Chris:
#### Short description of what this resolves:
For Android, tapping a push notification when app is in background recreates MainActivity, which may cause `pushNotificationActionPerformed` to not be fired on the expected Ionic view. Multiple webviews sometimes appear, but more importantly it will always appear that the app is restarting. It should just bring up the app from the background to the foreground and handle the tapped notification data immediately.

#### Changes proposed in this pull request:
If launchMode is not defined for MainActivity, then by default according to the [spec](https://developer.android.com/guide/topics/manifest/activity-element.html#lmode): "the system creates a new instance of the activity in the target task and routes the intent to it". I don't think that is the desired behaviour here, particularly given all the init/load code in BridgeActivity that will get fired every time a notification is tapped (or some other Intent gets sent its way).

**Capacitor Version**: 1.0.0-beta.22

**Fixes**: #1475, specifically what I wrote in [this long-winded comment](https://github.com/ionic-team/capacitor/issues/1475#issuecomment-490338020).


jcesarmobile:
Thanks!

PR #1481, opened 2019-05-08
All async

manucorporat:


core/src/definitions.ts
```diff
@@ -1,6 +1,4 @@
-declare global {
-  export interface PluginRegistry {}
-}
+import { PluginRegistry } from './core-plugin-definitions';
 
 export interface Plugin {
   addListener(eventName: string, listenerFunc: Function): PluginListenerHandle;
```
jcesarmobile: this should also return a Promise, right?

manucorporat:
Bad idea for today :)

PR #1483, opened 2019-05-09
Test if tests fail

jcesarmobile:
Tests should fail


jcesarmobile:
failed as expected, closing

PR #1510, opened 2019-05-15
feat(local-notifications): use app icon for default

mburger81:
waiting for custom icons we suggest to use at least default application icon before using ic_dialog_info

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotification.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotification.java
@@ -179,7 +179,13 @@ public class LocalNotification {
 
   public int getSmallIcon(Context context) {
     // TODO support custom icons
-    int resId = android.R.drawable.ic_dialog_info;
+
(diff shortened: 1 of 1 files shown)


jcesarmobile:
the problem with using app icon is, most of them are not valid notification icons and will just show a white circle or square

mburger81:
On what is this depending? The icon it self, or the android platform? In our test case the app icon is simple better then the default info icon.

We could also do a PR for a custom small icon like cordova does. But we are not sure where should we should configure it? 
Do you want it As I a dedicated main property or in the extra property? For what is the extra property? There is no documentation about that. 
We are not native programmers so sorry if this this is a "stupid" question

RoderickQiu:
> 
> 
> the problem with using app icon is, most of them are not valid notification icons and will just show a white circle or square

But on my own project, it shows the app's icon, if action bar icons are not set. (After using the code)

So it should be OK to merge that code?

RoderickQiu:
However, I think request icon as a parameter seems to be better. 
But I'm not experienced in native coding...

PR #1513, opened 2019-05-16
WIP feat(local-notifications): add Inbox and BigText Style for android

mburger81:
This is a suggestion to use same behavior as cordova plugin to handle android InboxStyle and BigTextStyle.
I would like to ask you, how do you would like to implement features like this and we will try to do the implementation. Or at least we'll try.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
@@ -166,6 +166,28 @@ public class LocalNotificationManager {
 
     mBuilder.setSmallIcon(localNotification.getSmallIcon(context));
     createActionIntents(localNotification, mBuilder);
+    
+    String text = localNotification.getBody();
+
+    if (text != null) {
+      if (text.contains("\n")) {
+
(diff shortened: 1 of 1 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
```diff
@@ -166,6 +166,28 @@ private void buildNotification(NotificationManagerCompat notificationManager, Lo
...
+        for (String line : text.split("\n")) {
+          style.addLine(line);
+        }
+
+        mBuilder.setStyle(style);
+
+
+      } else {
+
+        mBuilder.setStyle(new NotificationCompat.BigTextStyle().bigText(localNotification.getBigText()));
```
wslaghekke: CI is failing because of missing method localNotification.getBigText()
should it be the following?:
mBuilder.setStyle(new NotificationCompat.BigTextStyle().bigText(text));

dwieeb:
@mburger81 Thanks for the PR! I think we'd rather lean towards explicit behavior. Additionally, we're about to begin work on Capacitor 3 which will involve a lot of refactoring, so let's wait on adding this functionality. Local and push notifications may exist separate from the Capacitor core in an official plugin.

PR #1527, opened 2019-05-18
Removes npm-scope from package name.

parveenkhtkr:
`npx cap update` generates wrong pod name when the package is using npm-scope, causing a mismatch and failing `npx cap update` as generated plugin **podspec** filename does not include the scope name.

A plugin package named `@codetrixstudio/capacitor-google-auth`, generates pod named `CodetrixstudioCapacitorGoogleAuth` instead of `CapacitorGoogleAuth`

This fixes the issue by removing npm-scope from the package name.

### DIFF
--- a/cli/src/plugin.ts
+++ b/cli/src/plugin.ts
@@ -90,6 +90,7 @@ export function getDependencies(config: Config): string[] {
 
 export function fixName(name: string): string {
   name = name
+    .replace(/^@[^/]*\//g, '')
     .replace(/\//g, '_')
     .replace(/-/g, '_')
     .replace(/@/g, '')


mlynch:
Thanks! I vaguely remember a similar issue and PR that we ended up reverting, so will see what @jcesarmobile thinks before merging

jcesarmobile:
This is not an issue, the `CodetrixstudioCapacitorGoogleAuth` name is good and your .podspec should be called like that. 
There was a bug on the plugin generator that generated the .podspec without the scope, but it's fixed now.
Plugins that were created without the scope will need to be updated by renaming the .podspec and the name inside the file too. Plugins that were created without scope but were updated to use scope will also need to update the same way.

We decided to keep the scope to avoid name collisions.
In example, there is already a `@byteowls/capacitor-sms` and `capacitor-sms`, if we ignore the scope they will collide.

parveenkhtkr:
Makes sense, wasn’t sure which fix is preferable 👍🏻

PR #1538, opened 2019-05-21
[Docs] Updating API supported platforms

dotNetkow:
Went through [web folder](https://github.com/ionic-team/capacitor/tree/master/core/src/web) and [electron folder](https://github.com/ionic-team/capacitor/tree/master/electron/src/electron) and updated each API's supported platforms.

### DIFF
--- a/site/docs-md/apis/accessibility/index.md
+++ b/site/docs-md/apis/accessibility/index.md
@@ -1,4 +1,4 @@
-<plugin-platforms platforms="ios,android"></plugin-platforms>
+<plugin-platforms platforms="pwa,ios,android"></plugin-platforms>
 
 # Accessibility
 
--- a/site/docs-md/apis/accessibility/index.md
+++ b/site/docs-md/apis/accessibility/index.md
@@ -1,4 +1,4 @@
-<plugin-platforms platforms="pwa,ios,android"></plugin-platforms>
+<plugin-platforms platforms="pwa,ios,android,electron"></plugin-platforms>
 
 # Accessibility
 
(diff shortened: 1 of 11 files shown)


site/docs-md/apis/accessibility/index.md
```diff
@@ -1,4 +1,4 @@
-<plugin-platforms platforms="ios,android"></plugin-platforms>
+<plugin-platforms platforms="pwa,ios,android"></plugin-platforms>
```
dotNetkow: @jcesarmobile - you mentioned to me the other day that "if it supports web/pwa, then it also supports electron". Is that the case? Hard to tell in some instances. For example, does Browser support electron even though there's no implementation file for it under electron folder?

mlynch: While it's true web support means basic electron support, for some APIs that doesn't qualify as true electron support (because what could be supported in electron is much more full featured than many web apis, for example)

mlynch: I think you can add electron support for local notifications, network, and app as recent additions

dotNetkow: done!

jcesarmobile: Some of the pwa APIs are also limited compared with iOS/Android, but I think, as long as there is a working implementation we should mark it as supported, meaning, all the marked as pwa should also be marked as electron.

site/docs-md/apis/splash-screen/index.md
```diff
@@ -1,4 +1,4 @@
-<plugin-api index="true" name="splash-screen"></plugin-api>
+<plugin-platforms platforms="pwa,ios,android"></plugin-platforms>
```
jcesarmobile: plugin-api is a different tag user for other thing, don't remove it
as there is no plugin-platforms tag, just add it, but without removing the other one.
Also, electron seems to be supported on electron too.

PR #1543, opened 2019-05-21
Remove removeListener function from plugins

jcesarmobile:


mrahmadt:
Hi

Why was "removeListener" removed? Any other alternative?

PR #1554, opened 2019-05-22
Fix Typo in sound option

digitalcrafted:
Fix Typo in sound option for Push notifications appearance in foreground

[old] `sound`: the devide will ring/vibrate when the push notification is received

[new] `sound`: the device will ring/vibrate when the push notification is received


jcesarmobile:
thanks!

PR #1557, opened 2019-05-22
Move npm option over yarn on prompt

jcesarmobile:


nklayman:
Personally, I feel like yarn should come first as it is only listed if it is installed on the user's system. If someone has yarn installed on their computer, it is likely they use it for most of their projects.

jcesarmobile:
I have it installed and I use npm in most projects

PR #1580, opened 2019-05-26
Update gitignore file in android-template

jonz94:
Update the file according to latest https://github.com/github/gitignore/blob/master/Android.gitignore


mlynch:
Thanks!

PR #1594, opened 2019-05-29
Docs: Add more community guides to APIs and comm guides pages

dotNetkow:
- Community Guides page: adding *APIs* as a new subsection and link Paul’s Network post. Others in the future.

- Individual API pages: If content is explicitly about an API (like Paul’s on Network - https://developer.school/posts/ionic-4-network-detection-with-capacitor/), then link to it here too. Add header *Example Guides* to existing ones that could use it.

We get the best of both worlds and account for multiple ways that devs could be viewing the docs, for the lowest effort.

### DIFF
--- a/site/docs-md/apis/camera/index.md
+++ b/site/docs-md/apis/camera/index.md
@@ -64,9 +64,9 @@ async takePicture() {
 }
 ```
 
-## Angular example
+## Example Guides
 
-[Follow this guide](../../guides/ionic-framework-app) to implement the Camera API in an Ionic Angular app.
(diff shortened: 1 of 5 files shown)


jcesarmobile:
I don't really like linking external guides from the APIs section, nor linking our own internal guides from the Community section.

mlynch:
I think this is reasonable. Let's try it and see how it feels for a bit. If we find the content gets out of date or whatnot we will need to rethink.

dotNetkow:
Thanks Julio!

PR #1596, opened 2019-05-29
fix(splash): prevent centerCrop splash from rescaling

bodinaren:
When the plugin calls `.show()` the splash screen re-scale a little bit from the initial splash screen when using `centerCrop`. Changing width and height from -1 to 0 fixes this. I added the `if`-cases just for extra safety, but in my experience it's always -1.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Splash.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Splash.java
@@ -163,6 +163,9 @@ public class Splash {
         // Required to enable the view to actually fade
         params.format = PixelFormat.TRANSLUCENT;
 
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR, but what I'm seeing with this change is that the second splash screen doesn't appear at all. So, it looks like there is no resize because there is no second splash. It's not very noticeable if the app launch fast enough, but if if takes some time (or if you set `launchShowDuration` to a big value like `"launchShowDuration": 10000`), then you'll see a white screen for a few seconds.

bodinaren:
Ah, I didn't notice that. That's also an explanation why it didn't resize any longer xD  I'll need to keep investigating then.

PR #1618, opened 2019-06-01
fix(electron): Prevent promises both resolving and rejecting

27pchrisl:
This change adds returns after all rejects, so that promises don't resolve after a rejection.


27pchrisl:
Hi, I should have put this in the original ticket, but the specific way I ran into a problem was this code:

```
      this.NodeFS.stat(options.path, (err:any, stats:any) => {
        if(err)
          reject(err);    <== "err" exists, and so a rejection occured

        resolve({type: 'Not Available', size: stats.size, ctime: stats.ctimeMs, mtime: stats.mtimeMs, uri: lookupPath});    <== JS engine moves onto this line, where because of the err "stats" is undefined and so throws due to stats.size accessing a property of an undefined.
      });
```

jcesarmobile:
Ah, yeah, in that case it was a bug. Good catch!

PR #1624, opened 2019-06-02
feat(Filesystem): Add move implementation

27pchrisl:
Implementation for all platforms, and uses the same move semantics across all.
It also preserves mtime across all platforms for both files and directories, (ctime is not available in some platforms, or is inconsistently handled by the underlying filesystem).
Moving a file to a directory does not imply moving it into that directory (ie "mv <file> <directory>" does not imply "mv <file> <directory>/<file>" and will fail

```
Missing arguments -> fails
Move into a subdirectory of itself -> fails
Move a file to an nonexistent location -> fails
Move a file to "overwrite" a directory -> fails
Move a file into a nonexistent directory -> fails
Move a nonexistent file -> fails
Move a file into a subdirectory that is a file -> fails

Move a file onto itself -> succeeds (with no on-disk change)
Move a file into a directory -> succeeds
Move an empty directory into a directory -> succeeds
Move a file to overwrite an existing file -> succeeds (same as writeFile)
Move a directory tree into a directory -> succeeds
Move a directory tree out of a directory -> succeeds
```

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/PluginRequestCodes.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/PluginRequestCodes.java
@@ -20,5 +20,5 @@ public class PluginRequestCodes {
   public static final int FILESYSTEM_REQUEST_DELETE_FOLDER_PERMISSIONS = 9017;
   public static final int FILESYSTEM_REQUEST_URI_PERMISSIONS = 9018;
   public static final int FILESYSTEM_REQUEST_STAT_PERMISSIONS = 9019;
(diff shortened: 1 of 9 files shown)


27pchrisl:
This also merges in other dependent branches, as those changes are needed for this branch to work correctly.

mlynch:
Thanks! Given that we're trying to follow node's API for fs, should we call this `rename` instead?

27pchrisl:
> Thanks! Given that we're trying to follow node's API for fs, should we call this `rename` instead?

Sure thing, I have changed it.

PR #1626, opened 2019-06-02
Fix(android): remove incorrect trailing newline in readFile when using encoding

27pchrisl:
Using this test:
```
(async() => {
  let test = async (content, encoding) => {
    await Capacitor.Plugins.Filesystem.writeFile({ path: "fa", data: content, directory: "DOCUMENTS", encoding })
    return content === ((await Capacitor.Plugins.Filesystem.readFile({ path: "fa", directory: "DOCUMENTS", encoding })).data);
  }

  console.log(await test("\ra b\r\nc d\n\r\n", "utf8") ? 'ok' : 'ko');
})();
```

iOS and PWA implementations returned 'ok', Android returned 'ko'. This PR causes Android to return 'ok'.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Filesystem.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Filesystem.java
@@ -17,7 +17,6 @@ import com.getcapacitor.PluginMethod;
 import com.getcapacitor.PluginRequestCodes;
 import org.json.JSONException;
 
-import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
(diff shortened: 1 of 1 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/Filesystem.java
```diff
@@ -174,7 +166,7 @@ public void readFile(PluginCall call) {
           InputStream is = getInputStream(file, directory);
           String dataStr;
           if (charset != null) {
-            dataStr = readFileAsString(is, charset);
+            dataStr = readFileAsString(is, encoding);
```
jcesarmobile:
```suggestion
            dataStr = readFileAsString(is, charset.name());
```

27pchrisl: Done!

PR #1632, opened 2019-06-04
Added androidBackgroundColor to Splash Screen plugin.

trancee:
In cases where you use a different scale type, e.g. CENTER_INSIDE, it is possible that the splash image is smaller than the view itself. The missing parts around the splash image will then be transparent and showing the background of the app. You can now override this by specifying a background color (similar to the cordova splash screen plugin). The value is a simple #RGBA value.


mlynch:
Thanks for the PR! Not opposed to adding this, but are we going to want something similar for iOS and thus should look at a non-platform-prefixed config key?

trancee:
Thank you for your feedback. I am not sure if this config key makes sense on iOS, as it handles splash screens differently. It is necessary on Android because you can change the scale type and it may happen that not the full screen is covered by the splash screen. In those cases, it requires the background color to cover those areas or they will be transparent and showing off the web app which looks ugly.

Basically, I am just following the practice from the Cordova splash screen plugin.

Issue #560, opened 2018-05-25
Automatic detection of Android Studio on linux


erperejildo:
`npx cap open android`

this command tries to open Android Studio on a default path but I've got this installed on a different location

erperejildo:
Ok, found this:
https://github.com/ionic-team/capacitor/commit/f68ea10a1b3659e32d2029a771ba35fdfd96cd5f

Working now, but it's a bit annoying if we're working from different computers. There is no way to automatically find Android Studio like Cordova does?

jcesarmobile:
Found this, we can try it https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/getASPath.bat

jsdussanc:
Me too for ubuntu

darrenmothersele:
Android Studio installed from AUR on Arch Linux is also not detected. 

The configuration option to override the install path only works on Windows.

jcesarmobile:
@darrenmothersele the linux configuration option is linuxAndroidStudioPath, have you tried that one?

As the automatic detection is working on windows now and it's not a problem on macOS, edited the issue title for linux (but not sure if it's possible to detect it there)

darrenmothersele:
@jcesarmobile sorry it took a while for me to respond. I stopped using my linux machine for daily development work and I'm back on my hackintosh - so was unable to test this.

I can confirm that setting `linuxAndroidStudioPath` does work correctly.

gbhipolitoglyph:
sorry, where can I see and edit the `linuxAndroidStudioPath` config? thanks

OoDeLally:
@gbhipolitoglyph `linuxAndroidStudioPath` is a key-value in your `capacitor.config.json`.

jcesarmobile:
closing as won't do

In Capacitor 3 users can use `CAPACITOR_ANDROID_STUDIO_PATH` environment variable to configure the android studio path if installed in the non default place.

https://capacitorjs.com/docs/config#environment-variables

Issue #230, opened 2018-03-01
Move `windowsAndroidStudioPath` to a gitignored file?


janpio:
`windowsAndroidStudioPath` is by definition dependent on the machine you are using. Committing this configuration to git so doesn't really make sense, as you probably would have to change it on every machine you are checking out the project.

A better idea might be to have another configuration file for machine dependent settings that is ignored via `.gitignore`.

mlynch:
Hmm, good idea. Yea I hate that that's a thing.

mlynch:
I think a `.capacitorrc` in your home folder could work

janpio:
What does the `rc` at the end stand for?

jcesarmobile:
Now if the default or configured path doesn't exist, it will autodetect the path from windows registry. So there is no need to set the `windowsAndroidStudioPath` anymore. Closing.

andressoop:
> 
> 
> Now if the default or configured path doesn't exist, it will autodetect the path from windows registry. So there is no need to set the `windowsAndroidStudioPath` anymore. Closing.

Auto detect does not work when Android Studio is installed through JetBrains Toolbox (for me at least).

jcesarmobile:
The change is not released yet

bleuscyther:
same issue here , based on my attempts and this comment (https://github.com/desktop/desktop/issues/7195#issuecomment-479964589)  Jetbrain does not have a fixed path for its tools. 😅

PR #1633, opened 2019-06-04
chore(cli): Autodetect Android Studio path on Windows

ralscha:
Auto Detect Android Studio path on Windows by reading a Windows registry entry.

First checks if path in `windowsAndroidStudioPath` exist. 
If not it tries to read the Windows registry entry `HKEY_LOCAL_MACHINE\\SOFTWARE\\Android Studio`

Issues #560 and #230

Based on getASPath.bat 
https://github.com/apache/cordova-android/blob/master/bin/templates/cordova/lib/getASPath.bat

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -1,7 +1,8 @@
-import { accessSync, readFileSync } from 'fs';
+import { accessSync, existsSync, readFileSync } from 'fs';
 import { basename, join, resolve } from 'path';
 import { logFatal, readJSON } from './common';
 import { CliConfig, ExternalConfig, OS, PackageJson } from './definitions';
+import { execSync } from 'child_process';
 
 let Package: PackageJson;
 let ExtConfig: ExternalConfig;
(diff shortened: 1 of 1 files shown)


Genarito:
Excuse me, It would not be better to check first if the current platform is Windows before executing the retrieval of the path and `if` sentences?

```
private initWindowsConfig() {
    if (this.cli.os !== OS.Windows) {
        return;
    }

    // Your code (without checking cli again in the second if)...
}
```

ralscha:
I guess this could work too. In that case `this.windows.androidStudioPath` is undefined, but I guess this does not matter when you don't run it on Windows. 

The current code always sets `this.windows.androidStudioPath` with the value from `this.app.windowsAndroidStudioPath`

Current version:
```
this.windows.androidStudioPath = this.app.windowsAndroidStudioPath && this.app.windowsAndroidStudioPath;
```

cli/src/config.ts
```diff
@@ -193,7 +194,17 @@ export class Config implements CliConfig {
   }
 
   private initWindowsConfig() {
-    this.windows.androidStudioPath = this.app.windowsAndroidStudioPath && this.app.windowsAndroidStudioPath;
+    if (this.app.windowsAndroidStudioPath) {
+      if (!existsSync(this.app.windowsAndroidStudioPath) && this.cli.os === OS.Windows) {
```
jcesarmobile: can you put the `this.cli.os === OS.Windows` first? in that case it won't try to see if the file exists in non windows

cli/src/config.ts
```diff
@@ -193,7 +194,17 @@ export class Config implements CliConfig {
...
+        const buffer = execSync('REG QUERY "HKEY_LOCAL_MACHINE\\SOFTWARE\\Android Studio" /v Path');
```
jcesarmobile: can you put all this in a try catch?
If Android Studio is not installed it shows an ugly message because it doesn't find the registry.
In the catch you can set the path to empty string so it can be used later in the open command to not open if it's empty. I'll do that in a separate PR once this is merged.

cli/src/config.ts
ralscha: An alternative is to switch the comparison inside the if
If the first operand is false, JavaScript will short-circuit and not evaluate the second condition.
```
if (this.cli.os === OS.Windows && !existsSync(this.app.windowsAndroidStudioPath)) {
```

jcesarmobile: Yeah, that's what I meant when I said first, in the same if, but it's also ok to just do nothing if it's not windows as Genarito suggested.

Issue #1431, opened 2019-04-25
currently there is no way to disable logs


stewwan:
I think it's a must-have feature, we would be able to better isolate and debug issues on native part.

https://github.com/ionic-team/capacitor/blob/8158d910905790dd98bfffe3b927e445ecfb8c6e/ios/Capacitor/Capacitor/CAPBridgeDelegate.swift#L38

mlynch:
Marking this as high priority because it also has performance implications and we don't need to print out this stuff out by default but could be useful for debugging.

biesbjerg:
This looks like it was only solved for iOS? Or am I mistaken?

Boosten:
@biesbjerg you might be right :-) I don't see any config option to hide logs in the Android code from capacitor.

fifafu:
It would be really good if there was a flag for Android also. This is often an issue in enterprise builds because the security teams don't want the risk for any sensitive data being logged. (Which can happen easily, for example if you have a password store native plugin).

For now the only way seems to be to manually comment all the log statements.

PR #1640, opened 2019-06-06
feat(ios): add useLog config #1431

stewwan:
@mlynch 
@jcesarmobile

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridge.swift
+++ b/ios/Capacitor/Capacitor/CAPBridge.swift
@@ -454,7 +454,10 @@ enum BridgeError: Error {
   public func toJs(result: JSResult, save: Bool) {
     do {
       let resultJson = try result.toJson()
-      print("⚡️  TO JS", resultJson.prefix(256))
+
+      if (CAPConfig.useLog() as? Bool != false) {
+        print("⚡️  TO JS", resultJson.prefix(256))
+      }
       
       DispatchQueue.main.async {
         self.getWebView()?.evaluateJavaScript("""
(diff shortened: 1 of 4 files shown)


stewwan:
this PR provides a way to disable native logs #1431. you can control it on your `capacitor.config.json`

```json
{
  "useLog": false
}
```

stewwan:
I can add the android version, just let me know if the api fits good.

jcesarmobile:
I think this print should be a drop in replacement for regular print, so should look like

```
        if (self.useLog() as? Bool != false) {
            Swift.print(items, separator, terminator)
        }
    }
```
No need to have the `@objc` as print is Swift only (same for the other methods and class)

Not sure if we need this new `CAPLog` class, or if we can just put it inside the CAPBridge class, as it has a `modulePrint` function already, or move that one to the new CAPLog class.

Another thing, CAPConfig is no longer a singleton, so code should be changed.
First a static config should be declared
`public static let config = CAPConfig()`

And change the usage to
`let isProduction = config.getValue("production") as? Bool`
        
But I not sure about the `production` name of the preference, in my experience, people tend to think they are building a production app just by setting that value to true, so I would stick with showLogs, or hideLogs, or useLogs

Or maybe just don't make it configurable and hide all the logs when a release build is done with the DEBUG macro

```
#if DEBUG
  print(...)
#endif
```

@mlynch thoughts?

stewwan:
drop in replacement looks awesome. thanks for the ideas.

from my work experience macros won't work. what if I wanted to check capacitor logs with my app in production mode.

jcesarmobile:
As Max said we shouldn't show them by default, I suppose he means in production. 
So it should check both `isDevMode` method from CAPBridge and the configuration option in the `capacitor.config.json` to always show them in dev mode, but in production only if enabled with the preference.

PR #1645, opened 2019-06-06
fix(cli): Allow to filter incompatible plugins per platform

jcesarmobile:
and add `cordova-plugin-googlemaps` as iOS incompatible only.

### DIFF
--- a/cli/src/android/common.ts
+++ b/cli/src/android/common.ts
@@ -16,12 +16,13 @@ export function getAndroidPlugins(allPlugins: Plugin[]): Plugin[] {
 }
 
 export function resolvePlugin(plugin: Plugin): Plugin | null {
+  const platform = 'android';
   if (plugin.manifest && plugin.manifest.android) {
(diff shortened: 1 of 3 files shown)


wf9a5m75:
abusing?

jcesarmobile:
???
The plugin causes build failures on iOS, this allow people to still use it on Android

wf9a5m75:
At least, you should list here, not just blocking secretly.
Nobody can not found until inspecting the code.

https://github.com/ionic-team/capacitor/blob/29087ec682c29184efa2d25037cdf542d7c259ac/site/docs-md/cordova/known-incompatible-plugins.md

jcesarmobile:
When you run update command it tells you the plugin is incompatible, so it’s not a secret. But yeah, I’ll add it to the list

jcesarmobile:
documented on https://github.com/ionic-team/capacitor/pull/2072

PR #1648, opened 2019-06-07
feat(ios): add modelName to device plugin

stewwan:
```js
Device.getInfo();
```

```
{
  // ...
  modelName: "iPhone 6s"
}
```

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -400,9 +400,10 @@ export interface DevicePlugin extends Plugin {
 
 export interface DeviceInfo {
   /**
-   * The device model. For example, "iPhone"
+   * The device model. For example, "iPhone", "iPhone 6s"
    */
   model: string;
+  modelName: string;
   /**
    * The device platform (lowercase).
    */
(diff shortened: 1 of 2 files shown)


jcesarmobile:
I don't like having that big list of models that will have to be updated every time a new model comes out, and also don't like that it's iOS only.

Thoughts @mlynch?

stewwan:
actually this list is ios specific and dont need to be maintained, just be present to get specific names of older ios devices. other day I needed to hide a button for an `iPhone 6s` so I came up with this solution.

for android I dont think they have this problem of hiding the actual name of devices

stewwan:
maybe the `model` property could be fully replaced with what `modelName` is doing. it's basically the same but with some checks for older devices.

jcesarmobile:
Well, that's because you only want it to detect older devices, but if we add the feature it should work for all devices, not only for old ones. Some day the devices that are not on the list will be old too.
If I run the app on my iPad I get iPad7,3 because it's not on the list.

We could also just return the "official" name instead of the pretty name and then you check for `iPhone8,1` instead of `iPhone 6s`

stewwan:
the "official" name would work for me as well, just thought it would be useful for someone else.

nphyatt:
Thanks for the PR @stewwan but I think for the long term maintainability of capacitor it makes sense to use the official names returned by the OS. I can see value in having a map from the official names to more friendly ones but perhaps the place for that would be a third party JavaScript package that user could choose to install and use if they so desired. Cheers!

PR #1649, opened 2019-06-08
feat(splashscreen): Added backgroundColor to Splash Screen plugin.

trancee:
In cases where you use a different scale type, e.g. CENTER_INSIDE, it is possible that the splash image is smaller than the view itself. The missing parts around the splash image will then be transparent and showing the background of the app. You can now override this by specifying a background color (similar to the cordova splash screen plugin). The value is a simple `#RGB` or `#ARGB` value.

@mlynch Thank you for your feedback. I am not sure if this config key makes sense on iOS, as it handles splash screens differently. It is necessary on Android because you can change the scale type and it may happen that not the full screen is covered by the splash screen. In those cases, it requires the background color to cover those areas or they will be transparent and showing off the web app which looks ugly.

Basically, I am just following the practice from the Cordova splash screen plugin.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/SplashScreen.swift
+++ b/ios/Capacitor/Capacitor/Plugins/SplashScreen.swift
@@ -2,7 +2,7 @@ import AudioToolbox
 import Foundation
 
 @objc(CAPSplashScreenPlugin)
-public class CAPSplashScreenPlugin : CAPPlugin {
+public class CAPSplashScreenPlugin: CAPPlugin {
   var imageView = UIImageView()
   var image: UIImage?
   var call: CAPPluginCall?
@@ -27,7 +27,9 @@ public class CAPSplashScreenPlugin : CAPPlugin {
     self.call = call
 
     if image == nil {
-      call.error("No image named \"Splash\" found. Please check your Assets.xcassets for a file named Splash")
+      call.error(
+        "No image named \"Splash\" found. Please check your Assets.xcassets for a file named Splash"
+      )
       return
     }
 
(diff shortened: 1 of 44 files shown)


trancee:
@mlynch I have managed to add backgroundColor support for iOS. I renamed the configuration key to `backgroundColor` instead and it accepts either Hex `RGB` or Hex `ARGB` value.

ios/Capacitor/Capacitor/Plugins/SplashScreen.swift
```diff
@@ -1,149 +1,209 @@
...
+
     let showDuration = call.get("showDuration", Int.self, defaultShowDuration)!
     let fadeInDuration = call.get("fadeInDuration", Int.self, defaultFadeInDuration)!
     let fadeOutDuration = call.get("fadeOutDuration", Int.self, defaultFadeOutDuration)!
     let autoHide = call.get("autoHide", Bool.self, defaultAutoHide)!
-    
-    showSplash(showDuration: showDuration, fadeInDuration: fadeInDuration, fadeOutDuration: fadeOutDuration, autoHide: autoHide, completion: {
-      call.success()
-    }, isLaunchSplash: false)
+    let backgroundColor = call.get("backgroundColor", String.self)
```
jcesarmobile: It should be like this:
```suggestion
    let backgroundColor = getConfigValue("backgroundColor") as? String ?? nil
```
call.get is to get the value from the plugin call, not from the `capacitor.config.json`

trancee: OK, I have changed it according to your suggestion.

jcesarmobile:
Something went wrong with the rebase, looks like you didn't pull latest changes.

trancee:
> Something went wrong with the rebase, looks like you didn't pull latest changes.

Maybe I forgot to include all of the files. I am doing that now. Hopefully that works.

Issue #872, opened 2018-10-10
i18N for Capacitor plugins.


xmano:
As we have more capacitor plugins and also applications being developed for different locales,
having a fixed set of string resource values will make the app not "accessible"

Normally, this is done using storing the localisation strings in the resource table in the native code. I am proposing that the strings be passed as an array during the initialisation of capacitor plugin.

Another issue is that for supporting multiple languages would require having to spend time
on creating the resources, reviewing the correctness and committing the changes. This
would make the whole process time consuming and lengthy.

I am submitting a RFC for this.

**Proposal:**

```
1. A general mechanism for app developers to provide an array of localized strings that 
would be passed during the plugin initialization. 

2. When a plugin dialog is displayed use the localized strings that were provided to
   give a better accessible user experience

3. In case a localization string is not available, fallback to the default 'EN' string.

3.1 In case string was missed, future version of software can have the support added
    or provide better localizations.

```

By having the app developer control over the localised strings it would help develop apps
independently without having to depend on commit cycles of native plugin.

xmano:
I have some code changes for the proposal. It is a private repo though:

https://github.com/ionic-team/eauth-integ-test/commit/c9ef8405aa6969a46b3496f2b828bd0d99690b86

![dialogspanish](https://user-images.githubusercontent.com/8731418/46760240-25bbe900-cc97-11e8-832e-6e7059b4967c.jpg)

The idea is to pass a localise string array to the plugin initialisation code and have the platform render the strings properly when the UI is displayed.

TrayserCassa:
Any Updates on that? Because it is very deep interaction with capacitor, so I didn't want to make a pull request and do it on my own. But my Product Owner sad, 'ionic is dead when they can't support local languages'.

jcesarmobile:
Well, this only affects a few native messages on some plugins. Ionic apps can be totally localized except for those few native Capacitor plugins.

It’s still not clear if we will use some common json or if we will use the native implementation per platform.

As you said you thought about sending a PR, what would be your approach?

TrayserCassa:
@jcesarmobile 

It should be customizable over a File in the Project and not the native implementation. The cli could ask for a default language at the init function. The sync Function can then clone the data to the native string file. 

The Problem with directly Native is, that I don't want to care about implementing it twice (Android, iOS). If you answer is, well we do it native, I will be writing my own cli for i18n Transfer.

sandstrom:
**I'll propose an alternative solution to this problem**

[...]

PR #1659, opened 2019-06-13
feat(): labels for actionsheet options on camera plugin

jabiinfante:
We have found the need to translate the camera plugin action-sheet labels and we think you might find it usefull.

We came up with the following interface, but we are of course open to any suggestion or correction:
```ts

async takePicture() {
  const image = await Camera.getPhoto({
    quality: 90,
    allowEditing: true,
    resultType: CameraResultType.Uri,
    labels: {
      photo: 'Photo', // Only iOS (action-sheet title)
      fromPhotos: 'From my photo gallery',
      takePicture: 'Take a Picture',
      cancel: 'Cancel' // Only iOS (action-sheet cancel label)
    }
});
```

thanks for the great work ;) 
(we will be in production pretty soon)


jcesarmobile:
Thanks for the PR, we definitely want to make possible to localize the plugin texts, but I'm not sure about this approach. 
We have an issue open for debating, feel free to chime in and give your opinion
https://github.com/ionic-team/capacitor/issues/872

PR #1701, opened 2019-06-22
Use correct install command for @capacitor/cli in "Installing Capacitor"

toxsickcoder:
The `@capacitor/cli` package should be installed as a devDependency. Updated the "Installing Capacitor" page to use the correct commands for this.


jcesarmobile:
Thanks for the PR, it has been proposed before but we decided to keep it this way because of simplicity of install instructions.

Issue #984, opened 2018-11-20
Support blob/large data


mlynch:
Investigate passing blobs from native to JS and vice versa.

jcesarmobile:
There is this issue about improvements and one of the is about blobs. Should we close this one or better to have an issue per improvement?
https://github.com/ionic-team/capacitor/issues/974

thmclellan:
Great to see this on the roadmap. In the meantime, if working on a custom Capacitor plugin, are there any creative workarounds that might allow returning a blob/binary data without having to convert to JSON?  Based on https://capacitor.ionicframework.com/docs/plugins/ios, it seems like  the CAPPluginCall can only handle JSON right now, and that no amount of creative plugin writing could respond with a blob directly. Is that right? 

The background here is that we're trying to let the user import and upload 20MB PDF files with our app. The current workaround is converting the binary data to base64 using FileSystem.readFile, and then converting it back to a blob in the app, and uploading to a remote server as XHR. Trying to avoid memory issues, and unable to do a direct Fetch on a non-HTTP URI (e.g. file://etc on iOS). Thanks

jcesarmobile:
Fetch should work, but you have to use convertFileSrc on the file:// url

thmclellan:
@jcesarmobile thank you! I was using convertFileSrc while running ng serve and getting a CORS error on the "capacitor://localhost" URL, but I see now that fetch is working with convertFileSrc on a production build.  For ng serve, after doing convertFileSrc, replacing "capacitor://localhost" with window.location.origin gave me a URL that works just fine for fetch. Appreciate the quick help.

black-hawk85:
Sharing my workaround with you.

I'm using web-implementation of the Filesystem plugin.
Don't know if it's a good solution but it helps solving my problem.

```
import { FilesystemDirectory, FilesystemEncoding } from '@capacitor/core'
// const { Filesystem } = Plugins // => Doesn't work with blob in native
// Workaround: Use Web-Implementation of plugin
import { FilesystemPluginWeb } from '@capacitor/core/dist/esm/web/filesystem.js'
const Filesystem = new FilesystemPluginWeb()
```

digaus:
I was able to create a service for writing and reading large blobs. (For web, blobs are already supported for indexeddb)

Reading is simply done with a `fetch` call.

Writing is a bit more tricky because we can only pass strings to that native layer. 

[...]

PR #1705, opened 2019-06-25
Add support for Web Share API Level 2 in Share plugin

### LINKED ISSUE #1704: Add support for Web Share API Level 2
In Chrome 75 on Android, it is now possible to share files using the native share sheet. We can only assume that this will be available in other browser soon, since it builds on the current Web Share API standard. 

An example of how this is done can be found [here](https://developers.google.com/web/updates/2019/05/web-share-files).

This is particularly useful for anyone building PWAs.

jonathon-m:
Add 'files' field to ShareOptions interface. If this new field is not filled, the share plugin behaves as usual, otherwise it checks for availability of file sharing, and requests it from the global navigator.

Closes #1704

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -1387,9 +1387,14 @@ export interface ShareOptions {
    */
   url?: string;
   /**
+   * Set files object array to share
+   */
+  files?: [File];
+  /**
    * Set a title for the share modal. Android only
    */
   dialogTitle?: string;
+
 }
 
 //
--- a/core/src/web/share.ts
+++ b/core/src/web/share.ts
@@ -19,12 +19,23 @@ export class SharePluginWeb extends WebPlugin implements SharePlugin {
     if (!navigator.share) {
       return Promise.reject("Web Share API not available");
     }
-
+    if (!options.files) {
+      return navigator.share({
+        title: options.title,
+        text: options.text,
+        url: options.url
+      });
+    }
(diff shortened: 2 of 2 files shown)


jcesarmobile:
Thanks for the PR, I've tested it and works fine, the only problem is you can't pass a `File` object to the webview bridges, so this will never work in native apps.

I won't merge this for now. Will try to figure out how to handle the files in the native side.

jonathon-m:
Thanks @jcesarmobile. 
Any thoughts on how it could be handled for native? I know this feature was available in the [Cordova share sheet plugin](https://github.com/EddyVerbruggen/SocialSharing-PhoneGap-Plugin), so maybe we could borrow some ideas from there.

jcesarmobile:
That plugin uses the file paths, and I think the base64 representation of the files also works.

seopei:
I have a quick question regarding Web Share API Level 2: I am building an app for iOS and Android and I am using the newest versions of Ionic 5 and Capacitor 2. Do the native web views were my app will be rendered in support Web Share API Level 2?

dwieeb:
It sounds like this is depending on https://github.com/ionic-team/capacitor/issues/984 for blob support in Android and iOS.

dwieeb:
@jonathon-m Thanks for your PR. I'm going to close this for now because so much is changing for Capacitor 3. 

@seopei Feel free to create a discussion for your question (see the discussions tab above).

jcesarmobile:
I created and answered the questions about Web Share API Level 2 in WebViews
https://github.com/ionic-team/capacitor/discussions/3213

PR #1734, opened 2019-07-04
fix(android): splash spinner bar not properly handled in view

trancee:
I had to rearrange the code to accommodate the issue with removing the spinner bar from the view. Instead I am setting the visibility now and only remove it when the app is paused/destroyed.


jcesarmobile:
Can you describe the problem it fixes?

trancee:
> Can you describe the problem it fixes?

Yes, sure. I noticed that the first time the spinner appears, it works as intended. But if you try to manually show the splashscreen again using `SplashScreen.show()`, once it shows and is finished, the whole app crashes. The reason is because I added the spinner somewhere outside the current context of the Android UI and then want to remove it from another location. This code changes avoid this circumstance and resolves the issue. I hope I could explain it properly.

PR #1747, opened 2019-07-09
DO NOT MERGE (WIP) - FilesystemWeb inconsistency

martinkasa:
To stay consestent with iOS and Android do not use DEFAULT_DIRECTORY if it is not provided in readFile options. It was adding '/Data/' event when full there was full URI provided in path option.

### DIFF
--- a/core/src/web/filesystem.ts
+++ b/core/src/web/filesystem.ts
@@ -26,7 +26,6 @@ import {
 } from '../core-plugin-definitions';
 
 export class FilesystemPluginWeb extends WebPlugin implements FilesystemPlugin {
-  DEFAULT_DIRECTORY = FilesystemDirectory.Data;
   DB_VERSION = 1;
   DB_NAME = 'Disc';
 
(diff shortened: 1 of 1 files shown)


martinkasa:
I will have to update this PR, because there is DEFAULT_DIRECTORY used also in android plugin. But another problem I came to during development are more inconsistences. I will write short list:
- in iOS there is no exception when trying to create already created dir, on other platform there is exception
- there are no error codes in error callbacks and messages are different. I can imagine there could be some enum / error codes to be able to handle errors from different platforms the same way
- maybe add simple methods if file/directory exist?

martinkasa:
Sorry guys, for now Filesystem plugin is not usable for more complex task. After few days of trying I have to switch back to Cordova File plugin. There are too many inconsistencies. getUri/readdir in webplugin return different paths than in ios/android. I was not able to do scenario "read subdirectory and delete all files in it" without handling web separately.

Issue #1741, opened 2019-07-05
Provide better configuration management / better documentation for configuration management


macdja38:
In the iOS app build flow and android flow you can configure different variables to be used in Production / Staging / Dev builds.

As it stands at least the documented capacitor workflow for `capacitor.config.json` doesn't allow for using a different values for production vs staging, say you wanted to load from a different server the `server.url` would have to be edited by the user then copies to the environments.

There are also no best practice / documented recommendations on how to manage the `capacitor.config.json` file with source control / with staging / ci environments as far as I can find?

Bktechmasala:
watch this video to solved all problems capacitor camera plugin https://youtu.be/EPysW31oryk

dwieeb:
Closed in favor of https://github.com/ionic-team/capacitor/issues/3141

Issue #1478, opened 2019-05-08
A way to disable Capacitor Config.json Server Url in Production


agrosner:
Currently, when running the app with livereload enabled via:
```
ionic capacitor run ios --livereload
```
will insert into our `capacitor.config.json` a server url:
```
{
  "appId": "io.ionic.starter",
  "appName": "app",
  "bundledWebRuntime": false,
  "webDir": "www",
  "plugins": {
    "Keyboard": {
      "resize": "none"
    }
  },
****
  "server": {
    "url": "http://192.168.28.43:8100"
  }
****
}
```

This is problematic because if this is ever accidently committed to the repo, on launch from a production build from a CI server, it will still look for that URL, and users would be presented with a blank app screen that never loads.

This is a request to provide a way to turn off url searching, or provide a better way to set up live reloading without need for it to modify this file in the first place, possibly through the tooling in a different way.

jcesarmobile:
Some people is using the url to point to real websites, so we can't disable it for production as it will break some use cases.
What about using a new "devUrl" that is only used while in development mode?

Note that this will require Ionic CLI changes as it's Ionic CLI the one that sets the server url, so probably better report it there. https://github.com/ionic-team/ionic-cli/issues

agrosner:
good point, but yes, it should set a `devUrl` for live reloading, but wouldn't the url change based on a dev's ip address? I think this should be something if its set, is useable on any computer, since it could be checked in. Devs with different IP addresses would constantly be overwriting this line, cluttering git changelog. React Native, Flutter, and other tools do not require this field when live reloading, so maybe its just implementation details?

diachedelic:
my solution is to generate `capacitor.config.json` on build:

```javascript
// generate_capacitor_config.js

const ip = require('ip')
const fs = require('fs')
const path = require('path')

const capacitorConfig = {
  // ...
  server: {},
}

const isHotspotting = (
  // Samsung A5 + Macbook
  ip.address() === '192.168.43.101' &&
  process.env.NODE_ENV !== 'production'
)

if (isHotspotting) {
  // support HMR on device
  capacitorConfig.server.url = 'http://192.168.43.101:8013'
}

const configPath = path.resolve(__dirname, 'capacitor.config.json')

fs.writeFileSync(configPath, JSON.stringify(capacitorConfig))
```
```javascript
// package.json
{
  "scripts": {
    "build": "export NODE_ENV=production && npm run config && vue-cli-service build --mode development && cap copy",
[...]

PR #1754, opened 2019-07-10
capacitor.config.json with templates

macdja38:
The basic idea is to allow for pulling in variables from the platforms configuration systems.

Both iOS and Android allow users to have different configurations for dev / release they may choose from when running a build, along with ways of managing that configuration with respect to source control.

This PR allows utilization of those values from the capacitor.config.json through the syntax $(variable_name). 

This could be part of the solution for: https://github.com/ionic-team/capacitor/issues/1478

And provides better configuration as asked for by me here: https://github.com/ionic-team/capacitor/issues/1741

Note that this PR only includes Android / iOS implementations, web / electron is not supported by this PR at this time.

Future improvements may include default values and non string types.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Config.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Config.java
@@ -1,7 +1,10 @@
 package com.getcapacitor;
 
 import android.app.Activity;
+import android.content.Context;
+import android.content.pm.PackageManager;
 import android.util.Log;
+import android.content.pm.ApplicationInfo;
 
 import org.json.JSONException;
 import org.json.JSONObject;
@@ -10,6 +13,8 @@ import org.json.JSONArray;
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 /**
  * Management interface for accessing values in capacitor.config.json
@@ -17,6 +22,7 @@ import java.io.InputStreamReader;
 public class Config {
 
   private JSONObject config = new JSONObject();
+  private Context applicationContext = null;
 
   private static Config instance;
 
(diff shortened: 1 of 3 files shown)


jcesarmobile:
Can you explain how this works so I can test it?

Also, could you document it as part of the PR so everybody else can use it if merged?

macdja38:
@jcesarmobile yup, will work on that next week. It needs to be implemented for web as well though, and I'm not really sure how that should look as web doesn't have a ubiquitous config system similar to android / iOS.

A brief overview of how it works is:

It replaces $(variable_name) values in the capacitor.config.json with variables by the same name in the platforms configuration.

So for example you can have in your capacitor.config.json
{
  "url": "$(url_var)"
}

then in your AndroidManifest.xml you can add a meta tag for url_var and have it change based on which build config you have when you run builds. Similarly on iOS you can use different builds to pull variables in from the project config. This allows stuff like using a different server url / navigable domain whitelist in production / development / staging without weird scripts to over-write the capacitor.config.json depending on what build stage you run.

dwieeb:
@macdja38 Curious to hear your thoughts on an alternative implementation of this. Instead of using template syntax, we just allow JS config files. The Capacitor CLI would run the file to generate a config object which gets written into the native projets every time `cap sync` is run.

macdja38:
> @macdja38 Curious to hear your thoughts on an alternative implementation of this. Instead of using template syntax, we just allow JS config files. The Capacitor CLI would run the file to generate a config object which gets written into the native projets every time `cap sync` is run. 

So long as the js config file has access to the environment from .env files I’m completely for it. 

This is actually similar to the solution I ended up going with except I wrote a tiny external (to the capacitor project) tool to execute the js config file to generate the capacitor.config.json.

I think it would be ideal to have access to the environments on each platform but the js file compiled at compile time would be nice as well.

dwieeb:
@macdja38 Yeah it's just a file that will run with NodeJS so it will work with any config setup you want. Think webpack or rollup configs. I imagine the Capacitor CLI will provide some environment variables as well such as `CAPACITOR_PLATFORM`.

dwieeb:
An alternative implementation will be in Capacitor 3, track https://github.com/ionic-team/capacitor/issues/3141 for details.

PR #1763, opened 2019-07-15
fix(android): enable full screen video

### LINKED ISSUE #605: Allow full screen video on Android
I've created a video.js sample app and it doesn't show the full screen video button, nor opens full screen when using `playsinline: false` option.
Same app in Cordova has full screen video button and goes full screen when play is pressed if `playsinline: false` option is used.
We should also allow full screen video in Capacitor

Sample app:
https://github.com/jcesarmobile/playerjs-example

rgolea:
This allows capacitor to handle fullscreen. This is my first attempt to write code in Java and please keep in mind it's a really strange environment for me.

Thanks

This should close #605


rgolea:
Thank you @jcesarmobile!

PR #1776, opened 2019-07-19
fix(android-livereload-localhost):

luisvt:
* Use request proxy handler when livereload and appUrl localhost:8100
* Reconnect when connection has not been established

This will allow users run:

```bash
ionic capacitor run android -l
```

and let connect the app to `localhost:8100` without the need to use `--address=0.0.0.0`

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -173,10 +173,11 @@ public class WebViewLocalServer {
       return null;
     }
 
-    if (isLocalFile(loadingUrl) || loadingUrl.toString().startsWith(bridge.getLocalUrl())) {
+    if (isLocalFile(loadingUrl) || !bridge.getAppUrl().contains("localhost:8100") && loadingUrl.toString().startsWith(bridge.getLocalUrl())) {
       Log.d(LogUtils.getCoreTag(), "Handling local request: " + request.getUrl().toString());
       return handleLocalRequest(request, handler);
     } else {
+      Log.d(LogUtils.getCoreTag(), "Handling proxy request: " + request.getUrl().toString());
       return handleProxyRequest(request, handler);
     }
   }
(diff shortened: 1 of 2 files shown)


jcesarmobile:
Thanks, but this only works on simulators, on real devices it will try to proxy localhost:8100, and that will fail because "localhost" will be the device, but the server is in the computer.

I think Ionic CLI will be updated to use `--address=0.0.0.0` by default for Capacitor, so you won't need it in the future.

luisvt:
yes, the objective is also to let the user add the port fordwarding in the chrome inspector

luisvt:
When I run livereload using `--port=0.0.0.0` the `geolocation` plugin stop working

jcesarmobile:
Ah, didn't think about port forwarding.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -173,10 +173,11 @@ public WebResourceResponse shouldInterceptRequest(WebResourceRequest request) {
       return null;
     }
 
-    if (isLocalFile(loadingUrl) || loadingUrl.toString().startsWith(bridge.getLocalUrl())) {
+    if (isLocalFile(loadingUrl) || !bridge.getAppUrl().contains("localhost:8100") && loadingUrl.toString().startsWith(bridge.getLocalUrl())) {
```
jcesarmobile: this is not a good check, will stop working if the user changes the default port. It's probably better to just check `Config.getString("server.url") == null`

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -294,22 +295,32 @@ private WebResourceResponse handleProxyRequest(WebResourceRequest request, PathH
...
-        for (Map.Entry<String, String> header : headers.entrySet()) {
-          conn.setRequestProperty(header.getKey(), header.getValue());
-        }
-        conn.setRequestMethod(method);
-        conn.setReadTimeout(30 * 1000);
-        conn.setConnectTimeout(30 * 1000);
+
+        HttpURLConnection conn;
+        boolean isContentTypeNull;
+        do {
```
jcesarmobile: I'm not really confortable putting a while with Thread.sleep, might be useful for reconnection, but could cause problems if the networks is down.

luisvt: If there is a better way you could point me, I'll change it.

luisvt: Maybe I could throw an error after 10 attempts

PR #1781, opened 2019-07-21
feat(): Add recursive option for rmdir

27pchrisl:
Filesystem.rmdir has different behaviour on iOS to other platforms (deletes recursively on iOS, does not on other platforms). This addresses the inconsistency by making all platforms delete recursively.


27pchrisl:
Hi, I think that it should be possible to command the native code to do the rmdir on your behalf (either with or without an option), as the performance better than the JS code doing it over the bridge. That was my motivation for making that behaviour common across platforms for our app.

27pchrisl:
Behaviour updated, added 'recursive' option that defaults to false

27pchrisl:
Tests seem to be failing due to an internal CircleCI error

jcesarmobile:
I've merged your other PR and not this one have conflicts, can you fix them?

Hopefully tests will pass this time, they sometimes hang for no reason.

27pchrisl:
Conflicts fixed

PR #1782, opened 2019-07-21
fix(core/web): avoid appendFile/writeFile to overwrite existing directory entry

27pchrisl:
This fix corrects a problem where an appendFile/writeFile could corrupt a directory entry if given a directory entry path.


27pchrisl:
Not sure why that test failed - seems to be a problem in S3?

PR #1812, opened 2019-07-26
fix(electron): createIntermediateDirectories option not being used in mkdir

jcesarmobile:
Will only work in electron 5 or newer because it was added in node 10.12, but it's better than ignoring it.


D34THWINGS:
Any chance to have this released?

PR #1818, opened 2019-07-29
fix(core/Filesystem): allow readdir on root directories

### LINKED ISSUE #1775: Filesystem readdir() does not work on root of Data filesystem
Description of the problem:

Affected platform
- [x] Android
- [x] iOS
- [x] electron
- [x] web

OS of the development machine
- [ ] Windows
- [x] macOS
- [ ] linux

Other information:

Capacitor version: 1.1.0

node version: v10.14.2

npm version: 6.9.0

CocoaPods version:

Steps to reproduce:

I tried to use this code:
```
export const removeAllDataFiles = () => {
  return Filesystem.readdir({
    path: '',
    directory: FilesystemDirectory.Data,
  });
};
```

I also tried to use `'.'` and `'/'` for the value for `path`, but none of them worked. You always get this error back: `Error: Uncaught (in promise): Error: Folder does not exist.`

What this means is that you cannot list the files in the "root" directory of `Data`. That is equivalent to an operating system where you are given a home directory, but not allowed to actually do `ls` on it. You just have somehow know what files are there... 
Link to sample project:

jcesarmobile:
Closes #1775


adeel399:
I am also not able to read ios all albums on root using directory.document and path is ' '

PR #1830, opened 2019-07-31
feat(android): support custom local notification icon

### LINKED ISSUE #1752: Local Notifications icon cannot be changed
Description of the problem:
Cannot change local notifications icon (it stays a circle with an "i" by default) on Android even though I can change the push notifications icon by setting the required property on AndroidManifest.xml. On iOS it takes the app's own launcher icon by default.

Affected platform
- [x] Android
- [ ] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [ ] Windows
- [ ] macOS
- [x] linux

Other information:

Capacitor version: 1.1.0

node version:

npm version:

CocoaPods version:

Steps to reproduce:

Link to sample project:

nikosdouvlis:
Support for android custom local notification icon. 

Added a `smallIcon` property to `LocalNotificatoin`. Can accept values of both `res://icon_name` and `icon_name` types (to stay consistent with the attachments' `url` prop). Will only look inside the `res/drawables` directory since the `mipmap` one is supposed to be used only for launcher icons.

If no custom icon is set or if the custom one cannot be found, the standard android info icon will be used instead.

If this gets merged, the docs need to be updated to make clear that the custom icon needs to follow Android's guidelines (see the top answer here: https://stackoverflow.com/questions/45318614/why-is-my-smallicon-for-notifications-always-greyed-out) or else the it won't be displayed.

Closes #1752 
Closes #1510


core/src/core-plugin-definitions.ts
```diff
@@ -945,6 +945,7 @@ export interface LocalNotification {
   id: number;
   schedule?: LocalNotificationSchedule;
   sound?: string;
+  smallIcon?: string;
```
jcesarmobile: Maybe comment it's Android only?

nikosdouvlis:
My initial thought was to have an API similar to `cordova-plugin-local-notifications`. Since it was the most popular cordova local notifications plugin I guess most of the teams migrating from cordova to capacitor are already familiar with.

But I guess you're right, most apps (like ours) won't need to configure the icon for each notification individually - so a global config makes sense.

So I pushed a new commit, which allows configuring the smallIcon in `capacitor.config.json` like: 
```
  "plugins": {
    "LocalNotifications": {
      "smallIcon": "ic_stat_icon_config_sample"
    }
  }
```
but also keeps the previous functionality - allowing to override the default icon per notification.

PR #1844, opened 2019-08-04
Update electron to 6.0.0

eiymba:
### Changes

-  [package.json](https://github.com/ionic-team/capacitor/blob/master/electron-template/package.json) Electron version 4.0.0 -> 6.0.0

#### Description

This is more as a temporary solution for convenience so that new users don't get the depreciation warning from Electron 4.0.0 as I have. I'm not sure if this was ever meant to happen but if not, I will look to see if I could add a `postinstall` script which simply runs `npm install` and ensures Electron stays up-to-date (if that's of interest to anyone).

comments welcome! ✌️ 

| Before | After |
| :--: |:--:|
| Before| After|

### DIFF
--- a/electron-template/package.json
+++ b/electron-template/package.json
@@ -10,7 +10,7 @@
     "electron-is-dev": "^1.1.0"
   },
   "devDependencies": {
-    "electron": "^4.0.0"
+    "electron": "^6.0.0"
   },
   "keywords": [
     "capacitor",


jcesarmobile:
thanks for the PR, but I'm closing it since there is a newer PR that targets electron 7 https://github.com/ionic-team/capacitor/pull/2084

Issue #1848, opened 2019-08-05
The WebView doesn't load url with ErrorCode -6, net::ERR_CONNECTION_REFUSED


LukaszGrela:
Description of the problem:

Affected platform
- [x] Android
- [ ] iOS - not tried as not needed
- [ ] electron - not tried as not needed
- [ ] web - not tried as not needed

OS of the development machine
- [x] Windows

Other information:

Capacitor version:
1.1.1

node version:
8.13.0

npm version:
6.4.1

CocoaPods version:
n/a

Steps to reproduce:
Implement the app that is redirecting to some server and then back to localhost (e.g. the implicit flow), the webview throws an error net::ERR_CONNECTION_REFUSED and displays the android web page for that error.

With the Chrome Dev Tools on all you have to do is to click enter in the address bar with "failing url" and application works. Something is preventing valid url from working but I havent been able to investigate what or where.

What I was able to do for tests only is to implement the `onReceivedError` on `webView.setWebViewClient` in `Bridge`'s `loadWebView`, reload the `failingUrl` when `errorCode == -6`

```
  private void loadWebView() {
    appUrlConfig = Config.getString("server.url");
    String[] appAllowNavigationConfig = Config.getArray("server.allowNavigation");

    ArrayList<String> authorities = new ArrayList<String>();
    if (appAllowNavigationConfig != null) {
      authorities.addAll(Arrays.asList(appAllowNavigationConfig));
    }
    this.appAllowNavigationMask = HostMask.Parser.parse(appAllowNavigationConfig);

    String authority = Config.getString("server.hostname", "localhost");
    authorities.add(authority);
    localUrl = CAPACITOR_SCHEME_NAME + "://" + authority;

    if (appUrlConfig != null) {
      try {
        URL appUrlObject = new URL(appUrlConfig);
        authorities.add(appUrlObject.getAuthority());
      } catch (Exception ex) {
      }

      if (BuildConfig.DEBUG) {
        Toast.show(getContext(), "Using app server " + appUrlConfig.toString());
      }
    }

    final boolean html5mode = Config.getBoolean("server.html5mode", true);

    // Start the local web server
    localServer = new WebViewLocalServer(context, this, getJSInjector(), authorities, html5mode);
    localServer.hostAssets(DEFAULT_WEB_ASSET_DIR);


    appUrl = appUrlConfig == null ? localUrl : appUrlConfig;

    Log.d(LOG_TAG, "Loading app at " + appUrl);

[...]

PR #1855, opened 2019-08-07
Handle custom WebViewClient (#1848)

LukaszGrela:
Add means to set custom WebViewClient that will be called along with internal one. This PR is to help with #1848 

## Usage
In your `MainActivity` get instance of the `Bridge` and call `setCustomWebViewClient` passing new `WebViewClient` implementation.

**Note** This will NOT replace original implementation, it will be called by original implementation.

```JAVA
public class MainActivity extends BridgeActivity {
  @Override
  public void onCreate(Bundle savedInstanceState) {

    // Initializes the Bridge
    this.init(savedInstanceState, new ArrayList<Class<? extends Plugin>>() {{
      // Additional plugins you've installed go here
      // Ex: add(TotallyAwesomePlugin.class);
    }});

    // Setup custom WebViewClient to reload when net::ERR_CONNECTION_REFUSED occurs
    this.getBridge().setCustomWebViewClient(new WebViewClient(){
      @SuppressWarnings("deprecation")
      @Override
      public void onReceivedError(WebView view, int errorCode, String description, String failingUrl) {
        Log.e(LogUtils.getCoreTag("MainActivity"), "onReceivedError: "+errorCode+", "+ description+", "+failingUrl);
        if(errorCode == -6) {
          // errorCode: -6, description: net::ERR_CONNECTION_REFUSED
          try {
            view.stopLoading();
          } catch ( Exception e ) {
            // ignore
          } finally {
            // reload
            Uri failingUri = Uri.parse(failingUrl);
            if(failingUri.getQueryParameter("cap_reloaded") == null) {
              Uri.Builder reloadUri = failingUri.buildUpon();
              reloadUri.appendQueryParameter("cap_reloaded", "true");
              view.loadUrl(reloadUri.build().toString());
            }
          }
        }
      }
      @TargetApi(android.os.Build.VERSION_CODES.M)
      @Override
      public void onReceivedError(WebView view, WebResourceRequest req, WebResourceError error) {
        // Redirect to deprecated method, so you can use it in all SDK versions
        onReceivedError(view, error.getErrorCode(), error.getDescription().toString(), req.getUrl().toString());
      }
    });
  }
}
```

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -1,17 +1,31 @@
 package com.getcapacitor;
 
+import android.annotation.TargetApi;
 import android.app.Activity;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.ApplicationInfo;
+import android.graphics.Bitmap;
 import android.graphics.Color;
 import android.net.Uri;
+import android.net.http.SslError;
+import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
+import android.os.Message;
+import android.support.annotation.Nullable;
+import android.support.annotation.RequiresApi;
 import android.util.Log;
+import android.view.KeyEvent;
+import android.webkit.ClientCertRequest;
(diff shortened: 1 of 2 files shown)


basher:
Any update on this (or the related bug)?

jcesarmobile:
closing since there was no response

PR #1856, opened 2019-08-07
Web Camera - use file input

mlynch:
This PR adds support for the native file input method for picking/taking a photo. This makes it possible to select a photo from the filesystem or use the native prompt provided on mobile.

Because this experience is likely a bit closer to what users would expect on mobile/desktop browsers, I've also changed the default so that the PWA Elements system is _opt-in_ instead of the default if installed. Technically, this is a breaking change, but I don't think it warrants a major release.


mlynch:
I need to do one more thing: hide the input so it doesn't mess w/ any UI.

jcesarmobile:
I think you can’t click it if hidden, but you can probably use absolute position to put it outside of the visible area

core/src/web/camera.ts
```diff
@@ -18,36 +18,109 @@ export class CameraPluginWeb extends WebPlugin implements CameraPlugin {
...
+    if (options.source === CameraSource.Photos || options.source === CameraSource.Prompt) {
+      input.removeAttribute('capture');
+    } else if (options.direction === CameraDirection.Front) {
+      (input as any).capture = 'user';
+    } else if (options.direction === CameraDirection.Rear) {
+      (input as any).capture = 'environment';
+    }
+
+    input.addEventListener('change', (_e: any) => {
+      console.log('CHANGE', _e);
```
jcesarmobile:
```suggestion
     
```
Remove the log, the whole event it's not very nice.

core/src/web/camera.ts
```diff
@@ -18,36 +17,99 @@ export class CameraPluginWeb extends WebPlugin implements CameraPlugin {
...
+    });
+  }
+
+  private fileInputExperience(options: CameraOptions, resolve: any, reject: any) {
+    let input = document.querySelector('#_capacitor-camera-input') as HTMLInputElement;
+
+    if (!input) {
+      input = document.createElement('input') as HTMLInputElement;
+      input.type = 'file';
+      input.accept = 'image/*';
```
muuvmuuv: Wouldn't it make more sense to use `accept=".png, .jpg, .jpeg, .gif"` and remove the if stements below? In the comment `core/src/core-plugin-definitions.ts:356` you mentioned that only those are possible, so I think restrict the file input to exactly those makes more sense.

aaronksaunders:
when will this be merged?

core/src/web/camera.ts
AntoninBeaufort: Personally, when I specify extensions in the "accept" attribute the file selector doesn't even show up it would be a good idea to keep `image/*` until most browsers and WebView support the specific extensions.

rdlabo:
This is great function . I looks forward to merge this🎉

bijanmmarkes:
When will this be merged? :+1: @mlynch

rdlabo:
I noticed that iOS Safari may not work input of createElement...
: https://stackoverflow.com/questions/47664777/javascript-file-input-onchange-not-working-ios-safari-only

User must prepare real DOM in index.html.

```
<div style="width: 0; height: 0; overflow: hidden;">
  <input id="browserPhotoUploader" type="file" accept="image/*" />
</div>
```

I think this is not smart. Is there any good way?

alexcroox:
Any movement on this 10 month PR? I'd like to use file input for both web and native app.

core/src/core-plugin-definitions.ts
```diff
@@ -310,6 +310,12 @@ export interface CameraOptions {
    * iOS only: The presentation style of the Camera. Defaults to fullscreen.
    */
   presentationStyle?: 'fullscreen' | 'popover';
+
+  /**
+   * Web only: whether to use the PWA Element experience or file input. The default
+   * is to use file input experience. Learn more about PWA Elements: https://capacitor.ionicframework.com/docs/pwa-elements/
```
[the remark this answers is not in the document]
mlynch: I think having the option still makes sense since the experience is different in both cases. I'd like to rename the flag, however

mlynch:
Alright this is updated, working, and ready for review

core/src/web/camera.ts
```diff
@@ -16,29 +18,104 @@ export class CameraPluginWeb extends WebPlugin implements CameraPlugin {
...
-      await cameraModal.componentOnReady();
-      cameraModal.addEventListener('onPhoto', async (e: any) => {
-        const photo = e.detail;
-
-        if (photo === null) {
-          reject('User cancelled photos app');
-        } else if (photo instanceof Error) {
-          reject(photo.message);
+      if (options.webUseInput === false && options.source === CameraSource.Camera) {
+        const cameraModal: any = document.createElement('pwa-camera-modal');
```
dwieeb, line 23: `cameraModal` can be undefined if PWA Elements isn't installed, do we want to add a message about it?

dwieeb:
Looks good! @jcesarmobile want to approve?

core/src/web/camera.ts
```diff
@@ -16,29 +18,109 @@ export class CameraPluginWeb extends WebPlugin implements CameraPlugin {
...
+          } as CameraPhoto);
+        } else if (options.resultType === CameraResultType.Base64) {
+          const b64 = (reader.result as string).split(',')[1];
+          resolve({
+            base64String: b64,
+            format
+          } as CameraPhoto);
+        } else {
+          reject('Unsupported result type for this platform');
+        }
```
jcesarmobile, line 107:
```suggestion
        }
```
No need for this else as the reader.readAsDataURL(file); (line 113) is only called for DataUrl and Base64

core/src/core-plugin-definitions.ts
```diff
@@ -393,7 +399,7 @@ export interface CameraPhoto {
    */
   exif?: any;
   /**
-   * The format of the image. Currently, only "jpeg" is supported.
+   * The format of the image, ex: jpeg, png, gif.
```
jcesarmobile, line 402: iOS and Android only allow `jpeg`, should be noted that `png` and `gif` are web only, and I think `gif` will only work if using the file input, not pwa elements.

PR #1888, opened 2019-08-20
[FEATURE] Android send intent support

tavosansal:
The motivation for this PR is that handling send intents with cordova is a pain. There are some plugins out there but they are all convoluted and hard to understand. Also the API is usually not very good leading to confusion.
The `App` plugin already handles the intent lifecycle right now for handling `VIEW` intents for deep linking. So instead of creating a plugin with another listener I figured this would be a good place to handle a `SEND` intent as well.
If this is not the right place or if I miss anything please let me know. This is my very first PR so feedback is welcome.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/App.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/App.java
@@ -4,6 +4,7 @@ import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.net.Uri;
+import android.os.Bundle;
 import android.util.Log;
 import com.getcapacitor.JSObject;
 import com.getcapacitor.NativePlugin;
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/App.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/App.java
@@ -19,6 +19,7 @@ public class App extends Plugin {
   private static final String EVENT_URL_OPEN = "appUrlOpen";
   private static final String EVENT_STATE_CHANGE = "appStateChange";
   private static final String EVENT_RESTORED_RESULT = "appRestoredResult";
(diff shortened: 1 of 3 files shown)


ryanholte:
I could really use this functionality. Will this ship?

jcesarmobile:
Thanks for the PR, but since this is Android only and not a very common feature, I won't be adding it to Capacitor.
I've heard you created a plugin with the same functionality, I think it's better that way.
You can send a PR to add your plugin to the community plugins list
https://github.com/ionic-team/capacitor/blob/master/site/docs-md/community/plugins.md

tavosansal:
@jcesarmobile ok I understand. I would say though that sharing things in android is crucial to the OS and native experience. So I disagree that this is not a very common use case since receiving information from other apps is super common for Android. 
I did think about the iOS use case but iOS is not straight forward since there is an extra layer of complexity where you need to build a share extension. There are also plenty of APIs in capacitor that are either iOS or Android only. So not sure why that would be a requirement for a new feature.
I will create a plugin for this but my only issue with that is that Capacitor is already listening to new `Intents` so the plugin would have to also listen which duplicates the work.

ropstah:
> Thanks for the PR, but since this is Android only and not a very common feature, I won't be adding it to Capacitor.
> I've heard you created a plugin with the same functionality, I think it's better that way.
> You can send a PR to add your plugin to the community plugins list
> https://github.com/ionic-team/capacitor/blob/master/site/docs-md/community/plugins.md

@jcesarmobile I have to agree with @tavosansal in that the sharing-mechanism is crucial. Also his statement about already having different API implementations per platform makes a lot of sense. 

Seeing [recent progressions](https://w3c.github.io/web-share-target/) in the PWA ecosystem around sharing it might be good to re-evaluate.

dwieeb:
There's a community proposal for this FYI: https://github.com/capacitor-community/proposals/issues/46

PR #1902, opened 2019-08-26
[FEATURE] Add openPackage function for android

tavosansal:
Currently there is no way to open the main activity for an android app. If that app does not handle a `VIEW` intent from a url then it is impossible to open the app at all (for example the instagram app). 
This PR adds a function to the `App` plugin so that it is able to open an app via its package name. If the app is not installed, the call to the plugin will fail just like `App.openUrl`

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/App.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/App.java
@@ -104,6 +104,24 @@ public class App extends Plugin {
     }
   }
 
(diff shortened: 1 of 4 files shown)


jcesarmobile:
Thanks for the PR, but that was actually a bug, I've fixed it in the linked PR.
openUrl was already accepting packages, but not handling what you mention about apps that don't handle VIEW intents. If you check canOpenUrl, it was returning true for instagram package.

PR #1905, opened 2019-08-27
fix(electron): correctly read/write file with no encoding

D34THWINGS:
I was working with the Filesystem plugin and some data URL files encoded in base64 and I stumbled upon the fact that Electron bridge is not implemented like the Android and iOS one. I simply cannot store a PDF file because it always encrypt as UTF8 and break the file.

On iOS and Android, the default encoding when writing files is `base64`. If a data URL is detected it deletes its header before converting it to `binary`. When reading, if no encoding is given, the same thing happens the other way, it converts `binary` to `base64`.

This is a breaking change for Electron users but ATM, Electron filesystem plugin does not behave at all like iOS and Android.

### DIFF
--- a/electron/src/electron/filesystem.ts
+++ b/electron/src/electron/filesystem.ts
@@ -57,21 +57,22 @@ export class FilesystemPluginElectron extends WebPlugin implements FilesystemPlu
 
   writeFile(options: FileWriteOptions): Promise<FileWriteResult> {
     return new Promise((resolve, reject) => {
-      if(Object.keys(this.fileLocations).indexOf(options.directory) === -1)
+      if (Object.keys(this.fileLocations).indexOf(options.directory) === -1)
         reject(`${options.directory} is currently not supported in the Electron implementation.`);
       let lookupPath = this.fileLocations[options.directory] + options.path;
-      let data = options.data.indexOf(',') >= 0 ? options.data.split(',')[1] : options.data;
+      let data: (Buffer | string) = options.data;
       if (!options.encoding) {
-        data = Buffer.from(data, 'base64')
+        const base64Data = options.data.indexOf(',') >= 0 ? options.data.split(',')[1] : options.data;
+        data = Buffer.from(base64Data, 'base64');
       }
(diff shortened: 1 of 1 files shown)


electron/src/electron/filesystem.ts
```diff
@@ -44,30 +44,35 @@ export class FilesystemPluginElectron extends WebPlugin implements FilesystemPlu
...
       let lookupPath = this.fileLocations[options.directory] + options.path;
-      this.NodeFS.readFile(lookupPath, options.encoding, (err:any, data:any) => {
+      this.NodeFS.readFile(lookupPath, options.encoding || 'binary', (err:any, data:any) => {
         if(err) {
           reject(err);
           return;
         }
 
-        resolve({data});
+        resolve({ data: options.encoding ? data : data.toString('base64') });
```
jcesarmobile: write is fine now, but read is returning the converted string, not the original base64 written string.
In example, if I write "e", in the file it's written as "hello", and if I read it, I get "hello" too, but I should get "e" as that's what I wrote. On Android and iOS it works like that.
So I think it should be something like this:
```suggestion
        resolve({ data: options.encoding ? data : Buffer.from(data).toString('base64') });
```

D34THWINGS: It bugs me because `fs.readFile` returns a Buffer to the callback so I shouldn't have wrap it again in a Buffer 🤔

jcesarmobile: Looks like it returns a string if you pass `'binary'` as encoding, but returns a Buffer if you pass `null` as encoding. At least that's what I'm getting.

D34THWINGS: The trick was to add `'binary'` as encoding when using `Buffer.from`. Works now! But I also saw that the `data:application/pdf;base64,` is not stripped on `web`. Should I fix it in this PR? Or should I make another one?

jcesarmobile: better a new one

PR #1911, opened 2019-08-29
Release configs are for distribution

josh-m-sharpe:
My fastlane builds have been failing with nefarious code signing issues, and it seems all related to this config. Excluding me and fastlane all together, It also seems that xcode will automatically make this change the first time any release provisioning profile is selected.


jcesarmobile:
I have no idea about fastlane, but I don't think this change is required, as far as I know, fastlane should be able to handle this. 
Can you provide more information about the commands you are running with fastlane?

josh-m-sharpe:
For whatever reason, fastlane's `match` and `update_project_provisioning` don't update this config - so I don't see a way to automate this change.  The only way I know to update it is to manually open xcode and select a distribution profile for the release build.

That wouldn't be necessary if this change was in place though.  It seems to make sense to me though, since why would you want a development profile in place, as default, for the Release build?

jcesarmobile:
`iPhone Developer` should work when `automatically manage signing` is selected (which is enabled by default in Capacitor project).

The possible values are `iPhone Development` (for development), `iPhone Developer` (should work for both development and distribution if automatic signing is enabled), and `iPhone Distribution` for distribution only.

So, can you answer the question I asked so I can look into it?

josh-m-sharpe:
We're using manual signing:
```
    app_id = 'my.app.id'

    disable_automatic_code_signing(
      path: "ios/App/App.xcodeproj",
    )

    match(
      app_identifier: app_id,
      type: 'development'
    )

    update_project_provisioning(
      xcodeproj: "ios/App/App.xcodeproj",
      build_configuration: 'Debug',
      profile: ENV["sigh_#{app_id}_development_profile-path"],
      code_signing_identity: 'iPhone Development'
    )

    match(
      app_identifier: app_id,
      type: 'appstore'
    )

    update_project_provisioning(
      xcodeproj: "ios/App/App.xcodeproj",
      build_configuration: 'Release',
      profile: ENV["sigh_#{app_id}_appstore_profile-path"],
      code_signing_identity: 'iPhone Distribution'
    )
```

The end result here is that both debug and release builds work, but only if I also make the change in this PR since these commands don't correctly make that change.

steel:
I handle this currently with Fastlane without having to do the changes in this PR. There's a subtle difference in my script than yours.

```
    update_project_provisioning(
      xcodeproj: PROJECT,
      build_configuration: BUILD_CONFIGURATION,
      target_filter: TARGET,
      profile: profile_path
    )

    provisioningProfilesMapping = {}
    provisioningProfilesMapping[app_identifier] = profile_name

    build_app(
      workspace: WORKSPACE,
      scheme: TARGET,
      configuration: BUILD_CONFIGURATION,
      output_directory: "../build",
      export_method: "app-store",
      export_options: {
        provisioningProfiles: provisioningProfilesMapping
      },
      codesigning_identity: CODESIGNING_IDENTITY
    )
```

I think the difference may be the `export_options` in `build_app`. I manually pass in name of the profile that `sigh` provides.

`ENV["sigh_com.<app-id>_appstore_profile-name"]`

jcesarmobile:
thanks for your input steel

going to close since `iPhone Developer` is a valid value for automatic signing and Apple recommends automatic signing, so we are not changing the value to `iPhone Distribution`. You can try steel's configuration

PR #1916, opened 2019-08-29
fix(app-template): use correct node_modules value in .gitignore

vially:


electron/.gitignore
```diff
@@ -1,2 +1,2 @@
 dist/
-node-modules/
\ No newline at end of file
+node_modules/
```
jcesarmobile: This one should be removed as it's already in the root `.gitignore`, so changing it to the correct one has no effect. I think it's better to just remove it.

vially: Fixed

PR #1925, opened 2019-09-04
fix(ios/cordova): Replace UIWebView with WKWebView in CDVUserAgentUtil

### LINKED ISSUE #1912: Apple will stop accepting apps using deprecated UIWebView api
Description of the problem:
I just received a mail from Apple notifing me that they will stop accepting apps using the deprecated UIWebView api.

> Dear Developer, 
>We identified one or more issues with a recent delivery for your app,  [redacted]. Your delivery was successful, but you may wish to correct the following issues in your next delivery:
>
> ITMS-90809: Deprecated API Usage - Apple will stop accepting   submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.
>
> After you’ve corrected the issues, you can use Xcode or Application Loader to upload a new binary to App Store Connect. 
>
> Best regards, 
> The App Store Team
--

Right now we are pushing our app only to TestFlight and not yet to the public app store, so I can't tell if this rule is already being enforced for submissions to the app store (despite the warning the app was still accepted by TestFlight), but this is definitely a sign that the UIWebView should be replaced with a WKWebView in capacitor as soon as possible.

Affected platform
- [ ] Android
- [x] iOS
- [ ] electron
- [ ] web

Capacitor version:
1.1.1

jcesarmobile:
Closes #1912

### DIFF
--- a/ios/CapacitorCordova/CapacitorCordova/Classes/Public/CDVUserAgentUtil.m
+++ b/ios/CapacitorCordova/CapacitorCordova/Classes/Public/CDVUserAgentUtil.m
@@ -32,6 +32,38 @@ static NSInteger gNextLockToken = 0;
 static NSInteger gCurrentLockToken = 0;
 static NSMutableArray* gPendingSetUserAgentBlocks = nil;
 
+#import <WebKit/WebKit.h>
+
+ @interface WKWebView(SynchronousEvaluateJavaScript)
+- (NSString *)stringByEvaluatingJavaScriptFromString:(NSString *)script;
+@end
+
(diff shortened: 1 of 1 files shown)


seanyang1984:
@jcesarmobile we are using ionic 3.16.0, this bug will be fixed in it ? can you please help me with the steps to be followed to use this change?

dotNetkow:
Update to the latest version of Capacitor. Full details [here](https://ionicframework.com/blog/understanding-itms-90809-uiwebview-api-deprecation/).

PR #1950, opened 2019-09-11
fix(core): return proper Error objects from native

diachedelic:
If I receive news of a Capacitor plugin error via sentry, there are no stack traces because `fromNative` (in `native-bridge.js`) throws a deserialized JSON object, not an `Error`. It's not even clear that the error is from a Capacitor plugin, as it's usually something like `{ message: "File not found" }`.

image

image

This pull request transforms the error objects into actual `Errors` before they are thrown (well, passed to the callback/rejected). Now, at least one can see that the error originated from the native bridge:

image

Note that this PR uses `Object.assign` - I'm not sure whether Capacitor supports platforms without it. This is the MDN compatibility table:

image

### DIFF
--- a/core/native-bridge.js
+++ b/core/native-bridge.js
@@ -217,6 +217,11 @@
       if (storedCall) {
         // looks like we've got a stored call
 
+        if (result.error) {
+          // ensure stacktraces
+          result.error = Object.assign(new Error(), result.error);
+        }
+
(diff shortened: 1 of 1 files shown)


diachedelic:
I figured `Object.assign` not being supported by Android WebView was probably a problem, so I replaced it with an old school solution.

Issue #1930, opened 2019-09-06
Type error: Illegal Invocation on plugins addListener


LukaszGrela:
Description of the problem:

Affected platform
- [x] Android

OS of the development machine
- [x] Windows

Other information:

Capacitor version:
1.2.0

node version:
8.13.0

npm version:
6.4.1

Steps to reproduce:
With the Capacitor 1.2.0 set some listeners to a plugin e.g. `backButton` or `networkStatusChange`. Build your project and update android app `npx cap copy`. 
Build Android app and run on the Android 5.1.1 emulator (API22). Notice in Logcat the `E/Capacitor/Console: File: https://localhost/ - Line 203 - Msg: TypeError: Illegal invocation` entries for each `addListener` on plugins.

You can also remotely debug the WebView chrome://inspect/#devices and in the console You will notice the error: 
>  TypeError: Illegal invocationmessage: "Illegal invocation"stack: "TypeError: Illegal invocation↵    at Object.capacitor.logToNative (http://localhost/:389:11)↵    at Object.toNative (http://localhost/:189:23)↵    at Object.capacitor.nativeCallback (http://localhost/:289:22)↵    at Object.capacitor.addListener (http://localhost/:305:32)↵    at Object.t.addListener (http://localhost/:857:22)↵    at Object.<anonymous> (http://localhost/static/js/main.794d4e58.chunk.js:1:993)↵    at a (http://localhost/:972:1111)↵    at Object.<anonymous> (http://localhost/static/js/main.794d4e58.chunk.js:1:226)↵    at a (http://localhost/:972:1111)↵    at c (http://localhost/:972:980)"get stack: function () { [native code] }set stack: function () { [native code] }__proto__: Errorlocalhost/:203 toNative

you can try to run following snippet in console which is also throwing the `TypeError: Illegal Invocation` error

```
Capacitor.Plugins.App.addListener('backButton', function(){console.log("back");});
```

**Note** to make the capacitor work on Android 5.1 you need to manually change line 70 of native-bridge.js to be:

```
  const consoleKeys = ["debug", "error", "info", "log", "warn", "dir", "dirxml", "table", "trace", "group", "groupCollapsed", "groupEnd", "clear", "count", "countReset", "assert", "profile", "profileEnd", "time", "timeLog", "timeEnd", "timeStamp", "context", "memory", "reactStack", "reactStackEnd"];
    // Object.keys(win.console)
    consoleKeys.forEach(function (level) {
```
[...]

PR #1956, opened 2019-09-13
Bug fix #1860, #1930

### LINKED ISSUE #1860: Uncaught TypeError: Cannot read property 'call' of undefined, not custom plugin related.
Description of the problem:

Affected platform
- [x] Android 5.0.2
- [ ] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [x] Windows
- [ ] macOS
- [ ] linux

Other information:
Run on emulator, the same app run on Android 6+ works fine.

Capacitor version:
1.1.1

node version:
8.13.0

npm version:
6.4.1

Steps to reproduce:
Create a project for android, 
run it on the Android 5.0 (device or emulator), 
use the `addListener` on `App` plugin
Inspect the WebView and notice the error in console: `Uncaught TypeError: Cannot read property 'call' of undefined`

Try to call following from the dev tools console:
`Capacitor.Plugins.App.addListener('appStateChange', function handler() {console.log("App State Changed");})` You will notice the `TypeError: Cannot read property 'call' of undefined` error. debug and it will point to the 
```
    } catch (e) {
      orgConsole.error.call(win.console, e); // this line
    }
```

further debug will show that `orgConsole` is empty! and was attemptet to populate with a line starting `Object.keys(win.console)`

Attempt to console log the `Object.keys(window.console)` will show taht only `memory` key is lsited (where is `log`, `error`?) It seems that on the WebView available for Android 5.0 the console is not iterable.

The same call on the dev tools on the Android 6+ lists all of available console methods. (e.g. Android 8 - 24 items listed)

Possible solution could be to create a static list of console props

```
const consoleLevels = [    
    'debug',
    'error',
    'info',
    'log',
    'trace',
    'warn',
    // anything else needed
];
// then
consoleLevels.forEach(function(level){
   // rest of the code
});
```

Link to sample project:
N\A

LukaszGrela:
This will fix problems reported as Issue #1860 and #1930


jcesarmobile:
Thanks for the PR.
As I said on  #1860, we have removed the console overwrite on android, but I missed a few things that were causing this issues you reported on #1930. That have been fixed in #1991, so your changes are not needed.

Also, as a rule of thumb, we [don't support](https://capacitor.ionicframework.com/docs/getting-started/dependencies) old webview/chrome versions, so issues on Android 5-6 emulators are not a priority and might not get fixed.

> Capacitor requires an Android WebView with Chrome version 50 or later. On Android 5 and 6, the Capacitor uses the System WebView. On Android 7+, Google Chrome is used.

PR #1967, opened 2019-09-16
fix(ios): call webview.load on main thread on setServerBasePath

jcesarmobile:


naranjamecanica:
Hi,

I know you're working very hard and I'm very grateful for this project, but could you write some (short) explanation **why** these changes where made? If I find strange behaviour I walk trough the history of commits and see if I can find anything why something is happening. 

thanks again and cheers! 👍 
Martijn

naranjamecanica:
Thank you!! If it's possible to give every commit some context would be super!

PR #1972, opened 2019-09-17
fix(ios): Change CAPBridgeViewController class access to open

playerx:
## Motivation:
I need to create new child class `CustomCAPBridgeViewController` from `CAPBridgeViewController` to extend its functionality. In my case I need to override `viewWillTransition`

## Problem:
When I created new `CustomCAPBridgeViewController` and inherited it from `CAPBridgeViewController` I got error from xcode. It seems that if you want to inherit class it should be declared as open (this change was introduced in Swift 3)

## Sources:
* https://stackoverflow.com/questions/39072300/xcode-8-cannot-inherit-from-non-open-class?answertab=votes#tab-top

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -7,7 +7,7 @@ import UIKit
 import WebKit
 import Cordova
 
-public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScriptMessageHandler, WKUIDelegate, WKNavigationDelegate {
+open class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScriptMessageHandler, WKUIDelegate, WKNavigationDelegate {
   
   private var webView: WKWebView?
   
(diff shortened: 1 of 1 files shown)


jcesarmobile:
I'm not really sure if we want people to subclass CAPBridgeViewController.

For extending it you can create an extension like this:

```
import Capacitor

extension CAPBridgeViewController {
    override public func viewWillTransition(to size: CGSize,
                            with coordinator: UIViewControllerTransitionCoordinator) {
        print("view will transition")
    }
}
```

jcesarmobile:
Will close for now as the extension is enough for the use case.

Might reconsider in the future if we find a real use case for subclassing CAPBridgeViewController that can't be done with an extension.

codeincontext:
In my case I need access to `webView(_ webView: WKWebView, didFailProvisionalNavigation navigation: WKNavigation!, withError error: Error)`. I'm not the best with swift but I see the following options:

- Fork capacitor for the time being
- Capacitor gets an API to forward WKNavigationDelegate methods
- Capacitor opens CAPBridgeViewController and WKNavigationDelegate methods
- Swizzling?

For now I'm going to open up CAPBridgeViewController locally and subclass in my project

fkirc:
I believe that subclassing is helpful for "embedded capacitor".

In my app, I have two different CAPBridgeViewController's for two different features of the app.
Furthermore, I have a custom plugin that behaves differently depending on which CAPBridgeViewController is currently shown.
It would be helpful to subclass CAPBridgeViewController in order to customize this behavior at runtime.

Note that a simple extension like above does not work for my case because I want to differentiate between different CAPBridgeViewControllers.

Overally, I believe that "embedded capacitor" is still missing some features and helper functions (e.g. configure a URL postfix dynamically).
Moreover, I believe that more openness is the only short-term action that makes embedded capacitor possible (except of forking or code-copying, which are cumbersome ways of openness).
Therefore, do you consider to re-open and merge this PR?
After all, it is much less work for you to enable customizations instead of having to implement those customizations in the core library.

PR #1975, opened 2019-09-18
fix(android template): drop orphaned file provider authority variable

zyniq82:
This re-instates what was done in 94b72ccfec, undone in e048d2b96d and later "fixed" in 1642584d19, which orphaned the `fileprovider_authority` variable in `strings.xml`.

As I'm not an Android developer, I'm not sure if the proper fix is to do what I just did now, re-instating the reference to the `fileprovider_authority` variable, or if it's better to keep `AndroidManifest.xml` unchanged and instead just remove the `fileprovider_authority` variable from `strings.xml`.

### DIFF
--- a/android-template/app/src/main/AndroidManifest.xml
+++ b/android-template/app/src/main/AndroidManifest.xml
@@ -33,7 +33,7 @@
 
         <provider
             android:name="android.support.v4.content.FileProvider"
-            android:authorities="${applicationId}.fileprovider"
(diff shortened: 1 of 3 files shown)


zyniq82:
I've updated as requested, so now the only change is simply dropping the orphaned `fileprovider_authority` variable from `strings.xml`.

I've also updated the pull request title to better reflect the change.

PR #1857, opened 2019-08-08
docs: improve Android custom code section

RoderickQiu:
- Android custom code guide modified. 
> I was stuck on that the other day and I've solved it. And today I saw this article again and decide to modify it.
I think it's now better for common users to comprehend.
- Some style changes.

PR #2015, opened 2019-10-01
docs: Added JS example to iOS custom-code section

nip3o:
It's already well-documented, but adding a JS example means that you can just copy-paste instructions and start playing around right away.


nip3o:
Yeah, I agree.

We just released our (Capacitor-based) app so I have limited time for contributing right now, but I'll have a look at my PR:s this or next week.

nip3o:
Turns out a similar section was added to the Android docs in https://github.com/ionic-team/capacitor/pull/1857

PR #2023, opened 2019-10-03
fix(core): improve Filesystem parameter descriptions

nip3o:


nip3o:
I skimmed through all docstrings in the entire file, seems ReaddirOptions and StatOptions were the only incorrect ones. Also slightly changed the ReaddirOptions text, see if you like it better now.

Issue #1097, opened 2019-01-04
How to set additional wkWebView settings


timboozle:
I am using Capacitor without Ionic and I want to set additional WKWebView settings like:
- allowsBackForwardNavigationGestures
- allowsLinkPreview
- scrollView.bounces

How would this be possible?

jcesarmobile:
It's not possible at the moment

timboozle:
Thanks, so it's not possible at all to set those variables? Or is there a workaround maybe? Sorry, no experience with Swift but would like to have those things in my app

jcesarmobile:
the only "workaround" is to manually edit the code to set the values you want

chris-carrington:
Thank you ionic team for all your contributions to the community, allowing the ability to "allowsBackForwardNavigationGestures" would be much appreciated

mlynch:
Thanks, certainly something we could explore adding down the road

realityfilter:
On our project we had the need to register a custom url scheme handler. This can only
be done before the webview is initialized. I think one could easily refactor `CAPBridgeViewController`
to have factory methods for setting up the webview and the webview configuration and make the 
class and the factory methods open. Within a subclass `CAPBridgeViewController` one would be able to override as needed and specify the controller in the main storyboard.

If this is a viable solution I could make a PR for this. In the meantime we swizzle `WKWebViewConfiguration`.

macdja38:
What does everyone think of adding a static method that plugins can implement to allow them to modify the WKWebViewConfiguration?

Something like this
```swift
    public static override func configureWKWebView(_ configuration: WKWebViewConfiguration) {
        CAPLog.print("Wouldn't it be cool if this got called?")
        configuration.ignoresViewportScaleLimits = true
    }
```
My only issue would plugins might step on each others toes, but I'm not sure how to possibly get around that.

jcesarmobile:
since the issue is about the webview settings, those can be changed at any time, and as plugins have a reference to the webview, they can change any of the settings, so I'll just close.

@realityfilter @macdja38 as the `WKWebViewConfiguration` is a different issue, maybe you should create a new issue about it, or we can just keep @macdja38 PR as reference. Or you can send a separate PR since you mentioned a different idea.

jcesarmobile:
In capacitor 3 you can use your own WebView, so you can configure it the way you want

for changing values during runtime, better create a plugin of your own that exposes the properties you need

mb21:
[...]

PR #2028, opened 2019-10-04
Initial support for allowing plugins to modify the Webview Configuration

macdja38:
Solves https://github.com/ionic-team/capacitor/issues/1097

Allows a plugin to modify the WKWebViewConfiguration as follows:
```
import Foundation
import Capacitor

@objc(IgnoreViewPortScaleLimitsPlugin)
public class IgnoreViewPortScaleLimitsPlugin: CAPPlugin {
    public static override func configureWKWebView(configuration: WKWebViewConfiguration) {
        configuration.ignoresViewportScaleLimits = true
    }
}
```

Drawbacks at the moment is nothing stops two conflicting modifications from conflicting, no simple way of warning the user that a plugin is overwriting another plugin, not sure if plugins are loaded consistently so it's hard to say if things will fail in consistent ways.

Not sure where to document this, seems it should be documented in the same place as `load` but I can't find where load is documented… Maybe It should be here? https://capacitor.ionicframework.com/docs/plugins/ios But that's a guide not really documentation…

If someone from the capacitor team could let me know where to document it along with if they're interested in merging this / know of a better way to go about this please let me know then I'll finish it up and remove the draft status.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridge.swift
+++ b/ios/Capacitor/Capacitor/CAPBridge.swift
@@ -240,6 +240,20 @@ enum BridgeError: Error {
   func reset() {
     storedCalls = [String:CAPPluginCall]()
   }
+
+  /**
+   * Give all plugins that will be registered a chance to modify the Web View configuration before creating it
+   */
+  static func runWKWebViewConfigurationHooks(configuration: WKWebViewConfiguration) {
+    var numClasses = UInt32(0);
+    let classes = objc_copyClassList(&numClasses)
+    for i in 0..<Int(numClasses) {
+      let c: AnyClass = classes![i]
+      if class_conformsToProtocol(c, CAPBridgedPlugin.self) {
+        c.configureWKWebView?(configuration)
+      }
+    }
+  }
   
   /**
    * Register all plugins that have been declared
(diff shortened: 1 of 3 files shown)


eliellis:
I’m currently experiencing an issue that could be totally fixed by this! If anyone from the team could respond to this that’d be awesome.

skyleguy:
would love if the ignoresViewportScaleLimits field was configurable from my codebase!

dwieeb:
Thanks for the PR! This doesn't feel like the right way to solve this use case. We'd be open to looking at a PR that adds a configuration value for Capacitor that would alter `ignoresViewportScaleLimits` in WKWebView, but maybe it's best to wait. We're about to start major refactors for Capacitor 3.

I created https://github.com/ionic-team/capacitor/issues/3179 to track this.

johannds:
> Thanks for the PR! This doesn't feel like the right way to solve this use case. We'd be open to looking at a PR that adds a configuration value for Capacitor that would alter `ignoresViewportScaleLimits` in WKWebView, but maybe it's best to wait. We're about to start major refactors for Capacitor 3.
> 
> I created #3179 to track this.

Allowing this will allow some people to make bad choices and break things. But it also allows people to have a huge amount of control over the webview when they know what they're doing.

Issue #1871, opened 2019-08-13
Improvement: Splash with LaunchScreen Storyboard and use Splash when app not active


benharpers:
Currently, the Splash Screen plugin uses it's own ViewController that only shows the background image. A developer may have other items like logos or text on their LaunchScreen, so it should use the LaunchScreen Storyboard instead of making its own.

Secondly, since there is no way for javascript to intercept the platform.pause or app.state changes fast enough to display the splash screen, it would be helpful if the Splash plugin could handle showing itself when the app becomes inactive. This is for apps where you don't want to show the app when it's in the background or immediately after it becomes active in case you have some authentication code you want to run first.

Here's the updated code to accomplish both of the above, on iOS at least:

**SplashScreen.swift**
```
import Foundation
import AudioToolbox

@objc(CAPSplashScreenPlugin)
public class CAPSplashScreenPlugin: CAPPlugin {
    
  var viewController = UIViewController();
    
  var spinner = UIActivityIndicatorView()
  var showSpinner: Bool = false
  var call: CAPPluginCall?
  var hideTask: Any?
  var isVisible: Bool = false

  let launchShowDuration = 3000
  let launchAutoHide = true

  var pauseAutoShow = false
  var resumeAutoHide = false
  var resumeShowDuration = 3000
  
  let defaultFadeInDuration = 200
  let defaultFadeOutDuration = 200
  let defaultShowDuration = 3000
  let defaultAutoHide = true

  public override func load() {
    buildViews()
    showOnLaunch()

    NotificationCenter.default.addObserver(self, selector: #selector(pause), name: UIApplication.willResignActiveNotification, object: nil)
    
    NotificationCenter.default.addObserver(self, selector: #selector(resume), name: UIApplication.didBecomeActiveNotification, object: nil)
  }
  
  // Handle pause notification
  @objc func pause(_ notification: Notification) {
    
    let showOnPause = getConfigValue("pauseAutoShow") as? Bool ?? pauseAutoShow

    if showOnPause {
      showSplash(showDuration: defaultShowDuration, fadeInDuration: 0, fadeOutDuration: 0, autoHide: false, backgroundColor: nil, spinnerStyle: nil, spinnerColor: nil, completion: { }, isLaunchSplash: false)
    }
  }
  
  // Handle resume notification
  @objc func resume(_ notification: Notification) {
    
    let hideOnResume = getConfigValue("resumeAutoHide") as? Bool ?? resumeAutoHide

    if hideOnResume {

[...]

PR #2033, opened 2019-10-08
Allow contentMode to be configured in SplashScreen

typeoneerror:
It's not currently possible to use SplashScreen plugin with any size constraints but aspect fill. This adds a new config param to SplashScreen called "contentMode" which is an Int representing a `UIView.ContentMode` enum.

First time writing swift, so forgive me if this is off. Happy to be enlightened.


jcesarmobile:
I would prefer what was proposed on [this issue](https://github.com/ionic-team/capacitor/issues/1871).
Using the `LaunchScreen.storyboard` view instead of having a separate view that mimics it, so whatever change you make in the storyboard (adding text, changing contentMode of the image view, adding a background color, etc) will also be done in the splash view as it's the same.

PR #2044, opened 2019-10-12
fix(electron): Check whether the splash screen is already destroyed on close

### LINKED ISSUE #2010: Splash Screen is destroyed on macOS
Description of the problem:
When closing the main window on macOS, the following error is displayed.
```
A JavaScript error occurred in the browser process
Error: Object has been destroyed
    at BrowserWindow.<anonymous> (/Users/.../node_modules/@capacitor/electron/index.js:115:25)
    at BrowserWindow.emit (events.js:205:15)
```
This error does not occur on Windows. The method in question is 
```js
this.mainWindowRef.on('closed', () => {
    this.splashWindow.close();
});
```
On Windows, `this.splashWindow.isDestroyed()` returns false before close is called. On macOS, `isDestroyed()` returns true. The result is a popup window visible to the user.

Affected platform
- [ ] Android
- [ ] iOS
- [x] electron
- [ ] web

OS of the development machine
- [ ] Windows
- [x] macOS
- [ ] linux

Capacitor version: 1.2.0

node version: 12.4.0

npm version: 6.9.0

walkingriver:
Resolves #2010 on MacOS. 

The issue stems from the splash screen being destroyed already, so I'm simply adding a check to ensure that the splash screen reference has a value and that it isn't already destroyed. This seems to solve the problem in a straightforward manner.

### DIFF
--- a/electron/index.js
+++ b/electron/index.js
@@ -113,7 +113,9 @@ class CapacitorSplashScreen {
     `;
 
     this.mainWindowRef.on('closed', () => {
-      this.splashWindow.close();
+      if (this.splashWindow && !this.splashWindow.isDestroyed) { 
+        this.splashWindow.close(); 
+      }
     });
 
     this.splashWindow.loadURL(`data:text/html;charset=UTF-8,${splashHtml}`, {baseURLForDataURL: `file://${rootPath}/splash_assets/`});
--- a/electron/index.js
+++ b/electron/index.js
@@ -113,7 +113,7 @@ class CapacitorSplashScreen {
     `;
 
     this.mainWindowRef.on('closed', () => {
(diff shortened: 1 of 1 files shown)


electron/index.js
```diff
@@ -113,7 +113,9 @@ class CapacitorSplashScreen {
     `;
 
     this.mainWindowRef.on('closed', () => {
-      this.splashWindow.close();
+      if (this.splashWindow && !this.splashWindow.isDestroyed) {
```
benjcallaghan: `isDestroyed` is a function, not a property. This condition will always return false.

walkingriver: Fixed with latest commit. Thanks, @benjcallaghan!

PR #2071, opened 2019-10-20
docs: reword the Periodic Maintenance section

jcesarmobile:
The Periodic Maintenance section looks confusing to me, it sounds like the project gets borked or something, and also says update command updates dependencies, which it doesn't. 
So renaming the section and rewording a little to not mislead.


jcesarmobile:
No, section 5 explains how to update Capacitor itself (the capacitor packages), section 4 tells the users they need to run update command after installing a Cordova or Capacitor plugin

PR #2077, opened 2019-10-22
Update index.md

bibyzan:
It isn't made clear enough that android returns to you a FCM token when iOS doesn't. I've watched several people in the community have to figure this inconsistency out for themselves when a note could be made here and save all of them time, so let me know if I would need to go about including this information another way.


jcesarmobile:
I would put it in the "[guide](https://capacitor.ionicframework.com/docs/guides/push-notifications-firebase/)", not in the push docs.

PR #2086, opened 2019-10-23
feat(device): Add operatingSystem field

Southgarden116:
- Updateed the Documnetation to show possible return values.
- When calling device.getInfo(), platform returns 'web' when using the mobile browser on iOS or Android. However, I expect to get the actual platform.


Southgarden116:
@jcesarmobile isn't that what Capacitor.isNative is for to check for native feature stuff?

jcesarmobile:
Sorry, but platform is the Capacitor platform. iOS, Android are operating systems, not platforms. 
If you check cordova device plugin it also returns browser for browser platform.

If you want to add a new OS or operatingSystem or similar property I'm ok with that, but we won't be changing platform value.

PR #2087, opened 2019-10-23
docs(Filesystem): use await in writeFile example

bvx89:
Updated the example-code for the `Filesystem.writeFile({ ... })`-method to include the `await` keyword.


bvx89:
[https://capacitor.ionicframework.com/docs/apis/filesystem#method-writeFile-0](https://capacitor.ionicframework.com/docs/apis/filesystem#method-writeFile-0)

It says in the documentation that it returns a promise. It's not _strictly_ necessary to await, but the example code uses a try-catch that would never catch anything.

Issue #1611, opened 2019-05-31
Make writeFile on iOS create intermediate directories


mrodal:
On android the writeFile function creates intermediate directories. On iOS looks like it doesnt.
Can we at least have an option to make it create the intermediate directories?

If not, could this be detailed on the docs? That on android it creates them and on iOS it doesnt?
Thanks

PR #2105, opened 2019-10-28
Add note to point out iOS folder requirement

bitflower:
I ran into this trap and this hint would have saved me.


jcesarmobile:
thanks, but it's really a bug, writeFile should create intermediate directories on iOS, as default or with an option, already reported https://github.com/ionic-team/capacitor/issues/1611

PR #2117, opened 2019-10-30
capacitor-admob plugin currently support iOS to remove android only

KSayed91:
based on it's [github repo](https://github.com/rahadur/capacitor-admob)


jcesarmobile:
I don't see any iOS working code on the repo

PR #2124, opened 2019-10-31
feat(LocalNotifications): Add support for count to work with every on LocalNotificationSchedule

dotrub:
The notification schedule already has support for repeating on a period of time, via the `every` option. However, it only allows the user to specify one of the options in the enum, with no support for repeating at multiples of the unit of time. This adds a `count` option to LocalNotificationSchedule (inspired by the structure of the notification trigger in https://github.com/katzer/cordova-plugin-local-notifications), which works in concert with the `every` option to set repeating notifications on multiples of the units of time. If no `count` is provided, it defaults to one, so this change should be backwards compatible. Implemented support for iOS and Android, but the web plugin didn't have support for `every`, so I didn't add any support for `count` there.

I'm open to suggestions on the exact structure. I prioritized a structure that would not break any code for people who have already implemented reminders with the `every` option.

- [x] Add `count` to type definition of LocalNotificationSchedule
- [x] Support `count` on iOS
- [x] Support `count` on Android
- [x] Add button to example app for testing `count`
- [ ] Add documentation for `count` option

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationSchedule.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationSchedule.java
@@ -44,7 +44,7 @@ public class LocalNotificationSchedule {
   }
 
   private void buildCountElement(JSObject schedule) {
-    this.count = schedule.getInteger("count");
+    this.count = schedule.getInteger("count", 1);
   }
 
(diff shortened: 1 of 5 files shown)


dotrub:
@jcesarmobile Sure, I'll add your changes in. Sorry about the permissions; the "Allow edits from maintainers" checkbox is checked, so I'm guessing it's the whole repo you can't write to? Can I fix that after that fact?
But I'll add those changes now anyway. Thanks for the review!

PR #2153, opened 2019-11-09
Clarify Toast duration Documentation.

ignacio-chiazzo:
The documentation doesn't clarify how duration works. 

Link: https://capacitor.ionicframework.com/docs/apis/toast/

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -1612,6 +1612,9 @@ export interface ToastPlugin extends Plugin {
 
 export interface ToastShowOptions {
   text: string;
+    /**
+   * It can be 'short' (3000 ms) or 'long' (5000 ms). The default value is 'short'.
+   */
   duration?: 'short' | 'long';
 }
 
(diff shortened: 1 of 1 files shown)


ignacio-chiazzo:
Thanks for the quick review @danielsogl.    The document you referenced has a link to 
`<plugin-api name="toast"></plugin-api>` so I edited the schema.  I based on this code: https://github.com/ionic-team/capacitor/blob/4a1c19f361641df8ebee4b624d1f9c1f6de50548/core/src/web/toast.ts#L13-L22.

core/src/core-plugin-definitions.ts
```diff
@@ -1612,6 +1612,9 @@ export interface ToastPlugin extends Plugin {
 
 export interface ToastShowOptions {
   text: string;
+    /**
```
jcesarmobile: Can you align the * in the comments?

```suggestion
  /**
```

jcesarmobile:
Just checked the source code and looks like every platform if using different values. 

For android it's 2000-3500 (and can't be changed because those are Android constants)
For iOS it's 1500-3000
For web/electron it's 3000-5000.

So I think first the iOS and web values should be changed to match Android ones (as those can't be changed) and then document those values.

Sorry for requesting changes before checking.

jcesarmobile:
closing until they all have the same value https://github.com/ionic-team/capacitor/issues/2175

PR #2182, opened 2019-11-16
docs(ce-plugins): + capacitor-plugin-file-transfer

waptik:
Added { FileTransfer } from (https://github.com/areo/capacitor-plugin-file-transfer)[https://github.com/areo/capacitor-plugin-file-transfer]


jcesarmobile:
Thanks for the PR, but community plugins should be submitted by the plugin authors, if they want the plugin displayed. This one doesn't belong to you.

This is because we review the plugins and might need to tell the authors to do some required changes in the plugin before being published on the docs. In this case the plugin lacks of install instructions and that's a common reason for not publishing the plugin.

PR #2198, opened 2019-11-20
fix(cordova/ios): post CDVPluginHandleOpenURLNotification

jcesarmobile:
Post `CDVPluginHandleOpenURLNotification` with the url that triggered it so cordova plugins depending on it can listen and to their things


lm3m:
looks good to me
:shipit:

PR #2224, opened 2019-12-05
feat(android-template): allow HTTP for debug builds

tlancina:
This means livereload from localhost will work with no additional
configuration.

### DIFF
--- a/android-template/app/src/debug/AndroidManifest.xml
+++ b/android-template/app/src/debug/AndroidManifest.xml
@@ -1,0 +1,4 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android">
+    <application android:usesCleartextTraffic="true"></application>
+</manifest>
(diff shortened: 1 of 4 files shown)


tlancina:
Added it only for `localhost`, as this is what the starters currently do: https://github.com/ionic-team/starters/blob/master/integrations/cordova/resources/android/xml/network_security_config.xml.

Since a nice out of the box `--livereload` experience is what I'm going for, there is some work that needs to be done in the CLI to forward the ports as well as some documentation (if you're using `--external` you'll need to update `network_security_config.xml`.  I feel like this is a good balance between default development experience and not shooting yourself in the foot by having to explicitly opt in to override the network config.

jcesarmobile:
I've sent a PR with what I proposed, WDYT? https://github.com/ionic-team/capacitor/pull/2395

this way, Ionic CLI should have to add this to the `capacitor.config.json` when doing `ionic capacitor run android -l --external`

```
"server": {
    "cleartext": true,
    "url": "http://local.ip.here:8100"
  }
```

PR #2239, opened 2019-12-12
Fix browserFinished not fired on Android

### LINKED ISSUE #2202: Browser.addListener browserFinished not fired on android
browserFinished not fired when close Browser on android. on iOS works fine

Justin-Credible:
This adds the missing `browserFinished` event not being fired on the Android version of the Browser plugin.

This also uses a flag to ensure that the event isn't fired when the browser is open, but the user backgrounds the entire app.

Resolves #2202

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Browser.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Browser.java
@@ -31,6 +31,7 @@ public class Browser extends Plugin {
 
   private CustomTabsClient customTabsClient;
   private CustomTabsSession currentSession;
+  private boolean appIsPaused = false;
 
   @PluginMethod()
   public void open(PluginCall call) {
@@ -117,6 +118,7 @@ public class Browser extends Plugin {
   }
 
   protected void handleOnResume() {
+    appIsPaused = false;
     boolean ok = CustomTabsClient.bindCustomTabsService(getContext(), CUSTOM_TAB_PACKAGE_NAME, connection);
     if (!ok) {
       Log.e(getLogTag(), "Error binding to custom tabs service");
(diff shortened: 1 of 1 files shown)


Justin-Credible:
Hmm. The `appIsPaused` flag works fine on the emulator, but behaves differently on physical devices...

Justin-Credible:
I removed the `appIsPaused` flag check that I had in previously.

As the PR stands now, this fixes the issue for the 99% use case; the `browserFinished` is now fired when the browser is closed.

However, there is an edge case; `browserFinished` will be fired when the user opens the Android OS task switcher by tapping the square button (or equivalent) on the device. This causes `TAB_HIDDEN` to fire, which then fires `browserFinished`. The user can then switch back to the app, but the browser of course is still open. The `browserFinished` shouldn't fire in this case.

A similar thing happens when the user invokes the Android OS home button, but in this case the end result is fine; since the browser is closed automatically in this case it is okay that `browserFinished` is fired.

jcesarmobile:
Since in your approach the browserFinished was firing when tapping the square button, despite Browser wasn't closed, I've come with a different approach that seems more reliable. Can you give it a try?
https://github.com/ionic-team/capacitor/pull/2332

jcesarmobile:
Since there has been no response or feedback, I'm going with my approach.

Justin-Credible:
Hello again @jcesarmobile ! Sorry to bump this old thread, but I'm still having an issue with the change that was made. It doesn't appear to work on the Pixel 2 or 3 emulators because the `handleOnResume` and `handleOnPause` run in a different order than they do on a physical device. I believe I ran into the same issue with my first commit https://github.com/ionic-team/capacitor/pull/2239/commits/4a5eac884a6f20a5bb57af68bce5a58d8b61dd50

I can create a new issue if needed, but wanted to check with you here first. Thanks!

PR #2266, opened 2019-12-17
fix(android): implement getDeliveredNotifications and removeDeliveredNotifications

p7g:
This PR implements the `PushNotifications` plugin `getDeliveredNotifications` and `removeDeliveredNotifications` methods.

The fields available on the objects returned by `getDeliveredNotifications` are a subset of those available on iOS, but otherwise the behaviour should be the same.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -119,7 +119,14 @@ public class PushNotifications extends Plugin {
         if (notification != null) {
           jsNotif.put("title", notification.extras.getCharSequence(Notification.EXTRA_TITLE));
           jsNotif.put("body", notification.extras.getCharSequence(Notification.EXTRA_TEXT));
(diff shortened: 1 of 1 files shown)


trytuna:
This is 23 days old. Can we please merge that?

jcesarmobile:
How I'm supposed to use the removeDeliveredNotifications? If I first use getDeliveredNotifications and pass the result to removeDeliveredNotifications, all the ids are 0, so doesn't remove the notifications.

For local notifications it works since we set the id, but it's confusing since the PR is about Push notifications

p7g:
@jcesarmobile In my use-case I have a subclass of `FirebaseMessagingService` that sets the IDs of incoming notifications. I understand that not everyone will want to have this, but I think the problem of IDs  being 0 is a different issue than the one this PR addresses.

jcesarmobile:
yeah, but on a regular Capacitor app this will happen, and I'm concerned about users complaining about it

p7g:
I think that having these methods implemented is better than not.

I agree that users may encounter the issue you mentioned, but we can make a new issue to find out why the IDs of the notifications are 0.

Issue #2230, opened 2019-12-07
feat: provide WebViewClient accessor on bridge


jcesarmobile:
Some plugins or scripts might want to modify or augment the WebViewClient.

Android webView has a getWebViewClient method, but requires API 26.

Would be good if the bridge could provide a getWebViewClient method itself that returns the WebViewClient being set by Capacitor.

Also we should use a WebViewClient subclass that can be subclassed to make it easier to override unimplemented methods (or the implemented ones if needed).

cepages:
@jcesarmobile is there any documentation that helps me to use this feature? Thanks

PR #2271, opened 2019-12-19
feat(android/Bridge): allow custom ssl certs in dev

nklayman:
This PR will set the webview to ignore ssl certificate errors in development. This will allow people to use https localhost servers with a self signed certificate.


jcesarmobile:
Thanks for the PR, but we don't want to disable SSL checks, not even for dev builds because it's insecure and having different behaviors in dev and release modes have caused a lot of problems to Cordova users in the past and we don't want to repeat the same mistakes.

We have few issues and PRs with different possibilities of what could be done to allow users or plugins to change the WebViewClient's behavior such as https://github.com/ionic-team/capacitor/issues/2230

PR #2276, opened 2019-12-22
fix(cli): use proper kebab-case name for npm package

alappe:
The recommended case of the plugin-name is actually
[kebab-case][wikipedia] not snake-case.

[wikipedia]: https://en.wikipedia.org/wiki/Letter_case#Special_case_styles

### DIFF
--- a/site/docs-md/plugins/index.md
+++ b/site/docs-md/plugins/index.md
@@ -34,7 +34,7 @@ This starts a wizard prompting you for information about your new plugin. For ex
 ```bash
 npx @capacitor/cli plugin:generate
 ✏️  Creating new Capacitor plugin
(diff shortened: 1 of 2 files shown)


site/docs-md/plugins/index.md
```diff
@@ -34,7 +34,7 @@ This starts a wizard prompting you for information about your new plugin. For ex
 ```bash
 npx @capacitor/cli plugin:generate
 ✏️  Creating new Capacitor plugin
-? Plugin NPM name (snake-case): my-plugin
+? Plugin NPM name (kebab-case): my-plugin
```
jcesarmobile: good catch, can you also change it in line 47?

PR #2280, opened 2019-12-24
docs(usage): change example function definition

korzio:
As it is in the example session, it most probably should be a function, not a method of a class notation


korzio:
Or an async arrow function, what fits more in a code style?

jcesarmobile:
thanks for the PR

At the moment the code usage in the Example section is a bit inconsistent, I've raised an issue for that https://github.com/ionic-team/capacitor/issues/2313

won't be merging yours because it adds a new inconsistency since none of the current examples is used as a function

PR #2282, opened 2019-12-27
fix(ios): fire appStateChange in incoming calls

kajinka13:
I changed the monitoring event because with the previous one it is not possible to intercept the action of entry on the task manager or the arrival of telephone calls, with this event instead it is possible to have the various events (for the purposes I have undertaken with the above plugin is essential to make this change).


313hemant313:
How to detect incoming call and call has ended if the ionic app is running in background ? 

Currently in my music app if user picks up the incoming call, music doesnt pause or stops.

Please help me resolve this issue.

PR #2302, opened 2020-01-07
feat(android): add splashFullScreen configuration option

unitree-czk:
For android, it is useful to hide the status bar right from the start. 
The user can add following setting:
    "SplashScreen": {
      "splashFullScreen": true
    }


jcesarmobile:
would be good if you could send another PR documenting the preference

PR #2307, opened 2020-01-08
Allow plugins to override navigation

sbannigan:
Allows plugins to override the webview navigation. This functionality was present in cordova so attempting to reimplement it for capacitor.


vgb1993:
Hello, in my company we need this behaviour as well, any updates?

PR #2318, opened 2020-01-10
docs(core): show array type instead of any

### LINKED ISSUE #2305: docs: unclear what BrowserPrefetchOptions can accept
BrowserPrefetchOptions
```
interface BrowserPrefetchOptions {
urls : any;
}
```
can we pass a single URL as a string? Can we pass multiple URLs as a string?

`any` without any further information isn't very helpful 😢

jcesarmobile:
closes #2305

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@capacitor/core",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


mesqueeb:
@jcesarmobile I had no idea the type information of the docs is manually typed. Isn't it better to auto-generate that portion of the docs for each page ? I think it'll be less error prone that way. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

jcesarmobile:
It uses typedoc to generate json of the types and the json is used to create the docs with this script.
So it’s not manual, but not 100% automated neither.

But yeah, there is probably a lot of room for improvement.

mesqueeb:
@jcesarmobile I see now why you're flipping your table. 
(ノಠ益ಠ)ノ彡┻━┻

PR #2249, opened 2019-12-14
Adds the ability to see whether the Push Notifications permission was allowed/declined

Livijn:
This pr adds the ability to see whether the Push Notifications permission was allowed/declined. Basically this adds an event at the `requestAuthorization()`. This event is *only* fired on iOS since Android allows notifications by default.

This could be useful to determine whether or not the device has notifications enabled, so we don't send notifications to the APNS-token / FCM-token.

Closes #2247

PR #2324, opened 2020-01-13
feat(PushNotifications): Make register method to return if permission was granted

chrisweight:
Rather than simply calling `call.success()` on iOS as soon as a user requests permissions (regardless of result), this adds a completion that allows the user to `await` the result in their app and passes back the response or error.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -94,7 +94,9 @@ public class PushNotifications extends Plugin {
         sendError(e.getLocalizedMessage());
       }
     });
-    call.success();
+    JSObject result = new JSObject();
+    result.put("granted", true);
+    call.success(result);
   }
 
   @PluginMethod()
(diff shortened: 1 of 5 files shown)


chrisweight:
NOTE: This possibly conflicts with the proposed approach in https://github.com/ionic-team/capacitor/pull/2249.

Preferring this approach personally as it allows a consuming app to simply 'await' the response as per the standard approach documented in most Capacitor native APIs, rather than having to register / maintain event listeners.

corysmc:
For what it's worth, I'd prefer this method as well ☝️

chrisweight:
> Can you move the unregister functionality to a separate PR?
> 
> And it would need to be added in the types and implemented for Android too.
> 
> And about the register change, the same, the types should be updated too, and in Android, as it doesn't ask for permission, I think it should just always return granted.

@jcesarmobile - Ok, `unregister` removed, Android, types and docs updated....

chrisweight:
@jcesarmobile - Anything else you want changing on this, or is this good to go?

jcesarmobile:
Since it changes the API it's a breaking change, so won't be merging until I start working on Capacitor 2.0

You can send a separate PR with the unregister implementation, that could be merged before 2.0 since it's a new API

jcesarmobile:
I've added a new `requestPermission` method and that's the one returning granted now, as the PR also changes register method to just call `UIApplication.shared.registerForRemoteNotifications()` and not prompt.

https://github.com/ionic-team/capacitor/pull/2516

PR #2337, opened 2020-01-15
feat(ios/android): allow to position Toast

### LINKED ISSUE #1621: [Feature request] Positioning the toast element
Hi,

I would like to be able to customize the toast position instead of forcing it to the bottom. In my case the user submits forms with a bottom button, and in case of error the toast appears under its finger, that's a bit weird I think.

I saw you were waiting for more requests on this (https://github.com/ionic-team/capacitor/pull/549#issuecomment-389980807). #crossingFingers 😄 

Thank you,

veloce:
Implements feature request #1621 

It can be one of 3 positions: 'top', 'center' or 'bottom' (the default).

closes #1621

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Toast.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Toast.java
@@ -11,9 +11,6 @@ public class Toast extends Plugin {
 
   private static final int GRAVITY_TOP = Gravity.TOP|Gravity.CENTER_HORIZONTAL;
   private static final int GRAVITY_CENTER = Gravity.CENTER_VERTICAL|Gravity.CENTER_HORIZONTAL;
-  private static final int GRAVITY_BOTTOM = Gravity.BOTTOM|Gravity.CENTER_HORIZONTAL;
-
-  private static final int BASE_TOP_BOTTOM_OFFSET = 20;
 
   @PluginMethod()
   public void show(PluginCall call) {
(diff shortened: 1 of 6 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/Toast.java
```diff
@@ -23,7 +30,22 @@ public void show(PluginCall call) {
...
 
+
     android.widget.Toast toast = android.widget.Toast.makeText(getContext(), text, duration);
+
+    String position = call.getString("position", "bottom");
+
+    if("top".equals(position)) {
+      toast.setGravity(GRAVITY_TOP, 0, BASE_TOP_BOTTOM_OFFSET);
+    } else  if("bottom".equals(position)) {
+      toast.setGravity(GRAVITY_BOTTOM, 0, BASE_TOP_BOTTOM_OFFSET);
```
jcesarmobile: This also changes the behaviour, so I wouldn't do anything if set to bottom, to keep the old behaviour.

core/src/web/toast.ts
```diff
@@ -18,6 +18,7 @@ export class ToastPluginWeb extends WebPlugin implements ToastPlugin {
     const toast = document.createElement('pwa-toast') as any;
     toast.duration = duration;
     toast.message = options.text;
+    toast.position = options.position;
```
jcesarmobile: Since web toast doesn't really support the position, better not to add code that might look like it's implemented but broken
```suggestion
    
```

android/capacitor/src/main/java/com/getcapacitor/plugin/Toast.java
veloce: Is it not weird to have a margin at the top but no at bottom? IIRC on android the toast sticks to the bottom, I found it not as pretty as with a little margin.

jcesarmobile: I don't know if it's weird or not, but this PR changes the previous behaviour, moving the default toast position.

veloce: Ok I changed it, and put also 40 as top offset.

jcesarmobile:
Just one last change, now on Android, if you pass an invalid position it will use bottom, but on iOS it will return an error.
Both should behave the same way.
I'll let you decide which one you prefer, I'm ok as long as they are the same.

veloce:
I removed the position check on iOS, since it's consistent with the `duration` argument that is also not verified.

PR #2340, opened 2020-01-15
fix(toast): unify duration across platforms

### LINKED ISSUE #2175: bug: toast duration inconsistencies
Just checked the source code and looks like every platform if using different values for short/long duration.

For android it's 2000-3500 (and can't be changed because those are Android constants)
For iOS it's 1500-3000
For web/electron it's 3000-5000.

So I think first the iOS and web values should be changed to match Android ones (as those can't be changed) and then document those values.

pwespi:
* use same duration for long and short toasts on ios, android and web (closes #2175 )
* make `short` the default on all platforms
* make android and ios actually use the duration option
* I used this test app to test the toast duration: https://github.com/pwespi/toast-test/

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Toast.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Toast.java
@@ -16,7 +16,7 @@ public class Toast extends Plugin {
       return;
     }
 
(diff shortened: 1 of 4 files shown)


pwespi:
@jcesarmobile done

Issue #2020, opened 2019-10-02
bug: Original image is lost if choosing not to edit for Camera plugin on Android


domoosterloo:
# Bug Report

## Capacitor Version
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 1.2.1

  @capacitor/core: 1.2.1

  @capacitor/android: 1.2.1

  @capacitor/ios: 1.2.1

Installed Dependencies:

  @capacitor/cli 1.2.1

  @capacitor/ios 1.2.1

  @capacitor/android 1.2.1

  @capacitor/core 1.2.1

## Affected Platform(s)

- [x] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

## Current Behavior
On Android device:
Open camera using Camera plugin.
Take picture
Accept picture -> navigates to edit screen
Try and continue without editing picture
No image is returned

## Expected Behavior
If no edit is performed, original image should be returned

## Sample Code or Sample Application Repo

const image = await Camera.getPhoto({
    quality: 90,
    allowEditing: true,
    resultType: CameraResultType.Uri,
   source: "CAMERA"
  });

## Reproduction Steps

On Android device:
Open camera using Camera plugin.
Take picture
Accept picture -> navigates to edit screen
Try and continue without editing picture
No image is returned

## Other Technical Details

Issue is in com.getcapacitor.plugin.Camera 
handleOnActivityResult
if (requestCode == REQUEST_IMAGE_EDIT)
....
Needs to check result code is either Activity.RESULT_OK or Activity.RESULT_CANCELLED.
and only go to processImage if resultCode == Activity.RESULT_OK.
If Activity.RESULT_CANCELLED || data == null && originalImageUri exists
return original image.

JhonArlex:
Good day.
I will work on this issue. I would like to know from which branch I must clone to perform the task. @jcesarmobile

jcesarmobile:
From master

jcesarmobile:
closing since the PR was merged

vycoder:
Hi, I badly needed this fix. Just wondering on when will this fix be available. Thanks!

PR #2358, opened 2020-01-22
fix(android): return original camera image if edition was canceled 

JhonArlex:
Returns the original image in case of cancellation of the image edition


JhonArlex:
close #2020

PR #2374, opened 2020-01-27
add notifications for wknavigationdelegate

JoaquimJunque:
Add notifications for wknavigationdelegate so plugins can react to the navigation events like failed navigation, navigation finished


dwieeb:
What's the use case for this?

JoaquimJunque:
We currently use this in our custom plugins. Our plugin navigates to an internal connection problems page when a navigation in the wkwebview fails to load a page instead of a white page the wkwebview provides. 

We have multiple micro frontends that are hosted. So navigations between them requires a wkwebview navigation so this situation is somthing we must certainly be able to react to.

With the current code only capacitor core itself can react to this situation (since there can only be one delegate). My PR allows everyone to react. 

We don't develop for android but i guess their api is better then apples and it is probably already possible to react to webview events. 

Another use case (but we did another fix for this) that you have apps hosted that can be used in the browser. But you want to add some custom stuff if it is loaded in capacitor. If you have a navigation done event, you can then manipulate DOM or JS wihtout having to addapt all the micro apps. (there are over a 100).

I can think of many more but the first use case should be more than enough reason.

ikeith:
Thanks for the PR! However, cross platform parity is definitely a concern. At present, we don't have a good mechanism on the Android side that could be built upon to support similar functionality. That is something that should be investigated and I have created an issue to track that effort [#3232](https://github.com/ionic-team/capacitor/issues/3232). I'm closing this PR in preparation for 3.0 work but please follow that issue and/or submit an amended request when we can match behavior.

PR #2378, opened 2020-01-28
docs(ce-plugins): add capacitor-voice-recorder

tchvu3:
Added voice recording plugin for ios and android


tchvu3:
moved to the required section

PR #2385, opened 2020-01-29
feat(android): add group and groupSummary to LocalNotifications

p7g:
This PR adds `group` and `groupSummary` options in `LocalNotifications.schedule()`, which allows you to group notifications together on Android.

I noticed a TODO comment in `LocalNotificationManager.Build` for groups, which says to implement `setGroup`, `setGroupSummary`, and `setNumber`. This PR only implements the first 2, should I also add `setNumber`?

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/PushNotifications.java
@@ -122,6 +122,8 @@ public class PushNotifications extends Plugin {
         if (notification != null) {
           jsNotif.put("title", notification.extras.getCharSequence(Notification.EXTRA_TITLE));
           jsNotif.put("body", notification.extras.getCharSequence(Notification.EXTRA_TEXT));
+          jsNotif.put("group", notification.getGroup());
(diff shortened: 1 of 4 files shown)


jcesarmobile:
I'm not able to make this work, can you maybe provide a sample app where I can easily test it?

Also, core-plugin-definitions.ts have now conflicts (probably because I merged your iOS PR). Can you fix it and also document that those are Android only and a brief description (for the 4 new options)

p7g:
@jcesarmobile Here is a demo app I threw together: https://github.com/p7g/capacitor-android-group-demo

Just run it in the android emulator, send some notifications with a group name, and then send a groupSummary notification with the same group (I put a text box and some buttons on the page). That's essentially the use-case I have.

jcesarmobile:
Thanks for the sample app, I got it working.

You have removed `group` and `groupSummary` from `PushNotification` types, but you are putting them in `getDeilveredNotifications` in `PushNotifications.java`. `getDeilveredNotifications` returns a `PushNotification`, so they types should have `group` and `groupSummary`, or make `getDeilveredNotifications` not return the `group` and `groupSummary`

p7g:
Whoops, didn't mean to remove those, I've added them back.

jcesarmobile:
can you also add the `Android only` comments as on the LocalNotification ones?

Issue #2381, opened 2020-01-28
bug: Slow taps on active elements will not fire underlying event


brian-g:
# Bug Report

## Capacitor Version
`npx cap doctor` output:
yarn run v1.21.1
$ ./node_modules/.bin/cap doctor
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 1.4.0

  @capacitor/core: 1.4.0

  @capacitor/android: 1.4.0

  @capacitor/ios: 1.4.0

Installed Dependencies:

  @capacitor/cli 1.4.0

  @capacitor/android 1.4.0

  @capacitor/ios 1.4.0

  @capacitor/core 1.4.0

[success] Android looking great! 👌
  Found 5 Capacitor plugins for ios:
    cordova-background-geolocation-lt (3.4.2)
    cordova-plugin-background-fetch (5.6.1)
    cordova-plugin-badge (0.8.8)
    cordova-plugin-cocoalumberjack (0.0.4)
    cordova-plugin-insomnia (4.3.0)
[success] iOS looking great! 👌
Done in 3.76s.

## Affected Platform(s)

- [ ] Android
- [x] iOS
- [ ] Electron
- [ ] Web

## Current Behavior
If you have an active element on a page of a capacitor app, either an `<a href="#" onclick="func()">Link</a>` or a `<button onclick="func()">Button</button>`, under iOS 13, a longer tap or press of these elements results in no events firing. 

This was made very unusable by Apple in iOS 13, when they decreased the timer that fired for these long-press events to be very short.

## Expected Behavior
Expect that a normal click() event is fired. 

## Sample Code or Sample Application Repo
Use the iOS IonicRunner project

## Reproduction Steps
1. Press the link for 'Telephone_Test' for longer than 200 ms. 
2. Release

Expected: Telephone_link activates
Results: Nothing happens

PR #2387, opened 2020-01-29
docs: add allowsLinkPreview information

brian-g:
Per #2381, a new parameter was added to the capacitor.config.json. This PR is the update to the documentation for this parameter.


site/docs-md/basics/configuring-your-app.md
```diff
@@ -117,7 +117,11 @@ The current ones you might configure are:
     // Default is 11.0
     "minVersion": "11.3",
     // Some Cordova plugins require to configure the linker flags
-    "cordovaLinkerFlags": ["-ObjC"]
+    "cordovaLinkerFlags": ["-ObjC"],
+    // Turn off link previews for hyperlinks. `true` by default. Turn this off if your
```
jcesarmobile: I find the turn off and being true default a bit confusing, sounds like it’s off by default, but it’s the opposite.
I think it’s better to stick with the official docs “A Boolean value that determines whether pressing on a link displays a preview of the destination for the link.”
https://developer.apple.com/documentation/webkit/wkwebview/1415000-allowslinkpreview

brian-g: Updated.

PR #2388, opened 2020-01-29
Android LocalNotifications: Add extra data to notification

p7g:
There is an `extra` property on the `LocalNotification` class, but it is not used anywhere. This PR adds it to the `extras` bundle when building a Notification object.

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -1031,7 +1031,7 @@ export interface LocalNotification {
   smallIcon?: string;
   attachments?: LocalNotificationAttachment[];
   actionTypeId?: string;
-  extra?: any;
+  extra?: Record<string, string>;
   /**
    * iOS only: set the thread identifier for notification grouping
    */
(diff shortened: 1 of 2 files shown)


core/src/core-plugin-definitions.ts
```diff
@@ -1031,7 +1031,7 @@ export interface LocalNotification {
   smallIcon?: string;
   attachments?: LocalNotificationAttachment[];
   actionTypeId?: string;
-  extra?: any;
+  extra?: Record<string, string>;
```
jcesarmobile: At the moment it's fetched as a JSObject, so I think it should be 
```suggestion
  extra?: Record<string, any>;
```
I've tested passing object and they are correctly fetched, so I don't think we should limit it to just strings

p7g:
I think that I was using `PushNotifications.getDeliveredNotifications()` to get all notifications including local ones, and I needed access to the data in extra. Without adding the data to extra here I don't think it would be available from `getDeliveredNotifications`.

I already committed the change you suggested, but now that I think about it, I don't think that the code in this PR would handle `any`. Should I try to write a method that converts a nested JSObject structure into a bundle?

jcesarmobile:
Going to close for now since the type change is a breaking change and for Capacitor 3 we are moving some plugins out of core and local notifications is probably going one of them, so you can send the PR again once we have a separate repo for it.

PR #2389, opened 2020-01-29
iOS LocalNotifications: Added threadIdentifier and summaryArgument options

p7g:
Similar to #2385, this PR adds options to control the grouping of local notifications.


jcesarmobile:
I've tried to make some changes on the types text but wasn't able to push to your branch, so I've sent a separate PR that will close this one.

The PR uses your commit, so you'll still get attribution

PR #2411, opened 2020-02-06
chore(android): remove unused launch_splash.xml

dwieeb:
This file is no longer used. Instead, configure the splash screen image with the `androidSplashResourceName` configuration option.


IljaDaderko:
@jcesarmobile @dwieeb by the way

```
<item name="android:background">@drawable/NAME</item>
```

setting this in style xml file also sets this image as background for native modals on android. Is it possible to scope it just to splash screen?

PR #2419, opened 2020-02-07
docs(ce-plugins): Remove or replace deprecated plugins

AntoninBeaufort:
Hi,

Simply remove deprecated plugins add the category set of plugins with the replacement of the deprecated plugins.


AntoninBeaufort:
Ok if you want, I made this change because it's the author itself that said that the previous plugins was replaced by this set of plugins : 
```
 ***********************************************
  *                                             *
  * This capacitor plugin has been deprecated   *
  *                                             *
  * and added to the Capacitor Plugin Library   *
  *                                             *
  *             @jeepq/capacitor                *
  *                                             *
  ***********************************************

  Development will continue from there
```
Like said here : 
- [https://github.com/jepiqueau/capacitor-video-player](https://github.com/jepiqueau/capacitor-video-player)
- [https://github.com/jepiqueau/capacitor-data-storage-sqlite](https://github.com/jepiqueau/capacitor-data-storage-sqlite)

But I will remove this section so.

PR #2420, opened 2020-02-07
feat(ios): change native location accuracy values

### LINKED ISSUE #1893: Geolocation on iOS is slow because it is always using high accuracy
The geolocation plugin on iOS is taking a long time to get the users location because it is always using `kCLLocationAccuracyBest` or `kCLLocationAccuracyBestForNavigation`. Pull request #1773 added the option for `enableHighAccuracy`, but according to [Apple's developer documentation](https://developer.apple.com/documentation/corelocation/kcllocationaccuracybestfornavigation) on `kCLLocationAccuracyBestForNavigation`: "Because of the extra power requirements, use this level of accuracy only while the device is plugged in.". This level of accuracy seems like it is too high to be used for `enableHighAccuracy`.

The [cordova plugin equivalent](https://github.com/apache/cordova-plugin-geolocation/blob/6570cf9fd2c546620f6aa8b2a07cf0b8b073a95c/src/ios/CDVLocation.m#L140) uses `kCLLocationAccuracyThreeKilometers` by default and `kCLLocationAccuracyBest` for high accuracy.

Ideally the user could choose the desired accuracy from all of the iOS options listed in the documentation, but these might not line up with android very well. Regardless the current implementation takes a long time to get the current location, especially when high accuracy is not required by the app.

Note that the long wait time to fetch the current location is not as noticeable on the simulator and a physical device should be used to see the true wait time.

jcesarmobile:
If `enableHighAccuracy` is set, use `kCLLocationAccuracyBestForNavigation` only for `watchPosition()` and `kCLLocationAccuracyBest` for `getCurrentPosition()`

If `enableHighAccuracy` is not set, use `kCLLocationAccuracyThreeKilometers` for fastest results.

closes #1893

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Geolocation.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Geolocation.swift
@@ -14,7 +14,7 @@ public struct GeolocationCoords {
 class GetLocationHandler: NSObject, CLLocationManagerDelegate {
   var locationManager = CLLocationManager()
   var call: CAPPluginCall
-  
+
   init(call: CAPPluginCall, options: [String:Any]) {
     self.call = call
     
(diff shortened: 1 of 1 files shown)


naranjamecanica:
Any chance that this and https://github.com/ionic-team/capacitor/pull/2409 will soon land in a `1.x` release? Or is the `2.0` release imminent? 👍

PR #2436, opened 2020-02-12
fix(android): allow Share plugin to provide text or url only

veloce:
Currently, without this fix, either text only or url only sharing does not work.

I added two more buttons on example app to show that now it is working.


krimsonkla:
Any word on when this will launch? This was working and is blocking our migration launch from Cordova to capacitor. Thanks for the awesome work on this project!

PR #2455, opened 2020-02-20
feat:toast timer duration

AndreyMartinez:
feat toast time duration.


jcesarmobile:
Thanks for the Pull Request, but you have only changed the code for web, so you are breaking Android and iOS.
And it's not possible to set the Toast duration on Android, so we will keep the code as is, sorry.

PR #2476, opened 2020-02-24
Add missing instructions to install plugins

LouAdrien:
It seems if you don't use the "ionic cordova plugin add cordova-plugin-name" plugin won't be usable, it will fail with the following warning (example with AppVersion plugin). 

Native: tried calling AppVersion.getAppName, but the AppVersion plugin is not installed.

It seem adding this line of code solves the issue, its also what is stated in the plugin doc : https://ionicframework.com/docs/native/app-version but this document seems to be colliding as it implies just adding vi anpm will be enough

### DIFF
--- a/site/docs-md/cordova/using-cordova-plugins.md
+++ b/site/docs-md/cordova/using-cordova-plugins.md
@@ -16,6 +16,7 @@ Simply install your plugin of choice, sync your project, finish any required nat
 
 ```bash
 npm install cordova-plugin-name
+ionic cordova plugin add cordova-plugin-name
 npx cap sync
 ```
 


dotNetkow:
Thanks for the PR. I see now that while I updated the Usage section [here](https://ionicframework.com/docs/native/community#usage) for Ionic Native, it should be updated on each plugin page too, like the one you mention.

That said, you shouldn't need to run "plugin add" - "cap sync" will pull in the native bits. I'll review....

jcesarmobile:
`ionic cordova plugin add` is for Cordova projects, for Capacitor projects `npm install` is enough.
In fact that command doesn't work on Capacitor projects and will ask to integrate Cordova, so won't be merging this.

LouAdrien:
@jcesarmobile that's also what I understood from the general documentation was supposed to happen. But it does not seems to be the case. Installing only with npm will result in the plugin not working and the error ionic cordova plugin add cordova-plugin-name at launch. You can try with a default app and the AppVersion plugin.

Also the plugin documentation itself https://ionicframework.com/docs/native/app-version clearly state to run `ionic cordova plugin add`

Which solves the issue. So I think there is a conflict between your two answers, what is the correct one?

jcesarmobile:
I can't reproduce
commands are `npm install` for installing and then `npx cap update` (or sync) for copying the plugin files to the project.

As Matt said, those docs are for Cordova, that's why it uses cordova command, should be updated for covering Capacitor.

Note that you have to npm install both the cordova plugin and the ionic-native wrapper, so instead of
```
ionic cordova plugin add cordova-plugin-app-version
npm install @ionic-native/app-version
```
you have to do
```
npm install cordova-plugin-app-version
npm install @ionic-native/app-version
npx cap update
```

Issue #2130, opened 2019-11-02
feat: Include device name in device info


piotrblasiak:
# Feature Request

## Describe the Feature Request

Include [UIDevice currentDevice].name in the device info returned by the Device plugin.

## Platform Support Requested

- [x] Android
- [x] iOS
- [ ] Electron
- [ ] Web

PR #2491, opened 2020-02-26
feat(ios): add device name to Device.getInfo()

jstjnsn:
closes https://github.com/ionic-team/capacitor/issues/2130

### DIFF
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -409,7 +409,7 @@ export type OperatingSystem = 'ios' | 'android' | 'windows' | 'mac' | 'unknown'
 
 export interface DeviceInfo {
   /**
-   * Note: this property is iOS only, so it is optional.
+   * Note: this property is iOS only.
    * The name of the device. For example, "John's iPhone"
    */
   name?: string;
--- a/core/src/core-plugin-definitions.ts
+++ b/core/src/core-plugin-definitions.ts
@@ -411,7 +411,7 @@ export interface DeviceInfo {
   /**
    * The name of the device. For example, "John's iPhone"
    */
-  name: string;
+  name?: string;
   /**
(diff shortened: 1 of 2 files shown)


core/src/core-plugin-definitions.ts
```diff
@@ -408,6 +408,10 @@ export interface DevicePlugin extends Plugin {
 export type OperatingSystem = 'ios' | 'android' | 'windows' | 'mac' | 'unknown';
 
 export interface DeviceInfo {
+  /**
+   * The name of the device. For example, "John's iPhone"
+   */
+  name: string;
```
jcesarmobile: Since name is iOS only, it should be optional (add the ? at the end), that means it might or might not be returned.
```suggestion
  name?: string;
```

core/src/core-plugin-definitions.ts
```diff
@@ -408,6 +408,10 @@ export interface DevicePlugin extends Plugin {
 export type OperatingSystem = 'ios' | 'android' | 'windows' | 'mac' | 'unknown';
 
 export interface DeviceInfo {
+  /**
+   * The name of the device. For example, "John's iPhone"
```
jcesarmobile: Should mention that it's iOs only.

jstjnsn:
Changes should be all good now. Thanks for guiding me jcesarmobile, I'm learning stuff!

core/src/core-plugin-definitions.ts
```diff
@@ -408,6 +408,11 @@ export interface DevicePlugin extends Plugin {
 export type OperatingSystem = 'ios' | 'android' | 'windows' | 'mac' | 'unknown';
 
 export interface DeviceInfo {
+  /**
+   * Note: this property is iOS only, so it is optional.
```
jcesarmobile:
```suggestion
   * Note: this property is iOS only.
```

Issue #2366, opened 2020-01-22
bug: not possible to write empty string on Clipboard web plugin


jcesarmobile:
# Bug Report

## Capacitor Version

`npx cap doctor` output:

## Affected Platform(s)

- [ ] Android
- [ ] iOS
- [ ] Electron
- [x] Web

## Current Behavior
Android and iOS allow to write an empty string, but web doesn't
## Expected Behavior
all platforms to behave the same way

## Other Technical Details

`npm --version` output:

`node --version` output:

`pod --version` output (iOS issues only):

PR #2523, opened 2020-03-04
feat(clipboard): allow to write images on web plugin

jcesarmobile:
This allows clipboard plugin to write images
Also fixes read of images, which wasn't working

### DIFF
--- a/core/src/web/clipboard.ts
+++ b/core/src/web/clipboard.ts
@@ -8,6 +8,7 @@ import {
 } from '../core-plugin-definitions';
 
 declare var navigator: any;
+declare var ClipboardItem: any;
 
 export class ClipboardPluginWeb extends WebPlugin implements ClipboardPlugin {
   constructor() {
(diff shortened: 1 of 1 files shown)


core/src/web/clipboard.ts
```diff
@@ -22,10 +23,24 @@ export class ClipboardPluginWeb extends WebPlugin implements ClipboardPlugin {
       return Promise.reject('Clipboard API not available in this browser');
     }
 
-    if (options.string !== undefined || options.url) {
```
pwespi: @jcesarmobile Have you changed this line on purpose? This might have reintroduced issue #2366

jcesarmobile: Oops, sorry, will fix it tomorrow as I’m working in other clipboard changes

jcesarmobile: should be fixed again by https://github.com/ionic-team/capacitor/pull/2527

PR #2524, opened 2020-03-04
Android bridge fragment

mlynch:
Adds a Fragment for the bridge per our convo in slack

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
@@ -1,0 +1,163 @@
+package com.getcapacitor;
+
+import android.app.Activity;
+import android.content.Context;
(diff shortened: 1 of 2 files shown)


fkirc:
I was facing exactly the same issue and implemented a very similar bridge fragment.

This PR is not yet in a state where it can be merged, but it serves as a good basis for discussions.

The overall goal is to improve the applicability of Capacitor for existing native apps.
For Android, fragments are much better suited than Activities to run only portions of the UI via a Capacitor WebView.
Hence, supporting fragment would be a huge win for the gradual adoption of Ionic/Capacitor into native codebases.

I would like to contribute to this goal if I have something battle-tested.

mlynch:
One thing I neglected to add was support for configuring the `startDir` from XML. Will add

mlynch:
Closing in lieu of #2666

PR #2540, opened 2020-03-07
Fix ios 13 Exif data and return gps coordinate

oscarpedrero:
Currently, Capacitor does dont send Exif data in IOS 13.
To fix this bug, I implemented request permision to access gallery photo when open it.
Additional, I found useful get gps cordinates and send it in response if image content that info.


jcesarmobile:
Turns out exif was not working on any iOS version, not just iOS 13.

I've sent a PR where it fix it for all versions, not just iOS 13
https://github.com/ionic-team/capacitor/pull/2595

Also, this PR makes taking a picture from the Camera to crash and had threading issues on iOS 13. My PR fix those 2 issues too.


I'll close this, but thanks for contributing!

PR #2548, opened 2020-03-10
feat(android): added configurable icon color for local notifications

andysousa:
Good day,
I was finding that I needed to customize quite a few things with android local notifications so I decided to put my first PR in to see how it goes.  

This PR adds the ability to set an icon color in android.  You can do it with the plugin definition or globally in the config.

If this is something the community is interested in... i have already done something similar to make the notification expandable and can put that PR in at some point in the future as well.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
@@ -178,7 +178,12 @@ public class LocalNotificationManager {
 
     String iconColor = localNotification.getIconColor();
     if (iconColor != null) {
-      mBuilder.setColor(Color.parseColor(iconColor));
+      try {
+        mBuilder.setColor(Color.parseColor(iconColor));
+      } catch (Exception ex) {
+        call.error("The iconColor string was not able to be parsed.  Please provide a valid string hexidecimal color code.", ex);
+        return;
+      }
     }
 
     createActionIntents(localNotification, mBuilder);
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
@@ -7,6 +7,7 @@ import android.app.NotificationChannel;
 import android.app.PendingIntent;
 import android.content.Context;
 import android.content.Intent;
+import android.graphics.Color;
 import android.net.Uri;
 import android.os.Build;
 import android.os.Bundle;
@@ -174,6 +175,12 @@ public class LocalNotificationManager {
     mBuilder.setOnlyAlertOnce(true);
 
     mBuilder.setSmallIcon(localNotification.getSmallIcon(context));
+
+    String iconColor = localNotification.getIconColor();
+    if (iconColor != null) {
+      mBuilder.setColor(Color.parseColor(iconColor));
+    }
+
     createActionIntents(localNotification, mBuilder);
     // notificationId is a unique int for each localNotification that you must define
     Notification buildNotification = mBuilder.build();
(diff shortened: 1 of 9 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
```diff
@@ -174,6 +175,12 @@ private void buildNotification(NotificationManagerCompat notificationManager, Lo
     mBuilder.setOnlyAlertOnce(true);
 
     mBuilder.setSmallIcon(localNotification.getSmallIcon(context));
+
+    String iconColor = localNotification.getIconColor();
+    if (iconColor != null) {
+      mBuilder.setColor(Color.parseColor(iconColor));
```
jcesarmobile: this crash if the color is not valid, put it inside a try catch (check other plugins like Browser or Statusbar)

core/src/core-plugin-definitions.ts
```diff
@@ -1033,6 +1033,10 @@ export interface LocalNotification {
    * If set, it overrides default icon from capacitor.config.json
    */
   smallIcon?: string;
+      /**
```
jcesarmobile: can you properly align this?

site/docs-md/apis/local-notifications/index.md
```diff
@@ -33,7 +33,9 @@ LocalNotifications.schedule({
       sound: null,
       attachments: null,
       actionTypeId: "",
-      extra: null
+      extra: null,
+      smallIcon: null,
+      iconColor: null
```
jcesarmobile: since they are optional, don't add them to the sample, or if you add them, do it with a real value and not null.
In this case, null crash the app (see my previous try catch comment)

andysousa: is there a better way to show users that these values are optional anywhere in docs?  Feel like leaving it off doesn't let people know that they are there.

android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
```diff
@@ -174,6 +175,17 @@ private void buildNotification(NotificationManagerCompat notificationManager, Lo
     mBuilder.setOnlyAlertOnce(true);
 
     mBuilder.setSmallIcon(localNotification.getSmallIcon(context));
+
+    String iconColor = localNotification.getIconColor();
+    if (iconColor != null) {
+      try {
+        mBuilder.setColor(Color.parseColor(iconColor));
+      } catch (Exception ex) {
+        call.error("The iconColor string was not able to be parsed.  Please provide a valid string hexidecimal color code.", ex);
```
jcesarmobile: For consistency, better use the same error message as Status Bar plugin

```suggestion
        call.error("Invalid color provided. Must be a hex string (ex: #ff0000");
```

site/docs-md/apis/local-notifications/index.md
jcesarmobile: Most people will use typescript, so they can easily check the types.

Docs are generated from the types, but don't work very well in some cases, so checking the types directly is better than the website docs.

But as I said, if you want to change the example to use smallIcon and iconColor, just use valid values instead of null

android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java
andysousa: fixed

site/docs-md/apis/local-notifications/index.md
andysousa: removed for consistency

jcesarmobile:
BTW, can you document the plugin preferences you can add in the `capacitor.config.json` file, such as `iconColor`?
Similar to this keyboard configurations:
https://capacitor.ionicframework.com/docs/apis/keyboard#keyboard-configuration-ios-only-

andysousa:
sure thing, I will add small Icon too as I don't think that was added

PR #2550, opened 2020-03-10
fix(ios): return error if Cancel is selected from Camera.getPhoto() prompt

### LINKED ISSUE #2196: bug: Camera API getPhoto does not indicate if user cancelled the initial dialog
# Bug Report

## Capacitor Version

`npx cap doctor` output:
```
Latest Dependencies:
  @capacitor/cli: 1.3.0
  @capacitor/core: 1.3.0
  @capacitor/android: 1.3.0
  @capacitor/ios: 1.3.0

Installed Dependencies:
  @capacitor/cli 1.2.1
  @capacitor/ios 1.2.1
  @capacitor/android 1.2.1
  @capacitor/core 1.2.1
```

## Affected Platform(s)

- [ ] Android
- [x] iOS
- [ ] Electron
- [ ] Web

## Current Behavior

When using the Capacitor Camera API and calling `getPhoto` with the default source as "PROMPT", if you select "Cancel" from the first view (the `UIAlertController`), there is no way to detect that the user has cancelled photo selection.

## Expected Behavior

The behavior should be similar to what happens if they select "From Photos" or "Take Picture" in the `UIAlertController` and then choose "Cancel".  The same change that was made in [this PR](https://github.com/ionic-team/capacitor/pull/808)

## Other Information

Discussed [here](https://getcapacitor.slack.com/archives/C94SE07NJ/p1574202077119100)

christhompson05:
First time contributor.

Noticed that nothing was being returned from `Camera.getPhoto()` on iOS  if source is `CameraSource.Prompt` and 'Cancel' is selected. 

fixes #2196

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Camera.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Camera.swift
@@ -111,7 +111,7 @@ public class CAPCameraPlugin : CAPPlugin, UIImagePickerControllerDelegate, UINav
     }))
 
     alert.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: { (action: UIAlertAction) in
-      alert.dismiss(animated: true, completion: nil)
+      self.imagePickerControllerDidCancel(self.imagePicker!)
     }))
 
     self.setCenteredPopover(alert)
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/Plugins/Camera.swift
```diff
@@ -111,7 +111,7 @@ public class CAPCameraPlugin : CAPPlugin, UIImagePickerControllerDelegate, UINav
     }))
 
     alert.addAction(UIAlertAction(title: "Cancel", style: .cancel, handler: { (action: UIAlertAction) in
-      alert.dismiss(animated: true, completion: nil)
+      self.imagePickerControllerDidCancel(self.imagePicker!)
```
jcesarmobile: Better just call the error directly
```suggestion
      self.call?.error("User cancelled photos app")
```
`imagePickerControllerDidCancel` should only be called by the delegate, and also it dismiss the imagePicker, but at this point it has not been presented.

Issue #2325, opened 2020-01-13
bug: show time of when the notifications is launched and not when it was scheduled


MayKikuchiThrive:
## Describe the Bug

The problem I am having is that when I schedule a local notification for a future date, the notification will be displayed with the time the local notification was created as the timestamp, not the time the notification was scheduled.

## Platform 

- [x] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

jcesarmobile:
changing to bug, since I think it should work as you describe

jcesarmobile:
reopening since it was only fixed when the LocalNotification uses at, for every or on it actually crash, going to fix the crash, but will bring the old behavior.

PR #2551, opened 2020-03-10
fix(android): make sure scheduled time is shown in localnotification (#2325)

naranjamecanica:
Fixes the problem #2325  that when a local notification is schedules for a future date, the notification will be displayed with the time it was created instead of the scheduled time.


naranjamecanica:
I'm sending in a cleaner version

PR #2562, opened 2020-03-12
Add a note about plugin listener zoning

tva-TIS:
Added a note that points out, that plugin listeners are not executed within the ngZone. This might lead to consequences that are hard to find the origin of.


tva-TIS:
This might is probably valid for iOS Plugins too, but I haven't got the ability to test it.

jcesarmobile:
Thanks for the PR, but Capacitor is framework agnostic, so I don't think we need to add notes about specific frameworks.

PR #2563, opened 2020-03-12
Update push-notifications-firebase.md

outspokenmagazine:
Without enabling the Remote Notifications within the Background Modes Capability in Xcode there is no way to receive any push notification in ios, so this is a crucial step.


jcesarmobile:
Thanks for the PR, but I've followed all the tutorial steps and I can receive the push notifications just by enabling the capability.
Background mode of remote notifications is supposed to be used for silent push, regular push are received just fine without it.

PR #2568, opened 2020-03-13
docs: update Filesystem.writeFile sample

jcesarmobile:
it now returns the file path, so log it in the example.


aaronksaunders:
not seeing any output when running the sample code from the documentation, do I need to update something? Also where is the documentation saying exactly what the path is of where this content is being written... based on the new security in Android 10 I am seeing different results and looking for some guidance.

Thanks.

jcesarmobile:
This is on the beta only, we will release as final soon. For now you can install it like this

`npm install @capacitor/core@next` (same for iOS and Android)

PR #2578, opened 2020-03-14
Update plugins.md

jepiqueau:
capacitor-data-storage-sqlite and capacitor-video-player where on the list before. Now i have regrouped them in one npm package @jeepq/capacitor for easy maintenance


jepiqueau:
@jcesarmobile  thanks for your reply, i should have contacted you before to put the plugin in the same library, but i tried several time to contact you on slack like i did before but you were not answering so i conclude that you were too busy. So i will make a step back. One question from a user was how to have an electron plugin for capacitor-data-storage-sqlite as electron can access sqlite. there is no methodology described in the capacitor documentation but i know that is possible as you have done it for several capacitor standard plugins.

jcesarmobile:
closed because of https://github.com/ionic-team/capacitor/pull/2589

PR #2581, opened 2020-03-15
docs: Add Deep Links guide

dotNetkow:
Needed to implement this for a personal app, so turned it into a Capacitor guide.

Used my app for screenshots, video, etc examples - could switch to something more generic but thought it would be good to use a real example (instead of "yourwebsite.com" everywhere).

Still need to add React code, but wanted early feedback.


mlynch:
Nice! I think the link on the side should say "Deeplinking" which will be more obvious/identifiable by most people

andysousa:
nice, I am going to try this... noticed this am that I hadn't updated the config since moving to Capacitor and I would love to get rid of the Cordova plugin

site/docs-md/guides/deep-links.md
```diff
@@ -0,0 +1,311 @@
...
+                // example: 8L65AZE66A.com.netkosoft.beerswift
+                "appID": "TEAMID.BUNDLEID",
+                "paths": ["*"]
+            }
+        ]
+    }
+}
+```
+
+> Note: Some tutorials mention that the file must be served with content type `application/pkcs7-mime`. This is not needed. If your app runs in iOS 9 or later and you use HTTPS to serve the file, you can create a plain text file that uses the `application/json` MIME type.
```
jcesarmobile: I don't think this makes sense, if they are following your tutorial, why talk about other tutorials that are wrong?

site/docs-md/guides/deep-links.md
```diff
@@ -0,0 +1,311 @@
...
+    <category android:name="android.intent.category.BROWSABLE" />
+    <data android:scheme="https" android:host="beerswift.app" />
+</intent-filter>
+```
+
+The complete Activity should look similar to this:
+
+```xml
+<activity
+    android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale"
```
jcesarmobile: just updated this line the other day, so it won't look like this anymore

site/docs-md/guides/deep-links.md
```diff
@@ -0,0 +1,311 @@
...
+  - dotnetkow
+---
+
+# Deep Linking with Universal and App Links
+
+**Platforms**: Web, iOS, Android
+
+Universal links (iOS) and App Links (Android) offer the ability to take users directly to specific content within a native app (commonly known as deep linking).
+
+When users tap or click on a deep link, the mobile operating system sends the user directly into your app without routing through the device's web browser or website first. If the app isn't installed, then the user is directed to the website. If the user navigates directly to the website, they remain on the website.  This makes deep links an excellent feature for cross-platform apps built for the web, iOS, and Android.
```
jcesarmobile: Again, deep links don't work on web, they are regular links. I don't think we can say it's cross-platform if Capacitor's Web App plugin doesn't respond to `appUrlOpen`

site/docs-md/guides/deep-links.md
dotNetkow: good call, removed.

site/docs-md/guides/deep-links.md
```diff
@@ -0,0 +1,311 @@
...
+title: Deep Links
+description: Implement deep linking functionality in an iOS and Android app
+url: /docs/guides/deep-links
+contributors:
+  - dotnetkow
+---
+
+# Deep Linking with Universal and App Links
+
+**Platforms**: Web, iOS, Android
```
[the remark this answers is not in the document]
dotNetkow: true. I meant it more that it's a good cross platform approach because you can use deep links while still having a PWA too

site/docs-md/guides/deep-links.md
dotNetkow: that's ok - why i said "it'll look similar to this:" - because I recognize people's AndroidManifest may be different.

PR #2587, opened 2020-03-17
feat(ios): set status bar animation style when showing and hiding

sbannigan:
This adds an option to the status bar `show` and `hide` methods to change the animation style. The style will always default to `.slide` if not specified in the options (other options are `.fade` and `.none`. This is iOS only. I considered adding this as a plugin config that could be configured in `capacitor.config.json` but there could be use cases to change this on each show and hide.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/StatusBar.swift
+++ b/ios/Capacitor/Capacitor/Plugins/StatusBar.swift
@@ -35,12 +35,25 @@ public class CAPStatusBarPlugin: CAPPlugin {
     call.unimplemented()
   }
   
+  func setAnimation(_ call: CAPPluginCall) {
+    let animation = call.getString("animation", "SLIDE")
+    if animation == "FADE" {
+      bridge.setStatusBarAnimation(.fade)
+    } else if animation == "NONE" {
+      bridge.setStatusBarAnimation(.none)
+    } else {
+      bridge.setStatusBarAnimation(.slide)
+    }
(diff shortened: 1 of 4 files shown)


ios/Capacitor/Capacitor/Plugins/StatusBar.swift
```diff
@@ -35,12 +35,25 @@ public class CAPStatusBarPlugin: CAPPlugin {
...
   
+  func setAnimation(_ call: CAPPluginCall) {
+    let animation = call.getString("animation", "SLIDE")
+    if animation == "FADE" {
+      bridge.setStatusBarAnimation(.fade)
+    } else if animation == "SLIDE" {
+      bridge.setStatusBarAnimation(.slide)
+    } else if animation == "NONE" {
+      bridge.setStatusBarAnimation(.none)
+    }
```
jcesarmobile: If the use enters an invalid value, it will use latest animation set instead of the "SLIDE" default.
Change it to set it to `.slide` if the value is not `"NONE"` nor `"FADE"`

```suggestion
    } else if animation == "NONE" {
      bridge.setStatusBarAnimation(.none)
    } else {
      bridge.setStatusBarAnimation(.slide)
    }
```

sbannigan: 👍

sbannigan:
change made!

Issue #1876, opened 2019-08-14
Transparent status bar


UlisesCeca:
Hello,

is it possible to make the status bar transparent or translucent like the image below with the current plugin?

![image](https://user-images.githubusercontent.com/23300072/63000691-0fac2580-be72-11e9-8904-1f49dceb0962.png)

Thanks!

jcesarmobile:
It's possible to make the bar translucent as it accepts AARRGGBB as color, but it's not possible to show the webview behind the statusbar.

UlisesCeca:
Okay, thank you!

Will this ever be considered to be implemented in the future?

jcesarmobile:
Tagging with feature request tag:
Issues tagged with feature request are closed but tracked for reactions to gauge interest.
If you are interested in this feature, add the thumbs up emoji (+1) reaction on the issue so we can sort them based on it.

In case it's done, it would be android only method, should be like the cordova-plugin-statusbar `overlaysWebView` method.

Crylion:
I'm also really missing this :/
We're currently migrating a cordova app to capacitor and supporting phones with Notches correctly as an important part of the design, so we really relied on the cordova statusbar plugin being able to overlay the webview :(
Are there workarounds by changing the android code directly? The last time I developed for Android was about five years ago, so I'm a bit rusty x)

jcesarmobile:
this has been implemented on https://github.com/ionic-team/capacitor/pull/2597, removing the feature request label

sisoje:
@jcesarmobile what about ios support?

jcesarmobile:
iOS always overlays the status bar, you can avoid the status bar area by respecting safe areas
https://webkit.org/blog/7929/designing-websites-for-iphone-x/

PR #2597, opened 2020-03-18
feat(android): Add Statusbar.setOverlaysWebView method

Crylion:
Addition of a new method in the core Statusbar plugin, similar to the `overlaysWebView` method/preference of the cordova statusbar plugin. Updated example in the docs for the Statusbar plugin.

Works only for android (as discussed in [#1876](https://github.com/ionic-team/capacitor/issues/1876)) since on iOS it already handles it correctly on devices with notches.

I'm not confident in my android knowledge these days, so I might have overlooked something. But in my tests everything worked as expected, similar to the cordova plugin.

What I'm not sure about is the correct way to handle that the iOS version of the plugin doesn't implement this feature. I added it with a `"not implemented"` response in the `Statusbar.swift` file but I'm not sure if that is all that's needed here.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/StatusBar.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/StatusBar.java
@@ -1,6 +1,7 @@
 package com.getcapacitor.plugin;
 
 import android.graphics.Color;
+import android.support.annotation.ColorInt;
 import android.util.Log;
 import android.view.View;
 import android.view.Window;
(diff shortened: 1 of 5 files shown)


Crylion:
@jcesarmobile  Both are good points that I also thought about. 👍 
But it also seemed strange to me to just mirror the name of the method and call the option `"overlayWebview"` 😕
If I'm being honest, I would have preferred to not use an object at all and just take a boolean as a parameter, but looking at all the other official plugins I could not find any example of that 🤔
I'll go with your suggestion and call it `overlay`.

About the color: I also was not really satisfied with the way it is, but the cordova plugin actually does the same. I suspect that this is because it would be unusual to actually switch the overlaying behavior of the status bar back and forth during runtime? But I share your opinion that this isn't very clean that way. I'll look at restoring it!

(and i'm obviously fixing the typo ;D)

Thanks for your time!

Crylion:
Hopefully all done now 🙂
I went with `overlays` for the info attribute because I felt it would be clearer that the status bar is the one that overlays something rather than being overlaid itself.

Thank you for your time and help!

PR #2611, opened 2020-03-20
feat(App): Add getState method to check current app state

### LINKED ISSUE #2500: feat: Get current appState at any moment
# Feature Request

There is a way to [get a call back](https://capacitor.ionicframework.com/docs/apis/app#method-addListener-0) when the application goes background or foreground. But there is no way to get the application `isActive` value (a.k.a. foreground/background state) at any moment. The problem is that I don't know the application state until I subscribe to the `appStateChange` event AND the app goes background or foreground. JavaScript can start working while the app is in background (e.g. when you minimize the app during the splash screen) so I can't just say «it's foreground initially».

## Platform Support Requested

- [x] Android
- [x] iOS
- [x] Electron
- [x] Web

## Describe Preferred Solution

Add a method to the App plugin:

```ts
App.getAppState(); // returns an AppState instance
```

## Related Code

```ts
let appState: AppState = {
  isActive: ???
};

App.addListener('appStateChange', newState => appState = newState);
```

## Additional Context

https://stackoverflow.com/q/60435042/1118709

jcesarmobile:
closes #2500


Finesse:
Thank you!

PR #2507, opened 2020-03-02
fix(cordova): handle source-file with framework attribute

jcesarmobile:
don't try to read source-file tags with framework attribute as they might be directories

closes #2506

PR #2640, opened 2020-03-25
fix(cordova): handle source-file with framework attribute (#2507)

mion00:
cherry-picking this commit because I would like a fix for the mentioned bug in a project using Capacitor 1.5.1


jcesarmobile:
Thanks but we are not cherry picking fixes to v1.x unless they are really serious. 
You can use beta cli `npm install @capacitor/ios@next`

PR #2648, opened 2020-03-27
feat(LocalNotifications): Allow to create notifications without activity

### LINKED ISSUE #2626: feat(LocalNotification):  ability to generate localNotification when app is in background (from service)
# Feature Request

Not able to generate a local notification when the app is in the background because activity is needed to create [LocalNotificationManager](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java#L53) instance.

## Platform Support Requested

- [x] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

## Describe Preferred Solution
We should remove the activity as param and add context in the constructor. 
We can remove usage of activity from [LocalNotificationManager](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java#L249) and can be replace `activity` with `MainActivity.class`.

dwlrathod:
Fixes #2626

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java
@@ -37,7 +37,7 @@ public class LocalNotifications extends Plugin {
   public void load() {
     super.load();
     notificationStorage = new NotificationStorage(getContext());
-    manager = new LocalNotificationManager(notificationStorage, getActivity());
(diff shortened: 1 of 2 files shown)


BorntraegerMarc:
Fixes https://github.com/ionic-team/capacitor/issues/2626

PR #2656, opened 2020-03-28
Docs - State plugin import required for web implementation to be registereds

Suui:
It's a piece of data that evaded me until I found an answer in this issue:
ionic-team#749

It was stated that the docs would be updated, I couldn't find it so in case they haven't been I've written this small update.

It is to state that an import is required for the plugin web implementation to be registered to the Plugins accesible from @capacitor/core.


dwieeb:
@Suui Would you mind re-making this PR in the site repo if it's still relevant? https://github.com/ionic-team/capacitor-site

We are moving the docs out of this repo. Thanks!

PR #2666, opened 2020-03-31
feat(android): implement BridgeFragment for easier embedding using Fragments

mlynch:


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -58,7 +58,7 @@
 import java.util.ArrayList;
 
 
-/**
+/*
```
jcesarmobile:
```suggestion
/**
```

mlynch:
Ah good catch. Fixing

android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
```diff
@@ -0,0 +1,195 @@
...
+      cordovaInterface.restoreInstanceState(savedInstanceState);
+    }
+
+    mockWebView = new MockCordovaWebViewImpl(getActivity().getApplicationContext());
+    mockWebView.init(cordovaInterface, pluginEntries, preferences, webView);
+
+    pluginManager = mockWebView.getPluginManager();
+    cordovaInterface.onCordovaInit(pluginManager);
+
+    preferences = new CordovaPreferences();
```
jcesarmobile, line 108: Why do you do this? the preferences are being set on line 125, so this reset them

android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
```diff
@@ -0,0 +1,195 @@
...
+
+    if (startDir != null) {
+      bridge.setServerAssetPath(startDir);
+    }
+
+    if (savedInstanceState != null) {
+      bridge.restoreInstanceState(savedInstanceState);
+    }
+    this.keepRunning = preferences.getBoolean("KeepRunning", true);
+    // this.onNewIntent(getIntent());
```
jcesarmobile, line 119: is this needed or forgot to uncomment? If not needed just delete it

android/capacitor/src/main/java/com/getcapacitor/Splash.java
```diff
@@ -47,7 +48,14 @@ private static void buildViews(Context c) {
       String splashResourceName = Config.getString(CONFIG_KEY_PREFIX + "androidSplashResourceName", "splash");
 
       int splashId = c.getResources().getIdentifier(splashResourceName, "drawable", c.getPackageName());
-      Drawable splash = c.getResources().getDrawable(splashId, c.getTheme());
+
```
jcesarmobile, line 51: I guess this is to prevent some crash, but if the splash is not found it still crash on line 267
Can be prevented by changing to 

```
} catch (IllegalStateException | IllegalArgumentException ex) {
  Log.d(LogUtils.getCoreTag(), "Could not add splash view");
  return;
}
```

mlynch:
k cleaned it up and squashed

PR #2676, opened 2020-04-01
feat(LocalNotifications): add createChannel, deleteChannel and listChannels methods

### LINKED ISSUE #2675: feat(LocalNotification): multiple notificaiton channel for Android
# Feature Request
Allow local notification with multiple channels so we can have multiple priorities, sound, light etc for different notifications.

## Platform Support Requested

- [x] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

dwlrathod:
Starting in Android 8.0 (API level 26), all notifications must be assigned to a channel. For each channel, you can set the visual and auditory behaviour that is applied to all notifications in that channel. Then, users can change these settings and decide which notification channels from your app should be intrusive or visible at all.

This PR is for allowing channel feature to LocalNotification. The first user needs to create a Notification channel with `createChannel` method and then assign `channelId` at the time of scheduling notification. If no `channelId` is provided will use the default channel and if provided `channelId` does not exist, the notification will not be generated.
Fixes #2675

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java
@@ -1,5 +1,7 @@
 package com.getcapacitor.plugin;
 
+import android.app.NotificationManager;
+import android.content.Context;
 import android.content.Intent;
 
 import com.getcapacitor.JSArray;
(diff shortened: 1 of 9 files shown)


dwlrathod:
Fixes #2675

dwlrathod:
@jcesarmobile Good points. Thank you for your time and help!

PR #2690, opened 2020-04-04
fix(plugin/android):  =

rdlabo:
not use '=' at targetSDK version.


rdlabo:
Sorry, this is same as https://github.com/ionic-team/capacitor/pull/2689

PR #2698, opened 2020-04-05
Updates Push Notifications with Firebase Guide

javebratt:
Hey y'all, this PR pushes a few updates on the push notifications article:

- Most of the references where to Ionic 4. Changed it to reference to Ionic 5.
- The first part was prompting the user to install the old `ionic` CLI package. Changed to install the `@ionic/cli`.
- After we add the code to the `home.page.ts` there was no notice that the user needed to run a new build and `npx cap copy` to copy web assets to the android application, so that code wasn't getting executed.
- The instructions to add a new android app in Firebase worked only for new projects, existing projects didn't have that Icon so added some more details and a picture to show where to create it.


javebratt:
There's also a broken link here:

> Building and deploying iOS and Android applications using Capacitor requires a bit of setup. Please [follow the instructions to install the necessary Capacitor dependencies here](../getting-started/dependencies) before continuing.

I don't know how to fix it, because it works properly on localhost, but deployed it's adding the word 'guides' to the URL.

site/docs-md/guides/push-notifications-firebase.md
```diff
@@ -211,7 +220,9 @@ Name the project, accept the Firebase ToS and click **Create project** to contin
 
 This section more-or-less mirrors the [setting up Firebase using the Firebase console documentation](https://firebase.google.com/docs/android/setup?authuser=0). See below for specific Capacitor-related notes.
 
-Go to the Project Overview page for your Firebase project and, under the **Grow** section, click the **Cloud Messaging** option. At the top, click on the **Android** icon.
+Go to the Project Overview page for your Firebase project and at the top, click on the **Android** icon to add a new android application.
+
+![Add new Android Application in Firebase Console](/assets/img/docs/guides/firebase-push-notifications/add-android-app.png)
```
dotNetkow, line 225: sweet, love images. thanks

site/docs-md/guides/push-notifications-firebase.md
```diff
@@ -22,15 +23,16 @@ Building and deploying iOS and Android applications using Capacitor requires a b
 To test push notifications on iOS, Apple requires that you have [a paid Apple Developer account](https://developer.apple.com/) and a *physical* iOS device.
 
 ## Prepare an Ionic Capacitor App
-If you have an existing Ionic 4 app, skip this section. If not, let's create an Ionic app first. 
+
+If you have an existing Ionic 5 app, skip this section. If not, let's create an Ionic app first.
```
dotNetkow, line 27: just remove "5" here

site/docs-md/guides/push-notifications-firebase.md
```diff
@@ -22,15 +23,16 @@ Building and deploying iOS and Android applications using Capacitor requires a b
...
 
 In your preferred terminal, install the latest version of the Ionic CLI:
 
 ```bash
-npm install -g ionic
+npm install -g @ionic/cli
 ```
 
-Next, let's use the CLI to create a new Ionic 4 Angular app based on the **blank** starter project and call it **capApp**:
+Next, let's use the CLI to create a new Ionic 5 Angular app based on the **blank** starter project and call it **capApp**:
```
dotNetkow, line 35: here too

site/docs-md/guides/push-notifications-firebase.md
```diff
@@ -4,11 +4,12 @@ description: Learn how to get Firebase Cloud Messaging working on iOS and Androi
 url: /docs/guides/push-notifications-firebase
 contributors:
   - bryplano
+  - javebratt
 ---
 
 # Using Push Notifications with Firebase in an Ionic + Angular App
 
-**Web Framework**: Ionic 4 + Angular  
+**Web Framework**: Ionic 5 + Angular
```
dotNetkow, line 12: we actually want to get away from this. Couple reasons, because as of Ionic 4, "it's just Ionic", and like in the Deep Links guide, we (ideally) will make it open to all frameworks, have examples as separate headers (Angular, React, etc).

Keep the Platforms one though.

site/docs-md/guides/push-notifications-firebase.md
```diff
@@ -4,11 +4,12 @@ description: Learn how to get Firebase Cloud Messaging working on iOS and Androi
 url: /docs/guides/push-notifications-firebase
 contributors:
   - bryplano
+  - javebratt
 ---
```
dotNetkow, line 9: Under Required Dependencies section, can you add something to help avoid the issue where devs are using out of date firebase dependencies etc?  Because they get stale quickly, I'd like to avoid recommending "x version of node, y version of android.", unless it's a hard requirement like major versions, "iOS 13+" for example. Instead, how about something like "If you encounter issues like X, ensure that you're on the latest versions of Node, Android Studio, and Xcode".

Also, for one issue, the problem was that someone used `phonegap-plugin-push` which used FB messaging 2.x when latest is 4.x - any way to address that generically? like "Check that you're using the latest Firebase SDKs (list those that are used in this guide) as well as ensure that you're not using multiple copies of FB, that may be included in other plugins (especially Cordova ones like X, Y, Z (pg-plugin-push). 

If there's a bunch of details that would help others, we could add a Troubleshooting section at the very end of the guide instead, as well. up to you.

dotNetkow:
> There's also a broken link here:
> 
> > Building and deploying iOS and Android applications using Capacitor requires a bit of setup. Please [follow the instructions to install the necessary Capacitor dependencies here](../getting-started/dependencies) before continuing.
> 
> I don't know how to fix it, because it works properly on localhost, but deployed it's adding the word 'guides' to the URL.

You can ignore this, as it's rendering the link correctly when deployed. I think the difference is the path generated locally vs when deployed. Annoying but OK for now. good call out though :)

site/docs-md/guides/push-notifications-firebase.md
javebratt: Alright, I added a small note, don't want to give hard numbers as you say, but pushed people to use the latest **stable** version.

site/docs-md/guides/push-notifications-firebase.md
javebratt: Done

site/docs-md/guides/push-notifications-firebase.md
javebratt: Done

site/docs-md/guides/push-notifications-firebase.md
javebratt: I left it as Angular.

I was gonna remove it all together, but there's Angular specific code for the Capacitor push plugin usage.

Issue #2711, opened 2020-04-07
bug: Browser plugin in iOS-popover presentation freezes app after dismiss


chriswep:
Capacitor Version: 2.0.0

## Affected Platform(s)
- [ ] Android
- [x] iOS
- [ ] Electron
- [ ] Web

## Current Behavior
If opening an URL via the Browser plugin on iOS (13.4) with `presentationStyle: 'popover'`, the app freezes after dismiss of the Browser modal. 
The Safari remote inspector shows the opened webview and in can be inspected after dismissal. This indicates that something is not properly destroyed after dismiss.

`presentationStyle: 'fullscreen'` does not have this issue.

## Expected Behavior
Remove the Webview / Browser-modal completely after dismiss of the popover

chriswep:
thanks @jcesarmobile that was quick!

as far as i see, you disabled the popover-style for everything but iPads? On iPhones i used this to achieve the iOS13 modal style. I manually copied your changed code without that restriction, and it works fine on iPhones.

jcesarmobile:
In which devices does the popover works? I thought it was iPad only, doesn’t show different Style on my iPhone X

chriswep:
on iPhone it opens an animated modal card (iOS13 style) instead of a fullscreen window

jcesarmobile:
Ah, my device is on iOS 12.

jcesarmobile:
Can you test this code with your change in a real iPhone?

```
async openBrowserPop() {
    await Browser.open({ url: 'https://www.google.com/', presentationStyle: 'popover' });
    setTimeout(() => {
      Browser.close();
    }, 5000);
  }
```
The Browser window should close itself after 5 seconds.

The point of the tmpWindow is because some modal presentations were frozing the javascript execution, so the close was never called.
If I let the popover on my iOS 12 iPhone it still gets frozen, and I don't have an iOS 13 iPhone (I could update, but then I can't go back on testing on iOS 12).
But maybe the iOS 13 popover doesn't freeze the javascript execution and we can allow the popover there.

chriswep:
i don't see any freezing on a real iPhone iOS13. 
on iOS12 simulator it does freeze as soon as the browser opens.

chriswep:
@jcesarmobile do you want a new issue for this (enabling popover on iOS13-iPhone)?

mchambaud:
on iOS14 it crashes you **open + dismiss + open again**

works fine when using fullscreen

PR #2714, opened 2020-04-07
fix(ios): don't use the tmpWindow on popover presentation

jcesarmobile:
the tmpWindow doesn't work fine when using popover presentation, and since popover presentation doesn't freeze the javascript, changed the code to not use the tmpWindow when the presentation is popover

closes https://github.com/ionic-team/capacitor/issues/2711

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Browser.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Browser.swift
@@ -23,7 +23,7 @@ public class CAPBrowserPlugin : CAPPlugin, SFSafariViewControllerDelegate {
         self.vc = SFSafariViewController.init(url: url!)
         self.vc!.delegate = self
         let presentationStyle = call.getString("presentationStyle")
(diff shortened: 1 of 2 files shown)


ios/Capacitor/Capacitor/Plugins/Browser.swift
```diff
@@ -23,7 +23,7 @@ public class CAPBrowserPlugin : CAPPlugin, SFSafariViewControllerDelegate {
         self.vc = SFSafariViewController.init(url: url!)
         self.vc!.delegate = self
         let presentationStyle = call.getString("presentationStyle")
-        if presentationStyle != nil && presentationStyle == "popover" {
+        if presentationStyle != nil && presentationStyle == "popover" && UIDevice.current.userInterfaceIdiom == .pad {
```
chriswep, line 26: does this mean that popover presentation on iPhone is gone? is there a reason for this? i liked the iOS13 modal-style of it.

PR #2717, opened 2020-04-07
fix message: Cannot read property 'SplashScreen' of undefined

mandolyte:
If `capacitor.config.json` does not have any plugins defined, then the code throws an error message about a `SplashScreen` property, like this:
```
$ electron ./

(electron) The default value of app.allowRendererProcessReuse is deprecated, it is currently "false".  It will change to be "true" in Electron 9.  For more information please check https://github.com/electron/electron/issues/18397
Cannot read property 'SplashScreen' of undefined
Done in 12.30s.
$
```
The little change in this PR verifies that a plugin is defined before trying to read the property of it.

### DIFF
--- a/electron/index.js
+++ b/electron/index.js
@@ -65,9 +65,15 @@ class CapacitorSplashScreen {
 
     try {
       let capConfigJson = JSON.parse(fs.readFileSync(`./capacitor.config.json`, 'utf-8'));
+      let ss = {};
+      if ( capConfigJson.plugins !== undefined ) {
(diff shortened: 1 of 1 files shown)


mandolyte:
Done!

dwieeb:
@mandolyte Apologies this didn't make it in. We will be recommending the [Capacitor Community Electron platform](https://github.com/capacitor-community/electron) going forward. Feel free to create a PR over there! Thanks!

PR #2729, opened 2020-04-08
fix(cordova): Don't add as system library if it's a vendored library

### LINKED ISSUE #2642: bug: payPal plugin does't work on iOS when Capacitor is used for build
Paypal works fine only when using Cordova. But when I used the Capacitor it doesn’t work on iOs. I understand that it is not on the Ionic side. But I just want to inform you that
https://github.com/paypal/PayPal-Cordova-Plugin plugin works correctly only with Cordova.

This guide https://ionicframework.com/docs/native/paypal works only using Cordova. I created empty apps (only with PayPal plugin) using both Cordova and Capacitor and tested it with the latest library versions.

Error body from PayPal pop-up:
Invalid merchant
Payments to this merchant are not allowed (invalid clientId)

Error from console:
PayPal SDK: Request has failed with error: invalid_client - System error (invalid_client). Please try again later. (401) | PayPal Debug-ID: 61c53ad41d07b [mock, PayPal iOS SDK 2.17.0] | Details: ( { “error_description” = “Client Authentication failed”; } ).

The clientId is correct from the correct production environment and works only on Android using Capacitor and on both platforms using Cordova.

jcesarmobile:
Only system libraries should be added to libraries section, but since it's hard to know if the library is a system library or not, at least don't include a library in libraries section if it's also on vendored_libraries section

closes #2642

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -343,7 +343,7 @@ function filterARCFiles(plugin: Plugin) {
 
 function removeNoSystem(library: string, sourceFrameworks: Array<string>) {
   const libraries = sourceFrameworks.filter(framework => framework.includes(library));
-  return libraries.length == 0;
+  return libraries.length === 0;
 }
 
 async function getPluginsTask(config: Config) {
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -207,14 +207,15 @@ async function generateCordovaPodspec(cordovaPlugins: Plugin[], config: Config, 
       }
     });
   });
+  const onlySystemLibraries = systemLibraries.filter(library => removeNoSystem(library, sourceFrameworks));
   if (weakFrameworks.length > 0) {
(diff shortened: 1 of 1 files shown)


cli/src/ios/update.ts
```diff
@@ -340,6 +341,11 @@ function filterARCFiles(plugin: Plugin) {
   return sourcesARC.length > 0;
 }
 
+function removeNoSystem(library: string, sourceFrameworks: Array<string>) {
```
dwieeb, line 344: I think we should start preferring this syntax for TS arrays. Thoughts?

```suggestion
function removeNoSystem(library: string, sourceFrameworks: string[]) {
```

jcesarmobile: I have no preference on one syntax over the other, but yeah, we should decide one or the other and always use the same.
Won't change it for now, but we can do a refactor later that replaces all of them.

Issue #2732, opened 2020-04-08
feat: android - start animatable layers when splash resource is a layer drawable


tntwist:
# Feature Request

## Describe the Feature Request
The splashscreen plugin starts all animatable layers of a layer-list splash resource.

## Platform Support Requested

- [X] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

## Describe Preferred Solution
It would be great if the splashscreen plugin could start all animatable layers of a layer-list splash resource.

## Describe Alternatives
I have not considered any alternatives.

## Related Code
I have created a sample repo where this feature works:
https://github.com/tntwist/capacitor-layer-splash

## Additional Context
I already implemented this. I´m creating a pull request.

tntwist:
PR is merged. So this issue can be closed.

Thanks for the approve :)

PR #2733, opened 2020-04-08
feat(android): start animatable layers when splash drawable is layered

tntwist:
PR for #2732


carlpoole:
Thanks for your contribution! 🎉

PR #2734, opened 2020-04-09
docs(ce-plugins): Remove capacitor-apple-login

adam-h:
Remove `capacitor-apple-login` as the github package has been removed (unknown why)


hatsantos:
I'm currently migrating (cordova to capacitor) one of our main products from the company I work  that is currently in more than 60 clients. I was in the exact step of migrate the now mandatory Apple login (if others external logins exist in the App) and bammmm... broken link!

I made a few searchs and found this repo: https://github.com/rlfrahm/capacitor-apple-login

Are we talking about the same plugin?

adam-h:
@hatsantos Yes I believe so, I forked that myself for our apps. Seems to have been a copy of the original 'htorbov' version (mostly similar when I checked with my yarn cache version)

Issue #1682, opened 2019-06-18
How to subscribe to a topic in Firebase Push Notification


bubundas17:
Hi, I had an issue while implementing Push Notifications in an app. 

The push notifications functionality is working fine! But the issue is, how do I Subscribe and Then Unscribe user from a topic? There is no Documentation for that!

Please, guys, It Is a must-have feature.

minajacob:
Add this plugin
https://github.com/stewwan/capacitor-fcm

bubundas17:
> Add this plugin
> https://github.com/stewwan/capacitor-fcm

Yap, It helps. Thank you!

bubundas17:
But well. it should be also implemented in Capacitor's Native plugin also.

bubundas17:
@jcesarmobile Any Comments?

jcesarmobile:
PRs are always welcome, but if somebody want to add this feature, it has to be Android only as iOS APNS don't support topics.

bubundas17:
> Add this plugin
> https://github.com/stewwan/capacitor-fcm

Well, It seems to not work properly.

jwhijazi:
the solution worked for me.
Maybe you missed the line in the documentation for android devices:

on your MainActivity.java file add import io.stewan.capacitor.fcm.FCMPlugin; and then inside the init callback add(FCMPlugin.class);

orhan-swe:
For us this is only working on Android, there we are reading the FCM token. The issue is with iOS, there it seems the device is only giving us the APNS token which we can not user with FCM, has anyone seen or resolved this issue?
We have also tried with https://github.com/stewwan/capacitor-fcm but on iOS the "getToken" method only returns an empty object.
Any information on this would be greatly appreciated!

madmacc:
@orhan-swe

naveedahmed1:
@stewwan since you have already implemented similar functionality in your plugin https://github.com/stewwan/capacitor-fcm can you please create a pull request to implement this feature for the capacitor push notification plugin?

stewwan:
@naveedahmed1 @jcesarmobile explains above. That's why I created the plugin which serves both platform with no pain.

enjoy it

> PRs are always welcome, but if somebody want to add this feature, it has to be Android only as iOS APNS don't support topics.

naveedahmed1:
@stewwan yes its awesome, great job.

But @jcesarmobile who's a member of capacitor repository mentioned.

[...]

PR #2738, opened 2020-04-09
subscribe to & unsubscribe from a topic in Firebase Push Notification #1682

dibyendusaha:
Please review my code, and if possible allow to merge it to the base.
Subscribing and Unsubscribing from topic is now becoming an essential part of FCM, as with new API Endpoint if we need to send messages to multiple devices they are asking to create a topic as one of the options. So thought if we can have this directly in capacitor instead of using fcm-capacitor plugin from @stewwan 
Please let me know if there is any scope for improvement. @stewwan @dwlrathod

### DIFF
--- a/Capacitor.podspec
+++ b/Capacitor.podspec
@@ -9,8 +9,8 @@ Pod::Spec.new do |s|
   s.authors = { 'Ionic Team' => 'hi@ionicframework.com' }
   s.source = { :git => 'https://github.com/ionic-team/capacitor.git', :tag => s.version.to_s }
   s.source_files = 'ios/Capacitor/Capacitor/*.{swift,h,m}', 'ios/Capacitor/Capacitor/Plugins/*.{swift,h,m}', 'ios/Capacitor/Capacitor/Plugins/**/*.{swift,h,m}'
-  s.static_framework = true
   s.dependency 'CapacitorCordova', '2.0.0'
   s.dependency 'FirebaseMessaging'
   s.swift_version = '5.0'
+  s.static_framework = true
 end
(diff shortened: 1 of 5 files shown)


stewwan:
back in 2018 when I approached @jcesarmobile in Slack about this feature he told me that it would fit better in a plugin. That's why I didn't it before,  otherwise I would have sent the PR.

dibyendusaha:
> back in 2018 when I approached @jcesarmobile in Slack about this feature he told me that it would fit better in a plugin. That's why I didn't it before, otherwise I would have sent the PR.

I will suggest if you can also contribute and if we have this as a feature in capacitor

dibyendusaha:
> Thanks for the PR, but we are not going to add a Firebase dependency to `@capacitor/ios`, that's why I told stewwan in the past that this was better in a separate plugin.
> So, we might add this for Android only, but you'll still need a separate plugin for iOS.

Thanks for the reply. So is it fine to have this feature only in android not in iOS? Should I remove all the methods that I have implemented for iOS?
Another question, why we do not want Firebase dependency for @capacitor/ios, will it is break if someone not using Firebase for their project?

jcesarmobile:
on iOS they should be there, but return unimplemented error

dependencies increase the size of the apps, so adding firebase will increase the size for all user apps despite they don't use it

jcesarmobile:
I got a notification about you requesting me to review, but you have not addressed the changes I proposed on my latest comment.

stewwan:
This issue can be closed.

https://github.com/capacitor-community/fcm

dwieeb:
Thanks for the PR @dibyendusaha! Unfortunately this functionality might not make sense in Capacitor core.

dibyendusaha:
> Thanks for the PR @dibyendusaha! Unfortunately this functionality might not make sense in Capacitor core.

That's fine (Y)

PR #2741, opened 2020-04-10
fix(cli): use appName as package.json name on electron project

### LINKED ISSUE #2604: feat: name and description parameters for `npx cap add electron`
# Feature Request

When the command is run, it creates a package.json with hard-coded values for name and description:
```
$ cat package.json 
{
  "name": "capacitor-app",
  "version": "1.0.0",
  "description": "An Amazing Capacitor App",
... elided ...
```
And when the packager is run, it generates folders based on the “name” value. This is a request for  arguments to set these values so that they don’t have to be edited by hand (or sed in a Github Action).

So something like:
`npx cap add electron --name "myApp" --description "A bit of description"`

Thanks for your consideration.

## Platform Support Requested

- [ ] Android
- [ ] iOS
- [x] Electron
- [ ] Web

## Related Code
Note the `--web_dir` argument for `npx cap init` is similar in spirit.

Mike-Summ:
closes #2604

### DIFF
--- a/cli/src/electron/add.ts
+++ b/cli/src/electron/add.ts
@@ -9,7 +9,7 @@ export async function addElectron(config: Config) {
   await runTask(`Adding Electron project in: ${config.electron.platformDir}`, async () => {
     const copyReturn = await copyTemplate(config.electron.assets.templateDir, config.electron.platformDir);
     const capConfigName = require(join(config.app.rootDir, 'capacitor.config.json')).appName;
-    const packageJSONParse = JSON.parse(readFileSync(path.join(config.electron.platformDir, './package.json')) + '');
+    const packageJSONParse = require(join(config.electron.platformDir, 'package.json'));
     packageJSONParse.name = capConfigName;
     writeFileSync(path.join(config.electron.platformDir, './package.json'), JSON.stringify(packageJSONParse));
     return copyReturn;
--- a/cli/src/electron/add.ts
+++ b/cli/src/electron/add.ts
@@ -8,7 +8,7 @@ export async function addElectron(config: Config) {
 
   await runTask(`Adding Electron project in: ${config.electron.platformDir}`, async () => {
     const copyReturn = await copyTemplate(config.electron.assets.templateDir, config.electron.platformDir);
-    const capConfigName = JSON.parse(readFileSync(path.join(config.electron.platformDir, '../capacitor.config.json')) + '')['appName'];
+    const capConfigName = require(join(config.app.rootDir, 'capacitor.config.json')).appName;
     const packageJSONParse = JSON.parse(readFileSync(path.join(config.electron.platformDir, './package.json')) + '');
     packageJSONParse.name = capConfigName;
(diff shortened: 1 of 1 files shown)


cli/src/electron/add.ts
```diff
@@ -1,11 +1,18 @@
 import { exec } from 'child_process';
 import { Config } from '../config';
 import { copyTemplate, hasYarn, installDeps, runTask } from '../common';
+import { readFileSync, writeFileSync } from '../util/fs';
+import * as path from 'path';
```
jcesarmobile, line 5: Since you are only using join, no need to import *
```suggestion
import { join } from 'path';
```

cli/src/electron/add.ts
```diff
@@ -1,11 +1,18 @@
...
 
 export async function addElectron(config: Config) {
 
   await runTask(`Adding Electron project in: ${config.electron.platformDir}`, async () => {
-    return copyTemplate(config.electron.assets.templateDir, config.electron.platformDir);
+    const copyReturn = await copyTemplate(config.electron.assets.templateDir, config.electron.platformDir);
+    const capConfigName = JSON.parse(readFileSync(path.join(config.electron.platformDir, '../capacitor.config.json')) + '')['appName'];
```
jcesarmobile, line 11: There is no need to read and parse, you can just use require to a json file, also don't use relative urls, we can know the root dir with `config.app.rootDir`.
```suggestion
    const capConfigName = require(join(config.app.rootDir, 'capacitor.config.json')).appName;
```

cli/src/electron/add.ts
```diff
@@ -1,11 +1,18 @@
...
+    const packageJSONParse = JSON.parse(readFileSync(path.join(config.electron.platformDir, './package.json')) + '');
```
jcesarmobile, line 12: Same here, use require ad do't use `./`on the json name
```suggestion
    const packageJSONParse = require(join(config.electron.platformDir, 'package.json'));
```

cli/src/electron/add.ts
```diff
@@ -1,11 +1,18 @@
...
+    packageJSONParse.name = capConfigName;
+    writeFileSync(path.join(config.electron.platformDir, './package.json'), JSON.stringify(packageJSONParse));
```
jcesarmobile, line 14: Same here, don't use `./`
```suggestion
    writeFileSync(join(config.electron.platformDir, 'package.json'), JSON.stringify(packageJSONParse));
```

cli/src/electron/add.ts
```diff
@@ -1,11 +1,18 @@
 import { exec } from 'child_process';
 import { Config } from '../config';
 import { copyTemplate, hasYarn, installDeps, runTask } from '../common';
+import { readFileSync, writeFileSync } from '../util/fs';
```
jcesarmobile, line 4: readFileSync can be removed as require is used for getting the json files

```suggestion
import { writeFileSync } from '../util/fs';
```

Mike-Summ: Always forget about requiring JSON files

PR #2743, opened 2020-04-10
PR to change audio attribute from alarm to notification

dibyendusaha:
for commit #2676 : it is better to have audio attribute as USAGE_NOTIFICATION instead of USAGE_ALARM, as it will give control to device notification volume instead of alarm.
Please find my comment and related finding [here](https://github.com/ionic-team/capacitor/commit/d72e25d81b5b7bfafc6e10d8f1ac7fb35da9dcb8#r38351706)


carlpoole:
Thanks for your contribution. Agreed, this should be on the appropriate sound channel.

dibyendusaha:
> Thanks for your contribution. Agreed, this should be on the appropriate sound channel.

Thanks for accepting the PR

PR #2753, opened 2020-04-13
docs(ce-guides): Add tutorial for React Capacitor AdMob

rathoreabhijeet:
Tutorial with example for Ionic React Capacitor AdMob implementation. Tested on both iOS and Android.


jcesarmobile:
Thanks for the PR, but we don't want people to use old Capacitor versions and you recommend to use 1.2.0 in this guide.

dotNetkow:
Hi @rathoreabhijeet! Thanks for the PR. As Julio mentioned, with Capacitor 2.0 out now, we'd appreciate it focusing on that. Looking at your post, I think it's basically already there? You would know best of course, but appears to me that commands/code would be similar - just remove the Cap 1.x disclaimer etc...

Issue #456, opened 2018-04-12
Issue in IE11 when importing plugins from @capacitor/core


solojuve1897:
The following produces an error in IE11 about proxy not being defined:

```
import { Plugins } from '@capacitor/core'
const { Device } = Plugins
console.log(Device)
```
The error message in the console in my swedish IE11 says: "Proxy har inte definierats" (Proxy has not been defined).

Will this be fixed?

mlynch:
I responded in another issue, but we're strongly leaning towards not supporting IE 11. By supporting it, we'll have to invest time on IE 11 instead of making Capacitor better for all the other platforms we support. Also, IE 11 does not support any PWA APIs.

Various stats show IE 11 usage at < 3% worldwide, and once we get into production ready status that number should drop even more.

It's time to move on.

solojuve1897:
Any suggestion how to condition the import so it occurs only if not on IE11? I want the application to work on IE11 even if it is without the plugins from Capacitor. I played around with "require" instead of using "import". It will compile but the check won't in that case work on my apps for some reason. If you have any thoughts to guide me in the right direction that would be appreciated.

My check:

```
async function getDeviceInfo () {
      try {
        let response = await Device.getInfo()
        return response
      } catch (err) {
        console.log(err)
      }
    }
    getDeviceInfo().then((device) => {
      if (device.platform !== 'web') {
        this.isMobile = true
      }
    })
```

The field "isMobile" will then be used to show/hide some HTML-elements. All this works fine when using the import-statement:

```
import { Plugins } from '@capacitor/core'
const { Device } = Plugins
```

But as I said. Then the application won't start in IE11 at all.

ghenry22:
According to the below link the windows 10 webview is powered by edge so capacitor should still have full support for hybrid apps on the windows universal apps without any need to support IE11.

https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.WebView

moberwasserlechner:
What would I need to do to fix this problem? Could you give me a hint?

Angular supports IE 9 - 11 and beside Capacitor my Angular apps runs  - ok - with IE11 and although I share your opinion on IE especially in non IT companies this is easier said than done.

Some are still on Win7 and IE 11 is the default browser there.

uturnr:
I was able to conditionally require capacitor's Geolocation plugin as follows:
```
let Geolocation;
if (window.Proxy) {
  const capacitor = require('@capacitor/core');
[...]

PR #2759, opened 2020-04-14
fix(web): Gracefully degrade Proxy usage to fix IE11

laurentgoudet:
Fixes https://github.com/ionic-team/capacitor/issues/456. This guards the `Proxy` usage is order to ensures that loading Capacitor does not throw a Syntax Error in IE11 and other older browser engines.

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -59,8 +59,7 @@
     },
     "testMatch": [
       "**/__test__/*.(ts|tsx|js)"
-    ],
-    "testURL": "http://localhost/"
+    ]
   },
   "dependencies": {
     "tslib": "^1.9.0"
--- a/core/package.json
+++ b/core/package.json
@@ -59,7 +59,8 @@
     },
     "testMatch": [
       "**/__test__/*.(ts|tsx|js)"
-    ]
+    ],
+    "testURL": "http://localhost/"
   },
   "dependencies": {
     "tslib": "^1.9.0"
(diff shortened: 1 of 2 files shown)


core/package.json
```diff
@@ -59,7 +59,8 @@
     },
     "testMatch": [
       "**/__test__/*.(ts|tsx|js)"
-    ]
+    ],
+    "testURL": "http://localhost/"
```
laurentgoudet, line 63: FYI I had to add that to make the tests work as `jest` throws `SecurityError: localStorage is not available for opaque origins` otherwise, the reason being that `testURL` defaults to `about:blank` in [v0.22](https://jestjs.io/docs/en/22.x/configuration#testurl-string) whereas it has been fixed to `http://localhost/` in the [latest version](https://jestjs.io/docs/en/configuration#testurl-string).

This can be removed once `jest` has been upgraded, but this was beyond the scope of the PR.

27pchrisl:
For what it's worth, this change made our pretty complex Capacitor application work fine on IE11.

Perhaps although Ionic's (perfectly reasonable) position is that Capacitor isn't supported on IE11, they'll accept community pull requests like this that don't affect the features on engines that are officially supported?

mlynch:
Thanks @laurentgoudet, looks good to me. Curious what @jcesarmobile thinks

laurentgoudet:
> I agree that fixing the tests is out of the scope of this PR, so can you remove the `testURL`?

Done

Issue #1564, opened 2019-05-23
FileReader API not firing


ronwondaal:
Description of the problem:
In my project I'm trying to read and convert a Blob downloaded from the dropbox api to a text string. When building my project using Cordova, the FileReader fires onload and onprogress events. When building using Capacitor, filereader does not fire any events at all.

Affected platform
- [ ] Android
- [x ] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [ ] Windows
- [ x] macOS
- [ ] linux

Other information:
As suggested in other bug reports, I tried wrapping the FileReader in 
```
platform.ready().then(_ => {
/// perform file reading 
})
```
and in 
```
zone.run(() => {
/// perform file reading
})
```
None helped. Cordova has no issues reading the file.

When I use
```
var text = await (new Response(response.fileBlob)).text();
```
to read the file blob, it does work.

Capacitor version: 1.0.0-beta.24

jcesarmobile:
Can you provide a sample project?

ronwondaal:
Okay, so in the sample project it works. Guess the problem lies somewhere in the use of dropbox combined with promises and the filereader. I shall investigate further and when I find something I'll let you know!

ronwondaal:
@jcesarmobile, apparently the problem is in the cordova-plugin-file plugin.
https://github.com/ronwondaal/capacitorFileReaderTest

jcesarmobile:
But on the plugin itself or only when using it with Capacitor?

ronwondaal:
When using it with capacitor it prevents FileRead from firing events.

jcesarmobile:
Looks like some problem with zone.js, but not sure why it only affects Capacitor and not Cordova

If I log the FileReader object I get this results:

Capacitor with Ionic `{"__zone_symbol__originalInstance":{"_readyState":1,"_error":null,"_result":null,"_progress":0,"_localURL":"","_realReader":{}}}`

Cordova with Ionic `{"_readyState":1,"_error":null,"_result":null,"_progress":0,"_localURL":"","_realReader":{"__zone_symbol__originalInstance":{}}}`

Capacitor without framework `{"_readyState":1,"_error":null,"_result":null,"_progress":0,"_localURL":"","_realReader":{}}`

Cordova without framework `{"_readyState":1,"_error":null,"_result":null,"_progress":0,"_localURL":"","_realReader":{}}`

[...]

PR #2785, opened 2020-04-16
Update known-incompatible-plugins.md

yelhouti:
cordova-plugin-file doesn't work with angular because zonejs override changes made to FileReader by the cordova Plugin


jcesarmobile:
See https://github.com/ionic-team/capacitor/issues/1564#issuecomment-538200971

yelhouti:
@jcesarmobile already saw that, but is you use any library that uses FileReader it doesn't work so...

PR #2794, opened 2020-04-18
Add New plugin for capacitor : capacitor-snackbar-gfc

SASGeniusFlashConception:
We would like to make available to the community our new plugin which displays the native snackBar component


jcesarmobile:
It's been a month and the plugin have not been updated, so going to close
feel free to re submit in the future once they support iOS and have English docs

PR #2796, opened 2020-04-19
docs(browser): Updated information for close method

### LINKED ISSUE #2779: Documentation: Inaccurate documentation for Browser Plugin
The typescript documentation currently states that "Browser.close()" only works on iOS (https://github.com/ionic-team/capacitor/blob/master/core/src/core-plugin-definitions.ts#L249). However after trying it in the browser this also seems to work perfectly fine. Is the documentation out date or is it just a coincidence.
If it is outdated, I'd be happy to change it and create a PR.

pgerhard:
Closes #2779


pgerhard:
@jcesarmobile Done

PR #2799, opened 2020-04-19
PhotosPlugin - savePhoto Android implementation

danibranas:
Hi,

I was looking for a save-photo-to-gallery functionality and just found a core undocumented [Photos plugin](https://github.com/ionic-team/capacitor/blob/master/core/src/core-plugin-definitions.ts#L1351), with partial implementation on iOS and empty implementation on Android. That's what I need.

I think it could be interesting to complete this core plugin rather than creating an external one, maybe with a plugin refactoring. I've implemented and committed a functional `savePhoto` method for Android.


carlpoole:
Hi there, thanks for working on this. We appreciate your contribution. After discussing this as a team and with our goals in revamping the existing capacitor features and plugins, we decided we will be sunsetting this photos plugin in our update to Capacitor 3.0. The https://github.com/capacitor-community/media plugin supports saving photos on Android.

PR #2815, opened 2020-04-22
fix(android) : App can crash on clipboard.read if empty

jrdnp:
I encountered app crash when reading the content of the clipboard since 2.0.0+
It only occurs on Android platform and in my case on a Samsung with Android 7.
When my clipboard is empty (like after a reboot) and I attempt to read it, the plugin crashes the app because of a `NullPointerException`.
I checked the Android documentation and found the method [hasPrimaryClip()](https://developer.android.com/reference/android/content/ClipboardManager#hasPrimaryClip()).
So I implemented it in order to check if the primary clip exists. If it's the case, normal behavior happen otherwise the clipboard content will be set to an empty string.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Clipboard.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Clipboard.java
@@ -48,23 +48,35 @@ public class Clipboard extends Plugin {
     ClipboardManager clipboard = (ClipboardManager)
         c.getSystemService(Context.CLIPBOARD_SERVICE);
 
-    CharSequence value = "";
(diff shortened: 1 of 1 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/Clipboard.java
```diff
@@ -48,23 +48,35 @@ public void read(PluginCall call) {
...
+      if (value.toString().startsWith("data:")) {
+        type = value.toString().split(";")[0].split(":")[1];
+      }
     }
+    else {
+      ret.put("value", "");
+    }
+
     ret.put("type", type);
+
```
jcesarmobile, line 79: this changes are not really needed, since you set the value to null the existing
```
ret.put("value", value != null ? value : "");	
    if (value != null && value.toString().startsWith("data:"))
```
covers all cases

PR #2842, opened 2020-04-29
Rename applinks.json to assetlinks.json

dennisameling:
When creating `applinks.json` (as opposed to Google's requirement to call the file `assetlinks.json`, I get the following error when clicking "Test statement" in the [Google Statement List Generator and Tester](https://developers.google.com/digital-asset-links/tools/generator):

> No app deep linking permission found for nl.APP_NAME.app at app.APP_NAME.nl.

Renaming to `assetlinks.json` (like Google also recommends) [solves this issue](https://developer.android.com/training/app-links/verify-site-associations). Updated the docs to reflect this change.


dotNetkow:
Thanks @dennisameling! Slight mistake, good catch. It gets confusing since Apple and Google call them something slightly different :)

Issue #2850, opened 2020-04-30
bug: window is not defined when using server rendering


flogy:
# Bug Report

## Capacitor Version

`npx cap doctor` output:
```
Latest Dependencies:
  @capacitor/cli: 2.0.2
  @capacitor/core: 2.0.2
  @capacitor/android: 2.0.2
  @capacitor/electron: 2.0.2
  @capacitor/ios: 2.0.2

Installed Dependencies:
  @capacitor/cli 2.0.2
  @capacitor/core 2.0.2
  @capacitor/ios 2.0.2
  @capacitor/android 2.0.2
```

## Affected Platform(s)

- [ ] Android
- [ ] iOS
- [ ] Electron
- [x] Web

## Current Behavior
When using server rendering (e.g. with [Gatsby](https://www.gatsbyjs.org/)), a `window is not defined` error appears:

```
WebpackError: ReferenceError: window is not defined

  - web-runtime.js:12 new CapacitorWeb
    node_modules/@capacitor/core/dist/esm/web-runtime.js:12:1

  - global.js:8 
    node_modules/@capacitor/core/dist/esm/global.js:8:1

  - global.js:9 Module../node_modules/@capacitor/core/dist/esm/global.js
    node_modules/@capacitor/core/dist/esm/global.js:9:2

  - index.js:1 Module../node_modules/@capacitor/core/dist/esm/index.js
    node_modules/@capacitor/core/dist/esm/index.js:1:1
```

## Expected Behavior
Server rendering should work fine.

## Sample Code or Sample Application Repo

CodeSandbox: https://codesandbox.io/s/github/flogy/gatsby-starter-capacitor
GitHub Repo: https://github.com/flogy/gatsby-starter-capacitor

Run `npx gatsby build` to see the error in the sample project.

## Reproduction Steps

- Use Capacitor together with a SSR tool like Gatsby.
- Run the SSR process (e.g. build the Gatsby project).
- Error appears in the log.

## How to fix

The problem appears because a `window` variable was introduced in this PR: https://github.com/ionic-team/capacitor/pull/2759

To fix it, we could just add another check if `window` is actually defined:

```javascript
if (typeof window !== `undefined` && 'Proxy' in window) {
```

(This is actually how Gatsby recommends solving such problems: https://www.gatsbyjs.org/docs/debugging-html-builds/#how-to-check-if-window-is-defined)

laurentgoudet:
Uh right, sorry about that. The issue with:

```ts
[...]

PR #2851, opened 2020-04-30
fix(web): improve Proxy check to avoid SSR problems

flogy:
Fixed https://github.com/ionic-team/capacitor/issues/2850

Solves server rendering issues by checking if `window` is actually defined (which is not the case when doing server rendering).

### DIFF
--- a/core/src/web-runtime.ts
+++ b/core/src/web-runtime.ts
@@ -13,7 +13,7 @@ export class CapacitorWeb {
     // Gracefully degrade in non-Proxy supporting engines, e.g. IE11. This
     // effectively means that trying to access an unavailable plugin will
     // locally throw, but this is still better than throwing a syntax error.
-    if ('Proxy' in (typeof window === 'undefined' ? global : window)) {
+    if (typeof Proxy !== 'undefined') {
       // Build a proxy for the Plugins object that returns the "Noop Plugin"
       // if a plugin isn't available
       this.Plugins = new Proxy<any>(this.Plugins, {
--- a/core/src/web-runtime.ts
+++ b/core/src/web-runtime.ts
@@ -13,7 +13,7 @@ export class CapacitorWeb {
     // Gracefully degrade in non-Proxy supporting engines, e.g. IE11. This
     // effectively means that trying to access an unavailable plugin will
     // locally throw, but this is still better than throwing a syntax error.
-    if (typeof window !== 'undefined' && 'Proxy' in window) {
+    if ('Proxy' in (typeof window === 'undefined' ? global : window)) {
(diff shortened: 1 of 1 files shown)


IljaDaderko:
Just verified this with capacitor / nextjs build, all good 👍

jabczyk:
I ran across the same issue when using Capacitor and Gatsby.
After this change Gatsby's SSR works perfectly fine.
Please approve it in the next update, as it delays my deployment.

IljaDaderko:
Any eta on when we can merge this @jcesarmobile ?

core/src/web-runtime.ts
```diff
@@ -13,7 +13,7 @@ export class CapacitorWeb {
     // Gracefully degrade in non-Proxy supporting engines, e.g. IE11. This
     // effectively means that trying to access an unavailable plugin will
     // locally throw, but this is still better than throwing a syntax error.
-    if ('Proxy' in window) {
+    if ('Proxy' in (typeof window === 'undefined' ? global : window)) {
```
dwieeb, line 16:
```suggestion
    if (typeof Proxy !== 'undefined') {
```

flogy: Just tested it in the sample project (https://codesandbox.io/s/github/flogy/gatsby-starter-capacitor). Building works fine and it also still runs in IE 11 browser (which was the initial problem fixed by https://github.com/ionic-team/capacitor/pull/2759).

fildidi:
When should we expect this fix to be released? Because currently im having the same issue with Nuxt js, where Im on 'universal' mode, where since the first page need to be served from the server, the window is not defined.
It is for a site, which is suppose to go live within a week, so I urgently need this fix released, so we can us it
Do you have a time frame in mind, on when can that happen?

flogy:
Found an issue with this on IE 11:

The committed code is correct and functional. However, after it was built and released it again breaked IE 11 compatibility (which was originally fixed by #2759). It seems that  building the code adds `''` around the `Proxy` variable:

This is the line of code as committed in this PR:
https://github.com/ionic-team/capacitor/blob/7afc9eb81ad622282a27b6dd3fca30165bf2af95/core/src/web-runtime.ts#L16

This is the same line of code in my `node_module` dependencies:
```javascript
if (typeof 'Proxy' !== "undefined") {
```

When running the if statement in IE 11 browser it returns the following:

```
typeof 'Proxy' !== "undefined"
true

typeof Proxy !== "undefined"
false
```

It should be `false` but is not after packaging and distributing the `@capacitor/core` package.

Could someone experienced in the used packagers check on this issue?

Thanks in advance.

PR #2631, opened 2020-03-25
feat(camera): make prompt strings localizable

rdlabo:
Now, Prompt message in camera plugin is hard coding.

スクリーンショット 2020-03-25 22 21 19

I think this be able to change users label is better.
Thanks.

PR #2853, opened 2020-04-30
POC: native camera localization

jcesarmobile:
This is a POC for how the plugin localization is done in the native way.

This would require the user to add the native localizations in their project.

For iOS this is creating a `Localizable.strings` file with this content:

```
"From Photos" = "From Photos";
"Take Picture" = "Take Picture";
```
And then localize the file adding supported languages to the project and replacing the strings after the `=` (the ones before are keys that also act as default value)

For Android it requires adding in the user project localized `strings.xml` files and adding

```
<string name="take_picture">translation for Take Picture</string>
<string name="from_photos">translation for From Photos</string>
```


jcesarmobile:
closing in favor of https://github.com/ionic-team/capacitor/pull/2631

PR #2856, opened 2020-04-30
fixes issue where top level arrays were black holed

ProofOfKeags:
This fixes an issue where the capacitor plugin only accepts top level json objects as a response. According to the JSON spec, top level constructs must be either objects or arrays. This should permit both.


mlynch:
Thanks much! Adopted in https://github.com/ionic-team/capacitor-plugin-http/commit/48cc643a6422ef36d0731d44054cb228de2746f4

PR #2871, opened 2020-05-05
feat(ios): show toast when loading url in debug mode

Heerschop:
Added a **Using app server** toast, for iOS when running an app in debug mode on a custom `server.url`. This is functionally the same as the Android toast.


Heerschop:
I was afraid you would mention something like this :-) and I agree. I tried doing it that way, but I don't know anything about swift, this is basically the first time I'm using it.

When I tried I was unable to import a class from a different file. I can make the 'showToast' function nicer like you suggested and make it public.

Heerschop:
Really happy with this update thank you very much.

Issue #70, opened 2018-01-19
Add network request handling from cordova-plugin-ionic-webview


mlynch:
https://github.com/ionic-team/cordova-plugin-ionic-webview/blob/608d64191405b233c01a939f5755f8b1fdd97f8c/src/ios/CDVWKWebViewEngine.m#L609

PR #2872, opened 2020-05-05
feat: Allow plugins to override navigation

jcesarmobile:
variation of #2307 

changes:
Android method renamed from `shouldOpenExternalUrl ` to `shouldOverrideLoad` (more similar to the Android WebView method, and matches the iOS name). Logic changes, false load, true abort, null default.

ios method renamed from `shouldOverrideLoadWith` to `shouldOverrideLoad`, with a single `WKNavigationAction` param instead of two (you can get both request and navigationType from it).
Method return type changed from Bool to NSNumber so it can return nil.
And as it can return nil now, change logic to work as on Android (false load, true abort, nil default)

closes #2307 
closes https://github.com/ionic-team/capacitor/issues/70

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -224,6 +224,19 @@ public class Bridge {
   }
 
   public boolean launchIntent(Uri url) {
+    /*
+    * Give plugins the chance to handle the url
+    */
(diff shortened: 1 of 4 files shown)


dwieeb:
Only one plugin can do this in an app, right?

jcesarmobile:
There can be multiple plugins, but totally depend on them, if they only handle certain url and return null/nil for other cases, then another plugin can take care of that url, if they return true or false for all URLs then other plugins won’t be able to handle them

dwieeb:
Gotcha.

PR #2876, opened 2020-05-06
fix: set launchShowDuration to 0 on new projects only

jcesarmobile:
Instead of setting the default native value to 0 as done on https://github.com/ionic-team/capacitor/commit/b29346b2bf9d41ec3be621f80435b306c1027941

set the default values back to 3000 and make new projects have 
```
"plugins": {
    "SplashScreen": {
        "launchShowDuration": 0
    }
}
```


dwieeb:
Should it just be `showDuration` to match the plugin options at runtime?

image

jcesarmobile:
It was called `launchShowDuration` as it's only applied on the launch of the app, while `showDuration` is for API calls to show method.
We can't change the name now.

Issue #2764, opened 2020-04-14
feat: way to call plugin method from native code


dwlrathod:
# Feature Request
We are having an existing Ionic app with Cordova, we have already started Capacitor conversion. There was a custom plugin we were using with Cordova and converted it to native code right now. Inside it, some of the plugin methods are being used in both FE and native part. Is there any way to call PluginMethod from native code or any alternative to do it?

## Platform Support Requested

- [x] Android
- [x] iOS
- [ ] Electron
- [ ] Web

## Additional Context

Like scheduling Local Notification from Native Code by calling this [method](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java#L70)

jcesarmobile:
No, the plugins can't be called from native code and we don't plan to make that possible.

You can probably mock a plugin call to pass it to the plugin method, or use the webview to run the javascript method that triggers the native code.

BorntraegerMarc:
@jcesarmobile could you explain a bit more why we don't plan to make it possible?

To provide a bit more info: Our use case is for local notifications. We have custom code to download some content on FCM push notifications. Prior to display the notifications. We would like to use the official capacitor plugin to trigger notifications in our custom code & from JS.

PR #2908, opened 2020-05-13
LocalNotificationManager changed few methods to protected

dwlrathod:
As there is no way to call LocalNotification methods from Native Code (https://github.com/ionic-team/capacitor/issues/2764)
We have decided to add our own plugin (with some minor customizations) to do that and for that purpose we want our plugin to extend class [LocalNotificationManager](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java), so we don't miss any update from Capacitor.

This PR will help anyone who wants to do the same in Future with LocalNotificaiton.


dwieeb:
@dwlrathod Thanks for the PR! Push and local notifications are being moved out of the core for Capacitor 3. The use case for doing this will likely look drastically different. For this reason, I'm closing this.

PR #2909, opened 2020-05-13
Update firebaseMessagingVersion

SergioSuarezDev:
Update firebaseMessagingVersion to make Firebase Push Messaging work in the latest version of Capacitor


jcesarmobile:
Thanks for the PR, but we are not updating the template for now.

Anyway, the whole point of the variables file is to make it easier for users to update the versions without waiting for us to update, capacitor just creates it when the android platform is added, but it's managed by the user.

BTW, latest version is 20.1.7

PR #2910, opened 2020-05-13
fix(dependencies): peer dependencies to ~2.1.0

mburger81:
npm WARN @capacitor/android@2.1.0 requires a peer of @capacitor/core@~2.0.0 but none is installed. You must install peer dependencies yourself.

Not sure if you wan't ^ or ~ but, I think ~ should be fine

### DIFF
--- a/android/package.json
+++ b/android/package.json
@@ -16,7 +16,7 @@
     "access": "public"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


danielsogl:
Duplicates #2894 and can be closed

PR #2913, opened 2020-05-13
In Android Studio (and Java in general), 4 space indentation is standard.

smpeters:


jcesarmobile:
thanks for the contribution, but we won't be changing this, once the project is created, users can use the indentation they wish

PR #2917, opened 2020-05-13
Cleanups and fixes to the Android Java code

smpeters:
The Java code in the Android section turned up a good number of code smells between Android Studio and SonarLint.   Some of the changes are stylistic, but several are to remove code that is outdated by Java 8 standards.  There are some bug fixes since there were a few places where the code was comparing Strings with `==` instead of `equals()`.  I've limited the scope of the changes by commit, but the changes to replace anonymous classes with lambdas is pretty big.


dwieeb:
Thank you @smpeters! I merged this into our 3.x branch.

PR #2918, opened 2020-05-13
Update doc for transparency

Raven0us:
I think `register` sounds a bit confusing, and also felt the need to point out that `onNewToken` is actually passed to `registration` listeners. I was expecting it to be a separate event.


jcesarmobile:
thanks for the PR, but I don't think those notes are necessary

PR #2919, opened 2020-05-13
Updated Capacitor Admob Advanced Plugin Name

DTX-Elliot:
Updated the Community plugin page to reflect plugin name


jcesarmobile:
sorry, we won't be changing plugin names

DTX-Elliot:
It's my plugin?

PR #2922, opened 2020-05-14
feat(android): add NavigationBarPlugin

SanLeen:


jcesarmobile:
thanks for the contribution, but this belongs to a separate 3rd party plugin, not to Capacitor core

PR #2924, opened 2020-05-14
fix(android): incorrect keyboard height

duryno:
At the moment, the Keyboard plugin returns incorrect keyboard height on Android phones with cutouts. This fixes it by subtracting the size of the bottom cutout from the total screen height. 

I also removed obsolete code ensuring backwards-compatibility with Android SDK versions <21 since Capacitor doesn't support these versions.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Keyboard.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Keyboard.java
@@ -9,6 +9,7 @@ import android.util.DisplayMetrics;
 import android.view.Display;
 import android.view.View;
 import android.view.ViewTreeObserver;
+import android.view.WindowInsets;
 import android.view.inputmethod.InputMethodManager;
 
 import com.getcapacitor.JSObject;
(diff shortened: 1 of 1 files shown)


jcesarmobile:
how can I reproduce this? 
I don't have any device with cutout, but if I simulate it in a real device or simulator the keyboard size doesn't change with your code

duryno:
> how can I reproduce this?
> I don't have any device with cutout, but if I simulate it in a real device or simulator the keyboard size doesn't change with your code

Hey! 

You can reproduce this with a simulated cutout by looking at the value of `pixelHeightDiff` in native code or `event.keyboardHeight` that you get via e.g. `keyboardWillShow` event in javascript. 

For example, in Pixel 2 simulator with a simulated 'tall cutout' I get `keyboardHeight = -48`  when the keyboard is closed (I expect it to be 0 in this case) and `keyboardHeight = 234` when the keyboard is open. With the fix in this PR I get 0 when closed and 282 when the keyboard is open. The behaviour is the same on my Samsung Galaxy Note 10 with a real cutout.

To further clarify, it's this value I'm referring to, not the height of the actual keyboard (which is unrelated). Hope this makes it clear but do let me know if you need more information.

PR #2927, opened 2020-05-14
Docs - Improvements in capacitor migration doc (From <= 1.5.1 to 2.0.0 - Android)

bipoza:
When migrating Capacitor 1.5.1 to 2.0.0 I have had several errors which I have solved with this.

I have listed them in case it helps. Maybe it is not necessary, but it has helped me.

Thank you!

### DIFF
--- a/site/docs-md/android/updating.md
+++ b/site/docs-md/android/updating.md
@@ -118,4 +118,16 @@ Recommended changes:
 
   In `android/app/src/main/res/xml/file_paths.xml` add `<cache-path name="my_cache_images" path="." />`.
 
+* Change FileProvider to prevent the app crash at startup.
+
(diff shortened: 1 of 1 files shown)


jcesarmobile:
maybe you skipped this step?

> From Android Studio do Refactor -> Migrate to AndroidX. Then click on Migrate button and finally click on Do Refactor.

That does the things you have added

bipoza:
I have followed all the steps in the documentation.
I have checked again with capacitor 1.5.0 and in my case it has not added it.
The same thing happened to me when I migrated several Capacitor projects.

I have tested it with version 3.6.1 and 3.6.3 of Android Studio. But it is surely an Android Studio bug that skips those steps.

Thanks for your time!

jcesarmobile:
I'm using Android Studio  3.6.3 and it's updating those files when I do
> Refactor -> Migrate to AndroidX. Then click on Migrate button and finally click on Do Refactor.

So I won't be adding this changes, but thanks for contributing.

PR #2936, opened 2020-05-15
fix(android): Filesystem.requestPermissions() not working

jcesarmobile:


skr-maag:
Hi, friend 
I have this problem in my project, Is it possible that I explained a little bit of cal was your solution?
  you could request it on ionic or directly on android?
I'm work in Ionic-react with capacitor

PR #2939, opened 2020-05-16
docs(ce-plugins) add @byrds/capacitor-contacts plugin

Idrimi:
Hi,

finally had the time to add Android Support to my Capacitor-Contacts Plugin.
Updated the docs as well. You can find a link to a working demo as well. 

As Contacts fetching is a commonly used native function (and i guess there will be more people creating Contact Plugins) I allowed myself to create a section.


Idrimi:
Hi @jcesarmobile,

I would prefer having it in another section. As those in the "misc" section are rather small and/or visual plugins. Bluetooth, Sensors, Date/Calendar or Deep Links have their own section as well.
Storage Section would make sense as well and I would be fine with that too. What do you think?

jcesarmobile:
I disagree, I already told you to put it in the Misc section in the first PR you sent

Idrimi:
Ok, cool. Then we will put it in the "misc" section. 😜

PR #2952, opened 2020-05-19
Update to androidx.exifinterface.media.ExifInterface

smpeters:
Android Studio complains that `android.media.ExifInterface` is insecure and should be updated to `androidx.exifinterface.media.ExifInterface`.


smpeters:
Sorry, my mistake.  Android lint is what's complaining, not Android Studio.

carlpoole:
Thanks for updating this. I made a small change so the last stable version is used, and for some reason the unit tests for capacitor-android builds were failing until I updated the mockito dependency to a more recent version.

PR #2954, opened 2020-05-20
docs(StatusBar): Fix 'Configuring iOS' link

PawanKolhe:
This PR fixes the 'Configuring iOS' link under iOS notes on the [Status Bar](https://capacitor.ionicframework.com/docs/apis/status-bar/#ios-note) docs page.

Capture


jcesarmobile:
this is a known issue on the docs, if you load `https://capacitor.ionicframework.com/docs/apis/status-bar` the link works, if you load `https://capacitor.ionicframework.com/docs/apis/status-bar/` then it doesn't, so your change fixes it for a case, but breaks it for the other case

PR #2971, opened 2020-05-22
[Docs] Change line wrap inside the JSON block.

smpeters:
The current line wrapping in this block causes part of the comments to be hidden at https://github.com/ionic-team/capacitor/blob/master/site/docs-md/basics/configuring-your-app.md and https://capacitor.ionicframework.com/docs/basics/configuring-your-app/


dwieeb:
@smpeters Apologies this didn't make it in. We split out the Capacitor site to its own repo: https://github.com/ionic-team/capacitor-site

Would you mind recreating this there?

PR #3004, opened 2020-05-28
Update Gradle Wrapper to 5.6.4

friederbluemle:
This fully updates all 4 Gradle Wrappers contained in this repo to `5.6.4`.

As you can see in the diff, it does not actually change the `distributionUrl` in `gradle-wrapper.properties` (it was already set to `5.6.4`). However, according to the sha256sum of the `gradle-wrapper.jar` files, they had not been touched since Gradle 2.10 🤯 (initially added to this repo on Nov 19 2017 - 947aab28b8bd8512b92feaf7610176e0367dceea) The wrapper scripts (i.e. `gradlew`/`gradlew.bat`) were also missing many updates that went into newer releases of Gradle.

With the changes here, all files associated with Gradle Wrapper are updated to match the version `5.6.4` distribution of Gradle.


friederbluemle:
Rebased onto latest `master` branch. Please let me know if there are any questions!

friederbluemle:
Rebased onto latest master branch.

android-template/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,5 @@
-#Tue Jan 30 13:14:22 CST 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
```
dwieeb, line 3: Do you know why we need the `all` distribution (with gradle source code) vs `bin`?

cc @jcesarmobile

friederbluemle: As you mentioned, `all` includes additional sources that makes it easier to work with Gradle scripts. On top of that, it's the default distribution that is used by new projects generated with Android Studio, as well as new projects initialized with the React Native CLI. That makes it much more likely to already exist in the local cache on a developer's machine, saving an additional download.

dwieeb:
Thank you @friederbluemle!! 🍻 This will make it into Capacitor 2.3.0.

Issue #1572, opened 2019-05-24
Unable to inject Capacitor, Plugins won't work


mrodal:
Description of the problem:
Hey, I just got started with capacitor, and Im getting this error on the Android studio console when running the app:
`E/Capacitor: Unable to inject Capacitor, Plugins won't work`

Is it normal?

Affected platform
- [X] Android
- [ ] iOS
- [ ] electron
- [ ] web

OS of the development machine
- [X] Windows
- [ ] macOS
- [ ] linux

Other information:

Capacitor version:
4.12

node version:
10.15

npm version:
6.9

CocoaPods version:
-

Steps to reproduce:

Link to sample project:

jcesarmobile:
Can you provide a little more information? like what steps did you follow

A sample app would be helpful

mrodal:
Of course, its a vue app, so the type in ionic config is custom.
I created the app with vue-cli
I installed ionic as described here https://github.com/ionic-team/ionic/tree/master/vue
Executed `ionic init`
Then `ionic capacitor add android`
Added `"ionic:serve": "vue-cli-service serve",` to the package.json scripts
Executed `ionic capacitor run android --livereload`
Then ran the app on an emulator on android studio

Heres the app:
https://github.com/mrodal/reported-issues/tree/unable-inject-capacitor

mrodal:
Ok, I looked for the code raising the error, and its on JSInjector.java:
![image](https://user-images.githubusercontent.com/20667544/58345359-86ff8980-7e2e-11e9-9cf5-6f76ec186518.png)

This is called first with the site's html, where the script is injected succesfully, and then again a couple of times with content like this:
`{"websocket":true,"origins":["*:*"],"cookie_needed":false,"entropy":1469886054}`

So it is actually working, but it is trying to also inject the scripts where it shouldnt

jcesarmobile:
Yeah, Capacitor injects itself into html files searching for the head tag, but it's trying to inject itself into json files too showing that error message. Will fix it.

mrodal:
If you think its a good solution, I can make a PR checking for the content type on WebViewLocalServer.java:

![image](https://user-images.githubusercontent.com/20667544/58347380-23785a80-7e34-11e9-8d98-0fbf90172e21.png)

jcesarmobile:
Thanks, but I've already sent it.

theunreal:
I am facing this issue, what is the workaround?

jcesarmobile:
[...]

PR #3024, opened 2020-06-01
Document injection into head element

brodybits:
Document that Capacitor assumes it can inject itself into the `head` element

as discussed in:

- https://github.com/ionic-team/capacitor/issues/1572#issuecomment-495720601
- https://github.com/ionic-team/capacitor/issues/1572#issuecomment-495726425
- https://forum.ionicframework.com/t/unable-to-inject-capacitor/178713

I ran into this when migrating a small test app from Cordova to Capacitor, and it took me some research to find the solution.

I will completely understand if this proposal needs to be cleaned up, just hoping we can help someone else avoid this kind of a pitfall.

### DIFF
--- a/site/docs-md/getting-started/index.md
+++ b/site/docs-md/getting-started/index.md
@@ -27,7 +27,7 @@ make sure you update CocoaPods using `pod repo update` before starting a new pro
 
 ## Adding Capacitor to an existing web app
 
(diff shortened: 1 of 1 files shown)


dwieeb:
@brodybits Apologies this didn't make it in. We split out the Capacitor site to its own repo: https://github.com/ionic-team/capacitor-site

Would you mind recreating this there?

PR #3028, opened 2020-06-01
chore: simplify ios publishing

jcesarmobile:
the publish script was creating a capacitor-ios folder and moving ios files inside before publishing to npm

since we no longer need the Capacitor.podspec and CapacitorCordova.podspec in the root of the repository, they can be moved to the iOS folder, so we can remove the templates we use for injecting the version, same for the package.json.template.

This will allow users to install ios platform from a local folder, it has been requested by users and wasn't possible because of all this file copying.


dwieeb:
😍

PR #3031, opened 2020-06-01
Update plugins.md

epicshaggy:


dwieeb:
@epicshaggy We no longer have this table on the Capacitor site (https://capacitorjs.com/docs/community/plugins). Instead, we are encouraging community members to [create proposals](https://github.com/capacitor-community/proposals) for plugins and work with us ([read more](https://github.com/capacitor-community/.github)). This is all very new, so please bear with us. Feel free to ping @mlynch or myself.

Thanks!

PR #3038, opened 2020-06-02
fix(ios/keyboard): update frame immediately when keyboard hides

sbannigan:
Currently, when the keyboard hides, if the user is scrolling in the webview the `keyboardWillHide` event won't fire and the frame won't update until the scrolling stops. See this app that hides the keyboard when scrolling starts:

Current Behavior:
ezgif com-video-to-gif

Fixed Behavior:
ezgif com-video-to-gif (1)


jcesarmobile:
can you provide a sample app where the issue can be reproduced?

sbannigan:
https://github.com/sbannigan/capacitor-ios-keyboard-scrolling-bug

PR #3040, opened 2020-06-02
Find plugins recursively when calling "cap update" or "cap sync"

SmileyJames:
Currently, plugins are only found if they are listed in the package.json as a dependency. Therefore, if one of my dependencies depends on a capacitor plugin, the plugin is not found or copied into the native application. I propose finding plugins in any dependency in node_modules, using “read-installed” a package used by npm. That way any capacitor plugin you depend on is used by the native apps.


imjacobclark:
This is super helpful for Capacitor architectures where you have some Capacitor app build that pulls in some externally compiled JavaScript app and puts it inside the Capacitor `dist` folder. 

It decouples Capacitor from the rest of the application architecture, enabling a developer to build a standalone, decoupled component that is able to selectively use Capacitor plugins where appropriate by platform detection, and falling back to other implementations when the component is not within a Capacitor runtime.

This also essentially enables existing, legacy websites to be bundled into Capacitor without forcing the existing website to go all-in on a Capacitor PWA deploy, they can simply extend their existing codebase to pull Capacitor plugins where appropriate without disrupting their existing PWA development workflows where selecting Capacitor for app store distribution might come after the fact.

Diagram of usecase attached.

cap-runtime-resolution

SmileyJames:
@dwieeb Thank you for taking the time to take a look at this - much appreciated.

> How would a user opt-out of using a particular plugin? Today, they'd simply uninstall the plugin.

If you are depending on package A which depends on a capacitor plugin: package B. If you don't want to depend on the capacitor package B, I think you will need to stop depending on package A. I can't see a simple way around this. Perhaps it is worth investigating how this PR treats "optionalDependencies" and "peerDependencies".

> Also, are there benchmarks for this? I have to imagine this slows down update substantially.

I assume this is slower than the current implementation. However, the cli command still feels responsive - hard for a human to notice a difference I think. The performance could be improved, currently both android and ios call `getPlugins` - memoizing the output of `getPlugins` would roughly half run-time when using `cap update`. After implementing memoizing, I'd happily provide some benchmarks on an example repo with a significant number of dependencies.

dwieeb:
I am tempted to suggest that this be solved with `peerDependencies`. With this strategy, for example, the JS App would depend on a JS Component, which in turn depends on a Capacitor Plugin. Since peer dependencies aren't installed (rather the user is warned about them missing), the user would explicitly install the Capacitor Plugin in their app (meaning it shows up in the app's `package.json`). This makes sense, because only one version of a Capacitor Plugin can be used in an app at a time. This keeps things explicit and requires no code changes in the Capacitor CLI.

What are your thoughts?

SmileyJames:
@dwieeb

My most recent commits address your concerns regarding plugins with multiple versions and the speed of the command.

The most recent plugin version will be used and warnings printed about the other ignored plugin versions.

All cli commands now only call getPlugins once, improving the performance. getPlugins is wrapped in a runTask, which adds a spinner and times the task logging runtime, for our project (with a substancial number of dependencies) it takes less than 4 seconds to find all plugins.

### Use Case

**TL;DR** Enable Capacitor to take _any_ JS module that distributes a PWA and generate Android and iOS apps in a completely automated fashion - no typing, no human error. It's so close at the moment, but you've still got to edit MainActivity.java and update your peerDeps.

Unfortunately, peerDependencies do not solve the use case outlined by @imjacobclark.

We aim to produce and maintain a number of npm modules that compile to PWAs - and that is their sole responsibility - we have multiple apps across the estate each will be a PWA. Each app uses our reusable components, many of which are capacitor plugins, or depend on capacitor plugins.

Separately we have a "capacitor app shell", this is another npm module which depends on one of those PWA modules at a time, it consumes the compiled output (a PWA in the "dist" folder) and observes the dependency tree to find and add the various required plugins. Everything concerning shipping the PWAs to platforms lives here (signing ios apps for example), perhaps consider it a build pipeline. We only need to maintain 1 "app shell", but many PWAs. What we are trying to acheive here is an automated process, where we can change the "app shell" to depend on one of many PWAs in our estate and then output Android and iOS apps with a single command. We need to avoid maintaining duplicate "app shells" or "build pipelines" for each of those PWAs. There is only one pipeline, but it can consume *any* PWA.

Using peerDependencies would require installing those dependencies in the "capacitor app shell". Each time we point it to a different PWA those dependencies change and there is not automated way to update them.

dwieeb:
@SmileyJames Thanks for the explanation!

SmileyJames:
Sorry for the delay in replying

>   The CLI no longer is able to locate Cordova plugins (I believe that is due to a typo [here](https://github.com/ionic-team/capacitor/pull/3040/files#diff-9e79e4f5b899f91f9cada4c5c0b84675R107), but after fixing it, more errors occurred.)

Nice spot, I need to try this out with Cordova plugins, haven't been using any

>   Class generation seems incorrect (not every Capacitor plugin's java class has the same name as the plugin, and technically it's possible to have multiple plugin classes per npm package)

This is really interesting, **can you explain more about how js plugins find java classes (or point to some code/docs)**? I don't fully understand how a js plugin call get's mapped to a Java method - I was assuming it was through the naming of the plugin and java class.

> Additionally, it may be worth playing with alternative implementations of `getInstalled`. read-installed seems to have [a lot of logic](https://github.com/npm/read-installed/blob/master/read-installed.js) surrounding dependency resolution. Maybe just a vanilla readdirp operation looking for `package.json` and `plugin.xml` files would be even faster. An operation that used to take ~1ms now takes over 1.5s on my 2019 16" macbook pro. I just want to ensure the CLI stays as fast as possible.

I think this may be a faster approach, worth investigating.

dwieeb:
> can you explain more about how js plugins find java classes (or point to some code/docs)

Right now, `init()` is called on the `BridgeActivity` with a list of additional initial plugins supplied by the user. For each of these plugins, `registerPlugin()` is called where it gets the "simple" class name [here](https://github.com/ionic-team/capacitor/blob/b4815a5/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L444). Today, plugin class names have to be unique in an app or they'll clash. The name of the plugin itself isn't used--just the plugin class and methods that have the `@PluginMethod` annotation.

dwieeb:
@SmileyJames Any chance of getting this updated this week or next week? We are (possibly) going to do a feature freeze on Capacitor 2.x to focus on Capacitor 3 (see https://github.com/ionic-team/capacitor/issues/3133).

SmileyJames:
@dwieeb I think I could wrap this up in a week or so.

So android you have to specify what classes you use in MainActivity's init. That is not the case on iOS right? It's automatic?

Have you seen [this PR](https://github.com/ionic-team/capacitor/pull/3054) that builds on this one to make android's class imports automatic aswell? I'd like to hear your thoughts on it. In the meantime I will just remove console logging out android classes on this branch.

I'm happy to get this working with Cordova. Also I will try your suggestion to speed up getInstalled (just read files - no read-installed dependency).

It's great to hear that you are interested in merging this in and thank you for all the feedback.

dwieeb:
@SmileyJames Makes perfect sense! Yes, I did take a look at the other PR and I love the concept, but I'm not convinced we can reliably gather the Android class names without a more complex algorithm.

SmileyJames:
@dwieeb I've addressed your requested changes:
- No android classes are printed anymore
- Using readdirp not read-installed has sped up finding plugins from ~4s to ~2s (for the project I was testing at least)
- Now works with cordova plugins

dwieeb:
@SmileyJames It feels faster! 😄 

I noticed for official Cordova plugins it detects the `tests/plugin.xml` that they ship inside the package.

```
  Found 2 Cordova plugins for ios
    cordova-plugin-device (2.0.3)
    cordova-plugin-device-tests (2.0.3)
```

Cordova plugins should only be detected if the `plugin.xml` is at the root of the package. This can be accomplished a number of ways, one being checking if the path segments match a certain pattern:

- `node_modules/<package>/plugin.xml`
- `node_modules/<@scope>/<package>/plugin.xml`

There should be no other valid sequence of path segments for Cordova plugins. This works for recursive deps, too.

SmileyJames:
@dwieeb Test plugins are no longer found - only package.json / plugin.xml at root of a package

This may need testing on Windows, which uses backslash not forward slash. Hopefully node fs coerces them into a unix style?? If not the regex pattern will not match.

I don't have access to a Windows machine personally.

dwieeb:
Yeah, it won't work on Windows. You might be able to get away with constructing the regex string using [`path.sep`](https://nodejs.org/dist/latest-v12.x/docs/api/path.html#path_path_sep), e.g. `node_modules' + path.sep + '...`

dwieeb:
@SmileyJames I pushed a few commits to demonstrate what I mean.

I noticed the tests are failing legitimately. We didn't account for monorepos. Because of how node resolves modules, it is possible to have a dependency in the app's `package.json` that is installed in a higher directory. I should've caught it sooner--it may throw a wrench in our current approach.

dwieeb:
@SmileyJames @imjacobclark Our Capacitor 2 feature freeze is approaching (probably Monday). Given the scope of this change and recent concerns (re: monorepos), I'm tempted to mark this and https://github.com/ionic-team/capacitor/pull/3054 as features for Capacitor 3.

dwieeb:
Closing for now, but I made https://github.com/ionic-team/capacitor/issues/3265 to track this use case and I had an idea so please see my comment there. We may bring back this implementation, but right now we want a clean slate for Capacitor 3. Lots of improvements coming! 

Thanks for the PR!

PR #3054, opened 2020-06-04
Java reflection to automatically import android plugins into MainActivity.java

SmileyJames:
This branch builds on my other pull request: https://github.com/ionic-team/capacitor/pull/3040

Plugins are recursively found (other PR) then their java class paths are written to a json file. BridgeActivity then reads the json file and uses Java reflection to import and add the android plugin to the build.

The advantage of using Java reflection is that no manual editing of MainActivity.java is never required, as it currently is as described here: https://capacitor.ionicframework.com/docs/plugins/android/#export-to-capacitor. Capacitor would be able to convert a PWA to an android and ios app in a **completely automated** fashion.

This change is completly backwards compatible. Overriding onCreate in MainActivity.java will "opt out" of Java reflection, if you want to manually add plugins as before you can. Migration path is, keep doing what you are doing as before and no reflection will be used. Nothing changes for existing users.

It seems wise to seperate the concerns of these two PRs - if only to seperate discussion of the impact of these changes.


dwieeb:
Closing this as well (see https://github.com/ionic-team/capacitor/pull/3040#issuecomment-658411318). Please track this feature here: https://github.com/ionic-team/capacitor/issues/3266

Thank you!

PR #3055, opened 2020-06-04
feat(android): move Config to be per-instance rather than a singleton

jcesarmobile:
equivalent of https://github.com/ionic-team/capacitor/commit/e9640689af31c62e6c2416f1f289b07231368cb8 for Android

Used a JSONObject instead of a JSON formatted string (what iOS does) because JSONObject can be created from a JSON formatted string already, and can also put values in a simpler way than a JSON formatted string.

thoughts @carlpoole @dwieeb?


dwieeb:
I don't know either, other than passing in `config` for every log method call 🤮

carlpoole:
I think it would be messy if we had a logger instance per Config instance so I'm thinking have `shouldLog` as a static global variable on Logger that can be set by doing something like

`Logger.enable()`

and making it an all or nothing thing. I don't like the idea of tying the logger to the config when there is a possibility of having multiple configs.

jcesarmobile:
I've passed the config to logger on bridge init and now it works.

I've noticed that there is a "Starting BridgeActivity" log that is always displayed since it's logged before the config is set, but was also happening with previous implementation.

carlpoole:
Looks good!

PR #3056, opened 2020-06-04
chore(github): issue template cleanup

dwieeb:


jcesarmobile:
We should probably remove the platforms checkboxes, most people get it wrong and try to put an x on creation and ends not displaying correctly, and also makes it hard to search for, in example, an android bug, since all issues have android word on it.
And I think we should require a sample project because in most cases it’s a misconfiguration or using mismatching version or problems on the user code.

dwieeb:
@jcesarmobile 👍

PR #3057, opened 2020-06-04
chore(ci): switch to Github Actions

dwieeb:


jcesarmobile:
Cli test should run on macos since it adds iOS platform, so some things won’t be tested on Ubuntu.

Is it possible to specify a Xcode version? Or know which one was used? We only support Xcode 11+, but looks like they have a lot of versions, including Xcode 10.3

dwieeb:
We can use `xcode-select` to use [one of the installed versions](https://github.com/actions/virtual-environments/blob/master/images/macos/macos-10.15-Readme.md#xcode). Looks like the default today is 14.4.1. I think it's fine to let CI use whatever is the default Xcode. We can also use a [matrix strategy](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) to test multiple versions of Xcode.

jcesarmobile:
I think one version is enough

PR #3068, opened 2020-06-08
feat(Permissions): allow microphone check

KyDenZ:
Add checkMicrophone in permission plugin


KyDenZ:
Ok it's good.

Are you planning the Microphone in capacitor ? (Recording, access request) ?

jcesarmobile:
not planning on adding a microphone plugin for now, but a few things can trigger the microphone permission at the moment, like some cordova plugins, input file with capture attribute, getUserMedia on Android/browser, etc. so having the microphone check might be helpful

PR #3091, opened 2020-06-11
feat(cli): Allow to use capacitor commands in custom platforms

jcesarmobile:
This is a POC

It allows to have custom platforms in Capacitor

Example platform https://github.com/jcesarmobile/test-platform

Steps to test:
1. Install custom platform: `npm install https://github.com/jcesarmobile/test-platform`
2. Run capacitor CLI commands by passing the platform name:  examples `npx cap add @jcesarmobile/test-platform`, `npx cap copy @jcesarmobile/test-platform`, `npx cap update @jcesarmobile/test-platform`, `npx cap sync @jcesarmobile/test-platform`, `npx cap open @jcesarmobile/test-platform`.

Custom platforms need 5 scripts in scripts section of `package.json` (not all mandatory if the platform don't need them, but will show an error)

```
"scripts": {
    "capacitor:add": "script to run on npx cap add",
    "capacitor:copy": "script to run on npx cap copy",
    "capacitor:update": "script to run on npx cap  update",
    "capacitor:sync": "script to run on npx cap  sync",
    "capacitor:open": "script to run on npx cap  open"
  }
```

At first I also did a `npx cap add customPlatorm` command that just ran npm install on the custom platform (could be http url, npm package, local folder, etc. whatever npm install supports), but then thought it might not be a good idea as it install the package and might install the wrong thing if the user type it wrong, so I think it's better not to wrap it and just tell the user to use npm install


cli/src/tasks/copy.ts
```diff
@@ -10,15 +10,25 @@ import { getCordovaPlugins, handleCordovaPluginsJS, writeCordovaAndroidManifest
...
-    return;
-  }
-  try {
-    await allSerial(platforms.map(platformName => () => copy(config, platformName)));
-  } catch (e) {
-    logError(e);
+  if (selectedPlatformName && !config.isValidPlatform(selectedPlatformName)) {
+    const platformFolder = resolveNode(config, selectedPlatformName);
+    if (platformFolder) {
+      const result = await runCommand(`cd "${platformFolder}" && ${await hasYarn(config) ? 'yarn' : 'npm'} run copy`);
```
dwieeb, line 16:
```suggestion
      const result = await runCommand(`cd "${platformFolder}" && ${await hasYarn(config) ? 'yarn' : 'npm'} run capacitor:copy`);
```

cli/src/tasks/sync.ts
```diff
@@ -1,28 +1,38 @@
...
-    const now = +new Date;
-    const diff = (now - then) / 1000;
-    log(`Sync finished in ${diff}s`);
-  } catch (e)  {
-    logFatal(e);
+export async function syncCommand(config: Config, selectedPlatformName: string, deployment: boolean) {
+  if (selectedPlatformName && !config.isValidPlatform(selectedPlatformName)) {
+    const platformFolder = resolveNode(config, selectedPlatformName);
+    if (platformFolder) {
+      const result = await runCommand(`cd "${platformFolder}" && ${await hasYarn(config) ? 'yarn' : 'npm'} run sync`);
```
dwieeb, line 15: We should probably not do this. sync = copy + update, so we should just run `npm run capacitor:copy` followed by `npm run capacitor:sync`. What do you think?

cli/src/tasks/update.ts
```diff
@@ -2,29 +2,39 @@ import { Config } from '../config';
...
-  }
-  try {
-    await check(
-      config,
-      [checkPackage, ...updateChecks(config, platforms)]
-    );
+  if (selectedPlatformName && !config.isValidPlatform(selectedPlatformName)) {
+    const platformFolder = resolveNode(config, selectedPlatformName);
+    if (platformFolder) {
+      const result = await runCommand(`cd "${platformFolder}" && ${await hasYarn(config) ? 'yarn' : 'npm'} run update`);
```
dwieeb, line 13:
```suggestion
      const result = await runCommand(`cd "${platformFolder}" && ${await hasYarn(config) ? 'yarn' : 'npm'} run capacitor:update`);
```

dwieeb:
I'm thinking we actually _should_ implement `add`. Here are my thoughts:

`npx cap add <platform>` should resolve `<platform>` on npm using this strategy:

1. `@capacitor/<platform>` (official platforms)
1. `@capacitor-community/<platform>` (community platforms)
1. `<platform>` (third-party platforms, and we should probably have a prompt to confirm installation)

Once the package is resolved,
- run `npm install <pkg>` in the project
- run the `capacitor:add` script in the package's `package.json`

dwieeb:
@jcesarmobile @IT-MikeS please review

IT-MikeS:
Haven't been able to dig a ton but this current state works for me in testing after `this.initElectronConfig();` in `setCurrentWorkingDir(....) {....}` in `config.ts` of the CLI is commented out.

Only issue I found in the little time I got was that we lose the console output from the script that is run via `capacitor:add` hook

dwieeb:
@IT-MikeS I noticed that, too. If you don't mind, I'd like to tackle that at a later time. It may require some CLI refactoring.

IT-MikeS:
@dwieeb Yeah thats fine, it functions, and the logs do show after the script ends so if an error is thrown the user can see it and deal with it.

IT-MikeS:
This PR (jcesarmobile/capacitor#1) uses spawn instead of exec to run commands and thus doesn't hide live console output.

dwieeb:
To sum up, we want this feature to go into 2.3.0 so people can start using @IT-MikeS's new electron platform. We can't remove the existing electron platform without a breaking change (see https://github.com/ionic-team/capacitor/issues/3140).

So, the caveat is that people will need to use the commands like this until Capacitor 3:

```
npx cap add @capacitor-community/electron
npx cap sync @capacitor-community/electron
```

Instead of simply using `electron`. 

@IT-MikeS's PR is a quick fix to achieve more advanced output from the platform (we talked about it). For Capacitor 3, I would like to refactor how the Capacitor CLI handles subprocesses.

jcesarmobile:
should we also add an open command?
people asked for it in the current electron platform and what it does is to run `npm run electron:start`, so maybe custom platforms can also have a custom open script

dwieeb:
Yes. If platforms can't handle it, they can just print a message.

IT-MikeS:
I agree with adding it. 

I'll put one in the community platform

cli/src/tasks/open.ts
```diff
@@ -1,26 +1,35 @@
...
-  if (platforms.length === 0) {
-    logInfo(`There are no platforms to open yet. Create one with "capacitor add".`);
-    return;
-  } else if (platforms.length === 1) {
-    platformName = platforms[0];
+export async function openCommand(config: Config, selectedPlatformName: string) {
+  if (selectedPlatformName && !config.isValidPlatform(selectedPlatformName)) {
+    const platformFolder = resolvePlatform(config, selectedPlatformName);
+    if (platformFolder) {
+      const result = await runCommand(`cd "${platformFolder}" && ${await hasYarn(config) ? 'yarn' : 'npm'} run capacitor:open`);
```
IT-MikeS, line 11: Shouldn't this be `runPlatformHook` instead?

dwieeb: Ah yeah, I forgot to merge before doing this. 😂

jcesarmobile:
I can't approve since the PR was originally mine, but looks good, updated my test platform and all commands work fine.

PR #3093, opened 2020-06-11
chore: remove package-lock.json files

jcesarmobile:
from core and cli, other packages don't have it or don't commit it


dwieeb:
Please add a local `.npmrc` files to each package so the npm CLI doesn't re-add the lock files. (see https://github.com/ionic-team/ionic-cli/blob/develop/packages/%40ionic/cli/.npmrc for an example)

PR #3095, opened 2020-06-11
feat(ios): Add ios capacitor config to use a custom wkwebview

rtrap95:
I needed I way to hide the context menu when text is selected on ios which it seems to be impossible in iOS13 (css doesn't work). I created this pull request to give users the possibility to create a custom wkwebview and use it. This way I can write a class like this (always extend WKWebView): 
```swift
class CustomWKWebView: WKWebView{
    override func canPerformAction(_ action: Selector, withSender sender: Any?) -> Bool {
        return false
    }
}
```
and use it to hide the context menu. 
User could add this in capacitor.config.json to enable this feature otherwise normal WKWebView is used:
```json
{
  "ios":{
    "customWebView":"NameOfTheCustomWebView"
  }
}
```

Let me know what do you think about this solution

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -72,7 +72,13 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKScr
       webViewConfiguration.applicationNameForUserAgent = appendUserAgent
     }
 
-    webView = WKWebView(frame: .zero, configuration: webViewConfiguration)
+    let moduleName = Bundle.main.infoDictionary!["CFBundleName"] as! String
+    if let customWebView = capConfig.getString("ios.customWebView"), let webViewClass = NSClassFromString(moduleName.replacingOccurrences(of: " ", with: "_")+"."+customWebView) as? (WKWebView).Type{
+        webView = webViewClass.init(frame: .zero, configuration: webViewConfiguration)
+    } else {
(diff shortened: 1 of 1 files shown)


ikeith:
@rtrap95 thanks for the PR! However, we don't feel that adding custom class names to the config file is the route we want to go down. Our aim is to support this use case in 3.0 by allowing you to subclass CAPBridgeViewController in your app to accomplish this type of override.

I created [#3191](https://github.com/ionic-team/capacitor/issues/3191) to track that issue.

rtrap95:
Thank you @ikeith ! I'm looking forward to this feature! Let me know if I can help in some way.

joewoodhouse:
@ikeith Not sure I totally follow how sublassing CAPBridgeViewController and extending WKWebView are the same thing? I have a similar requirement where I need to override a function of WKWebView (`point()`) but I've got not obvious way to hook into that. Even if I hack/subsclass CAPBridgeViewController that doesn't let me do this, does it?

ikeith:
They aren't the same thing, but making CAPBridgeViewController explicitly subclassable provides ways to address a lot of different use-cases. In 2.x, the instantiation and configuration of the webview and bridge communication all happens in one big block so overriding it isn't useful; it was never designed for that. But when it's refactored we can provide individual methods that subclasses could override to change how it works. To address this particular example, it could have a method like [UIView's layerClass](https://developer.apple.com/documentation/uikit/uiview/1622626-layerclass) to supply the class to use instead of hardcoding WKWebView. And it could have methods to modify properties on the WKWebViewConfiguration or the web view after it's been created.

Issue #3106, opened 2020-06-15
feat: allow setting of hostname/startpage via code


crixx:
### Description

In Cordova, it was possible to set the hostname dynamically from code. This was used to allow testers and designers to verify, that a certain app build is compatible with all versions of the SPA deployed to different envs, without the need to build and install each env separately.

My use case is as follows:
- after a fresh install a default hostname is loaded
- then the tester can switch in the settings to a different env; the view is reloaded and shows the SPA from the selected env
- the env is persisted at selection, such that on a fresh app start, the env from the settings is loaded
- in our case, since it's the only Settings.bundle in the app, we can exclude it from the production-build of the app

### Platform(s)

- ios
- android?

### Preferred Solution

OPEN FOR SUGGESTIONS!

My idea but I'm not a swift dev:
CAPBridgeViewController.swift:
- private property `dynamicHostname`
- setter for private property `dynamicHostname`
- if `dynamicHostname` is set, use it to load the webView instead of the hostname from the settings/local

### Alternatives

- override-able configuration hook (! CAPBridgeViewController is not open)
- extendable protocol

### Usage

Plugin authors can set the `dynamicHostname` in their Plugin's load() method:
- retrieve environment from the settings and compare it to the current env
- if mismatch, set the dynamicHostname property via the setter
```swift
 public override func load() {
        if(ENVIRONMENT.name != EnvironmentName.default){
            let vc = self.bridge.viewController as! CAPBridgeViewController
            vc.setDynamicHostname(ENVIRONMENT.backend.absoluteString)
        }
    }
```

dwieeb:
In Capacitor 3, we plan to allow multiple instances of Capacitor running within the native app. This should address your use case. You can track this here: https://github.com/ionic-team/capacitor/issues/3182

crixx:
Hi @dwieeb

If I got it right, dynamic configuration is already in the roadmap for Capacitor v3 with #3141 ? I think that would be the nicer way for this exact use case, since it would allow a single web view as in "production". All we need is a hook that can be used to edit/generate the configuration before the web-views are loaded.

What do you think?

dwieeb:
I'm not sure I understand. The intended usage would be that the native app manages Capacitor instances, so you could have one Capacitor instance for a particular hostname and another for a different one and switch between the instances at runtime.

crixx:
Hi @dwieeb 

[...]

PR #3107, opened 2020-06-15
feat(ios): add setter to allow setting the hostname programmatically

crixx:
Draft implementation for #3106 

Open Questions:

- [ ] are there already ways to do it that I overlooked?
- [ ] is this the right way?
- [ ] are there tests that need to be adapted?!
- [ ] I'm not Swift developer, feel free to give any feedback that comes to mind!


crixx:
@dwieeb any news / feedback  on this?

dwieeb:
@crixx What does this look like for Android?

dwieeb:
@crixx Thanks for the PR! We won't be adding this at this time (see [my comment in your issue](https://github.com/ionic-team/capacitor/issues/3106#issuecomment-651932681)).

PR #3112, opened 2020-06-16
fix(android): LocalNotification action not dismissing notification

jcesarmobile:
I've noticed some actions are not dismissing the notification when you choose an action, all of them should, so removing the check.


tomkahn:
@jcesarmobile Shouldn't the notification only be dismissed when either "destructive" (of the LocalNotificationAction) or "autoCancel" (of the LocalNotification) is set to true? That's how I thought it would work from reading [the documentation](https://capacitorjs.com/docs/apis/local-notifications). At the moment it seems that no matter what I set for these props, the notification is always dismissed when I tap on it or on one of its buttons. The app is also always brought to the foreground (ignoring foreground: false). I only tested this on Android 9 & 10, though.

PR #3119, opened 2020-06-17
Filesystem plugin methods

27pchrisl:
When developing a plugin that works with the filesystem (eg zip plugin), it's really handy to be able to pass the same { directory, path } syntax that the Filesystem methods use. This change pulls the helper methods that can convert a directory/path to a URL (iOS) or File (Android) so a plugin can really easily get hold of a native object representing the target file.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
@@ -7,6 +7,7 @@ import android.content.pm.PackageInfo;
 import android.content.pm.PackageManager;
 import android.net.Uri;
 import android.os.Bundle;
+import android.os.Environment;
 
 import androidx.appcompat.app.AppCompatActivity;
 import androidx.core.app.ActivityCompat;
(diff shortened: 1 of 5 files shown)


ikeith:
@27pchrisl thanks for the PR! After discussing this, we agreed that while it makes sense to pull these functions out of the Filesystem plugin so they can be shared, the Plugin superclass isn't the right place for them. Their functionality is self-contained and would be better suited to be in a file helper utility. Can you refactor the functions into a separate file?

jcesarmobile:
In Android there is a FileUtils.java class with some file helper methods, and on iOS there is CAPFile.swift also with similar methods, so maybe those classes can be used instead of new ones.

ikeith:
@27pchrisl With 3.0, we will be breaking many of the plugins out of the core into separate packages (see #3227). So some of this refactoring could happen naturally as part of that process. But we need to close this PR to prepare for that work. Please track that issue and/or submit an updated PR once the plugin lands in its new home.

PR #3131, opened 2020-06-22
Clarify "missing web assets directory" error message

dotNetkow:
Update the error message displayed when a user tries to add a native platform to their project, before having built their web code.

Changes:
- Fix the More Info link to point to the correct page (/basics/building-your-app instead of /basics/configuring-your-app).
- Help the user by telling them _which_ property to change in capacitor config (webDir).
- Added build details for both Ionic apps and non-Ionic apps.

Wasn't sure how to format correctly for newlines - this function doesn't use log("") like the others.

### DIFF
--- a/cli/src/common.ts
+++ b/cli/src/common.ts
@@ -29,9 +29,12 @@ export async function checkWebDir(config: Config): Promise<string | null> {
   }
   if (!await existsAsync(config.app.webDirAbs)) {
     return `Capacitor could not find the web assets directory "${config.app.webDirAbs}".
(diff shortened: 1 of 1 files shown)


dwieeb:
It's using template literals. Newlines are printed as-in onto the screen. It actually preserves your spaces too so it prints kinda weird, like this:

```
Capacitor could not find the web assets directory "./www".
    Please create it and make sure it has an index.html file. You can change
    the path of this directory in capacitor.config.json (webDir option). You may need ...
```

Issue #3134, opened 2020-06-22
bug: Switch must be exhaustive (Xcode 12)


danielsogl:
## Bug Report

Compiling an Capacitor app with Xcode 12 causes some build error like not exhaustive switch statements in the `Permissions.swift` file

jcesarmobile:
was fixed in https://github.com/ionic-team/capacitor/pull/3160

Abhi0725:
For others having this issue run:

`npm install @capacitor/ios@latest`

then sync the project:

`npx cap sync ios`

gubbigubbi:
> For others having this issue run:
> 
> `npm install @capacitor/ios@latest`
> 
> then sync the project:
> 
> `npx cap sync ios`

Confirmed - fixed for me

PR #3135, opened 2020-06-22
fix(capacitor): make switch statements exhaustive

danielsogl:
fixes: #3134


ios/Capacitor/Capacitor/Plugins/Permissions.swift
```diff
@@ -61,6 +61,8 @@ public class CAPPermissionsPlugin: CAPPlugin {
         ret = "denied"
       case .authorized:
         ret = "granted"
+    case .limited:
+        ret = "granted"
```
danielsogl, line 65: @jcesarmobile Maybe we should return `limited` as result?

ios/Capacitor/Capacitor/Plugins/Permissions.swift
```diff
@@ -97,6 +99,8 @@ public class CAPPermissionsPlugin: CAPPlugin {
         ret = "denied"
       case .notDetermined:
         ret = "prompt"
+      case .ephemeral:
+        ret = "granted"
```
danielsogl, line 103: @jcesarmobile Maybe we should return `ephemeral` as result?

dwieeb:
@danielsogl It looks like the iOS build is failing for this PR, would you be able to look into it?

danielsogl:
@dwieeb  I think the error is exclusive to iOS 14 because the new given options are not available for iOS 13.

ikeith:
Yes, the additional cases will break compilation under previous versions of Xcode. This needs some` #available` checks to conditionally reference the new cases only when building with the beta(s).

danielsogl:
I will update the PR

ios/Capacitor/Capacitor/Plugins/Permissions.swift
```diff
@@ -97,8 +101,11 @@ public class CAPPermissionsPlugin: CAPPlugin {
         ret = "denied"
       case .notDetermined:
         ret = "prompt"
+      #if swift(>=5.3)
+      case .ephemeral:
+        ret = "ephemeral"
       }
-      
+      #endif
```
ikeith, line 108: This `#endif` needs to be moved up. When compiled out, it's breaking the brackets on the switch statement.

ikeith:
After some examination, the problem is a bit more complicated. I've opened a new PR, #3160, to clean things up with the enums. Please feel free to comment on that if you have any concerns.

PR #3148, opened 2020-06-24
chore(android): improve error message when Filesystem.copy fails

asztal:
In my case, this failed on Android 10 due to not having `android:requestLegacyExternalStorage="true"`
in `AndroidManifest.xml`. At least if there's a useful error message, it's a starting point for debugging.

I noticed the documentation for the FileSystem plugin does mention this for the FileSystemDirectory.Documents
enum member, but actually this can also be required when copying a file produced by the Camera plugin into the Data
directory. 

Presumably this was introduced when upgrading the target SDK version from 28 to 29, so it might be worth calling
out in the 2.0 upgrade notes [here](https://capacitorjs.com/docs/android/updating#from-1-5-1-to-2-0-0).

I'm also a little concerned about what I read about Android 10+. If Capacitor doesn't switch to using the Storage Access Framework, will copying photos from the camera stop working entirely at some point?
https://commonsware.com/blog/2019/06/07/death-external-storage-end-saga.html


mepc36:
Hey @jcesarmobile! Does your comment quoted below mean that all we have to do to be able to safely remove `android:requestLegacyExternalStorage="true"` from our AndroidManifest.xml file is save the media files created by our apps to somewhere other than public folders like `Documents` or `ExternalStorage`? Thanks!

>About the copying problems you mention, I had no problem copying a photo taken from Camera plugin to Data directory. Data, Cache and External should be safe to use in future Android releases as they belong to your app.
The folders that were restricted in Android 10 and that could go away in a future release are public folders like the ones used by Documents or ExternalStorage.

Thanks for all your hard work on the repo! If I haven't supplied you with enough context to answer my question, just let me know and I'll fill it in ASAP 👍

PR #3149, opened 2020-06-24
Make switch statements exhaustive

assertchris:
When building with Xcode 14 (beta), these switch statements generate errors and warnings; as they aren't exhaustive.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/Permissions.swift
+++ b/ios/Capacitor/Capacitor/Plugins/Permissions.swift
@@ -37,12 +37,12 @@ public class CAPPermissionsPlugin: CAPPlugin {
     
     var ret = "prompt"
     switch (authStatus) {
-    case .notDetermined:
-      ret = "prompt"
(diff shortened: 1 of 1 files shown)


dwieeb:
@assertchris Would you be able to comment on the differences between this and https://github.com/ionic-team/capacitor/pull/3135?

assertchris:
@dwieeb I would not, as I don't understand what that PR does/is for. This is just correcting an issue currently being reported. I don't see why these two bits of code can't co-exist.

dwieeb:
Whoops, I assumed they were changing the same method 😂 That's what I get for not expanding the diff.

ikeith:
`@unknown default` should be preferred for these switches since that will trigger a compiler warning when new enum cases are publicly added. That, plus the desire to explicitly handle all known cases, means that I've opened a new PR, #3160, to unify the handling of all the enums. Please feel free to comment on that if you have any concerns.

PR #3160, opened 2020-06-25
fix(iOS): Making permissions switch statements exhaustive & supporting new iOS 14 cases

ikeith:
This PR supersedes [#3135](https://github.com/ionic-team/capacitor/pull/3135) and [#3149](https://github.com/ionic-team/capacitor/pull/3149). It unifies the handling of permission enums and fixes the compiler errors under Xcode 12 with the iOS 14 SDK.


ios/Capacitor/Capacitor/Plugins/Permissions.swift
```diff
@@ -52,31 +54,42 @@ public class CAPPermissionsPlugin: CAPPlugin {
...
-      case .authorized:
-        ret = "granted"
+    case .denied, .restricted:
+      ret = "denied"
+    case .authorized:
+      ret = "granted"
+    #if swift(>=5.3)
+    case .limited:
+      // TODO: address this new case properly
+      #warning(".limited != .authorized, authorization status should be revisted for iOS 14")
```
dwieeb, line 67:
```suggestion
      #warning(".limited != .authorized, authorization status should be revisited for iOS 14")
```

ios/Capacitor/Capacitor/Plugins/Permissions.swift
```diff
@@ -52,31 +56,44 @@ public class CAPPermissionsPlugin: CAPPlugin {
...
-        ret = "prompt"
-      case .denied, .restricted:
-        ret = "denied"
-      case .authorized:
-        ret = "granted"
+    case .denied, .restricted:
+      ret = "denied"
+    case .authorized:
+      ret = "granted"
+    #if swift(>=5.3)
```
MortenGregersen, line 66: The new enum cases are not bound to Swift 5.3. They are added in iOS 14 and the availability check should check for os version instead.

jcesarmobile: As far as I know, they are only available on iOS 14 if you use SDK 14 to compile (Xcode 12), if you use SDK 13 (Xcode 11) then they won't be available because it works in a sort of "compatibility mode".

The `#if swift(>=5.3)` is only true on Xcode 12, it's false in Xcode 11 because it ships with swift 5.1 to 5.2.4.

I agree that using `if #available(iOS 14, *)` would be more clear, but sadly, it can't be used inside the switch.

Maybe is less confusing if we use `#if compiler(>=5.3)` instead of `#if swift(>=5.3)`?

ikeith: Yes, it is true that language version is not synonymous with the SDK version but in practice it's not an issue. The only way it would be a problem is if you downloaded a newer version of the Swift compiler and replaced the toolchain in Xcode 11. If you did that, you should know what you're doing and can expect side effects in many places.

Not only is `#available` invalid inside a switch statement (so we'd need to have duplicated switches to use it), it doesn't actually solve our problem because that's a hook to insert a runtime check, meaning the new enum case would fail to compile under Xcode 11. There is no compiler flag that I'm aware of to strip code based on the SDK or OS version being targeted.

At this time, we need to be able to build cleanly with Xcode 11 _and_ 12. Ultimately, this is only temporary for some months until 12 becomes the default.

buda2570:
@ikeith   Thanks for the contribution to solve this problem of verisions, I have compiled my application in Capacitor again.

Issue #1147, opened 2019-01-23
Persistent local notification (Android)


nathantaal:
I want to be able to make notifications on android persistent (So it cannot be swiped away) is this already possible?

jcesarmobile:
It's not possible.
As far as I know, the setOngoing should be true to be persistent, but right now it's not configurable
https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/LocalNotificationManager.java#L150

nathantaal:
There are lots of todo's there but it should be as simple as putting ongoing in the localNotification object and reading it instead of setting it hardcoded to false. I might do that I have have some spare time at the end of this week.

jcesarmobile:
Issues tagged with feature request are closed but tracked for reactions to gauge interest

nathantaal:
A few hours before closing I actually opened capacitor on GitHub to look at the statues of my issue (I looked how far my issue was away from the last page, I figured low-priority issues coming closer to the last page would eventually be resolved.)
In general, I think for capacitor to succeed, it's needs a few full-featured API's and not many API's that are all implemented half. I'm not saying local notifications needs to be this API but last time I checked (And yes, that is a half year ago) most a API where's just half-done.

DasNiels:
Would love to have this aswell, a simple boolean value shouldn't be too hard to implement no?

jcesarmobile:
Removing feature request tag since it was implemented in https://github.com/ionic-team/capacitor/pull/3165

PR #3165, opened 2020-06-27
(Android)Added option to make a notification ongoing(so it can't be swiped away)

DasNiels:
Found this issue(https://github.com/ionic-team/capacitor/issues/1147) while looking for a solution myself, didn't find one so decided to add my own based on the information given in that issue.

One can specify whether they want the notification to be ongoing(so it can't be swiped away) from the javascript side now.

Example(copy-pasted mostly from the docs):
```javascript
const notifs = await LocalNotifications.schedule( {
    notifications: [
        {
            title: "Title",
            body: "Body",
            id: 1,
            schedule: { at: new Date( Date.now() + 1000 ) },
            sound: null,
            attachments: null,
            actionTypeId: "",
            extra: null,
            ongoing: true, // << The new bit
            autoCancel: false // << see edit below
        }
    ]
} );
```

EDIT: Also added option to set autoCancel manually. default is still set to true as it was before.


carlpoole:
Thank you for your contribution. I merged in master and tested.

PR #3184, opened 2020-06-30
3.x

dwieeb:
The branch for working on Capacitor 3.

This branch will become the default branch soon, once the other PRs are mostly merged/closed.

**TODO**

- [ ] Create a `2.x` branch
- [x] Squash and merge changes from embedded Capacitor fork
- [x] Squash and merge https://github.com/ionic-team/capacitor/pull/3215
- [ ] Rebase `3.x` into `master` to retain history


android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
```diff
@@ -48,17 +48,12 @@ public void onShowCustomView(View view, CustomViewCallback callback) {
     callback.onCustomViewHidden();
     super.onShowCustomView(view, callback);
   }
-  
-  @Override
-  public void onHideCustomView() {
-    super.onHideCustomView();
-  }
```
jcesarmobile, line 55: this method is need for full screen video button to appear in players

dwieeb: This was a change from this PR: https://github.com/ionic-team/capacitor/pull/2917

My understanding is a method that only calls the parent class's method can simply be removed and it will use the parent class's implementation.

jcesarmobile: I only know that if you remove it, the button disappear, if you add it back, the button appears, so despite it only calls the super method, looks like it's needed for the full screen button to appear and work.

PR #3201, opened 2020-07-02
chore(cli): pin cli version instead of latest on plugin generation

jcesarmobile:
When generating a new plugin, pin the core, ios and android dependencies to the cli version used to build the plugin

### DIFF
--- a/cli/src/tasks/new-plugin.ts
+++ b/cli/src/tasks/new-plugin.ts
@@ -97,6 +97,7 @@ export async function newPlugin(config: Config) {
     const pluginPath = removeScope(answers.name);
     const domain = answers.domain;
     const className = answers.className;
+    const cliVersion = config.cli.package.version;
 
     if (await existsAsync(pluginPath)) {
       logFatal(`Directory ${pluginPath} already exists. Not overwriting.`);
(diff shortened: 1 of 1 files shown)


cli/src/tasks/new-plugin.ts
```diff
@@ -234,13 +235,13 @@ function generatePackageJSON(answers: NewPluginAnswers) {
     author: answers.author,
     license: answers.license,
     dependencies: {
-      '@capacitor/core': 'latest'
+      '@capacitor/core': `${cliVersion}`
```
dwieeb, line 238: We should allow a version range, otherwise it will install many different `@capacitor/core` packages.

```suggestion
      '@capacitor/core': `^${cliVersion}`
```

cli/src/tasks/new-plugin.ts
```diff
@@ -234,13 +235,13 @@ function generatePackageJSON(answers: NewPluginAnswers) {
...
     },
     devDependencies: {
       'rimraf': '^3.0.0',
       'typescript': '^3.2.4',
-      '@capacitor/ios': 'latest',
-      '@capacitor/android': 'latest'
+      '@capacitor/ios': `${cliVersion}`,
+      '@capacitor/android': `${cliVersion}`
```
dwieeb, line 244:
```suggestion
      '@capacitor/ios': `^${cliVersion}`,
      '@capacitor/android': `^${cliVersion}`
```

PR #3204, opened 2020-07-03
feat(IOS/Android/Web): Camera Permission Request

KyDenZ:
Add Camera permissionRequest :

* [x] IOS
* [x] Android
* [ ] Web

I suggest removing requestPermissions from all plugins and adding this if the plugin needs permission.

```typescript
const { Camera } = Plugins;

await Camera.requestPermission();
```

There is a confusion between requestPermission and requestPermissions, I suggest renaming requestPermission for any request for permission from a plugin.

I need help with requestPermission on the web.


dwieeb:
Thanks for the PR @KyDenZ! There is a lot happening with [plugins for Capacitor 3](https://github.com/ionic-team/capacitor/issues/3227). This is one of the things we're going to be changing, likely drastically. I'm closing this for now, as it will be addressed in Capacitor 3. Thanks again!

PR #3233, opened 2020-07-09
feat(android): share both text and url at the same time

nikosdouvlis:
Hello everyone,

This is a minor refactor that allows to share a file (a url that starts with `file:`) with text on android. I noticed that the ios version already supports this functionality.

I'm not sure if this is considered a "breaking change" but if it gets merged it **will** affect what is shared on the following scenario: 
1. android device
2. share plugin called with both `url` and `text`

Before fix: It will only share the file
After fix: it will share the file and will also have text if the receiving app supports it.

Demo:
https://drive.google.com/file/d/141Q0G3hA2tf8n5k_rNMc7WkhJClrle6i/view?usp=sharing

All feedback is welcome :)

Have fun


dwieeb:
Works great, thank you! 🎉

Issue #3228, opened 2020-07-09
Plugin Template updates


dwieeb:
The following can be addressed before the feature freeze:

- [x] the CLI prompt recommends a class name suffix of `Plugin`, which results in a `PluginPlugin` suffix for web
- [ ] ~class names are inconsistent between native and web "I entered MyPlugin and that created a plugin that's accessible via Capacitor.Plugins.MyPlugin"~ (moved to https://github.com/ionic-team/capacitor/issues/3267)
- [x] Podspec has a bunch of whitespace issues
- [x] `rollup.config.json` is present, but rollup is neither installed nor configured (also should be updated to use [the new node-resolve](https://github.com/rollup/plugins/tree/master/packages/node-resolve))
- [ ] ~provide an initial separation between functionality and bridge communication (see https://github.com/ionic-team/capacitor/discussions/3222)~ (moved to https://github.com/ionic-team/capacitor/issues/3267)
- [x] Provide prettier, swiftlint etc. and whatever the Capacitor Community settles on for consistent tooling (cc @priyankpat)
- [x] [strict mode](https://www.typescriptlang.org/docs/handbook/compiler-options.html), [strict mode](https://www.typescriptlang.org/docs/handbook/compiler-options.html), [strict mode](https://www.typescriptlang.org/docs/handbook/compiler-options.html)! and clean up `tsconfig.json`

PR #3241, opened 2020-07-10
Plugin Template updates

dwieeb:
These changes are for Capacitor 2.3. See planned changes to the plugin template for Capacitor 3 here: https://github.com/ionic-team/capacitor/issues/3267

You can see a list of files the CLI generates from this PR here: https://github.com/ionic-team/capacitor-plugins/pull/1/files

resolves https://github.com/ionic-team/capacitor/issues/3228


cli/src/tasks/new-plugin.ts
```diff
@@ -167,24 +167,24 @@ async function createIosPlugin(config: Config, pluginPath: string, domain: strin
...
+  s.version = package['version']
+  s.summary = package['description']
+  s.license = package['license']
+  s.homepage = package['repository']['url']
+  s.author = package['author']
+  s.source = { :git => package['repository']['url'], :tag => s.version.to_s }
+  s.source_files = 'ios/Plugin/**/*.{swift,h,m,c,cc,mm,cpp}'
+  s.ios.deployment_target  = '${config.ios.minVersion}'
+  s.dependency 'Capacitor'
+  s.swift_version = '5.0'
```
ikeith, line 185: The swift version can be bumped since 5.1 is the oldest version supported by Xcode 11.
```suggestion
  s.swift_version = '5.1'
```

jcesarmobile:
After this changes, building a plugin takes an unreasonable amount of time (around 5 minutes on my computer) that makes it look like the process hanged. 

And build command output this, not sure if it might be a real problem
> (!) `this` has been rewritten to `undefined`
> https://rollupjs.org/guide/en/#error-this-is-undefined

cli/src/tasks/new-plugin.ts
```diff
@@ -46,19 +46,19 @@ export async function newPlugin(config: Config) {
...
       name: 'name',
-      message: 'Plugin NPM name (kebab-case):',
+      message: 'Plugin npm name (kebab-case. ex: capacitor-plugin-example):',
       validate: requiredInput
     },
     {
       type: 'input',
       name: 'domain',
-      message: 'Plugin id (domain-style syntax. ex: com.example.plugin)',
+      message: 'Plugin id (domain-style syntax. ex: com.example.plugins.example)',
```
jcesarmobile, line 55: isn't too much example?

cli/src/tasks/new-plugin.ts
```diff
@@ -221,47 +221,61 @@ function generateAndroidManifest(domain: string, pluginPath: string) {
...
+    'keywords': [
       'capacitor',
       'plugin',
       'native'
     ],
-    capacitor: {
-      ios: {
+    'capacitor': {
+      'ios': {
         src: 'ios',
```
jcesarmobile, line 271: since you are wrapping everything with `''`, shouldn't src be wrapped too? 
```suggestion
        'src': 'ios',
```

plugin-template/android/Plugin.java
```diff
@@ -6,10 +6,10 @@
 import com.getcapacitor.PluginCall;
 import com.getcapacitor.PluginMethod;
 
-@NativePlugin()
+@NativePlugin
```
jcesarmobile, line 9: Not related to the PR itself, but if the `()` is not really needed (tested and seems to work fine without it), we should probably remove it from all our core plugins, they all have it

dwieeb: Prettier will remove them automatically as it did for the plugin template

PR #3253, opened 2020-07-14
fix: requestPermission reflects user settings

### LINKED ISSUE #3252: bug: LocalNotifications.requestPermission always returns success on android
### Capacitor Version
```
Latest Dependencies:

  @capacitor/cli: 2.2.1
  @capacitor/core: 2.2.1
  @capacitor/android: 2.2.1
  @capacitor/electron: 2.2.1
  @capacitor/ios: 2.2.1

Installed Dependencies:

  @capacitor/cli 2.2.1
  @capacitor/android 2.2.1
  @capacitor/core 2.2.1
  @capacitor/ios 2.2.1

  Found 7 Capacitor plugins for ios:
    @capacitor-community/native-audio (0.1.3)
    cordova-plugin-advanced-http (2.5.1)
    cordova-plugin-badge (0.8.8)
    cordova-plugin-file (6.0.2)
    cordova-plugin-insomnia (4.3.0)
    cordova-plugin-network-information (2.0.2)
    cordova-plugin-purchase (10.2.0)
```

### Platform(s)
Android

### Current Behavior
`LocalNotifications.requestPermission` is [hardcoded to return success](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java#L95).
```
{
  "granted": true
}
```

### Expected Behavior
when I turn notifications off in settings, I should get:
```
{
  "granted": false
}
```

### Code Reproduction
https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java#L95

### Other Technical Details

`npm --version` output:
6.14.5

`node --version` output:
v12.17.0

### Additional Context
an easy fix would be to copy from [`areEnabled`](https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java#L120). will submit a PR now to do so.

kheftel:
fixes #3252 

I understand if this is working as intended, if so feel free to disregard. I felt it could be nice to unify the api so that one could call `requestPermission` as a way to check for notifications being permitted/enabled on both iOS and Android.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/LocalNotifications.java
@@ -2,6 +2,8 @@ package com.getcapacitor.plugin;
 
 import android.content.Intent;
 
+import androidx.core.app.NotificationManagerCompat;
+
 import com.getcapacitor.JSArray;
 import com.getcapacitor.JSObject;
 import com.getcapacitor.NativePlugin;
@@ -91,9 +93,9 @@ public class LocalNotifications extends Plugin {
 
   @PluginMethod()
   public void requestPermission(PluginCall call) {
-    JSObject result = new JSObject();
-    result.put("granted", true);
-    call.success(result);
+    JSObject data = new JSObject();
+    data.put("granted", manager.areNotificationsEnabled());
+    call.success(data);
   }
 
   @PluginMethod()


jcesarmobile:
Thanks for the PR

It's not documented for LocalNotifications plugin, but on PushNotifications same method says it just returns true on Android, so it's technically a breaking change as should be changed on both plugins. And since it's a breaking change, it should be included in Capacitor 3, but Capacitor 3 will move some plugins out of core, including Push/Local notification ones, so I'm closing for now and you can re send once we have a repo for the plugins we move out

kheftel:
no worries, thanks for the update! Yeah, I noticed `areEnabled`, I will need to check the iOS native code to see if there's any difference between that and `requestPermission` on that platform, on first check that shows the user a dialog box, vs later checks that don't. Mainly just looking for a cross-platform way to occasionally bug the user to enable notifications :).

Capacitor is a great platform and I hope to use / contribute more in the future!

kheftel:
> PushNotifications same method says it just returns true on Android

apologies, @jcesarmobile, I don't see in [the docs](https://capacitorjs.com/docs/apis/push-notifications) where it says that about Android. I checked the source code though and yeah, it does definitely return a hard-coded true. It also looks like local notifications has the `areEnabled` call, but push notifications does not.

No matter, now that I know how it works I can make it work on both platforms. thanks for the reply!

PR #3263, opened 2020-07-14
chore(cli): add deprecation notice for electron

dwieeb:


dwieeb:
Yeah, and for now I'm just linking to the repo because the installation instructions there will always remain current.

PR #3264, opened 2020-07-14
feat(web): support PWA Elements 2.x

mlynch:


mlynch:
Closing, no need to change the event names, was able to just fix the issue in pwa elements directly.

PR #3283, opened 2020-07-17
chore: remove electron remaining files

jcesarmobile:
Looks like the package.json got back or not removed, and also was present in lerna.json


dwieeb:
We also need to remove Electron from the CLI code.

Issue #1980, opened 2019-09-19
feat: add .gitignore to plugin template


danielsogl:
Description of the problem:

Running the `npx @capacitor/cli plugin:generate` command will not generate a `.gitignore` but a `.npmignore` file that duplicates the `files` configuration inside the `package.json` file.
Also the command should init the plugin with git.

OS of the development machine
- [ ] Windows
- [x] macOS
- [ ] linux

Steps to reproduce:
run `npx @capacitor/cli plugin:generate`

mburger81:
Error still persists, does someone have a usefull template for .gitignore?

jcesarmobile:
2.4.0 should generate the `.gitignore` file
was fixed in https://github.com/ionic-team/capacitor/pull/3292

PR #3285, opened 2020-07-17
fix(cli): avoid npm gitignore rename on new plugins

jcesarmobile:
npm renames .gitignore files to .npmignore files. That makes the plugin generator lose the .gitignore file the template has (only happen when running from npm installed package, not present if using locally linked/installed CLI)

This PR renames the .gitignore to gitignore and the CLI on plugin generation renames the file back to .gitignore

closes https://github.com/ionic-team/capacitor/issues/1980


dwieeb:
I noticed this as well. Would it make sense to create an archive of each CLI asset directory and then unpack it?

jcesarmobile:
Can this renaming cause problems to prefer the archive and unpack?
We do this renaming at the moment for android-template and ios-template too, so if we do the pack/unpack those should work the same way.

dwieeb:
I just personally think renaming/moving files from an original template adds complexity. We might want to reevaluate this for Capacitor 3.

We can ship this quick, though.

dwieeb:
See https://github.com/ionic-team/capacitor/pull/3292.

Issue #1952, opened 2019-09-11
Options for max width/height or keep aspect ratio for Camera


zyniq82:
I have a feature request about adding options to the Camera plugin for setting max `width`/`height`, alternatively a new option for keeping aspect ratio (e.g. `keepRatio`).

This is somewhat related to #217, in which @mlynch wrote:
> _Just added this feature to master and will release tomorrow. you can set `width` and/or `height`. If you set one but not the other, it does an aspect ratio scale to fit that dimension._

What I need is a further extension of the scaling capabilities, I want to be able to limit the _longest_ side of the photo, yet keep the aspect ratio. I've used the [Cordova Camera plugin](https://cordova.apache.org/docs/en/latest/reference/cordova-plugin-camera/index.html) before, in which this was possible.

The problem now is that
- if I specify both `width` _and_ `height`, I'm sure no side is too long, but then the aspect ratio of the resulting image is changed.
- if I specify only `width` _or_ `height`, the aspect ratio is retained, but then I don't know if my prescribed value is applied to the longer or shorter side of the image, since I don't know beforehand if the image is in portrait or landscape format.

I don't know if it's good that `width` and `height` work like in the Cordova case, always retaining aspect ratio, or if it's better to have an additional option enforcing that aspect ratio is kept (e.g. `keepRatio`). The latter alternative gives more flexibility, but maybe makes things too complicated.

Let me know what you think of this!
Is this a feature that can be added?

zyniq82:
For reference, it seems that the problem @Cclown98 is experiencing is not the problem for which I've opened this issue. However I think @Cclown98's issue may be fixed by pull request #2123 (correct me if I'm wrong).

PR #3297, opened 2020-07-20
Add preserveAspectRatio to CameraOptions while resizing an image

iphilgood:
Hi guys

First of all, this is my very first PR to the capacitor repo.

This PR is a retry of PR #2050 and should also fix ionic-team/capacitor-plugins#2 (also the discussion #1952). I've used the Android implementation of @sandstrom (shout-out) and also did an implementation for iOS.

There's now a `preserveAspectRatio` flag on `CameraOptions` which is set to `true` by default.

I've also fixed a bug where the app would crash while saving an image to the gallery on Android. This only happens on certain devices but I was able to reproduce it by creating 32 images until it crashed. A random UUID gets now appended to the file name to make sure no name clashes happen anymore. 

If there's anything I need to change or enhance, please let me know.


sandstrom:
@iphilgood Interesting with the Android crashes. We've had an issue like that which we couldn't track down (couldn't reliably reproduce). Could be something else, but eager to try this if this ends up being merged.

Minor, but I'd argue that `preserveAspectRatio` is what everyone want, so I wouldn't even add that as an option.

The only reason for adding an option would be backwards compatibility, but if that's the case I'd name it `legacyIgnoreAspectRatio`, make the default value true (to avoid a breaking change), but mention that 99% of people would want to have it set to false. Then only keep it until the next major release where that option would be dropped.

dwieeb:
@iphilgood @sandstrom I'm guessing you want this before Capacitor 3? If so, you will need to switch the base branch of this PR to `2.x`. `main` is for Capacitor 3 now that we're in a feature freeze. We could probably squeeze this in as a "fix" (even though it's really a feature). 

We do want to maintain backwards compatibility, even for the weird behavior in this case. `preserveAspectRatio` will need to be defaulted to `false`. In Capacitor 3 we will change the default and perhaps even remove the option altogether.

iphilgood:
@sandstrom Yeah it took me some time to really find out what the issue was. I initially thought it has something to do with an inserted SD Card for the device. But after taking 32 images while testing I was able to reproduce it and I hope your crashes are gonna be fixed with this as well 😊 

Regarding the aspect ratio I totally agree with you. But in my case I'm just happy to prevent distorted images in any way. In the meantime I've created my own Camera Plugin which prevents the distorted images. As soon as this gets hopefully merged, I'm gonna remove my custom plugin again. And thanks for your part of the implementation 👍🏽 

@dwieeb Do you already know when Capacitor 3 is going to be released? Otherwise I think it would be nice if you could squeeze it in as a fix.

I'm going to close this PR and create a new one which is based on the `2.x` branch. Just to maintain the history. I'm also going to set the default of `preserveAspectRatio` to `false`.

PR #3305, opened 2020-07-21
Plugin Import updates

dwieeb:
This is an implementation of the [plugin import proposal](https://github.com/ionic-team/capacitor/issues/3306). To achieve consistent imports in each platform, I recommend we move the `Plugins` proxy from user code to plugin author code.

`registerPlugin` should be used to register implementations of plugins in each platform
`registerWebPlugin` will continue to work so people can continue using the plugin proxy and the web implementation (we can avoid a breaking change)

TODO
- [ ] ~add a `compat.ts` for legacy (PluginRegistry, etc.)~

closes: https://github.com/ionic-team/capacitor/issues/3306


dwieeb:
`registerWebPlugin` got moved here: https://github.com/ionic-team/capacitor/pull/3305/files#diff-1cb8bd56af1537f63fe48eb4fd61aad0R85-R104

Existing plugins should continue working, with a console warning. We'll have to add an upgrade guide for plugins to switch to the new `registerPlugin` method so that the user can use the new imports across all platforms.

mlynch:
I saw that but wasn't sure where the actual registration was happening anymore now that this line is gone: https://github.com/ionic-team/capacitor/pull/3305/files#diff-8aa3e4905a661a3f822d1f6cdf593c15L22

dwieeb:
Oops, updated!

mlynch:
I'm still missing where the actual merge/register is being called for the core web plugins? If it's not called somewhere explicitly then the web plugins won't be registered, right?

dwieeb:
I can add it in, but we're going to be splitting out the plugins anyway in the next few days.

mlynch:
Right that makes sense

dwieeb:
Okay this is good to go. @jcesarmobile?

Issue #2747, opened 2020-04-10
bug: LocalNotification `on` repeats reminders every hour after first reminder


dotrub:
# Bug Report

## Capacitor Version
`npx cap doctor` output:

  @capacitor/cli 1.5.0

  @capacitor/core 1.5.0

  @capacitor/android 1.5.0

  @capacitor/ios 2.0.0

## Affected Platform(s)

- [x] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

## Current Behavior
Set up a LocalNotification. I want to set up daily repeating reminders at a time the user has set. The `on` option seems best suited, so I collect a time, extract the hours and minutes, and set those values as `on: { hour: ${time.hour}, minute: ${time.minute} }`. At the scheduled time, the reminder triggers; I tap it and do what it wanted me to. An hour later, when the minutes of the time match the `on` field again, I get another reminder, even though the hour shouldn't match. I continue to get reminders for every hour after the original.

Worth noting that I tried to see if it was because `hour` wasn't being respected, by setting my reminder to about an hour before now and waiting for the minutes to match. The reminder does not pop up in this case; this bug is only present after the reminder has triggered the first time.

## Expected Behavior
`on` should only trigger when all the fields match. If I set hour and minute, I expect the reminder to only fire when the hour and minute match.

## Sample Code or Sample Application Repo

Not much code to provide for this one. Here's most of the reminder object I'm setting:

```
const reminder = {
    id: 100,
    schedule: { 
        on: {
            hour: 12,
            minute: 30,
        },
    },
};
```

With this reminder, my reminder triggers at 12:30pm, and then at 1:30pm, 2:30pm, and so on...

## Reproduction Steps

Set a reminder like the one above and wait for it to trigger. Then wait an hour and see it trigger again.

## Other Technical Details

`npm --version` output: 6.9.0

`node --version` output: v10.16.3

IT-MikeS:
I gave this a shot, and I wasn't able to reproduce this with a clean `ionic start myApp tabs --type=react`

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

dwieeb:
There's an open PR that fixes this, so I'll reopen.

jcesarmobile:
closing since the linked PR was merged. Should be fixed in 2.4.0

PR #3307, opened 2020-07-21
fix(android): LocalNotification `on` not working properly

kheftel:
I've read the contributing docs and I hope I'm doing this properly, please feel free to educate me on how to best approach this.

I needed to use the cron-like `on` notifications on a project of mine to make daily notifications happen at a certain time (like every day at 8:30 am) and noticed the Android implementation didn't work properly. On iOS, it all appears to be handled by `DateComponents`, but on Android it's got a more complex custom implementation.

It's a bit hard to find documentation, but after examining the native code, it appears that the `on` style notifications are supposed to let you:

- pass a minute and have a notification repeat each hour (passing 15 would trigger every hour at :15)
- pass an hour and a minute and have a notification repeat each day (passing 14 and 30 would result in a notification at 2:30pm each day, exactly my use case and what led me to investigate this)
- pass a day, hour, and minute and have a notification repeat for example on the 5th of every month at 8:30am.
- pass a month, day, hour, and minute and have a notification repeat each year for example on May 1st at 3pm.

I could be wrong in my understanding, of course.

On the iOS side, Apple's [documentation for `DateComponents`](https://developer.apple.com/documentation/foundation/datecomponents) isn't very clear, tbh.

Googling turned up some problems people were having ([link](https://github.com/ionic-team/capacitor/discussions/2752), [link](https://github.com/ionic-team/capacitor/issues/2747)), and I ran into my own issues as well. When trying to have a notification fire on the hour, it sometimes fired not on the hour, and if I passed hour and minute, I would sometimes get a notification that triggered at the next whole hour instead of the next day, and when I passed hour and omitted minute, I ended up with a notification that repeated ad infinitum as soon as I dismissed it, etc.

The existing logic appeared to be incomplete. It calculates the next time the notification should fire by first taking the current moment, setting the provided hour, minute etc, seeing if the resulting trigger time is in the past, and then bumping it by a certain amount. But it was only comparing to see if one unit was less than another (i.e. if hour < hour or if minute < minute), and sometimes it bumped by the wrong amount. Now it compares the timestamp of the two dates and explicitly remembers how much each iteration of the notification should be advanced by, according to the largest unit specified.

I've fixed it for my use case, i.e. I can pass hour and minute, and if device time is 2:15pm and I schedule it for 2:30pm, it goes off 15 min later, and if device time is 2:30 pm and I schedule it for 2:15pm, it schedules for the next day at 2:15pm. I left a logging statement commented out in the code for now that helped me test it and verify it's working - I expect to remove it before merge. The logic should be sound for the longer-repeating cases as well, though I haven't tested those on-device yet (I can certainly do that if I'm going about this right).

Please let me know how to proceed next, I am very impressed with this platform and am excited to be building my app on it, and if I can also contribute something back, so much the better. Thanks!

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
@@ -108,14 +108,9 @@ public class DateMatch {
         next.set(incrementUnit, next.get(incrementUnit) + 1);
       }
     }
-    // SimpleDateFormat sdf = new SimpleDateFormat("yyyy/MM/dd HH:mm:ss");
-    // Logger.info(Logger.tags("LN"), "Next notification set to " + sdf.format(next.getTime()));
+    
     return next.getTimeInMillis();
   }
-
-  // private boolean matchesUnit(Integer unit, Calendar current, Calendar next) {
-  //     return next.get(unit) < current.get(unit);
-  // }
 
   private Calendar buildNextTriggerTime(Date date) {
     Calendar next = buildCalendar(date);
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
@@ -108,7 +108,7 @@ public class DateMatch {
         next.set(incrementUnit, next.get(incrementUnit) + 1);
       }
     }
-    
+    next.set(Calendar.SECOND, 0);
     return next.getTimeInMillis();
   }
 
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
@@ -73,6 +73,8 @@ public class DateMatch {
   private Calendar buildCalendar(Date date) {
     Calendar cal = Calendar.getInstance();
     cal.setTime(date);
+    cal.set(Calendar.MILLISECOND, 0);
+    cal.set(Calendar.SECOND, 0);
     return cal;
   }
 
@@ -108,7 +110,6 @@ public class DateMatch {
         next.set(incrementUnit, next.get(incrementUnit) + 1);
       }
     }
-    next.set(Calendar.SECOND, 0);
     return next.getTimeInMillis();
   }
 
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
@@ -92,46 +92,52 @@ public class DateMatch {
    * Postpone trigger if first schedule matches the past
    */
   private long postponeTriggerIfNeeded(Calendar current, Calendar next) {
-    int currentYear = current.get(Calendar.YEAR);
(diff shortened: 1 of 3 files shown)


kheftel:
Ok, I tested up to monthly recurring events and everything seems to be working.

My testing repro is available [here](https://github.com/kheftel/test-cap-localnotifs). In case you want to try it out, you'd have to install `@capacitor/core` and `@capacitor/android` into it from the `core` and `android` folders in a local cloned version of my capacitor fork. I don't if it's possible to install monorepo modules from github, and my testing was not automated, and on a device, so, sorry if it's messy. I basically set up a bunch of notifications at various repeating intervals in the past and future and then played with the device clock.

Let me know if there's anything else I can do, if I should re-base on main or on v2.3.0. I have an app that's going into beta that will be using this feature branch for now, and when/if this gets merged/resolved I'll switch it back to the main repo again. Thanks!

dwieeb:
@carlpoole Care to take a look?

@kheftel Yeah, please change the base branch to `2.x`, not `main`. No need to recreate for `main`, we'll merge fixes in after each 2.x release. Thanks!

kheftel:
I borked it with an accidental force push, changing it to a draft while i fix it, sorry!

kheftel:
ok, based it on 2.x branch, testing didn't seem to break anything, should be ready to look at now.

carlpoole:
I tested the changes and this appears to be working as expected. I went ahead and removed the test log and the commented out method for minor cleanup. Thanks for contributing! 🎉 👍

dwieeb:
@jcesarmobile Here's a fix for 2.x. Good to go?

kheftel:
thanks @carlpoole!

Thanks for changing the logs to `debug`, I assume that's what the other capacitor modules/plugins do. I also didn't see any commented-out things in other parts of the code, so, that makes sense that you would do that as part of capacitor's workflow. I didn't want to remove them before reviewers had seen them tho.

android/capacitor/src/main/java/com/getcapacitor/plugin/notification/DateMatch.java
```diff
@@ -92,46 +92,47 @@ public long nextTrigger(Date date) {
...
+        incrementUnit = Calendar.DAY_OF_MONTH;
+      } else if (unit == Calendar.MINUTE) {
+        incrementUnit = Calendar.HOUR_OF_DAY;
+      }
+
+      if (incrementUnit != -1) {
+        next.set(incrementUnit, next.get(incrementUnit) + 1);
+      }
     }
+
```
jcesarmobile, line 111: I don't think it's related to your change, but I've noticed that on Android, when using on, the notification is fired at the same second it was scheduled, while on iOS it's fired on second 0.
Example, I want to fire a notification at 19:10
On iOS it fires at 19:10:00
On Android it fires at 19:10:XY, where XY can be any second based on the second it was scheduled, so can be any value from 00 to 59, there could be almost 1 minute of difference.

I think it makes more sense iOS behavior, if you set it to "minute 10" it should fire at exact 10 minute and 0 seconds, not to an unpredictable second. 

I've accomplished that by adding this line

```suggestion
    next.set(Calendar.SECOND, 0);
```

If you think it makes sense, accept the suggestion.

kheftel:
absolutely @jcesarmobile! Actually, I just pushed a new commit that zeroes out both second and millisecond, and does it in a slightly different place.

dwieeb:
Thanks, @kheftel! Great work 💪

PR #3309, opened 2020-07-21
add preserveAspectRatio to CameraOptions while resizing an image

iphilgood:
This PR is a re-opening of #3297 because of wrong base branch and a retry of #2050 and should fix ionic-team/capacitor-plugins#2 (also the discussion #1952). I've used the Android implementation of @sandstrom (shout-out) and also did an implementation for iOS.

There's now a `preserveAspectRatio` flag on CameraOptions which is set to `false` by default.

I've also fixed a bug where the app would crash while saving an image to the gallery on Android. This only happens on certain devices but I was able to reproduce it by creating 32 images until it crashed. A random UUID gets now appended to the file name to make sure no name clashes happen anymore.

@dwieeb is there anything I can do to provide some testing or how do you test this fix?

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
@@ -343,12 +343,7 @@ public class Camera extends Plugin {
     if (saveToGallery && (imageEditedFileSavePath != null || imageFileSavePath != null)) {
       try {
         String fileToSave = imageEditedFileSavePath != null ? imageEditedFileSavePath : imageFileSavePath;
-        MediaStore.Images.Media.insertImage(
-            getActivity().getContentResolver(),
-            fileToSave,
-            "IMG_" + UUID.randomUUID().toString(),
-            ""
-        );
+        MediaStore.Images.Media.insertImage(getActivity().getContentResolver(), fileToSave, "", "");
       } catch (FileNotFoundException e) {
         Logger.error(getLogTag(), IMAGE_GALLERY_SAVE_ERROR, e);
       }
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
@@ -38,6 +38,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.text.SimpleDateFormat;
 import java.util.Date;
+import java.util.UUID;
 
 /**
  * The Camera plugin makes it easy to take a photo or have the user select a photo
@@ -182,6 +183,7 @@ public class Camera extends Plugin {
     settings.setHeight(call.getInt("height", 0));
     settings.setShouldResize(settings.getWidth() > 0 || settings.getHeight() > 0);
     settings.setShouldCorrectOrientation(call.getBoolean("correctOrientation", CameraSettings.DEFAULT_CORRECT_ORIENTATION));
+    settings.setPreserveAspectRatio(call.getBoolean("preserveAspectRatio", false));
     try {
       settings.setSource(CameraSource.valueOf(call.getString("source", CameraSource.PROMPT.getSource())));
     } catch (IllegalArgumentException ex) {
@@ -410,7 +412,12 @@ public class Camera extends Plugin {
     }
 
     if (settings.isShouldResize()) {
-      final Bitmap newBitmap = ImageUtils.resize(bitmap, settings.getWidth(), settings.getHeight());
+      final Bitmap newBitmap = ImageUtils.resize(
+          bitmap,
+          settings.getWidth(),
+          settings.getHeight(),
+          settings.getPreserveAspectRatio()
+      );
       bitmap = replaceBitmap(bitmap, newBitmap);
     }
     return bitmap;
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
@@ -38,7 +38,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.UUID;
 
 /**
  * The Camera plugin makes it easy to take a photo or have the user select a photo
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
@@ -343,7 +343,12 @@ public class Camera extends Plugin {
     if (saveToGallery && (imageEditedFileSavePath != null || imageFileSavePath != null)) {
       try {
         String fileToSave = imageEditedFileSavePath != null ? imageEditedFileSavePath : imageFileSavePath;
-        MediaStore.Images.Media.insertImage(getActivity().getContentResolver(), fileToSave, "", "");
+        MediaStore.Images.Media.insertImage(
+            getActivity().getContentResolver(),
+            fileToSave,
+            "IMG_" + UUID.randomUUID().toString(),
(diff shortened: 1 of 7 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/camera/ImageUtils.java
```diff
@@ -17,6 +17,21 @@
...
 
+  /**
+   * Resize an image to the given width and height considering the preserveAspectRatio flag.
+   * @param bitmap
+   * @param width
+   * @param height
+   * @param preserveAspectRatio
+   * @return a new, scaled Bitmap
+   */
+  public static Bitmap resize(Bitmap bitmap, final int width, final int height, final boolean preserveAspectRatio) {
```
dwieeb, line 28: Will this be a breaking change? You've added a new parameter to this public method in the Android platform. Will calls such as `ImageUtils.resize(bitmap, 100, 100)` continue to work?

You can mitigate this by adding the following method:

```java
  public static Bitmap resize(Bitmap bitmap, final int width, final int height) {
    resize(bitmap, width, height, false);
  }
```

core/src/core-plugin-definitions.ts
```diff
@@ -321,6 +321,13 @@ export interface CameraOptions {
    * The height of the saved image
    */
   height?: number;
+  /**
+   * Wheter to preserve the aspect ratio of the image.
+   * If the this flag is true the width and height will be used as max values
+   * and the aspect ratio will be preserved
+   * Default: false
```
dwieeb, line 328:
```suggestion
   * Whether to preserve the aspect ratio of the image.
   * If this flag is true, the width and height will be used as max values
   * and the aspect ratio will be preserved.
   * Default: false
```

dwieeb:
@iphilgood To test we usually just add a test button in the example app in the 2.x branch: https://github.com/ionic-team/capacitor/tree/2.x/example

With Capacitor 3 we'll be doing more unit testing. 😁

iphilgood:
@dwieeb I'm gonna fix your reviews tomorrow 👍🏽 Thanks for having a look at it. Are you gonna add the test button or is it me? And do you know why my build failed? I ran the command on my local machine and except for some warnings I didn't see an error 🤔

Hahaha alright 😄 Some tests are always a good idea 😉

dwieeb:
@iphilgood If you could add the button, that would be great! Just making sure there's a new button for using `preserveAspectRatio: true` and that the other buttons work as expected. 

Sometimes the `pod lib lint` can hang. I just restarted the jobs so we'll see if it passes.

android/capacitor/src/main/java/com/getcapacitor/plugin/camera/ImageUtils.java
iphilgood: Actually I think it's not. There's only one call of this static method and it's [here](https://github.com/ionic-team/capacitor/pull/3309/files/79863659188fcd59bd219a004f37e2226abf0bcb#diff-436e54d5c31c5e7863a079db6374c887R420-R425). But I've added an overload anyway. 👍🏽

iphilgood:
@dwieeb I've resolved your comments and added a new button to test my code (somehow the menu is not working in the example app. I had to set the RootPage accordingly) 😄 

I think the other buttons still work as expected but it would be nice if someone else could have a look at it.

dwieeb:
@iphilgood There isn't a button for the menu but you can swipe from the left of screen because it's just a sidemenu.

This works great on Android and iOS! Nice work!

iphilgood:
@dwieeb Hahahaha I'm sorry for that 😅 Now everything makes sense 😂 Thank you!

sandstrom:
Awesome work @iphilgood 🎉

core/src/core-plugin-definitions.ts
```diff
@@ -321,6 +321,13 @@ export interface CameraOptions {
    * The height of the saved image
    */
   height?: number;
+  /**
+   * Whether to preserve the aspect ratio of the image.
+   * If this flag is true, the width and height will be used as max values
+   * and the aspect ratio will be preserved.
+   * Default: false
+   */
+  preserveAspectRatio?: boolean;
```
jcesarmobile, line 330: Aspect ratio is already respected if only height or width is passed, so this should make more clear that is for when both height and width are passed

android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
```diff
@@ -341,7 +343,12 @@ private void returnResult(PluginCall call, Bitmap bitmap, Uri u) {
     if (saveToGallery && (imageEditedFileSavePath != null || imageFileSavePath != null)) {
       try {
         String fileToSave = imageEditedFileSavePath != null ? imageEditedFileSavePath : imageFileSavePath;
-        MediaStore.Images.Media.insertImage(getActivity().getContentResolver(), fileToSave, "", "");
+        MediaStore.Images.Media.insertImage(
+            getActivity().getContentResolver(),
+            fileToSave,
+            "IMG_" + UUID.randomUUID().toString(),
+            ""
+        );
```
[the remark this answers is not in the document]
dwieeb: I think this prompted for a permission on iOS, so let's be careful here.

core/src/core-plugin-definitions.ts
sandstrom: I'd suggest a name like `legacyPreserveAspectRatio` since this is more of a bug then a feature. That will make it more obvious that this will be removed in 3.0.

android/capacitor/src/main/java/com/getcapacitor/plugin/Camera.java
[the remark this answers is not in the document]
iphilgood: @jcesarmobile you're right. I'm going to create a separate PR for this fix.

core/src/core-plugin-definitions.ts
iphilgood: So what I'm supposed to do now? 😅 Leave it like that? Update the comment? Rename the flag?

sandstrom: The name itself probably isn't that important. Could just leave it as-is and clarify in the comment (that's probably what jcesarmobile ment).

Perhaps something like this?

```js
  /**
   * Whether to preserve the aspect ratio of the image.
   * If this flag is true, the width and height will be used as max values
   * and the aspect ratio will be preserved. This is only relevant when
   * both a width and height is passed, when only one of them are provided
   * the aspect ratio is always preserved.
   * 
   * A future major version will change this behavior to be default,
   * and may also remove the flag altogether.
   * Default: false
   */
```

iphilgood: Thank you @sandstrom. I think your comment describes it very well. What do you think @dwieeb?

dwieeb: Agreed, that's perfect.

iphilgood:
Okay I've resolved all the comments. Thanks everybody 🤗

core/src/core-plugin-definitions.ts
```diff
@@ -321,6 +321,18 @@ export interface CameraOptions {
    * The height of the saved image
    */
   height?: number;
+  /**
+   * Whether to preserve the aspect ratio of the image.
+   * If this flag is true, the width and height will be used as max values
+   * and the aspect ratio will be preserved. This is only relevant when
+   * both a width and height is passed. When only width or height is provided
```
jcesarmobile, line 328:
```suggestion
   * both a width and height are passed. When only width or height is provided
```

dwieeb:
2.4.0 was just released with this!

sandstrom:
Awesome, thanks @iphilgood and @dwieeb 🎉

sandstrom:
@dwieeb Will there be an update to CHANGELOG.md to include the 2.4.0 changes?

PR #3310, opened 2020-07-21
chore: Update gradle tooling on plugin template

jcesarmobile:
gradle plugin from 3.6.1 to 4.0.1
gradle wrapper from 5.6.4 to 6.1.1

the jar and gradlew files are automatically updated by `./gradlew wrapper --distribution-type all` command


dwieeb:
Did we want to update gradle in the android-template and `@capacitor/android` as well? Or not yet?

PR #3343, opened 2020-07-27
Add Nullable annotations to JSObject

cwoolum:
This adds a Nullable Annotation for the getString method. When using Kotlin, the existing functionality can lead to some confusion because the method actually does return nulls but Android Studio will try to optimize out the null checks.

image

After adding `@Nullable`, the incorrect warnings go away.


dwieeb:
Hi @cwoolum! Thanks for the PR. Care to run `npm run fmt` and commit the changes? I'm assuming it's the import order the linter is complaining about.

cwoolum:
Aaah, got it! I missed the fmt command and prettier wasn't being helpful with what was actually wrong.

carlpoole:
Is this the same for `getInteger` and `getBool` as well? Just thinking if we should add it to all 3.

cwoolum:
That's a great point! I added the `@Nullable` annotations for all calls related to fetching configuration. 

In my own plugin project, it helped uncover some bugs where the compiler assumed that a value could never be null

android/capacitor/src/main/java/com/getcapacitor/JSObject.java
```diff
@@ -40,32 +41,37 @@ public static JSObject fromJSONObject(JSONObject obj) throws JSONException {
...
         return getString(key, null);
     }
 
-    public String getString(String key, String defaultValue) {
+    @Nullable
+    public String getString(String key, @Nullable String defaultValue) {
         try {
             String value = super.getString(key);
-            if (!super.isNull(key) && value != null) {
+            if (!super.isNull(key)) {
```
cwoolum, line 53: I remove the null check here because the `super` call actually throws an exception if the value is null so it was a redundant call. Android studio was showing a warning for it.

dwieeb:
Thanks @cwoolum! Great work. 🎉

PR #3348, opened 2020-07-28
fix(core): use own type for backButton event result

### LINKED ISSUE #3346: bug: Minor typing error
## Bug Report
[The following typing (data: AppUrlOpen) seems to be wrong](https://github.com/ionic-team/capacitor/blame/25313a6087da5eba67d28de2619cf432d4172859/core/src/core-plugin-definitions.ts#L143)

```addListener(eventName: 'backButton', listenerFunc: (data: AppUrlOpen) => void): PluginListenerHandle;```

the data is very likely something else but **AppUrlOpen**, probably a copy paste error of the line `129`

```addListener(eventName: 'appUrlOpen', listenerFunc: (data: AppUrlOpen) => void): PluginListenerHandle;```

### Capacitor Version
From 1.0.0-alpha.11 still there on capacitor 3.0.0-alpha.0

### Platform(s)
N/A

### Current Behavior
N/A

### Expected Behavior
N/A

I've no idea if the back button really emits a special type.

### Code Reproduction
N/A

jcesarmobile:
sending to main since the type change is probably a breaking change, apps using the `AppUrlOpen` import would need to change to `BackButtonResult`

closes #3346

There are a few listeners that just return `{}`, should we just return nothing for such cases?


dwieeb:
Yeah, it wouldn't break the app's compilation to add parameters later, so I'm not sure why we do that. We can merge this though because I imagine we'll be cleaning up the types as we move these definitions into their respective plugin packages.

PR #3375, opened 2020-08-03
Add 2.4.0 to changelog

sandstrom:


dwieeb:
@sandstrom We are going to merge all changes from `2.x` into `main` each time there's a 2.x release, so this isn't necessary. I just forgot to do it. Thanks, though!

PR #3381, opened 2020-08-03
Merge 2.x for 2.4.0 release

dwieeb:


dwieeb:
@sandstrom Here's the full PR.

dwieeb:
@iphilgood @kheftel Would you mind glancing over this quick for your respective changes? We're pulling them into the main branch with this PR.

ios/Capacitor/Capacitor/Plugins/Camera.swift
```diff
@@ -232,7 +234,7 @@ public class CAPCameraPlugin: CAPPlugin, UIImagePickerControllerDelegate, UINavi
         }
 
         if settings.shouldResize {
-            guard let convertedImage = resizeImage(image!) else {
+            guard let convertedImage = resizeImage(image!, false) else {
```
iphilgood, line 237: @dwieeb this should consider the flag from the setting (`settings.preserveAspectRatio`), right?

jcesarmobile: yeah, good catch, should be
```suggestion
            guard let convertedImage = resizeImage(image!, settings.preserveAspectRatio) else {
```

dwieeb: Thanks, these were not easy conflicts to resolve. 😂

PR #3384, opened 2020-08-04
chore(iOS): Cleaning up types for JavaScript <-> Native bridging

ikeith:
This branch updates the iOS code's type system to address several issues:

- The JSON serialization is not type-safe because the values are defined as `Any`, which includes Swift value-types such as structs and enums that cannot be bridged to the JS environment.

- The convenience accessors on `CAPPluginCall` support default values but always return an optional. That leads to many force unwraps and noisy complaints from the linter.

- The generic `get` convenience accessor cannot justify its existence because it offers no benefit over raw access to the underlying dictionary and generates confusion by defining an additional way to retrieve values. (and only really exists as a historical artifact according to @mlynch)

- `JSArray` is defined as an array of dictionaries, which is insufficient since it cannot include numbers, strings, etc that make up valid JSON.

To that end, this branch does 4 things:

1) Defines a `JSValue` "type" (enforced via an empty, specializing protocol) to ensure that only expected types are used for bridging communication. This moves potential runtime errors to compile time errors instead.

2) The convenience `get` accessors are now split into two separate methods for each type: one to return an optional and one that supports a default value but returns a non-optional. (the simple `get` accessor has been removed) This should improve clarity at the point of use and reduce confusion over the use of optionals.

3) Protocols! The new accessors are built using composable protocols so they can be inherited by other objects by conforming to the `JSValueContainer` protocol. (this would allow them to be leveraged in the future for things like configuration... 🤔)

4) Rather than mixing Obj-C and Swift support such that it is hard to tell what is expected to work where, the new protocols are purely swift (as is required by the underlying types), and the Obj-C convenience accessors are split out into a separate file so only Obj-C code that needs it can import it.


dwieeb (approved):

ikeith:
This PR has been updated with unit tests to verify the expected behavior around converting to the new protocol-supported JS types.

ios/Capacitor/Capacitor/Plugins/LocalNotifications.swift
```diff
@@ -338,6 +338,9 @@ public class CAPLocalNotificationsPlugin: CAPPlugin {
 
         createdCategories.append(generalCategory)
         for type in actionTypes {
+            guard let type = type as? JSObject else {
+              continue
+            }
```
jcesarmobile, line 343: Can something like this be used in this 3 fors instead of the guards?

```suggestion
        for case let type as JSObject in actionTypes
```

jcesarmobile (changes requested):
Added a minor comment.

Also on line [144](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/Plugins/LocalNotifications.swift#L144) of the same file, I think this `guard let types = call.getArray("types", Any.self) as? JSArray else {` can be changed to this `guard let types = call.getArray("types", JSObject.self) else {`

ios/Capacitor/Capacitor/Plugins/LocalNotifications.swift
ikeith: That would work. I am little concerned, though, that that kind of syntax might be too esoteric just to save a couple of lines. It could be hard for someone less familiar with Swift to understand if they looked at the core plugins for examples of how to write a plugin.

ikeith:
> Also on line [144]

Good catch. That's inconsistent with the other code.

ios/Capacitor/Capacitor/Plugins/LocalNotifications.swift
jcesarmobile: I didn't propose it for saving lines, but because I find the guard with continue confusing.

Could be possible to indicate in some way that the whole JSArray contain elements of the same type so don't need to cast them inside the for?

In the change I proposed (`guard let types = call.getArray("types", JSObject.self)`), we are saying the elements are JSObject, but we still need to cast them when we go over the Array.

ikeith: Oh, I see. Your suggested change is functionally identical so I assumed it was just to be more concise. (I think a for loop with an inner guard is easier to understand but obviously that's subjective) I agree that bubbling the more exact types through the code is better but I was trying to avoid too many changes, especially since I knew the plugins were being split out and I didn't expect this change to get merged quickly. I will update the code to avoid casting types up to a more generic one.

jcesarmobile (approved):

PR #3389, opened 2020-08-04
Unsupported Browser Exception proposal

dwieeb:
For the web platform, many APIs may not work, especially for IE11, for example. We need a consistent way to report that a feature is unsupported. This will be used by official plugins, but all Capacitor plugins are welcome to use it. This is just a proposal to spark discussion.

Stems from a discussion in the Haptics PR: https://github.com/ionic-team/capacitor-plugins/pull/5#issuecomment-667739490

- Usage in plugins: https://github.com/ionic-team/capacitor-plugins/blob/4862a1997c0d1e6e65f4a16a7de976083acbf4d5/screen-reader/src/web.ts#L16-L20
- Usage in apps: https://github.com/ionic-team/capacitor-testapp/blob/84aaab694f05dadebb1666ff130f33130a279ec1/src/pages/ScreenReader.tsx#L38-L49

### DIFF
--- a/core/src/plugins.ts
+++ b/core/src/plugins.ts
@@ -1,6 +1,28 @@
 import { Capacitor, Plugins } from './global';
 import { WebPlugin } from './web';
 
+export enum ExceptionCodes {
+  UNSUPPORTED_BROWSER = 'UNSUPPORTED_BROWSER',
+}
+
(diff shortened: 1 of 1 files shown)


dwieeb:
@cameronrr Thoughts on this?

PR #3391, opened 2020-08-05
chore: Make native resolve return undefined

jcesarmobile:
When calling `call.resolve()` it will return `undefined` instead of `{}`
`call.success()` will keep returning `{}`


dwieeb:
We also wanted to deprecate `success()`, so I added annotations for that.

jcesarmobile:
I was planning doing that in a separate PR

dwieeb:
Okay I made https://github.com/ionic-team/capacitor/pull/3392 instead.

PR #3392, opened 2020-08-05
Deprecate redundant PluginCall methods

dwieeb:
- `success()` is now deprecated (use `resolve()`)
- `error()` is now deprecated (use `reject()`)

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/PluginCall.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/PluginCall.java
@@ -66,7 +66,7 @@ public class PluginCall {
      */
     @Deprecated
     public void success() {
-        this.success(new JSObject());
+        this.resolve(new JSObject());
     }
 
     public void resolve(JSObject data) {
--- a/android/capacitor/src/main/java/com/getcapacitor/PluginCall.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/PluginCall.java
@@ -51,16 +51,20 @@ public class PluginCall {
     }
 
     /**
-     * @deprecated Use `resolve()`.
+     * @deprecated
+     * Use {@link #resolve(JSObject data)}
      */
+    @Deprecated
     public void success(JSObject data) {
         PluginResult result = new PluginResult(data);
         this.msgHandler.sendResponseMessage(this, result, null);
     }
 
     /**
-     * @deprecated Use `resolve()`.
+     * @deprecated
+     * Use {@link #resolve()}
      */
+    @Deprecated
     public void success() {
         this.success(new JSObject());
     }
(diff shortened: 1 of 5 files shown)


jcesarmobile:
Cancel should also be deprecated, keep only reject and resolve

dwieeb:
Do you mean `error()`? Yes, we should prefer `reject()` over `error()`. I can't find `cancel()`.

jcesarmobile:
Yeah, sorry, I meant error

dwieeb:
Current method signatures for `error` and `reject` on Android and iOS:
```java
public void error(String msg, Exception ex)
public void error(String msg, String code, Exception ex)
public void error(String msg)
public void reject(String msg, Exception ex)
public void reject(String msg, String code)
public void reject(String msg)
```

```swift
func error(_ message: String, _ error: Error? = nil, _ data: PluginCallErrorData = [:])
func reject(_ message: String, _ code: String? = nil, _ error: Error? = nil, _ data: PluginCallErrorData = [:])
```

I don't know why they're inconsistent in Android (`reject` doesn't allow for both code and exception, but `error` does) or why iOS and Android are inconsistent with each other (`error` allows for both code and exception on Android, but not iOS). 

I think it's fixable, though. Let me try an implementation.

ios/Capacitor/Capacitor/CAPPluginCall.swift
```diff
@@ -62,10 +62,12 @@ public typealias PluginEventListener = CAPPluginCall
         return self.options.index(forKey: key) != nil
     }
 
+    @available(*, deprecated, message: "Use `resolve()`")
```
jcesarmobile, line 65: on Xcode the message appears as 
> 
```
'success' is deprecated: Use `resolve()`
```
So I think it would be better to use it like this 

```suggestion
    @available(*, deprecated, message: "Use 'resolve'")
```

so it appears as `'success' is deprecated: Use 'resolve'`

android/capacitor/src/main/java/com/getcapacitor/PluginCall.java
```diff
@@ -80,11 +86,28 @@ public void errorCallback(String msg) {
...
+     * @deprecated Use `reject()`.
+     */
     public void error(String msg, String code, Exception ex) {
+        reject(msg, code, ex);
+    }
+
+    /**
+     * @deprecated Use `reject()`.
+     */
+    public void error(String msg) {
```
jcesarmobile, line 106: Java allows to link the new method
Also the method should have the `@Deprecated` annotation, not just on the comment
```suggestion
    /**
     * @deprecated
     * Use {@link PluginCall#reject(String msg)} instead
     */
     @Deprecated
    public void error(String msg) {
```

ios/Capacitor/Capacitor/CAPPluginCall.swift
dwieeb: If I use `renamed: "resolve"` then this is the message that appears and there's an option to fix automatically:

image

android/capacitor/src/main/java/com/getcapacitor/PluginCall.java
```diff
@@ -50,11 +50,21 @@ public void successCallback(PluginResult successResult) {
...
         this.msgHandler.sendResponseMessage(this, result, null);
     }
 
+    /**
+     * @deprecated
+     * Use {@link #resolve()}
+     */
+    @Deprecated
     public void success() {
         this.success(new JSObject());
```
jcesarmobile, line 69:
```suggestion
        this. resolve(new JSObject());
```

PR #3397, opened 2020-08-07
fix(android): plugin template missed androidx appcompat dependency causing Android Studio issue

carlpoole:
Added the androidx appcompat dependency to the plugin template. When this isn't present, Android Studio does not identify that AppComptActivity has access to methods on the Activity class for some reason. We found this because `runOnUiThread` was not resolving as a valid method during plugin development.


plugin-template/android/build.gradle
```diff
@@ -1,5 +1,6 @@
 ext {
     junitVersion =  project.hasProperty('junitVersion') ? rootProject.ext.junitVersion : '4.12'
+    androidxAppCompatVersion = project.hasProperty('androidxAppCompatVersion') ? rootProject.ext.androidxAppCompatVersion : '1.1.0'
```
jcesarmobile, line 3:
```suggestion
    androidxAppCompatVersion = project.hasProperty('androidxAppCompatVersion') ? rootProject.ext.androidxAppCompatVersion : '1.2.0'
```

carlpoole: I think so. Without this dependency, Android Studio doesn't resolve basic `Activity.java` methods that can be overridden like `runOnUiThread` and it isn't clear how to fix that. Any plugin that tries to run code that impacts the UI will need to use this method. Dan came across this when working on a plugin being broken out of core capacitor.

android/capacitor/build.gradle
```diff
@@ -1,5 +1,5 @@
 ext {
-    androidxAppCompatVersion = project.hasProperty('androidxAppCompatVersion') ? rootProject.ext.androidxAppCompatVersion : '1.1.0'
+    androidxAppCompatVersion = project.hasProperty('androidxAppCompatVersion') ? rootProject.ext.androidxAppCompatVersion : '1.2.0'
```
jcesarmobile, line 2: this is a plugin fix, should only contain changes related to the plugin

We also need to update the default version here, but should be a separate PR that updates the version here, and in variables.gradle file, and should be documented that users should update their own variables.gradle file as it's managed by them.
And probably check all other dependencies and update all of them at once, but checking that nothing breaks.

carlpoole: So we want a PR that only updates the version in the plugin template here, and then a separate PR that only updates the version in capacitor android and variables.gradle?

jcesarmobile:
the default is just a default, if we don't update it on the template now we will have to update all plugins later, so do what you think is best

dwieeb:
@carlpoole You can reopen this PR here: https://github.com/ionic-team/create-capacitor-plugin

Issue #3133, opened 2020-06-22
Capacitor 3.0


imhoffd:
Welcome to the Capacitor 3 planning issue! This issue will be periodically updated up until the launch date (TBD). We've locked this issue, but we highly encourage you to [open a discussion](https://github.com/ionic-team/capacitor/discussions) if you have questions or suggestions.

> 🚨 This issue is meant to organize the scope and work for Capacitor 3 and is not intended to be used as a roadmap or guarantee of features.

---

The Capacitor project will focus on a few key areas of improvement for 3.0:
- More community involvement. 💖 
    - We opened [Discussions](https://github.com/ionic-team/capacitor/discussions) to provide a central place for community discussions and support.
    - We will be using public issues, milestones, and the new [Capacitor project board](https://github.com/orgs/ionic-team/projects/7) to manage tasks.
    - We created the [Capacitor Community org](https://github.com/capacitor-community/) to house high-quality Capacitor plugins and platforms and the [Proposals repo](https://github.com/capacitor-community/proposals) to gauge interest and track ideas.
- Re-invest in the idea that Capacitor can turn any web app into a native app.
    - We are re-evaluating the core plugins to provide the best experience for the most people. 
    - Refactors are underway that improve the performance and stability of the Capacitor core.
- Developer productivity and experience.
- Foster a stable, high-quality Public API for apps and plugins.

## High-Level Changes

- **Electron as a community-maintained platform.** We estimate over 95% of Capacitor devs are targeting iOS and Android, but not Electron. Given this and our limited resources, we decided it makes the most sense to focus on the iOS and Android platforms and leave the Electron platform in the very capable hands of the community. Capacitor 3 will proudly offer [the Electron Platform](https://github.com/capacitor-community/electron) as the first ever community maintained platform for Capacitor.
- **Rethinking Plugins.** Core plugins will be pulled out of Capacitor into their own packages (now "official plugins"). See #3227 for details.
- **Programmatic run command.** A new `npx cap run` command should be added that provides a CLI workflow for deploying debug builds to devices. This means IDE usage will no longer be required during day-to-day development. Please note: this is _not_ intended for release builds (https://github.com/ionic-team/capacitor/issues/324). See [this tracking issue](https://github.com/ionic-team/capacitor/issues/3142) and the corresponding PR: https://github.com/ionic-team/capacitor/pull/3599
[...]

PR #3405, opened 2020-08-10
Embedded Capacitor: Pass urlPath to Bridge

fkirc:
For Embedded Capacitor, it is often necessary to have multiple Capacitor fragments with different routes.
For example, I have two embedded Capacitor fragments like this:
`http://localhost/feature_1`
`http://localhost/feature_2`

This PR passes the `/feature_1` and `/feature_2` to the target bridges.
It is somewhat similar to the following iOS-issue: https://github.com/ionic-team/capacitor/issues/3106

My intended usage of this PR is as follows (although not the only possible usage):

```Kotlin
import android.os.Bundle
import com.getcapacitor.BridgeFragment

class MyFeatureBridgeFragment : BridgeFragment() {
    override fun onCreate(savedInstanceState: Bundle?) {
        super.onCreate(savedInstanceState)
        super.setUrlPath("/feature_1")
    }
}
```

## Why the URL path?

One could argue whether it makes sense to pass only URL paths instead of complete URLs.
However, I still want the host to be independent of the path in order to configure the host for "live reload debugging":
As described in https://github.com/ionic-team/capacitor/discussions/1478, live reload debugging leads to dangerous mis-configurations, but this is a story for another PR.

## Limitations and Further Work

Passing only the URL path might not be enough for all use cases of Embedded Capacitor.
For example, one might have different hosts instead of different URL paths.
Although this practice might be forbidden by app store guidelines, it might still be a valid use case:
`https://remote_host_1.com`
`https://remote_host_2.com`

More generally, I propose a system where the entire Capacitor-configuration can be dynamically modified/overriden by the native code.
In turn, this may or may not lead to the elimination of the native `capacitor.config.json`-files.
Nevertheless, I still submit this PR because I cannot estimate how long it would take to have a new configuration system.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -135,7 +135,8 @@ public class Bridge {
         CordovaInterfaceImpl cordovaInterface,
         PluginManager pluginManager,
         CordovaPreferences preferences,
-        JSONObject config
+        JSONObject config,
+        String urlPath
     ) {
         this.context = context;
         this.webView = webView;
@@ -163,10 +164,10 @@ public class Bridge {
         // Register our core plugins
         this.registerAllPlugins();
 
-        this.loadWebView();
+        this.loadWebView(urlPath);
     }
 
-    private void loadWebView() {
+    private void loadWebView(String urlPath) {
         appUrlConfig = this.config.getString("server.url");
         String[] appAllowNavigationConfig = this.config.getArray("server.allowNavigation");
 
@@ -199,6 +200,9 @@ public class Bridge {
             if (!scheme.equals(Bridge.CAPACITOR_HTTP_SCHEME) && !scheme.equals(CAPACITOR_HTTPS_SCHEME)) {
                 appUrl += "/";
             }
+        }
+        if (urlPath != null) {
+            appUrl += urlPath;
         }
 
         final boolean html5mode = this.config.getBoolean("server.html5mode", true);
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
@@ -76,7 +76,7 @@ public class BridgeActivity extends AppCompatActivity {
 
         pluginManager = mockWebView.getPluginManager();
         cordovaInterface.onCordovaInit(pluginManager);
-        bridge = new Bridge(this, webView, initialPlugins, cordovaInterface, pluginManager, preferences, this.config);
+        bridge = new Bridge(this, webView, initialPlugins, cordovaInterface, pluginManager, preferences, this.config, null);
 
         Splash.showOnLaunch(this, bridge.getConfig());
 
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
@@ -45,8 +45,7 @@ public class BridgeFragment extends Fragment {
     private PluginManager pluginManager;
     private CordovaPreferences preferences;
     private MockCordovaWebViewImpl mockWebView;
-    private int activityDepth = 0;
-    private String bridgeStartDir;
+    private String urlPath;
 
     private String lastActivityPlugin;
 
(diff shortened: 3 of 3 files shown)


carlpoole:
This is currently possible by using the following constructor for a new bridge fragment:

`public static BridgeFragment newInstance(String startDir)`

where `startDir` is the directory containing the app web assets. I think in the example given this would be `feature_1`

Although this does not suit the use case where you use the BridgeFragment by overriding and inflating via XML, as an example. The way this is currently written would allow you to call `newInstance` with a startDir and specify a path in `onCreate` and the Bridge would first load one then the other. It may be cleaner to streamline this.

fkirc:
I actually tried to modify `startDir`, but this is not exactly the thing that I want.
My goal is to only change the route for each fragment, while still keeping the same startDir and the same `index.html` for all fragments.

Imagine a typical React-app with react-router: I only want to build a single React-app with multiple routes, instead of building multiple React-apps.

fkirc:
One more note about the `newInstance`-method:

I would be fine to submit a solution that supports both XML-inflation and explicit fragment instantiation. If we agree on an API, then I would be willing to extend this PR.

carlpoole:
Ah I see, so it sounds like there are a couple of use cases:

- Web view in fragment 1 points to one app directory, and web view in fragment 2 points to a different app directory
- Web view in fragment 1 points to `/feature_1` route in a web app directory, and fragment 2 points to `/feature_2` route in the same app directory.

Am I understanding that correctly? If so then two different settings do make sense. One for the app directory and one optional for a route in that web app (default `/`)? 

Yeah, this reminded me of the plugin PR also :)

fkirc:
Yes this is correct, those are two independent functionalities.
 `startDir` is well-suited if you have two completely different web-apps, whereas the `urlPath` provides different entry-points into the same web-app.

fkirc:
Any updates on this PR?
Do you have a concrete solution to pass a `urlPath` at launch time?
In the worst case, I might have to write my own `BridgeFragment` and my own `Bridge.java`, or perhaps hack something together with reflection.

carlpoole:
Hi Felix, sorry for delay in response. We think this is definitely a good addition to embedded Capacitor and know there is opportunity to improve the embedded use case. While trying to think of a way to incorporate this now in context of upcoming potentially breaking changes and cleanup, we felt it would be best to wait and include changes to address this use case as part of the overall upcoming improvements for embedded Capacitor in v3 for both mobile platforms.

Thanks for bringing attention to this and helping improve Capacitor.

fkirc:
I have to admit that I am a little bit frustrated about the current merge-situation.
I neither know when Capacitor-V3 will be ready; nor do I know whether Capacitor-V3 will support this use case out-of-the-box.

One more sidenote:
I spotted a bug in `BridgeFragment` that causes some config in `capacitor.config.json` to be silently ignored.
However, it does not seem helpful to file another pull request in the current merge-situation.

dwieeb:
@fkirc It's never easy in transitions between major versions like this. Luckily, we are happily accepting bug fixes to the `2.x` branch. A new 2.4.1 release went out today, actually. We'd be happy to look at the bug fix for BridgeFragment for 2.x. We'd prefer if new features were added to Capacitor 3, though.

For what it's worth, we don't plan on increasing the scope of Capacitor 3 beyond what [this issue](https://github.com/ionic-team/capacitor/issues/3133) outlines, and we're making good progress.

fkirc:
Perhaps I am mistaken, but I thought that Capacitor 3.X is developed on the `main`-branch.
Therefore, isn't this also a PR for Capacitor 3.X?

dwieeb:
Yeah, that's correct. I believe @carlpoole is saying that there's a larger suite of changes being planned for the Embedded Capacitor use case. Although it may be possible today in Capacitor 2.x, it isn't a focus or a documented feature. It will be for Capacitor 3.x, though.

carlpoole:
The proposed addition in this PR is great and we are going to include a way to specify the path in 3.x. In its current form the changes will conflict with plans we are exploring in refining the configuration and Embedded developer experience. This development is expected to begin in the coming month and we would rather approach this as one larger improvement to both native platforms than merge a change for Android now that will change again soon. Please do continue to submit bug reports/fixes and suggest features as we are always interested in improving Capacitor. The help is greatly appreciated.

fkirc:
I agree that an Android-only change is not a viable long-term plan.
Therefore, I would recommend to merge https://github.com/ionic-team/capacitor/pull/1972 in order to have the exact same functionality for iOS. If #1972 is merged, then I could submit a small writeup on how to implement this use case for iOS.

PR #3420, opened 2020-08-13
chore: remove Storage plugin

dwieeb:


naimajro:
Hi, is Storage API in Capacitor still available, or it's officially removed? If yes, how can i differently use local storage?

jcesarmobile:
In Capacitor 3 all plugins are being removed from the core, but they will be available as independent packages `@capacitor/storage` https://github.com/ionic-team/capacitor-plugins/tree/main/storage

PR #3429, opened 2020-08-18
Added restartApp method

trancee:
Added the `restartApp` method that is similar to `exitApp` but instead it starts another instance and exits the old one.


jcesarmobile:
thanks for the PR, but we have removed App plugin from capacitor and moved it to https://github.com/ionic-team/capacitor-plugins
You can send the PR there if you want

PR #2882, opened 2020-05-08
fix(ios): Support CORS when using Live Reload

diachedelic:
`_capacitor_file_` HTTP requests fail CORS checks when server.url is set. This PR implements CORS support on iOS. (Sidenote: does not require a fix for Android as `_capacitor_file_` requests are sent to the same origin of the page).

# How to reproduce
```json
// capacitor.config.json
{
  "server": { "url": "http://192.168.15.26:6789" }
  ...
}
```

```javascript
// my-app.js
const url = Capacitor.convertFileSrc(uri)
await fetch(url) // throws
```

```
[Error] Origin http://192.168.15.26:6789 is not allowed by Access-Control-Allow-Origin.
[Error] Failed to load resource: Origin http://192.168.15.26:6789 is not allowed by Access-Control-Allow-Origin. (1588911030406.txt, line 0)
[Error] Fetch API cannot load capacitor://localhost/_capacitor_file_/Users/me/Library/Developer/CoreSimulator/Devices/E60456F7-630E-4574-B7BD-A486C85B06C3/data/Containers/Data/Application/FFE5A297-770F-418D-9331-5833B1CCA94C/Documents/1588911030406.txt due to access control checks.
[Error] TypeError: Origin http://192.168.15.26:6789 is not allowed by Access-Control-Allow-Origin.
  capacitorConsole (user-script:2:78)
  (anonymous function) (bundle.js:4179)
  promiseReactionJob
```

PR #3433, opened 2020-08-19
Support CORS when using Live Reload

diachedelic:
A repeat of https://github.com/ionic-team/capacitor/pull/2882, with some improvements.

Even if this will be superseded by Capacitor v3, I would still like to see this merged for v2. I am using `patch-package` for now :(

### DIFF
--- a/ios/Capacitor/Capacitor/CAPAssetHandler.swift
+++ b/ios/Capacitor/Capacitor/CAPAssetHandler.swift
@@ -47,14 +47,10 @@ class CAPAssetHandler: NSObject, WKURLSchemeHandler {
 
           headers["Content-Type"] = mimeType
           headers["Cache-Control"] = "no-cache"
+          headers["Content-Length"] = String(expectedContentLength)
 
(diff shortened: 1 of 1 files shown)


riderx:
@diachedelic are you still running cap 2?

diachedelic:
@riderx Yes, why do you ask?

riderx:
Because it's 2 years old maybe it was outdated, and not necessary :)

diachedelic:
I think this fix is still relevant to Capacitor 3. The relevant file was renamed to [WebViewAssetHandler.swift](https://github.com/ionic-team/capacitor/blob/b893a57aaaf3a16e13db9c33037a12f1a5ac92e0/ios/Capacitor/Capacitor/WebViewAssetHandler.swift), but was not substantially changed. This patch still applies.

riderx:
@diachedelic i understand, but there are no really point to stay in capacitor 2.
The gap is not huge, to migrate, and the benefice are huge.

PR #3436, opened 2020-08-19
chore(cli): remove cordova-plugin-googlemaps from skip list

jcesarmobile:
The plugin is being reworked to be compatible with Capacitor and people can't test it unless we remove it from the skip list.

Related: https://github.com/ionic-team/capacitor/discussions/3431


panteleevnikita:
This plugin wasn't reworked yet as in November 2021. (See here: https://github.com/mapsplugin/cordova-plugin-googlemaps/issues/2781 )

And when trying to use another maps plugin to supply this problem there are compatibility errors. :disappointed:
Could these change be reverted by any chance, @jcesarmobile ?

PR #3440, opened 2020-08-19
chore(cli): remove init command

dwieeb:
The steps in the `init` command should just be documented for explicitness and to avoid unnecessary maintenance (added https://github.com/ionic-team/capacitor-site/issues/105 as a reminder).

TODO:
- [x] remove shared utilities of init and create once https://github.com/ionic-team/capacitor/pull/3438 is merged


jcesarmobile:
What’s the reasoning for removing the init command? And how are users supposed to init the project then? The linked issue about documenting it only says it needs to be documented, but no mention about how it’s supposed to work now.
Also, without init, the config.xml to capacitor.config.json cordova preference migration is lost.

dwieeb:
There's a couple reasons to remove it:
- Capacitor CLI should be local only.
    - We could change documentation to install it locally first (`npm install -D @capacitor/cli && npx cap init`), however, one of the reasons to make it local only is to set clear boundaries on the responsibilities of the Capacitor CLI. Without `create`, `init`, and `plugin:generate`, the Capacitor CLI's responsibilities become very clear: it's an essential part of the tooling pipeline for your app.
- Prefer training & education over magic.
    - In my opinion, commands like this are a disservice to the user. If a user runs this command, they...
        - ...don't know the command just did or what changes were just made to their app. This is at best confusing and at worst scary and/or frustrating.
        - ...aren't educated about `capacitor.config.json`. It doesn't even mention it. Instead, it would be more educative to guide them through editing the file. This would teach users what each value is for and then they would know what to do in the future. For example, magically moving Cordova config into `capacitor.config.json` may seem helpful, but when the user needs to change a value (or configure Cordova newly installed plugins in the future), they may not know where to do it. If they were guided through the migration process, they would remember.
        - ...may be disillusioned by the "Your Capacitor project is ready to go!" message. The command gives the illusion that it successfully integrated Capacitor with your project no matter what, but even when run in a new CRA app, `capacitor.config.json` isn't configured correctly and `npx cap sync` results in failures. 

Instead of trying to maintain a magical command that integrates Capacitor with any project, we should write a handful of tutorials: "Add Capacitor to a Web App", "Migrate from Cordova to Capacitor", etc.

adamdbradley:
`npm init @capacitor/app` would be the replacement

jcesarmobile (approved):

dwieeb:
We have decided to keep `cap init`, but change it slightly. See https://github.com/ionic-team/capacitor/issues/3465

PR #3457, opened 2020-08-24
feat(ios): Add CAPFileUtils class with getCleanData function

jcesarmobile:
Removes the data part of a base64 file representation if present, to use it in Clipboard, Filesystem or other plugins that might need it.
Usage is like this `CAPFileUtils.getCleanData(data)`

### DIFF
--- a/ios/Capacitor/Capacitor/CAPFile.swift
+++ b/ios/Capacitor/Capacitor/CAPFile.swift
@@ -72,3 +72,13 @@ private class CAPFileResolverNotImplemented: CAPFileResolver {
         return nil
     }
 }
+
+@objc public class CAPFileUtils: NSObject {
+    public static func getCleanData(_ data: String) -> String {
+        let dataParts = data.split(separator: ",")
+        if let part = dataParts.last {
+            return String(part)
+        }
+        return data
+    }
+}


ikeith:
I'm personally of the opinion that it doesn't need to be centralized and shared from the core because it is so small. But if it is going to be moved from a private, internal method into a public one then I have several concerns with this specific implementation.

1. This function isn't a file method - it's a string manipulation function. It may be used in some file operations but that's beside the fact.

2. The name `getCleanData` is unclear because
   a) it has nothing to do with `Data` objects (it operates on `String),
   b) it is not idiomatic, and
   c) isn't descriptive about what it actually does

3. Adding multiple classes to the same file is a can be a problem because that makes it harder to find if the class name != file name. The exception would be private/fileprivate extensions or objects that are small and used internally by some parent object (like many of the other declarations in this file). Things that are public, though, should be in separate files or placed in an obviously generic container like "Utilities" or something.

So, my argument would be that this belongs as an extension on String, [as it is here](https://github.com/ionic-team/capacitor-plugins/blob/1abafb9ed92103d1462280194d182ccd48ba3fb6/clipboard/ios/Plugin/Clipboard.swift#L61), just moved into the core project instead of private to `Clipboard`.
- It operates on a string and returns a string.
- It can be named to better describe what it actually does.
- Making it public is [far less dangerous now with the psuedo namespace](https://github.com/ionic-team/capacitor/pull/3447).

So, instead of `CAPFileUtils.getCleanData(foo)`, the usage would be:
```
let foo: String = "whatever"
foo.capacitor.lastCommaSeparatedValue
```

PR #3467, opened 2020-08-25
chore: remove Geolocation plugin

carlpoole:


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -883,4 +883,8 @@ public BridgeWebViewClient getWebViewClient() {
     public void setWebViewClient(BridgeWebViewClient client) {
         this.webViewClient = client;
     }
+
+    public void registerGeolocationPlugin(Plugin geolocationPlugin) {
```
carlpoole, line 887: @dwieeb I added this method to the bridge so that it's possible to provide a Geolocation plugin for the `onGeolocationPermissionsShowPrompt` method in `BridgeWebChromeClient` to use. I'm unsure if we want to leave it at this, or make this method available on the `Plugin` class instead so that plugins don't need to touch the Bridge to do this. Thoughts?

dwieeb: Would it make more sense to have it be more of an event-based design?

jcesarmobile:
We were using the Geolocation plugin for prompting the geolocation permission because it was integrated and was easier to use that than duplicating the code, but now that we are moving plugins out the WebChromeClient should handle the geolocation permissions itself.
There is no point on adding a separate plugin for handling the permission, since if somebody chooses to use the WebView implementation instead of the plugin, then they are not interested in installing a geolocation plugin to begin with.
WebChromeClient already checks some other needed permissions without using plugins.

android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
```diff
@@ -37,12 +38,14 @@
  * WebView instance.
  */
 public class BridgeWebChromeClient extends WebChromeClient {
-    private Bridge bridge;
     static final int FILE_CHOOSER = PluginRequestCodes.FILE_CHOOSER;
     static final int FILE_CHOOSER_IMAGE_CAPTURE = PluginRequestCodes.FILE_CHOOSER_IMAGE_CAPTURE;
     static final int FILE_CHOOSER_VIDEO_CAPTURE = PluginRequestCodes.FILE_CHOOSER_VIDEO_CAPTURE;
     static final int FILE_CHOOSER_CAMERA_PERMISSION = PluginRequestCodes.FILE_CHOOSER_CAMERA_PERMISSION;
     static final int GET_USER_MEDIA_PERMISSIONS = PluginRequestCodes.GET_USER_MEDIA_PERMISSIONS;
+    static final int GEOLOCATION_REQUEST_PERMISSIONS = PluginRequestCodes.GEOLOCATION_REQUEST_PERMISSIONS;
```
jcesarmobile, line 46: Shouldn't use the same request code that is used on Geolocation plugin, the codes are used to identify who called the permission prompt, so if the Geolocation plugin is installed it could cause an undesired callback.

android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
```diff
@@ -261,11 +264,11 @@ public void onGeolocationPermissionsShowPrompt(String origin, GeolocationPermiss
         // Set that we want geolocation perms for this origin
         callback.invoke(origin, true, false);
```
jcesarmobile, line 265: This should only be called with the second param as true if the permissions are accepted, otherwise it should prompt for the permission and this shouldn't be called until the user accept (and pass true) or deny (and pass false).

carlpoole: thanks nice catch. I think this was always just saying true regardless 🙃

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -687,6 +687,11 @@ public void startActivityForPluginWithResult(PluginCall call, Intent intent, int
      */
 
     public void onRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
+        if (requestCode == BridgeWebChromeClient.GEOLOCATION_REQUEST_PERMISSIONS) {
```
jcesarmobile, line 690: the code should be unique, if you use the same on the geolocation plugin then only one of them will receive the response

carlpoole: Yeah. Will make sure it is different

jcesarmobile (changes requested):
Added a minor comment about the request code.

Also, for other permissions I used a "fake" cordova plugin, that's simpler than what you did  (https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java#L90-L107)
but what you did is ok too and works fine.

carlpoole:
> Added a minor comment about the request code.
> 
> Also, for other permissions I used a "fake" cordova plugin, that's simpler than what you did (https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java#L90-L107)
> but what you did is ok too and works fine.

This PR will use the existing code for the geo permissions but it will be adjusted in the permissions + geo-plugin PR's so they dont overlap.

The original way I coded this ensured nobody could override the request code for the browser requesting permissions, but the way using the Cordova plugin is a lot cleaner and simplified things on the bridge so I switched it to match.

jcesarmobile (approved):
LGTM

Issue #3465, opened 2020-08-25
Changes to init command


dwieeb:
- Make it clear it's _only_ for creating Capacitor configuration files 
- Prompt for information and be informative about each requested value 
- Ask for permission to do additional steps, such as copying preferences from `config.xml` in a Cordova app
- Print actionable next steps
- Change the final message (there may be more work to do)
- Do https://github.com/ionic-team/capacitor/pull/3470#discussion_r477234717

PR #3470, opened 2020-08-25
chore(cli): replace inquirer with prompts

dwieeb:
[prompts](https://github.com/terkelg/prompts) will reduce CLI dependencies by ~30 and save >20MB of disk space per install.

The `onCancel` emulates the previous behavior with inquirer. When ctrl+c is used with inquirer, it calls `process.exit` automatically.


cli/src/tasks/add.ts
```diff
@@ -93,16 +94,18 @@ export async function generateCapacitorConfig(config: Config) {
...
+      {
+        type: 'text',
+        name: 'webDir',
+        message:
+          'What directory are your web assets in? (index.html, built JavaScript, etc.):',
+        initial: 'www',
+      },
+    ],
+    { onCancel: () => process.exit(1) },
+  );
```
jcesarmobile, line 108: Looks like this code is unreachable, the whole `generateCapacitorConfig` could be deleted. The `capacitor.config.json` is generated on init, so it always exists on add. And if it doesn't exist add won't work anyway, so it never prompts for webDir.
Maybe it should be moved to the init?

dwieeb: Yeah, I can do that as part of https://github.com/ionic-team/capacitor/issues/3465.

This PR isn't trying to change behavior, just replace the prompt library

jcesarmobile: yeah, but I think a better way of replacing a prompt that is not used is deleting it 😛

jcesarmobile (approved):

PR #3486, opened 2020-08-29
refactor(cli): use output lib

dwieeb:
This PR introduces `@ionic/cli-framework-output` to the Capacitor CLI, which is a lib used by the Ionic CLI with spinners, a logger, and terminal output utilities.

We can also use `@ionic/utils-terminal`, which is already a dependency of `@ionic/cli-framework-output`, to detect terminal features and whether we're in CI or not (in which case we should always disable interactivity, even if the CI supports TTY). 

With it, we can replace `ora`, which has 21 dependencies (including chalk). The output lib has terminal spinners built-in (same ones as Ionic CLI). I recently updated the lib to include elapsed time output to exactly replicate the Capacitor CLI's output.

The output lib also comes with an intelligent logger, which provides automatic wrapping and proper indentation of log messages. 

before:
image

after:
image

TODO:
- [ ] Remove the manual word wrapping in CLI error messages (messages are now auto-wrapped)
- [ ] Replace direct usages of `kleur` with the new `COLORS` object (this will provide semantic meanings to color usage, which should improve consistency)


dwieeb:
@jcesarmobile Thoughts on this before I invest more time?

jcesarmobile (approved):

PR #3496, opened 2020-09-01
Added asset ignore string to allow dirs beginning with underscore in assets

carlpoole:
@dwieeb to answer your question from the other PR I made..

The presence of these items in the list means they *will* be ignored. (The default string includes the `<dir>_` pattern that ignores the directories, thus we override it here to omit that and allow them). The exclamation point here means not to print a warning about it (vs. meaning negation like in other code)

Source: https://android.googlesource.com/platform/frameworks/base/+/6d2c0e5/tools/aapt2/Files.h#88

### DIFF
--- a/android-template/app/build.gradle
+++ b/android-template/app/build.gradle
@@ -9,6 +9,9 @@ android {
         versionCode 1
         versionName "1.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        aaptOptions {
+            ignoreAssetsPattern '!.svn:!.git:!.ds_store:!*.scc:!CVS:!thumbs.db:!picasa.ini:!*~'
+        }
     }
     buildTypes {
         release {
--- a/android-template/app/build.gradle
+++ b/android-template/app/build.gradle
@@ -10,7 +10,7 @@ android {
         versionName "1.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         aaptOptions {
-            ignoreAssetsPattern '!.svn:!.git:!.ds_store:!*.scc:!CVS:!thumbs.db:!picasa.ini:!*~'
+            ignoreAssetsPattern '!.svn:!.git:!.ds_store:!*.scc:.*:!CVS:!thumbs.db:!picasa.ini:!*~'
         }
     }
     buildTypes {
--- a/android-template/app/build.gradle
+++ b/android-template/app/build.gradle
@@ -11,7 +11,7 @@ android {
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         aaptOptions {
              // Files and dirs to omit from the packaged assets dir, modified to accommodate modern web apps.
-             // Default: https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt/AaptAssets.cpp#61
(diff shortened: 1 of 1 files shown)


jcesarmobile:
just noticed the original default ignore also includes `.*` (hidden files) `!.svn:!.git:!.ds_store:!*.scc:.*:<dir>_*:!CVS:!thumbs.db:!picasa.ini:!*~`, should it be `!.svn:!.git:!.ds_store:!*.scc:.*:!CVS:!thumbs.db:!picasa.ini:!*~`? or we want to allow hidden files?

source of the default ignore pattern: https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt/AaptAssets.cpp#61

carlpoole:
> just noticed the original default ignore also includes `.*` (hidden files) `!.svn:!.git:!.ds_store:!*.scc:.*:<dir>_*:!CVS:!thumbs.db:!picasa.ini:!*~`, should it be `!.svn:!.git:!.ds_store:!*.scc:.*:!CVS:!thumbs.db:!picasa.ini:!*~`? or we want to allow hidden files?
> 
> source of the default ignore pattern: https://android.googlesource.com/platform/frameworks/base/+/master/tools/aapt/AaptAssets.cpp#61

Thanks, nice catch.

jcesarmobile (approved):
looks good

dwieeb:
> The presence of these items in the list means they will be ignored.

Yeah, I was just saying we don't need to ignore some of these. The default pattern may confuse web devs because they don't work with those sorts of files. My goal is to hard-code a default that has familiar values that they can change if need be.

We should also document the pattern syntax (or link to it).

android-template/app/build.gradle
```diff
@@ -9,6 +9,9 @@ android {
         versionCode 1
         versionName "1.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
+        aaptOptions {
+            ignoreAssetsPattern '!.svn:!.git:!.ds_store:!*.scc:.*:!CVS:!thumbs.db:!picasa.ini:!*~'
```
dwieeb, line 13:
```suggestion
            ignoreAssetsPattern '!.git:!.ds_store:.*:!thumbs.db:!*~'
```

carlpoole:
I understand what you are saying, but if we remove them then there is potential for these files to be included in android packages and I don't think that is good.

jcesarmobile:
Yeah, I agree with Carl, we should change the default as little as needed, just because we don't use CVS or SVN doesn't mean other people doesn't and that should not be part of the assets, scc files seem to be for source control too.

dwieeb:
Maybe I'm misunderstanding this option. It's for determining which files to include in the bundle from the Android assets, correct? Why would there be source control files inside the assets directory, especially for Capacitor projects where typically it's several directories up?

In any case, I'm fine with that, but we should at least link to the patterns syntax.

carlpoole:
If people drop in their own web apps or in the embedded use case, I think? I am going to add a comment next to it to help prevent any confusion.

jcesarmobile:
When linking a line I prefer linking to the line in the commit that added that line
https://android.googlesource.com/platform/frameworks/base/+/282e181b58cf72b6ca770dc7ca5f91f135444502/tools/aapt/AaptAssets.cpp#61

because in the future master's line 61 could be a different thing

jcesarmobile (approved):

PR #3501, opened 2020-09-02
refactor(cli): cleanup log messages

dwieeb:
TODO

- [x] use logger directly
- [x] replace direct `kleur` usage with more semantically meaningful styles, which should help with consistency in the future
- [x] replace `log()` usage with just `output.write()` where it was printing output to the screen vs logging a message
- [x] make log messages more consistent in terms of capitalization, voice, and style
- [x] fix typos, misspellings, incorrect information


cli/src/ios/update.ts
```diff
@@ -117,9 +118,10 @@ export function generatePodFile(config: Config, plugins: Plugin[]) {
   const capacitoriOSPath = resolveNode(config, '@capacitor/ios');
   if (!capacitoriOSPath) {
     logFatal(
-      `Unable to find node_modules/@capacitor/ios. Are you sure @capacitor/ios is installed? This file is currently required for Capacitor to function.`,
+      `Unable to find ${c.strong(
+        '@capacitor/ios',
+      )} in your project. Is it installed?`,
```
jcesarmobile, line 123: this message is different from the others, all should look the same

jcesarmobile (changes requested):
added two comments

cli/src/android/add.ts
```diff
@@ -20,7 +22,7 @@ export async function addAndroid(config: Config) {
     },
   );
 
-  await runTask(`Syncing Gradle`, async () => {
+  await runTask(`Syncing Gradle with ${c.input('./gradlew')}`, async () => {
```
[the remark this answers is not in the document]
dwieeb: Thanks, updated. 

I think we should show it because we do something similar for iOS (`Updating iOS native dependencies with pod install`), and it aligns with the goal that the Capacitor CLI should be more transparent/informative.

cli/src/android/add.ts
```diff
@@ -22,9 +22,14 @@ export async function addAndroid(config: Config) {
     },
   );
 
-  await runTask(`Syncing Gradle with ${c.input('./gradlew')}`, async () => {
-    return createLocalProperties(config.android.platformDir);
-  });
+  await runTask(
+    `Syncing Gradle with ${c.input(
+      process.platform === 'win32' ? 'gradlew.bat' : './gradlew',
+    )}`,
```
jcesarmobile, line 28: createLocalProperties does more than running gradlew, so if you want to specify what's going on, the new message is still not correct. Anyway, I don't see the point of extending this message and add this complexity to say that it's running gradlew

also, not related to the PR, but it made me realize that we run gradlew in several places, but that won't work on windows

dwieeb: Good point. Maybe "Syncing Gradle" isn't the best message? In any case, I'll revert for now.

> we run gradlew in several places, but that won't work on windows

Yeah, I noticed as well. We also aren't using `cross-spawn` which is necessary once subprocess invocations become more complex.

jcesarmobile (approved):

PR #3516, opened 2020-09-07
so capacitor shows the same launch screen as iOS

jayenashar:
Currently, the Splash Screen plugin uses it's own ViewController that only shows
the background image. A developer may have other items like logos or text on
their LaunchScreen, so it should use the LaunchScreen Storyboard instead of
making its own.

https://github.com/ionic-team/capacitor/issues/1871


jcesarmobile:
Thanks for the PR but we have moved splash plugin to https://github.com/ionic-team/capacitor-plugins, you can send the PR there if you want

PR #3523, opened 2020-09-08
better support for removing delivered notification

jurepurgar:
Currently `removeDeliveredNotifications` uses notification id to remove notification, however by default id is always 0. This adds support for removing notifications by tag which can be send via push. According to discussion in #2266 the current implementation only works by customizing `FirebaseMessagingService`, with this it is possible to use tag to achieve the same result, but without customizing messaging service.


jcesarmobile:
Thanks for the PR but we have moved push notifications plugin to https://github.com/ionic-team/capacitor-plugins, you can send the PR there if you want

PR #3533, opened 2020-09-10
refactor(cli): config refactor

dwieeb:
- `Config` class replaced with a function that returns a readonly object
- deleted unused properties
- `resolveNode` now just uses `require.resolve` with most statements changed to resolve to root of packages using each package's `package.json`


dwieeb:
@ptitjes Would you be able to review/test the changes around `resolveNode`? It is possible to use `require.resolve` if we use `<pkg>/package.json` as the resolving point for the root of a package. It also works for resolving any JS files within the package. My biggest concern is with `getFilePath`. Do you remember why this is using `resolveNode`? The function seems to be for joining paths of non-JS files within a plugin's directory. I don't understand why the code within the `path.startsWith('node_modules')` if-block is necessary.

cli/src/config.ts
```diff
@@ -1,354 +1,156 @@
...
+  const platformDir = resolve(rootDir, name);
+  const webDir = 'public';
+  const nativeProjectName = 'App';
+  const templateName = 'ios-template';
+  const pluginsFolderName = 'capacitor-cordova-ios-plugins';
+
+  return {
+    name,
     minVersion: '11.0',
     cordovaSwiftVersion: '5.0',
```
jcesarmobile, line 115:
```suggestion
    cordovaSwiftVersion: '5.1',
```
We updated the Capacitor version to 5.1, so this one should be probably updated too

jcesarmobile (commented):
Added a minor comment.

Also, the way for users to configure the android studio path was `windowsAndroidStudioPath` and `linuxAndroidStudioPath`. The PR seems to remove those config options (so breaking change? should be documented?).
In open.ts it's still mentioned that you should use them, so texts should be updated for the new way of configuring them.

And in `capacitor.config.json` the plugins object appears the first one, since it can grow a lot if users add a lot of configurations, the other values would be hard to find, probably better to print the plugins object the last

dwieeb:
I didn't mean to break the Android Studio path stuff at this time, so I fixed it. But I think we should change it before the final release.

Fixed the `plugins` appearing first in some cases. 👍

cli/src/common.ts
```diff
@@ -222,47 +206,28 @@ export function buildXmlElement(configElement: any, rootName: string): string {
...
+    };
+  }
 
-  // Store our newly created or found external config as the default
-  config.loadExternalConfig();
+  await writeJSON(
+    config.app.extConfigFilePath,
+    {
+      ...oldConfig,
+      ...extConfig,
```
jcesarmobile, line 227: It's still showing the plugins first because on first execution oldConfig is empty and just adds the plugin entry to it
```suggestion
      ...extConfig,
      ...oldConfig,
```

dwieeb: Unfortunately we want `extConfig` to overwrite keys on `oldConfig`, so this suggestion won't work

jcesarmobile (approved):

PR #3539, opened 2020-09-10
chore: integrate ESLint

dwieeb:
(almost) seamless Prettier + [ESLint](https://eslint.org/) integration using [`@ionic/eslint-config`](https://github.com/ionic-team/eslint-config)

[These lines](https://github.com/ionic-team/eslint-config/blob/main/index.js#L9-L10) disable the rules that conflict with Prettier. ~For new projects, we could also distribute a "recommended" set of lint rules as opposed to just the base set that this PR is using (see https://github.com/ionic-team/eslint-config/issues/1).~ (now using recommended rule set)

Notice the built-in support for GitHub Action annotations 😮


cli/src/ios/update.ts
```diff
@@ -96,7 +96,7 @@ export async function updatePodfile(
   const podfileLockPath = join(projectRoot, 'Podfile.lock');
   let podfileContent = await readFileAsync(podfilePath, 'utf8');
   podfileContent = podfileContent.replace(
-    /(Automatic Capacitor Pod dependencies, do not delete)[\s\S]*(# Do not delete)/,
+    /(Automatic Capacitor Pod dependencies, do not delete)[\s\S]*(# Do not delete)/,
```
dwieeb, line 99: @jcesarmobile I noticed the [no-irregular-whitespace](https://eslint.org/docs/rules/no-irregular-whitespace) rule triggered for this space. Was it intentional to not use a normal space here?

I believe this change will have downstream effects when the podfile is regenerated in users' apps. (They'll see a git difference, which is probably fine.)

jcesarmobile: I don't think it was intentional nor know how to create an "irregular" whitespace, but does the replace still work if you change it?

jcesarmobile: Since it's next to the # I guess I pressed the space bar while still holding alt key when I added that line in the Podfile, and then I copied from the Podfile to the update.ts

cli/src/ios/update.ts
```diff
@@ -96,8 +96,8 @@ export async function updatePodfile(
   const podfileLockPath = join(projectRoot, 'Podfile.lock');
   let podfileContent = await readFileAsync(podfilePath, 'utf8');
   podfileContent = podfileContent.replace(
-    /(Automatic Capacitor Pod dependencies, do not delete)[\s\S]*(# Do not delete)/,
-    '$1' + dependenciesContent + '\n  $2',
+    /(def capacitor_pods)[\s\S]+?(\nend)/,
+    `$1\n  # Autogenerated by Capacitor--DO NOT DELETE${dependenciesContent}\n  # DO NOT DELETE$2`,
```
dwieeb, line 100: @jcesarmobile Here's what I ended up doing. Now it replaces the content of the entire block. I think this is okay?

jcesarmobile (approved):

PR #3542, opened 2020-09-11
chore: Add request code for Share plugin

jcesarmobile:
I need a new request code for Share plugin, since we have all other plugin codes listed here I think it makes sense to add this new code.


dwieeb:
We can keep a list of request codes somewhere to make sure we're not reusing them. Third-party plugins also will need to be unique, so we could also add the request codes from Capacitor Community plugins to the list. Eventually, we could have tooling in CI that verifies every request code is unique across core and community plugins.

PR #3546, opened 2020-09-13
Added disable drag and drop in a WKWebView.

trancee:
There is an annoying issue on iPads and WKWebView when long pressing on images or text, it starts a drag and drop operation. This only happens on iPads in our tests so far. While researching the issue, I came across this StackOverflow post:
https://stackoverflow.com/questions/49911060/how-to-disable-ios-11-and-ios-12-drag-drop-in-wkwebview

It explains how to circumvent this issue. I took the basic idea and implemented it for use in Capacitor. As mentioned in the post, since iOS 12 it is being handled differently so I had to add another override method for `viewDidAppear` and added the code to disable the UIDragInteraction there. 

It solves the issue and makes sure that no drag and drop operations are happening inside the WKWebView. However, I am not sure if this should be configurable as some projects might need that feature.

Please kindly review this merge request and give feedback. Thank you!

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -167,6 +167,21 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKUID
         loadWebView()
     }
 
+    override public func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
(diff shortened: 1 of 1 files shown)


jcesarmobile:
thanks for the PR
I think some apps might still want the drag and drop for other purposes and this will prevent it from working

It probably makes more sense in a separate plugin.

While you can't get the `viewDidLoad` outside the view controller, I've verified that `UIApplication.didBecomeActiveNotification` it's called after `viewDidLoad` and putting the code there still seems to work. You can add the listener for it on the `load()` method of a plugin class.

PR #3550, opened 2020-09-14
Embedded Capacitor: Fix broken BridgeFragment configuration

fkirc:
### Platform(s)

Android embedded use case.

### Current Behavior

`BridgeFragment` ignores configuration in `android/app/src/main/assets/capacitor.config.json`.
For example, the following entry in `capacitor.config.json` will be ignored:

```
"server": {
    "url": "http://localhost/not-existing:8100"
  }
```

### Expected Behavior

If we add the following to `android/app/src/main/assets/capacitor.config.json`:

```
"server": {
    "url": "http://localhost/not-existing:8100"
  }
```

Then BrideFragment should show an error like `http://localhost/not-existing:8100 could not be loaded...`, instead of launching as normal.

### Code Reproduction

I created the following sample project to reproduce the error: https://github.com/fkirc/capacitor/tree/broken_bridgefragment_config_reproducer

For this sample project, I modified the Capacitor Android template to use `BridgeFragment` instead of `BridgeActivity`.
The modified `activity_main.xml` looks like this:

```
<?xml version="1.0" encoding="utf-8"?>
<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:tools="http://schemas.android.com/tools"
    android:layout_width="match_parent"
    android:layout_height="match_parent"
    tools:context=".MainActivity">
    <fragment android:name="com.getcapacitor.BridgeFragment"
        android:id="@+id/my_bridge_fragment"
        android:layout_width="match_parent"
        android:layout_height="match_parent" />
</LinearLayout>
```

The modified `MainActivity.java` looks like this:

```Java
import android.net.Uri;
import android.os.Bundle;
import androidx.appcompat.app.AppCompatActivity;
import com.getcapacitor.BridgeFragment;

public class MainActivity extends AppCompatActivity implements BridgeFragment.OnFragmentInteractionListener {
  @Override
  public void onCreate(Bundle savedInstanceState) {
    setContentView(R.layout.activity_main);
  }
  @Override
  public void onFragmentInteraction(Uri uri) {
    // Ignore
  }
}
```

Finally, I modified `android/app/src/main/assets/capacitor.config.json` to showcase the error as described above:

```
{
  "appId": "com.captemplate",
  "appName": "captemplate",
  "bundledWebRuntime": true,
  "npmClient": "npm",
  "webDir": "www",
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 0
    }
  },
  "server": {
    "url": "http://localhost/not-existing:8100"
  }
}
```


fkirc:
I have now resolved the merge-conflict with my own previous PRs.

jthoms1:
We have a separate effort underway for using Capacitor in Native applications. We will provide more info on the topic in coming months. This effort is a commercialization and will not be made available as open source. If you would like to learn more about it I am happy to chat about it. Thank you!

Issue #3544, opened 2020-09-12
feat: Build Safety - Detect outdated apps to increase release safety and prevent common mistakes


fkirc:
### Description

Now that Capacitor is becoming a mature project, I feel we should not only focus on features and bugfixes, but also on the day-to-day workflows of app developers.
One thing that has always annoyed me is the risk to produce _outdated apps_, that is, native apps where the web-assets originate from a different commit/branch than the current commit/branch.
For example, outdated apps can happen if a developer forgot to run `npx cap sync` after a `git pull`.
The damage of outdated apps can vary:

- If a developer does not discover the mistake immediately, then this might waste quite a lot of time (e.g. if manual tests have to be repeated).
- In the worst case, a broken release gets published to app stores.

### Platform(s)

iOS (template project), Android (template project)

### Preferred Solution

I created the tool [capsafe](https://github.com/fkirc/capacitor-build-safety) to solve the issue for all my projects: https://github.com/fkirc/capacitor-build-safety
You can give `capsafe` a try with:

`npm install --save-dev capacitor-build-safety`

`capsafe`'s [README](https://github.com/fkirc/capacitor-build-safety/blob/master/README.md) does not only describe how to integrate `capsafe`, but also describes my philosophy on how the issue should be solved.

However, although I believe that `capsafe` is a great fit for most development teams, `capsafe` is still too opinionated for a merge into Capacitor. To merge those safety checks into Capacitor, I aim to establish a "minimum-viable-compromise", that is, the least opinionated solution that still creates value.

### Proposed Solution: prebuild.sh scripts

I propose the following solution to be integrated into Capacitor's template projects (iOS/Android):

- Add two new scripts `prebuild.sh` to the iOS/Android template projects.
- During each iOS/Android-build, `prebuild.sh` checks whether a web-asset directory exists at the right location. If not, then `prebuild.sh` prints a meaningful warning message. This warning message will accompany the following error ("404: could not load from localhost...").
- `prebuild.sh` will be an extension point for developers who want to add actual safety-checks instead of warning messages (e.g. `capsafe verify-commit-evidence`).

See [README](https://github.com/fkirc/capacitor-build-safety/blob/master/README.md) for platform-specific details on how the `prebuild.sh` scripts could be executed.

### Alternatives

If we are not changing the iOS/Android template projects, then we might at least add some guidance to the documentation.

In theory, `capsafe` could be replaced with dead-simple shell scripts.
[...]

PR #3555, opened 2020-09-14
feat: Build Safety - Add prebuild scripts to prevent common mistakes

fkirc:
The goal of this PR is to promote safer and more efficient workflows for Capacitor development.
Please see https://github.com/ionic-team/capacitor/issues/3544 and https://github.com/ionic-team/capacitor/discussions/1478 for a detailed explanation of this PR.


fkirc:
Any updates on this?

After several months of experimentation, I have finally found a workflow that works well in practice: https://github.com/fkirc/capacitor-build-safety
As described in the README, the workflow involves a `capsafe disable`-command.

Nevertheless, `capsafe` would be too opinionated for the official Capacitor-repo, which is why I am proposing a "minimum-viable-compromise" with the lowest impact on already existing workflows.

jcesarmobile:
Thanks for the PR, but we are not going to add this feature.

The .sh scripts won't work for windows users and I don't think it's so common mistake as to add a script.

Also, as you created your own package, users that are worried about that problem can use it.

PR #3569, opened 2020-09-17
chore: Remove Modals plugin

jcesarmobile:
Android build fails because Camera is using Dialogs class, which is removed.
Camera plugin should be removed first.


dwieeb:
Camera plugin was removed from `main`

PR #3596, opened 2020-09-22
fix(android): move splash screen trigger before the webview render to prevent flicker

carlpoole:
This seems to fix the following flickering issue

https://github.com/ionic-team/capacitor/discussions/3368


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -143,6 +143,8 @@ public Bridge(
         this.config = new CapConfig(getActivity().getAssets(), config);
         Logger.init(this.config);
 
+        Splash.showOnLaunch((BridgeActivity) context, getConfig());
```
jcesarmobile, line 146: Since it's the `Bridge` class you can call `this.config` instead of the `getConfig()`

```suggestion
        Splash.showOnLaunch((BridgeActivity) context, this.config);
```

Also, I think this should go in the 2.x branch first

dwieeb:
LGTM, but please recreate on the 2.x branch.

dwieeb:
This will make it in when we merge 2.x into main.

PR #3621, opened 2020-09-28
Permissions improvements

carlpoole:
Changes:

- Introduces new `@CapacitorPlugin` annotation to replace `@NativePlugin` that allows a developer to pick an alias for a permission. Aliases are optional (otherwise you have to use the raw Android permission string). Aliases can be the same for two different permissions, for example: `Manifest.permission.ACCESS_COARSE_LOCATION` and `Manifest.permission.ACCESS_FINE_LOCATION` can be given an alias "location". All permissions mapped to a single alias must be granted for the helper methods to report back a fully granted state. If one is false, all are considered false. If this is not desirable, then multiple aliases are probably more suited for the use case.
- Backwards compatible with `@NativePlugin`
- Plugins should define their own constants for request and permission codes, and not use constants from within core (removed).
- `checkPermissions` and `requestPermissions` are available on the parent `Plugin` class to fit the standard request plan. `requestPermissions` existed before in Android and needs to be extended by the Plugin and return the payload. A new helper called `getPermissionStates ` will get a snapshot of the current permission states for easy returning by the plugin developer, example:
- Capacitor now caches the Android permission state for "denied" when it is detected that the user denied a permission and selected "Never ask again".

```
@Override
    protected void handleRequestPermissionsResult(int requestCode, String[] permissions, int[] grantResults) {
        super.handleRequestPermissionsResult(requestCode, permissions, grantResults);

        // .. some custom logic
        
        savedCall.resolve(getPermissionStates());
        savedCall.release(bridge);
    }
```


android/capacitor/src/main/java/com/getcapacitor/Permission.java
```diff
@@ -0,0 +1,22 @@
+package com.getcapacitor;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+/**
+ * Permission annotation for use with @CapacitorPlugin
+ */
+@Retention(RetentionPolicy.RUNTIME)
+public @interface Permission {
```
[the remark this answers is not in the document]
carlpoole: Yeah I an do that, would be more organized

core/src/web/local-notifications.ts
```diff
@@ -133,7 +132,7 @@ export class LocalNotificationsPluginWeb
     });
   }
 
-  requestPermissions(): Promise<PermissionsRequestResult> {
+  requestPermissions(): Promise<{ results: [any] }> {
```
dwieeb, line 135: `[any]` = 1-tuple whose element is of type `any`
`any[]` = array of `any`

```suggestion
  requestPermissions(): Promise<{ results: any[] }> {
```

android/capacitor/src/main/java/com/getcapacitor/NativePlugin.java
```diff
@@ -1,12 +1,16 @@
 package com.getcapacitor;
 
+import com.getcapacitor.annotation.CapacitorPlugin;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 
 /**
  * Base annotation for all Plugins
+ * @deprecated
+ * <p> Use {@link CapacitorPlugin} instead
```
dwieeb, line 10: Not sure if this was intentional? Does it need to be wrapped in `<p></p>` tags?

```suggestion
 * Use {@link CapacitorPlugin} instead
```

dwieeb (commented):
Android side is looking good. Notes:

- Calling `requestPermissions({ permissions: ['garbage'] })` resolves with `undefined` and should probably error.

Where are the changes for iOS and web?

android/capacitor/src/main/java/com/getcapacitor/NativePlugin.java
carlpoole: It's valid. It doesn't look right but it's how the IDE formatted it correctly, I think. https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#format

carlpoole:
> Where are the changes for iOS and web?

I think all the iOS changes are plugin-based and not in the core (since iOS doesn't have the core permission structure like Android does). I.E. the requestPermissions and checkPermissions functions are to be implemented directly in a plugin.

For web we cleaned up some of the existing code, and added the base object for the standard permissions states https://github.com/ionic-team/capacitor/pull/3621/files#diff-ed1172842ecd74e628f7d2329b4414fe1b5196f6440f9293a5f08c0b14860b9c

Side note: I missed the "prompt-with-rationale" state in Android - fixing...

I think that was all we needed per previous discussion but let me know if you want to discuss further.

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -13,8 +13,10 @@
 import android.os.Handler;
 import android.os.HandlerThread;
 import android.webkit.ValueCallback;
+import android.webkit.WebChromeClient;
```
jcesarmobile, line 16:
```suggestion
```

not used

android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
```diff
@@ -39,13 +39,12 @@
...
-    static final int FILE_CHOOSER_CAMERA_PERMISSION = PluginRequestCodes.FILE_CHOOSER_CAMERA_PERMISSION;
-    static final int GET_USER_MEDIA_PERMISSIONS = PluginRequestCodes.GET_USER_MEDIA_PERMISSIONS;
-    static final int GEOLOCATION_REQUEST_PERMISSIONS = PluginRequestCodes.GEOLOCATION_REQUEST_PERMISSIONS;
-    static final String[] geoPermissions = { Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION };
+    static final int FILE_CHOOSER = 9007;
+    static final int FILE_CHOOSER_IMAGE_CAPTURE = 9008;
+    static final int FILE_CHOOSER_VIDEO_CAPTURE = 9009;
+    static final int FILE_CHOOSER_CAMERA_PERMISSION = 9010;
+    static final int GET_USER_MEDIA_PERMISSIONS = 9011;
+    static final int GEOLOCATION_REQUEST_PERMISSIONS = 9004;
```
jcesarmobile, line 47:
```suggestion
    static final int FILE_CHOOSER = 9001;
    static final int FILE_CHOOSER_IMAGE_CAPTURE = 9002;
    static final int FILE_CHOOSER_VIDEO_CAPTURE = 9003;
    static final int FILE_CHOOSER_CAMERA_PERMISSION = 9004;
    static final int GET_USER_MEDIA_PERMISSIONS = 9005;
    static final int GEOLOCATION_REQUEST_PERMISSIONS = 9006;
```

better put codes in order for making easier to spot which ones are available in case we need to add more

carlpoole: Will these not conflict with the uses that already used the codes in that range? https://github.com/ionic-team/capacitor/pull/3621/files#diff-120f0e6b31193b391266aa177c18c6825b978c72c444be297115b31a061bb3a1

jcesarmobile: In the plugins that used those codes we will need to use different codes from this ones, but plugins need to be updated anyway since this PR removes the file with the constants.
We just need to make sure the codes in the plugins are unique.

carlpoole: I think this is good, but let's do it after we have fixed all the plugins to work with the new permissions. Then we can do all the code changes across all plugins in a single commit and centralize it in one place so we don't miss any.

dwieeb (approved):

android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
jcesarmobile: Then would be better to revert the removal of PluginRequestCodes.java and remove it once all plugins are migrated

carlpoole: Makes sense, reverted for now

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -449,15 +525,18 @@ public void requestPermissions(PluginCall call) {
      * Handle request permissions result. A plugin can override this to handle the result
      * themselves, or this method will handle the result for our convenient requestPermissions
      * call.
+     * @deprecated
```
jcesarmobile, line 528: add what should be used instead.

ikeith (approved):

jcesarmobile (changes requested):
Added some minor comments.



Also, the code wrongfully assumes that `handleRequestPermissionsResult` will only be used by plugins that have `NativePlugin` annotation and that plugins with `CapacitorPlugin` will overwrite `onRequestPermissionsResult`, but nothing prevents users from using `onRequestPermissionsResult` on `NativePlugin`.
I don't think we can do anything about it, but at least should be documented that `onRequestPermissionsResult` will be called for `CapacitorPlugin` and  `handleRequestPermissionsResult` for `NativePlugin`

And a question, `PermissionState` is android specific, should plugins extend it based what the possible return values are available on iOS?

jcesarmobile (approved):

PR #3641, opened 2020-10-02
refactor(android): assorted cleanups

smpeters:


android/capacitor/src/main/java/com/getcapacitor/Logger.java
```diff
@@ -21,7 +21,7 @@ public static void init(CapConfig config) {
     }
 
     private void loadConfig(CapConfig config) {
-        this.config = config;
+        Logger.config = config;
```
carlpoole, line 24: @jcesarmobile is this change correct, or should `config` be an instance variable instead of a static one?

android/capacitor/src/main/java/com/getcapacitor/cordova/CapacitorCordovaCookieManager.java
```diff
@@ -11,7 +11,7 @@
     public CapacitorCordovaCookieManager(WebView webview) {
         webView = webview;
         cookieManager = CookieManager.getInstance();
-        cookieManager.setAcceptFileSchemeCookies(true);
+        CookieManager.setAcceptFileSchemeCookies(true);
```
carlpoole, line 14: @smpeters do you recall what the specific complaint was from Android Studio was regarding this? At first glance, the original looks correct.

smpeters: The complaint is that the method is a static method so it should be called using the class and not an instance.

smpeters:
Prettier wanted a good number of minor cleanups with the most recent merge.

dwieeb:
@smpeters Can you add me as a contributor to your fork or fix the conflicts (just `git rm` the conflicting files)?

dwieeb:
@smpeters What is the purpose of adding all the gradle files in `android/capacitor`? The CLI is going to use `android/gradlew`, etc.

smpeters:
> @smpeters What is the purpose of adding all the gradle files in `android/capacitor`? The CLI is going to use `android/gradlew`, etc.

These files are removed.

dwieeb:
Thank you!

Issue #3670, opened 2020-10-13
Investigate errors and error codes (UNAVAILABLE, UNIMPLEMENTED)


dwieeb:
We should align the `unimplemented()` and `unavailable()` plugin calls with the new `UnsupportedBrowserException` type of errors, which allow for an error code and custom messages.

PR #3673, opened 2020-10-13
feat: unified errors and error codes

dwieeb:
This unifies the base class of Capacitor exceptions to a new `Capacitor.Exception` class, allowing the following:

```ts
err instanceof Capacitor.Exception; // true
```

Additionally, the `.unimplemented()`/`.unavailable()` helpers on iOS, Android, and Web can now accept a message argument and use the new exception codes to allow the following:

```ts
e.code === ExceptionCode.Unimplemented; // true
e.code === ExceptionCode.Unavailable; // false
```

These codes make it easier for people to check what went wrong with Plugin API calls.

Because of the new availability of `UNAVAILABLE` on iOS and Web, it makes sense to remove `UNSUPPORTED_BROWSER` and use unavailable instead. Usage example:

```diff
--- a/haptics/src/web.ts
+++ b/haptics/src/web.ts
@@ -1,4 +1,4 @@
-import { UnsupportedBrowserException, WebPlugin } from '@capacitor/core';
+import { WebPlugin } from '@capacitor/core';

 import type {
   HapticsPlugin,
@@ -70,9 +70,7 @@ export class HapticsWeb extends WebPlugin implements HapticsPlugin {
     if (navigator.vibrate) {
       navigator.vibrate(pattern);
     } else {
-      throw new UnsupportedBrowserException(
-        'Browser does not support the vibrate API',
-      );
+      throw this.unavailable('Browser does not support the vibrate API');
     }
   }
 }
```

closes https://github.com/ionic-team/capacitor/issues/3670


core/src/definitions.ts
```diff
@@ -72,7 +72,37 @@ export interface PluginConfig {
...
+  Unavailable = 'UNAVAILABLE',
+}
+
+export interface CapacitorException extends Error {
+  code: ExceptionCode;
+}
+
+declare const CapacitorException: {
+  prototype: CapacitorException;
+  new (message: string, code: ExceptionCode): CapacitorException;
```
jcesarmobile, line 101:
```suggestion
  new (message: string, code?: ExceptionCode): CapacitorException;
```
If we plan to make all errors thrown by Capacitor to be Capacitor.Exception, then code should be optional.
With current implementation and this changes, all native Capacitor errors thrown are Capacitor.Exception, but only unimplemented and unavailable have codes, other errors have undefined code. But on web we can't create Capacitor.Exception errors without a code

core/src/definitions.ts
```diff
@@ -72,7 +72,37 @@ export interface PluginConfig {
   name: string;
 }
 
+export enum ExceptionCode {
+  /**
+   * API is not implemented.
+   *
+   * This means the API can't be used because:
+   *   - it is not implemented at all
```
jcesarmobile, line 80: I don't think this will ever be the case (not implemented at all), when it's not implemented it's just because it's not possible in one or more platforms, but if it's not possible on any platform the method should just not exist.

jcesarmobile (changes requested):
just two minor comments, other than that looks and works good.

jcesarmobile (approved):
looks good

PR #3676, opened 2020-10-14
chore: remove App plugin

jcesarmobile:
Since on Android the App plugin was very coupled with the bridge I've created a new App class (that is not a plugin) to host that functionality and the App plugin will use it as an intermediary, i.e subscribe to its events.


android/capacitor/src/main/java/com/getcapacitor/App.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public interface AppRestoredListener {
+        void onAppRestored(PluginResult result);
+    }
+
+    @Nullable
+    private AppStatusChangeListener statusChangeListener;
+
+    private BackButtonListener backButtonListener;
+    private AppRestoredListener appRestoredListener;
```
dwieeb, line 32: Are these also `@Nullable`? or does it not matter?

android/capacitor/src/main/java/com/getcapacitor/App.java
```diff
@@ -0,0 +1,85 @@
...
+        this.appRestoredListener = listener;
+    }
+
+    protected void fireRestoredResult(PluginResult result) {
+        if (appRestoredListener != null) {
+            appRestoredListener.onAppRestored(result);
+        }
+    }
+
+    public void fireChange(boolean isActive) {
```
dwieeb, line 69: Would this be more descriptive?

```suggestion
    public void fireStatusChange(boolean isActive) {
```

dwieeb (approved):

dwieeb:
Would it make sense to allow multiple listeners to these events, instead of just one?

carlpoole:
> Would it make sense to allow multiple listeners to these events, instead of just one?

I wondered that too. I think it makes sense to allow for it

carlpoole (approved):

Issue #3182, opened 2020-06-30
Embedded Capacitor


imhoffd:
Allow for the use case of embedding Capacitor inside native apps.
Allow for multiple instances of Capacitor in an app (not necessarily concurrently).

fkirc:
I think that "Embedded Capacitor" already works today, but there are still some minor glitches like https://github.com/ionic-team/capacitor/pull/3280.
Nevertheless, it would be nice to have some additional documentation and tested sample code for embedded usage.

dwieeb:
@fkirc Great to hear. Please follow Capacitor 3 development! We will be doing dev releases once development picks up a bit. Your feedback would be very valuable.

JovanShandro:
Is there any plan to add more documentation into how to do this embedding? We are currently looking for possible ways to embed parts of a Capacitor app on a native application. I saw in #3280 that a class called BridgeFragment exists, but could not manage to find any example of how to use it, and also didn't manage to find a way of how to achieve this on Ios. Is this still a feature that is under development, or is it just lacking documentation?

fkirc:
`BridgeFragment` is already working and you can embed it into a native Android app like any other fragment (e.g. inflate it from an XML layout).

However, there are still some rough edges because this is not yet an official feature. 
For example, `BridgeFragment` seems to have a bug that makes `capacitor.config.json` partially broken. This bug can be easily fixed with a one-line change, but my point is that you should not trust the current `BridgeFragment` implementation.
If you demand a highly customized configuration, then you might need to write your own BridgeFragment (which is doable because the relevant code is short and easy to read).

fkirc:
For iOS: I managed to embed Capacitor into an existing native iOS app by embedding `CAPBridgeViewController.swift`.
In my case, I embedded `CAPBridgeViewController.swift` into an existing Storyboard via raw XML-editing in Xcode, but I am pretty sure that there are also other ways to do it.

carlpoole:
Allow embedded capacitor to load a specific path on an instance of embedded capacitor as described here: https://github.com/ionic-team/capacitor/pull/3405.

dwieeb:
For those following this issue, I removed the 3.0.0 milestone because it may be confusing. We are laying the foundation for "Embedded Capacitor" with the 3.0.0 release, but there will still be much work to do. Thanks for your patience!

jthoms1:
Closing this enhancement request.

PR #3678, opened 2020-10-14
chore(iOS): Redefining bridge API

ikeith:
This branch has two goals:

1) Isolate the bridge behind a protocol so that the publicly exposed API can be easily parsed and understood. The scope of the old API was extremely difficult to grasp because of the lack of organization and unclear rules for cross-language (Swift <-> Obj-C) importation. 

2) Update the exposed API to be more consistent and idiomatic while preserving backwards-compatibility wherever possible.

The first goal is accomplished through the declaration of the `CAPBridgeProtocol`, which is what plugins have access to, and making the actual bridge and related objects `internal`. Making the bridge internal meant it needed to be renamed so a new `CAPBridge` class could be exposed to provide a stub for any static methods, in order to preserve backwards compatibility.

The app delegate-related methods have been split into a separate object as they are independent functionality. And that new `AppDelegateProxy` object needs to be made public.


ios/Capacitor/Capacitor/CAPLog.swift
```diff
@@ -1,26 +1,19 @@
...
-        return {}
-    }()
-    
+
+    public static let config = CAPConfig()
+
     public static func print(_ items: Any..., separator: String = " ", terminator: String = "\n") {
-        oneTimeConfigCheck()
-        if enableLogging {
+        if !self.hideLogs() {
```
aparajita: I really don't understand why this is checked dynamically on every single log call — it's enormously wasteful, because the value can't change during runtime. It really should be a static value.

ikeith: This is reverting a workaround for a bug that appeared during testing (a different workaround is used for the time being). Eventually, it will be a static flag as configuration is revisited in support of https://github.com/ionic-team/capacitor/issues/3182

aparajita: Cool, thanks for letting me know.

ios/Capacitor/Capacitor/CAPBridgeProtocol.swift
```diff
@@ -0,0 +1,124 @@
...
+
+/*
+ Extensions to Obj-C protocols are not exposed to Obj-C code because of limitations in the runtime. Therefore these methods
+ are implicitly Swift-only.
+
+ The methods marked as deprecated are here because they can be declared without interfering with the synthesized Obj-C setters
+ for the respective properties (e.g. `setStatusBarVisible:` for 'statusBarVisible`).
+ */
+extension CAPBridgeProtocol {
+    // varidic parameters cannot be exposed to Obj-C
```
aparajita, line 72:
```suggestion
    // variadic parameters cannot be exposed to Obj-C
```

dwieeb:
image

Clicking "Fix" on these renames doesn't substitute the right thing. Any chance we could get this to work automatically? If they don't "just work" maybe we shouldn't use `renamed:`(?)

ikeith:
> Clicking "Fix" on these renames doesn't substitute the right thing. Any chance we could get this to work automatically? If they don't "just work" maybe we shouldn't use `renamed:`(in other words, we shouldn't provide an automatic fix if it doesn't work)

It looks like `renamed` can't do what's needed here. It works for things at the same level, e.g. a property on the same object, but since this is moving from an enum to an extension on a different type it won't substitute correctly. I can change the deprecation to a message so it won't offer the 'fix'.

ios/Capacitor/Capacitor/CAPBridgeProtocol.swift
ikeith: Oops

dwieeb (approved):

jcesarmobile (approved):

PR #3688, opened 2020-10-15
chore(iOS): Updating project build

ikeith:
This branch updates the project structure so it no longer requires CocoaPods to build independently:
- The CapacitorCordova .xcodeproj has been updated to compile all files and add a modulemap (and clean up some small compiler warnings).
- The Capacitor project has been similarly updated and its Podfile has been removed.
- The individual podspecs have been updated to reflect the project configuration, such that directly building the project(s) and importing them through CocoaPods should match.
- The Capacitor project can now run a test scheme.


ios/CapacitorCordova/CapacitorCordova/Info.plist
```diff
@@ -4,6 +4,8 @@
 <dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>$(DEVELOPMENT_LANGUAGE)</string>
+	<key>CFBundleDisplayName</key>
+	<string>Cordova</string>
```
dwieeb, line 8: Why not `CapacitorCordova`?

dwieeb (approved):

ios/CapacitorCordova/CapacitorCordova/Info.plist
ikeith: Because this matches how it is configured when importing via Cocoapods. The output bundle/module/filename are all `Cordova` despite what the project is named.

ikeith:
PR has been updated to work around a race condition between `CAPLog` and `CAPConfig` at initialization and to add an empty host application for Capacitor Tests. This avoids crashes of the type "Invalid parameter not satisfying: bundleProxy != nil" thrown by `UNUserNotificationCenter`. The unit test scheme is now fully functional.

ios/Capacitor/Capacitor/CAPLog.swift
```diff
@@ -1,19 +1,26 @@
...
+        // async dispatch, there is a window of time where the default value will be valid before the config
+        // value(s) can be loaded.
+        DispatchQueue.main.async {
+            let config = CAPConfig()
+            if let configFlag = (config.getValue("ios.hideLogs") as? Bool) ?? (config.getValue("hideLogs") as? Bool) {
+                enableLogging = !configFlag
+            }
+        }
+        return {}
+    }()
```
ikeith, line 16: This is a stop-gap fix as I don't think this is really the best way to solve the problem. Ultimately, I think the `enableLogging` flag should be set externally, probably by the BridgeViewController at the point of entry, rather than letting `CAPLog` load a configuration. That will be the best way to handle things for an embedded context where the configuration might be programmatically changed. But that can addressed when `CAPConfig` is redone and the view controller is refactored.

jcesarmobile (changes requested):
Added a minor comment.

Also, if CAPLog is going to be revisited, I think it shouldn't be changed for now.

To avoid the crash you can create a `capacitor.config.json` file in the sample app and put an empty object inside it `{}`, that's enough to skip the crash for now.

ios/Capacitor.podspec
```diff
@@ -11,6 +11,8 @@ Pod::Spec.new do |s|
   s.authors = { 'Ionic Team' => 'hi@ionicframework.com' }
   s.source = { :git => 'https://github.com/ionic-team/capacitor.git', :tag => s.version.to_s }
   s.source_files = 'Capacitor/Capacitor/*.{swift,h,m}', 'Capacitor/Capacitor/Plugins/*.{swift,h,m}', 'Capacitor/Capacitor/Plugins/**/*.{swift,h,m}'
+  s.exclude_files = 'Capacitor/Capacitor/Plugins/DefaultPlugins.h'
```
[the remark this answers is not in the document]
ikeith: It's excluded so the cocoapods build matches. Otherwise cocoapods will try to add it to the public headers but that will generate a warning since it's not included in the umbrella header. But, yes, it's probably easier to just remove it since it is empty.

jcesarmobile (approved):
LGTM

PR #3710, opened 2020-10-20
Support browser other then Chrome

garaboncias:
When Chrome not installed, plugin is failed as Chrome package name is hard-wired.


jcesarmobile:
thanks for the PR, but we are not accepting new features in Capacitor 2

I've made some change requests in your PR for Browser plugin
https://github.com/ionic-team/capacitor-plugins/pull/76

PR #3715, opened 2020-10-22
fix(android): avoid crash on input file capture

jcesarmobile:
With the Camera plugin removal, `isMediaCaptureSupported()` no longer works as Camera plugin is not present.

 `hasDefinedPermission` and `getManifestPermissions` are a copy of the same methods from `Plugin` class, maybe we should put them in a separate class that can be used by both the plugins and the WebChromeClient instead of duplicating.


carlpoole (approved):
Sounds good. I can add that in with the permissions update branch so the `Plugin` and `BridgeWebChromeClient` can share a permission helper.

PR #3719, opened 2020-10-22
fix(Splash): parse colors as RGBA

aparajita:
On iOS, colors are parsed as RGBA. I would argue that this is the way it **should** be, because that is the [standard CSS format](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#rgba()) (as opposed to ARGB). As web developers, when we see a hex color prefixed by "#", we expect it to be a standard CSS color, which means RGBA.

On Android, `Color.parseColor()` is backwards; it expects ARGB. This is especially surprising coming from google. In any case, this PR does two things to make colors consistent across platforms:

- Allows the "#" to be optional on Android (as it is on iOS).
- Parses the color as RGBA on Android.


aparajita:
Hey @dwieeb, any chance this can at least make it into 3.0?

dwieeb:
I believe so. The PR will need to be remade in our plugins repo.

aparajita:
Okay, I'll make it part of my splash screen PR then. Hopefully I'll get to it next week.

jcesarmobile:
There are a bunch of places where `Color.ParseColor` is used, so if we are going to change this, we should change it everywhere (which will probably need a class for handling it in @capacitor/android, so plugins can use it).
On the top of my head, status bar and browser plugins and even the bridge uses it.

aparajita:
@jcesarmobile I'll make sure to do a search and find everywhere it's being used.

PR #3733, opened 2020-10-27
chore: Update Xcode ci versions

jcesarmobile:
Use latest Xcode 11 (11.7) at the moment
Use latest Xcode 12 beta (12.2) at the moment

not sure if we want to switch to stable Xcode 12 or keep using betas


dwieeb:
> not sure if we want to switch to stable Xcode 12 or keep using betas

I think we should keep using betas because we'll learn about build issues with newer Xcode versions sooner.

dwieeb (approved):

Issue #2493, opened 2020-02-26
feat: Get linuxAndroidStudioPath out of capacitor.config.json


OoDeLally:
# Feature Request

## Describe the Feature Request

The location of Android Studio is specific to a particular machine, not to a project. Every collaborator may use a different Android Studio path.
Therefore `linuxAndroidStudioPath` should not be committed into the repo.

## Describe Preferred Solution
Machine-dependent settings should be in an environment variable, or in a gitignored file.

jcesarmobile:
We have an issue for automatic detection https://github.com/ionic-team/capacitor/issues/560
If it's possible and somebody implements it, then there is no need to use `linuxAndroidStudioPath `

OoDeLally:
Indeed, automatic detection is a hard thing to do, given the many setups to cover.
That is why an environment variable such as `$CAPACITOR_ANDROID_STUDIO_PATH`, or a gitignored file `.capacitor-local-config.json` solve the problem. It is easy to implement, and would cover all use cases.

mirkonasato:
I agree with @OoDeLally, the Android Studio location is not a project-level settings, it's a user-specific one.

Capacitor could simply try launching `studio.sh` by default on Linux, and assume it's in the `PATH`. Users can set their `PATH` as desired in their `.profile` or similar.

PR #3755, opened 2020-11-02
feat(cli): STUDIO_PATH environment variable

dwieeb:
This is a breaking change that removes the `windowsAndroidStudioPath` and `linuxAndroidStudioPath` configuration options, combining them into a `STUDIO_PATH` environment variable, which can be set on any OS.

The naming style of `STUDIO_PATH` comes from other Android Studio environment variables: https://developer.android.com/studio/command-line/variables

addresses https://github.com/ionic-team/capacitor/issues/230 and https://github.com/ionic-team/capacitor/issues/2493

### DIFF
--- a/cli/src/android/open.ts
+++ b/cli/src/android/open.ts
@@ -1,18 +1,29 @@
+import { pathExists } from '@ionic/utils-fs';
+import Debug from 'debug';
 import open from 'open';
 
 import c from '../colors';
 import type { Config } from '../definitions';
 import { logger } from '../log';
 
-export async function openAndroid(config: Config): Promise<void> {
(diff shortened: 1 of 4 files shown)


jcesarmobile:
Wasn’t the point of config.js/ts point to be able to use env variables within the files? That would make this PR redundant without the need to use an environment variable of our own

dwieeb:
This change is addressing the following in the config file proposal:

> Android Studio path (and other system paths) should not be a part of the Capacitor config because it is the configuration of the _app_ and not for system settings.

jcesarmobile (approved):

Issue #3141, opened 2020-06-23
Support dynamic configuration files


dwieeb:
Capacitor should allow a `capacitor.config.{ts,js}` file so that configuration can be generated based on environment.

Related: https://github.com/ionic-team/capacitor/issues/1741, https://github.com/ionic-team/capacitor/discussions/1478

buschtoens:
FWIW we patched exactly this feature into our app project, because we needed separate configurations (e.g. `server.url`, `appId`) for different countries. We can now run `cap sync` with some env variables which generates the `capacitor.config.json` on-the-fly, right before copying it over into the native apps. The `capacitor.config.json` files (generated source and the two copies in the native apps) are `.gitignore`d.

This is working quite nicely so far and I'm much in favor of making `capacitor.config.{ts,js}` official. But we found one major issue: whitelabeling / distinct apps per country.

When you update fields like `appName` and `appId`, these changes _will not_ sync into the native apps when running `cap sync`, because [`editProjectSettingsAndroid`](https://github.com/ionic-team/capacitor/blob/c8a0595e0f36bae69f0471be481c858266e6768b/cli/src/android/common.ts#L62-L172) and [`editProjectSettingsIOS`](https://github.com/ionic-team/capacitor/blob/c8a0595e0f36bae69f0471be481c858266e6768b/cli/src/ios/common.ts#L93-L128) are only called once during `cap add`.

Probably this would not be desirable _anyway_, as when you're running these types of multi-builds, you'll likely want to have a more sophisticated native build pipeline, with different product flavors on Android and build schemes on iOS.

I can't really speak for iOS, as I don't have a lot of experience with the platform, but at least on Android, you would likely prefer to run the full build matrix in one step instead of individually for performance & caching reasons. This however does not work with the current copy-one-config-at-a-time approach.

From my POV what we'd need ideally is being able to generate a whole set of configs, copy all of them over into the native projects (with different file names), and then have the native build pipeline select the right config for the right flavor. Alternatively the dependency could be inverted so that the native build pipeline calls `cap` to generate a config on demand instead of `cap sync` pushing it in.

The dependency inversion would have these additional neat side-effects:

- only required configs are generated, in case not the full build matrix is built
- the native build command is the definitive step that will determine the app flavors, so it likely has all necessary state to provide to the config generator

[...]

PR #3756, opened 2020-11-03
feat(cli): support TS/JS config files

dwieeb:
Devs can now have `capacitor.config.ts` (or `capacitor.config.js`) files, for example:

```typescript

  appId: 'com.capacitorjs.testapp',
  appName: 'capacitor-testapp',
  bundledWebRuntime: false,
  webDir: 'build',
  plugins: {
    SplashScreen: {
      launchShowDuration: 0,
    },
  },
};

export = config;
```

depends on https://github.com/ionic-team/capacitor/pull/3755
resolves https://github.com/ionic-team/capacitor/issues/3141

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -16,7 +16,7 @@ import { OS } from './definitions';
 
 const debug = Debug('capacitor:config');
 
-export const EXTERNAL_CONFIG_FILE = 'capacitor.config.json';
+const EXTERNAL_CONFIG_FILE = 'capacitor.config.json';
 
 export async function loadConfig(): Promise<Config> {
   const appRootDir = process.cwd();
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -30,7 +30,7 @@ export async function loadConfig(): Promise<Config> {
   const webDir = extConfig.webDir ?? 'www';
   const cli = await loadCLIConfig(cliRootDir);
 
-  return {
+  const config = {
     android: await loadAndroidConfig(appRootDir, extConfig, cli),
     ios: await loadIOSConfig(appRootDir, extConfig, cli),
     web: await loadWebConfig(appRootDir, webDir),
@@ -51,6 +51,10 @@ export async function loadConfig(): Promise<Config> {
       bundledWebRuntime: extConfig.bundledWebRuntime ?? false,
     },
   };
+
+  debug('config: %O', config);
+
+  return config;
 }
 
 async function loadCLIConfig(rootDir: string): Promise<CLIConfig> {
@@ -195,8 +199,6 @@ async function determineAndroidStudioPath(os: OS): Promise<string> {
     case OS.Linux:
       return '/usr/local/android-studio/bin/studio.sh';
   }
-
-  debug('No Android Studio path could be determined.');
 
   return '';
 }
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -2,7 +2,8 @@ import { pathExists, readJSON } from '@ionic/utils-fs';
 import Debug from 'debug';
 import { dirname, join, resolve } from 'path';
 
-import { runCommand } from './common';
+import c from './colors';
+import { logFatal, resolveNode, runCommand } from './common';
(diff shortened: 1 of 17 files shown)


aparajita:
> Still debating if TS config files should be supported

What are the technical issues in implementing that? I assume you would compile on the fly.

dwieeb:
> What are the technical issues in implementing that? I assume you would compile on the fly.

We'd have to ship typescript as a dependency and then do something like [what Stencil does](https://github.com/ionic-team/stencil/blob/master/src/compiler/sys/node-require.ts#L15-L61).

aparajita:
> We'd have to ship typescript as a dependency and then do something like what Stencil does.

Yeah, I figured it would be something like that. If you already have working code you can steal from Stencil, you might as well do it. You could add TypeScript as an optional dependency.

adamdbradley (approved):

cli/src/cordova.ts
```diff
@@ -527,7 +527,7 @@ export async function getCordovaPreferences(config: Config): Promise<any> {
     const answers = await logPrompt(
       `${c.strong(
         `Cordova preferences can be automatically ported to ${c.strong(
-          'capacitor.config.json',
+          config.app.extConfigName,
```
jcesarmobile, line 530: init only works on .json files, don't need to change this

jcesarmobile (changes requested):
Added a few comments on the definitions.
`ios.scrollEnabled` is missing.

Also, since init won't work if a .ts/.js file is present, don't need to change the code in a few places, I commented the places where it doesn't need to change.

dwieeb:
Added `scrollEnabled` (isn't documented currently, btw: https://capacitorjs.com/docs/config)

I reworded the bundle option and clarified the default value for `allowsLinkPreview` per your suggestions 👍 

> Also, since init won't work if a .ts/.js file is present, don't need to change the code in a few places, I commented the places where it doesn't need to change.

I chose not to use the config name in init. I think it might be easier when maintaining it in the future (easier to search for `capacitor.config.json`), but I don't feel strongly about it

> Since there is already a common confusion about the android package name and the application ID, I think we should not use package ID here, better use App ID.

For now I changed it back, but we should probably have a bigger discussion because Ionic, Capacitor, and Appflow all call it different things. 😂 I copied this description from [create-capacitor-app](https://github.com/ionic-team/create-capacitor-app/), but I can change it to App ID there too.

jcesarmobile (approved):

cli/src/config.ts
```diff
@@ -57,6 +54,86 @@ export async function loadConfig(): Promise<Config> {
...
+      );
+    }
+
+    const ts = require(tsPath); // eslint-disable-line @typescript-eslint/no-var-requires
+
+    return {
+      extConfigType: 'ts',
+      extConfigName,
+      extConfigFilePath: extConfigFilePath,
+      extConfig: requireTS(ts, extConfigFilePath) as any,
```
buschtoens, line 87: I think it'd be quite useful to optionally support exporting a `() => Promise<ExternalConfig>` instead of just `ExternalConfig`.

Would you accept a pull request?

dwieeb: Yes, I was thinking about this as well. I would be happy to look at a PR that adds that functionality.

cli/src/declarations.ts
```diff
@@ -1,22 +1,340 @@
...
+    /**
+     * Load an external URL in the Web View.
+     *
+     * This is intended for use with live-reload servers.
+     *
+     * **This is not intended for use in production.**
+     *
+     * @since 1.0.0
+     */
+    url?: string;
```
webuniverseio, line 315: Would it be possible to clarify why we say this `This is not intended for use in production`?
You can see under https://github.com/ionic-team/capacitor/discussions/5075 this question was asked. In https://github.com/ionic-team/capacitor/issues/4122 people still use it and it fixes problems like https://github.com/ionic-team/capacitor/issues/1373

webuniverseio: @imhoffd

PR #3759, opened 2020-11-03
feat(iOS): Refactoring configuration

ikeith:
This branch revisits configuration with multiple goals:

- Move the configuration store away from a key/value container to typed properties. This reduces the risk of "stringly-typed" parsing scattered around the project and improves type safety.

- Centralize all configuration defaults and file keys so they are not spread over the codebase. This also allows for easier testing and updating of the configuration file.

- Provide a programmatic structure so applications can configure the capacitor bridge without relying on injecting JSON.

- Add unit tests.

The configuration file(s) is parsed into an intermediate structure which can be modified (supported in a future refactoring of CAPBridgeViewController), and then transformed into an immutable store on the bridge.

CAPLog has been updated to use a simple boolean to control logging instead of loading a second instance of CAPConfig (and making assumptions about its location). The initialization has been reworked to set that flag based on the configuration before any logging might occur.


ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+extension InstanceConfiguration {
+    @objc public func getPluginConfigValue(_ pluginId: String, _ configKey: String) -> Any? {
```
dwieeb, line 4: Since this is now defined, should we deprecate `CAPPlugin.getConfigValue()` ?

ikeith: This has always been defined on `CAPConfig`. I moved it here for compatibility as `CAPPlugin.getConfigValue()` is just a pass-through to this method. (having two methods to do the same thing is potentially confusing but that's the way it's always worked)

Ideally, I think it'd be better if the plugin held its own configuration (which could be set at registration), at which point both of these methods could be deprecated. But if we pursue that approach it would be a follow-on piece of work.

aparajita: > Ideally, I think it'd be better if the plugin held its own configuration 

"held" meaning what exactly? What API do you propose for getting the plugin's config?

ikeith: Presently, the plugin's config is parsed from the file and held as a sub dictionary by the bridge's config. Like so:
```
{
"plugins": {
  "SplashScreen": {
    "launchShowDuration": 3000,
    "backgroundColor": "#ffffffff"
  }
}}
```
Which means that the plugins call `getConfigValue("backgroundColor")`, which calls `bridge.getPluginConfigValue("SplashScreen", "backgroundColor")` (or they can call `getPluginConfigValue` directly) and the bridge looks for that key path in its store to get the value. It's unnecessarily complicated.

Instead, the bridge could match up the name of the plugin and set the configuration at registration. So the plugin would have its own ivar that looks like:
```
{
  "launchShowDuration": 3000,
  "backgroundColor": "#ffffffff"
 }
```
So rather than reaching up the bridge for each and every value, it would function more like `CAPPluginCall`s `options`. And that opens the possibility for the plugin's config to adopt the same typed accessors like `getString`, `getBool`, etc.

aparajita: @ikeith That's what I figured, because that's exactly what I would do. 😁

One thing I did in my plugins was to implement a convenience API that looks first in a call's options and then falls back to the config. So:

```swift
func getString(from call: CAPPluginCall, orConfig fallback: Bool = false) -> String?
```

While you're rewriting so much API, have you given any thought to making the Swift API more iOS idiomatic? For example, this is idiomatic for JavaScript or Java, because there are no labeled parameters in those languages:

```swift
func getString(_ call: CAPPluginCall) -> String?
```

But this is idiomatic for iOS:

```swift
func getString(from call: CAPPluginCall) -> String?
```

ikeith: Well, the usage of `get` in the name isn't idiomatic for Swift in the first place so that isn't a great example. But, yes, moving toward a more fluent, idiomatic API is a goal but not an absolute one. It's a gradual process (some of which was done in #3678) and has to be balanced against backwards compatibility, unnecessarily disruption, and expectations of developers less familiar with the native conventions.

aparajita: You can always add idiomatic API that calls through to the non-idiomatic API. Or maybe vice versa is better since you may want to eventually deprecate the non-idiomatic API.

I guess the most idiomatic API would be to define functions on the config object on the plugin. Something like this:

```swift
class CAPPluginConfig {
  private config: [String: Any?]

  func string(withKey key: String) -> String? {
    return config[key] as? String
  }

  func int(withKey key: String) -> Int {
    return config[key] as? Int
  }

  // and so on...
}

class CAPPlugin {
  public config: CAPPluginConfig

  func getConfigString(_ key: String) -> String? {
    return config.string(withKey: key)
  }
}

public class MyPlugin: CAPPlugin {
  @objc func doSomething() {
    if let foo = config.string(withKey: "foo") {
      // hooray
    }
  }
}
```

Having wrestled with (and read all of the source code of) the current config API, I'd be happy to work on adding something like the above.

ikeith: In the example of CAPPluginCall, the idiomatic way to access the parameters would be `call.options["something] as? String ?? "default"`. The various `get` methods exist _solely_ as a convenience for developers who don't want/expect Swift idioms. So any change to the plugin would follow a similar pattern. Making Swift-idiomatic wrapper methods isn't the point.

aparajita: You are right, good point. I sometimes forget just how good Swift is.

ios/Capacitor/Capacitor/CAPLog.swift
```diff
@@ -1,19 +1,11 @@
...
     public static func print(_ items: Any..., separator: String = " ", terminator: String = "\n") {
-        if !self.hideLogs() {
+        if enableLogging {
             for (itemIndex, item) in items.enumerated() {
                 Swift.print(item, terminator: itemIndex == items.count - 1 ? terminator : separator)
             }
         }
     }
-
-    public static func hideLogs() -> Bool {
```
dwieeb, line 13: Not sure why this was public to begin with, but just noting that it is something being removed from the public API.

ios/Capacitor/Capacitor/KeyPath.swift
```diff
@@ -0,0 +1,62 @@
...
+        self.segments = segments
+    }
+
+    // returns a tuple of the first segment and the remaining key path. result is nil if the key path has no segments.
+    func headAndRemainder() -> (head: String, remainder: KeyPath)? {
+        guard !isEmpty else {
+            return nil
+        }
+        var paths = segments
+        let head = paths.removeFirst()
```
dwieeb, line 25: I don't know swift very well, does this modify `self.segments`?

ios/Capacitor/Capacitor/KeyPath.swift
```diff
@@ -0,0 +1,62 @@
...
+            case let (head, remainder)? where remainder.isEmpty: // reached the end of the path
+                return self[head]
+            case let (head, remainder)?: // we have at least one level to traverse
+                switch self[head] {
+                case let childObject as JSObject: // iterate down the next level
+                    return childObject[keyPath: remainder]
+                default: // not an object, can't go any deeper
+                    return nil
+                }
+            }
```
dwieeb, line 59: This is so functional 😍

ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
```diff
@@ -0,0 +1,17 @@
...
+        return (pluginConfigurations as? JSObject)?[keyPath: KeyPath("\(pluginId).\(configKey)")]
```
dwieeb, line 5: How come you can't do this like in other places?

```suggestion
        return (pluginConfigurations as? JSObject)?[keyPath: "\(pluginId).\(configKey)"]
```

ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    @available(*, deprecated, message: "Use direct property accessors")
+    @objc public func getValue(_ key: String) -> Any? {
+        return (legacyConfig as? JSObject)?[keyPath: KeyPath(key)]
```
dwieeb, line 10: same question

```suggestion
        return (legacyConfig as? JSObject)?[keyPath: key]
```

ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
```diff
@@ -0,0 +1,17 @@
...
+    }
+
+    @available(*, deprecated, message: "Use direct property accessors")
+    @objc public func getString(_ key: String) -> String? {
+        return (legacyConfig as? JSObject)?[keyPath: KeyPath(key)] as? String
```
dwieeb, line 15: same question

```suggestion
        return (legacyConfig as? JSObject)?[keyPath: key] as? String
```

dwieeb (commented):
- I found `@class CAPConfig;` in `CAPPlugin.h` -- is this a problem?

ikeith:
> * I found `@class CAPConfig;` in `CAPPlugin.h` -- is this a problem?

Good catch. It's an unused forward declaration so, no, it's not a problem but it should be removed anyway.

ios/Capacitor/Capacitor/CAPLog.swift
ikeith: Yeah, its replacement property is public as well but we should probably note that.

ios/Capacitor/Capacitor/KeyPath.swift
ikeith: Short answer: no.

Long answer: collections are value types in Swift (unless you jump through some hoops to make a reference type) so this is only modifying `paths` not `segments`. In addition, KeyPath itself is a struct (which is a copy-on-write value types) so if this method did change `segments` then it would need to be marked as `mutating` or else it wouldn't compile.

dwieeb: Thanks!

ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
ikeith: Because this is not using a string literal (which is fully defined at compile time) like "server.url"; this string is assembled from variables at runtime. The other way works through some implicit type conversions to create a KeyPath from the literal automatically.

I did consider making a second subscript operator for `JSObject` that accepted a string and converted it into a `KeyPath` but deferred to being more explicit. That felt like it was getting too clever for the sake of syntactic sugar but I don't feel that strongly about it either.

jcesarmobile (approved):

PR #3763, opened 2020-11-05
Update CAPBridgedPlugin.h

183959633:
There's an error on line 33, which should be inherited  CAPPlugin

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgedPlugin.h
+++ b/ios/Capacitor/Capacitor/CAPBridgedPlugin.h
@@ -30,7 +30,7 @@
 [methods addObject:[[CAPPluginMethod alloc] initWithName:@#method_name returnType:method_return_type]]
 
 #define CAP_PLUGIN(objc_name, js_name, methods_body) \
-@interface objc_name : NSObject \
+@interface objc_name : CAPPlugin \
 @end \
 @interface objc_name (CAPPluginCategory) <CAPBridgedPlugin> \
 @end \


ikeith:
Thank you for the PR!

However, this is not actually an error. Explaining the full context for that would require diving deeper into the details of the Obj-C runtime than makes sense to do here.

But the short version is that the interface declaration exists here solely so that the subsequent category declaration will compile. As such, the fact that it inherits from NSObject is the relevant part. This inserted interface declaration is not exposed outside of the `.m` file(s) where the macro is used and will be lost when the final binary is assembled. Inheriting from CAPPlugin instead simply makes the macro more fragile by coupling it tighter to other parts of the code.

183959633:
1)If not inherited from CAPPlugin, in Plugin.m, No "notifyListeners" method found;
2)in Plugin.swift,inherited from CAPPlugin,so I can use this method about "notifyListeners";
I want to use "notifyListeners" method in Plugin.m, what do I do?

ikeith:
If I understand correctly, you're trying to write a plugin in Obj-C instead of Swift? Your file imports and/or declarations are wrong. Your implementation should not be relying on the macro to define your plugin class - that macro should only be invoked in a separate `.m` file. You need 3 files; the general structure is included below. If something's not working, [please use a discussion to ask for help](https://github.com/ionic-team/capacitor/discussions).

MyPlugin.h:
```
#import <Capacitor/Capacitor.h>

@interface MyPlugin : CAPPlugin
- (void)pluginMethod:(CAPPluginCall *)call;
@end
```

MyPlugin.m:
```
#import "MyPlugin.h"
#import <Capacitor/Capacitor.h>
#import <Capacitor/Capacitor-Swift.h>

@implementation MyPlugin
- (void) pluginMethod:(CAPPluginCall *)call {
    // implementation
}
@end
```

MyPluginExport.m:
```
#import <Capacitor/Capacitor.h>

CAP_PLUGIN(MyPlugin, "MyPlugin",
    CAP_PLUGIN_METHOD(pluginMethod, CAPPluginReturnPromise);
)
```

Issue #3723, opened 2020-10-24
bug: Capacitor has so many console messages


Sampath-Lokuge:
Hi,

Why Capacitor has so many console messages? It directly affects the performance of the native build. i.e. poor performance. So it is very bad and against the clean code pattern and practices too. I hope you'll remove all of those in the future version.

![c](https://user-images.githubusercontent.com/5206526/97080027-10178300-1616-11eb-95d0-c5a59529bfee.png)

Ionic:

   Ionic CLI                     : 6.11.11 (C:\Users\Sampath\AppData\Roaming\npm\node_modules\@ionic\cli)
   Ionic Framework               : @ionic/angular 5.3.3
   @angular-devkit/build-angular : 0.1001.2
   @angular-devkit/schematics    : 10.0.8
   @angular/cli                  : 10.1.2
   @ionic/angular-toolkit        : 2.3.3

Capacitor:

   Capacitor CLI   : 2.4.1
   @capacitor/core : 2.4.1

Utility:

   cordova-res : 0.14.0
   native-run  : 1.2.1

System:

   NodeJS : v12.13.1 (C:\Program Files\nodejs\node.exe)
   npm    : 6.12.1
   OS     : Windows 10

EKashpersky:
There's `"hideLogs"` boolean key for capacitor.config.json

https://capacitorjs.com/docs/config#configuration-file

KevinKelchen:
Do we know if `hideLogs` is automatically disabled in release builds or is it just used in debug builds?

Looking at the source code I did not find evidence of it being disabled in release builds.

Sampath-Lokuge:
Yes, that is not working.

PR #3764, opened 2020-11-05
Update CAPLog.swift

183959633:
Security issues, make code more robust

### DIFF
--- a/ios/Capacitor/Capacitor/CAPLog.swift
+++ b/ios/Capacitor/Capacitor/CAPLog.swift
@@ -5,7 +5,9 @@ public class CAPLog {
     public static func print(_ items: Any..., separator: String = " ", terminator: String = "\n") {
         if !self.hideLogs() {
             for (itemIndex, item) in items.enumerated() {
+                #if DEBUG
                 Swift.print(item, terminator: itemIndex == items.count - 1 ? terminator : separator)
+                #endif
             }
         }
     }


ikeith:
Thanks for the PR! While it might be desirable to suppress logs in production, there are several issues to consider:

1) This would be a breaking change for existing applications where the log statements suddenly stop appearing as expected and no way to control it.
2) This change is only for the native iOS code. There is an existing issue about the number of console messages logged from javascript https://github.com/ionic-team/capacitor/issues/3723.
3) There is already an official way to turn off this logging: [the `hideLogs` flag in the configuration](https://capacitorjs.com/docs/reference/config).

aparajita:
I have a WIP console replacement that addresses all of these issues by allowing the user to set a log level (among other enhancements).

183959633:
oh,which one, and how to use?

aparajita:
@183959633 It's a work in progress, hopefully I'll get a chance to publish it next week.

183959633:
OK, thank you for your help. Congratulations!

183959633:
This link is invalid and cannot be found. Please send it again

aparajita:
I didn't send a link, it isn't public yet.

jcesarmobile:
There are multiple possible reasons for wanting logs in release builds, i.e. for release builds send to testers.

As ikeith said, it is already possible to turn it off with hideLogs configuration option, that allows the user to choose when to hide them instead of doing it automatically by release type.

PR #3775, opened 2020-11-07
fix(core): export PermissionState

dwieeb:
We need to export this so plugins can use it.


carlpoole (approved):

JorisDebonnet:
I see this made it into a version 3 alpha, but will it also make it into the 2 release line? I cannot currently install @capacitor/camera without this.

PR #3778, opened 2020-11-08
feat(Android): Refactoring configuration

carlpoole:
This branch seeks to improve the config use in Android similar to #3759.

- Config can now be initialized from capacitor.config.json and programmatically via `ConfigBuilder` for embedded use case
- The existing config getters like `getInt` and `getString` are deprecated. 
  - For plugins, use `getPluginInt` and `getPluginString`.
  - For core config values, use the new direct accessor like `getServerUrl` and `getBackgroundColor`
- Added tests for the Builder and JSON Reading use cases.


android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -9,21 +12,82 @@
...
+            Logger.error("Capacitor Config could not be created. Context must not be null.");
+            return;
         }
+
+        AssetManager assetManager = context.getAssets();
+        webContentsDebuggingEnabled = defaultDebuggable;
+
+        // Load capacitor.config.json
+        loadConfig(assetManager);
+        deserializeConfig();
```
dwieeb, line 58: Should all this be happening in the constructor? Maybe we can take this opportunity to practice some separation of concerns. The iOS side now has a few different classes that separate "loading from a JSON file" and "containing the configuration".

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -52,81 +116,341 @@ private void loadConfig(AssetManager assetManager) {
...
+            getBoolean(configJSON, "android.allowMixedContent", getBoolean(configJSON, "allowMixedContent", allowMixedContent));
+        captureInput = getBoolean(configJSON, "android.captureInput", captureInput);
+        webContentsDebuggingEnabled = getBoolean(configJSON, "android.webContentsDebuggingEnabled", webContentsDebuggingEnabled);
+        hideLogs = getBoolean(configJSON, "android.hideLogs", getBoolean(configJSON, "hideLogs", hideLogs));
+
+        // Plugins
+        pluginConfigurations = getObject(configJSON, "plugins");
+    }
+
+    public boolean html5mode() {
```
dwieeb, line 144: I think this should follow the getter pattern: `getHTML5Mode()` or `isHTML5Mode()`

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -141,7 +465,163 @@ private String getConfigKey(String key) {
...
+        public CapConfig create() {
+            return new CapConfig(this);
+        }
+
+        public Builder pluginConfigurations(JSONObject pluginConfigurations) {
+            this.pluginConfigurations = pluginConfigurations;
+            return this;
+        }
+
+        public Builder html5mode(boolean html5mode) {
```
dwieeb, line 567: I think these should follow the setter pattern: `setHTML5Mode()`, `setServerURL`(), etc.

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -52,81 +116,341 @@ private void loadConfig(AssetManager assetManager) {
...
+    /**
+     * Get a string value for a plugin in the Capacitor config.
+     *
+     * @param pluginId The ID of the plugin used in the config
+     * @param configKey The key of the value to retrieve
+     * @return The value from the config, if exists. Null if not
+     */
+    public String getPluginString(String pluginId, String configKey) {
+        return getPluginString(pluginId, configKey, null);
+    }
```
dwieeb, line 205: I don't think we should add these methods here. Is the intended usage `this.bridge.config.getPluginString()` in the Plugin class? If so, it would be a nicer API if it was just `this.config.getString()` (or `this.pluginConfig.getString()`), etc. We could make a `PluginConfig` class to keep classes smaller and purposeful.

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -52,81 +116,341 @@ private void loadConfig(AssetManager assetManager) {
...
+
+    public String getAppendedUserAgentString() {
+        return appendedUserAgentString;
+    }
+
+    public String getBackgroundColor() {
+        return backgroundColor;
+    }
+
+    public boolean allowMixedContent() {
```
dwieeb, line 176: I would phrase this as `isMixedContentAllowed()`, otherwise the `allow` verb makes it seem like this method allows mixed content in the config. Same with `captureInput()` and `hideLogs()`

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -141,7 +465,163 @@ private String getConfigKey(String key) {
             }
 
             return value;
-        } catch (Exception ex) {}
+        } catch (Exception ignore) {
```
dwieeb, line 468: Can this be narrowed to only JSON exceptions?

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -52,81 +116,341 @@ private void loadConfig(AssetManager assetManager) {
...
+
+    public Boolean getWebContentsDebuggingEnabled() {
+        return webContentsDebuggingEnabled;
+    }
+
+    public boolean hideLogs() {
+        return hideLogs;
+    }
+
+    public JSONObject getPluginConfigurations() {
```
dwieeb, line 192: Little nitpick: this method is returning the `plugins` configuration. `getPluginsConfiguration()` might be a more suitable name.

android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
```diff
@@ -167,6 +175,8 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
 
     @Override
     protected void onActivityResult(int requestCode, int resultCode, Intent data) {
+        super.onActivityResult(requestCode, resultCode, data);
```
dwieeb, line 178: Try to avoid adding unrelated changes to PRs. It makes it difficult to review and creates a less meaningful history. We should strive for single-purpose PRs with small scopes.

android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
```diff
@@ -39,7 +47,7 @@ protected void init(Bundle savedInstanceState, List<Class<? extends Plugin>> plu
      * use {@link BridgeActivity#bridgeBuilder}.
      */
     @Deprecated
-    protected void init(Bundle savedInstanceState, List<Class<? extends Plugin>> plugins, JSONObject config) {
+    protected void init(Bundle savedInstanceState, List<Class<? extends Plugin>> plugins, CapConfig config) {
```
dwieeb, line 50: This is a breaking change. Can you add back the other implementation that accepts JSON and deprecate it similar to how the other ones are?

android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
```diff
@@ -4,15 +4,23 @@
...
 
 public class BridgeActivity extends AppCompatActivity {
 
     protected Bridge bridge;
     protected boolean keepRunning = true;
-    private JSONObject config;
+    private ArrayList<PluginEntry> pluginEntries;
+    private PluginManager pluginManager;
+    private CordovaPreferences preferences;
+    private MockCordovaWebViewImpl mockWebView;
```
dwieeb, line 21: These appear to be unused. (Might have been my fault when `main` was merged in)

jcesarmobile (commented):
CapConfig was created to avoid a breaking change as we had the Config class and users were using it on plugins. Used CapConfig as the name to also match iOS name.
Since this PR changes the constructor, that's also a breaking change.
Not sure if instead of changing the constructor, we rename the class to `InstanceConfiguration` as Ian did on the iOS equivalent https://github.com/ionic-team/capacitor/pull/3759

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -9,21 +12,82 @@
...
+     * @param defaultDebuggable Enable dev mode flag
+     */
+    public CapConfig(Context context, boolean defaultDebuggable) {
+        if (context == null) {
+            Logger.error("Capacitor Config could not be created. Context must not be null.");
+            return;
         }
+
+        AssetManager assetManager = context.getAssets();
+        webContentsDebuggingEnabled = defaultDebuggable;
```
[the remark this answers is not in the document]
jcesarmobile: yeah, context is enough to get the debuggable value like this `(context.getApplicationInfo().flags & ApplicationInfo.FLAG_DEBUGGABLE) != 0;`

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
carlpoole: They dont need to, but it seems Android does that so I'll copy their style

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
carlpoole: > CapConfig was created to avoid a breaking change as we had the Config class and users were using it on plugins. Used CapConfig as the name to also match iOS name.
> Since this PR changes the constructor, that's also a breaking change.
> Not sure if instead of changing the constructor, we rename the class to `InstanceConfiguration` as Ian did on the iOS equivalent #3759

It is a breaking change, but is it expected that 3rd parties would be creating new instances of CapConfig? I updated the code in the core to use the new constructor. If it's not an expected use-case or an embedded/edge case, it might be an acceptable breaking change.

I can see the appeal in naming things similarly to iOS, but I don't think it should be the rule. Java has instances of classes so the name "InstanceConfiguration" feels unnecesarily verbose, but we are likely introducing a new class here to contain the plugin config, so we do need some name work.

carlpoole:
> CapConfig was created to avoid a breaking change as we had the Config class and users were using it on plugins. Used CapConfig as the name to also match iOS name.
> Since this PR changes the constructor, that's also a breaking change.
> Not sure if instead of changing the constructor, we rename the class to `InstanceConfiguration` as Ian did on the iOS equivalent #3759

It is a breaking change, but is it expected that 3rd parties would be creating new instances of CapConfig? I updated the code in the core to use the new constructor. If it's not an expected use-case or an embedded/edge case, it might be an acceptable breaking change.

I can see the appeal in naming things similarly to iOS, but I don't think it should be the rule. Java has instances of classes so the name "InstanceConfiguration" feels unnecesarily verbose, but we are likely introducing a new class here to contain the plugin config, so we do need some name work.

jcesarmobile (approved):
I think both platforms should be as close as possible, specially when two classes do the exact same thing. If not, when somebody not familiar with one of the two platforms needs to do some change in the other platform will have a hard time finding things.
But if you don't agree I'm ok with it.

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
carlpoole: Thanks for the feedback. I have removed the public constructor for CapConfig and replaced it with a method `fromFile` that can be used to create an instance of CapConfig from the JSON config file. 

I moved the JSON file loading to the FileUtils class since it involved reading a file and seemed appropriate there (and its also reused elsewhere in the core).

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
carlpoole: I split the Plugin JSON config into its own class and moved the methods there also. Since the CapConfig and PluginConfig class used similar JSON parsing methods, I centralized those in a new class called JSONUtils that both Config classes may use. Thoughts?

android/capacitor/src/main/java/com/getcapacitor/BridgeFragment.java
```diff
@@ -50,6 +51,10 @@ public void addPlugin(Class<? extends Plugin> plugin) {
         this.initialPlugins.add(plugin);
     }
 
+    public void configure(CapConfig config) {
```
dwieeb, line 54: This looks like a setter, should it be `setConfig()` ?

android/capacitor/src/main/java/com/getcapacitor/FileUtils.java
```diff
@@ -134,6 +137,26 @@ else if ("file".equalsIgnoreCase(uri.getScheme())) {
...
+    /**
+     * Read a plaintext file.
+     *
+     * @param context Used to get access to the asset manager to open the file.
+     * @param fileName The path of the file to read.
+     * @return The contents of the file path.
+     * @throws IOException Thrown if any issues reading the provided file path.
+     */
+    static String readFile(Context context, String fileName) throws IOException {
+        BufferedReader reader = new BufferedReader(new InputStreamReader(context.getAssets().open(fileName)));
```
dwieeb, line 147: `reader.close()` is never called. new code should use try-with-resources: https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html

android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java
carlpoole: After discussing, the breaking change is acceptable since the use case for using this method included Embedded and edge cases. Will document in upgrade guide. Added some javadoc to highlight that an instance of CapConfig is used instead of a JSONObject.

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -1,147 +1,384 @@
...
-        }
+    public boolean isMixedContentAllowed() {
+        return allowMixedContent;
+    }
+
+    public boolean isInputCaptured() {
+        return captureInput;
+    }
+
+    public Boolean getWebContentsDebuggingEnabled() {
```
dwieeb, line 176: Is this supposed to be `Boolean`?

Also, if other boolean getters use `is`, should this as well?

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -1,147 +1,384 @@
...
+            this.captureInput = captureInput;
+            return this;
+        }
+
+        public Builder setWebContentsDebuggingEnabled(boolean webContentsDebuggingEnabled) {
+            this.webContentsDebuggingEnabled = webContentsDebuggingEnabled;
+            return this;
+        }
+
+        public Builder setHideLogs(boolean hideLogs) {
```
dwieeb, line 379: Even though this not the same class, it's referring to the same property, essentially. Should we call it `setLogsHidden()` to match `isLogsHidden()`?

android/capacitor/src/main/java/com/getcapacitor/JSONUtils.java
```diff
@@ -0,0 +1,166 @@
+package com.getcapacitor;
+
+import org.json.JSONArray;
+import org.json.JSONException;
+import org.json.JSONObject;
+
+/**
+ * Helper methods for parsing JSON objects.
+ */
+public class JSONUtils {
```
dwieeb, line 10: Love this class, but should it go in `com.getcapacitor.util`?

android/capacitor/src/main/java/com/getcapacitor/PluginsConfig.java
```diff
@@ -0,0 +1,118 @@
+package com.getcapacitor;
+
+import org.json.JSONObject;
+
+/**
+ * Represents the configuration options for plugins used by Capacitor
+ */
+public class PluginsConfig {
```
dwieeb, line 8: I think we may have misunderstood each other. I was envisioning a reusable `PluginConfig` class that would correspond to a _single_ plugin's configuration within the Capacitor config file. 

Usage would look like this:

```java
config.getPluginConfig(TEST_PLUGIN_NAME).getString("var2")
```

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
dwieeb: Looking good so far--check out my latest comment: https://github.com/ionic-team/capacitor/pull/3778#discussion_r540597530

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -168,14 +167,32 @@ public PluginCall getSavedCall() {
         return this.savedLastCall;
     }
 
+    /**
+     * Get the config options for this plugin.
+     *
+     * @return a config object representing the plugin config options
+     */
+    public PluginConfig getConfig() {
+        return bridge.getConfig().getPluginConfiguration(handle.getId());
```
dwieeb, line 176: Would it make sense to return an empty `PluginConfig` to avoid possible null exceptions?

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -168,14 +167,32 @@ public PluginCall getSavedCall() {
...
+     *
+     * @param key the key for the config value
+     * @return some object containing the value from the config
+     */
+    @Deprecated
     public Object getConfigValue(String key) {
         try {
-            JSONObject plugins = bridge.getConfig().getObject("plugins");
-            if (plugins == null) {
+            PluginConfig pluginConfig = bridge.getConfig().getPluginConfiguration(handle.getId());
```
dwieeb, line 190:
```suggestion
            PluginConfig pluginConfig = getConfig();
```

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
carlpoole: Thanks

dwieeb (approved):

PR #3787, opened 2020-11-10
refactor(Android): deprecate savedLastCall on Plugin class to use Bridge instead, simplify permission result flow

carlpoole:
This PR deprecates the saving of PluginCalls on the Plugin object and instead centralizes the behavior using the `savedCalls` map on the Bridge. Both fields remain functional while deprecated.

Changes:

~~- Plugins overriding `onRequestPermissionsResult` should use `getPermissionCall()` to access the saved call and release the call when the permission response has been handled. `savedLastCall` is no longer used in conjunction with the new `@CapacitorPlugin` annotation and permissions updates.~~
- Plugins overriding `onRequestPermissionsResult` will have the saved call passed in from the Bridge to use.
- Calls in conjunction with permissions are saved in a queue now on the Bridge, which respects the order in which requests were received to the plugin and responds in the appropriate order. Previous `savedLastCall` had potential to be overwritten.
- `handleOnActivityResult` contains a new parameter `lastPluginCall` to reference the bridge `pluginCallForLastActivity` instead of it being saved to the plugin `savedLastCall`


dwieeb:
I think we should deprecate any behavior on `Plugin` that assumes the plugin will only be storing one `PluginCall` at a time, which would encourage plugin developers to use the map on the bridge or store their own calls their own way. This would align the behavior with iOS, which doesn't have these 'convenience' methods on the plugin class.

jcesarmobile (approved):

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -46,10 +46,24 @@
...
+     * {@link com.getcapacitor.Bridge#releaseCall(PluginCall)}
+     */
+    @Deprecated
     protected PluginCall savedLastCall;
 
+    /**
+     * The call ID of a saved plugin call on the Bridge. This can be used to persist a call
+     * between activity lifecycle changes or permission requests.
+     */
+    protected String savedLastCallId;
```
dwieeb, line 65: Do you think this still promotes an undesirable workflow where plugin devs are using a single variable to store multiple call IDs? I am almost of the opinion the plugin dev should be fully responsible for managing their call IDs. What do you think?

carlpoole: We still need to store an ID for the last call for permission to work, but I'm currently refactoring this slightly to indicate its more tied to permissions, like `permissionsCallId`. Do you think that would make a good distinction that this is just used internally for permissions and it shouldn't be used by plugin author?

Im deprecating the `freeSavedCall`, `saveCall` and `getSavedCall` methods in addition to that, and making specific ones tied to permissions so the intent is clearer.

Thoughts?

dwieeb: Yes, perfect!

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -184,7 +184,7 @@ public PluginCall getSavedCall() {
      * @return The saved plugin call
      */
     protected PluginCall getPermissionCall() {
```
dwieeb, line 186: We should document that this dequeues the call ID now. Since this is no longer idempotent, a rename may be warranted as well.

carlpoole:
After meeting with @dwieeb on Friday, this PR evolved. We realized that from moving the saved call to the bridge, we could also simplify some of the updated permissions code and by extension... the requirements plugin developers would have if they override `onRequestPermissionsResult` are also simplified.

These changes handle the `requestPermissions` callback, permission validating, and null check all outside of `onRequestPermissionResult` and the only thing a plugin author would need to do is write the code for the custom permissions handling in the callback. 

https://github.com/ionic-team/capacitor/blob/cdea1b0d83e0597649ab2b55e4e25743c91a7ea6/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L764

There is quite a bit of moving around so there might be a little more cleanup/rearranging to polish this off. Testing went well. I think this is a big improvement but please let me know if there is an edge case that might have been broken by this.

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -713,13 +758,136 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
...
             plugin.getInstance().handleRequestPermissionsResult(requestCode, permissions, grantResults);
         }
     }
 
+    /**
+     * Saves permission states and rejects if permissions were not correctly defined in
+     * the AndroidManifest.xml file.
+     *
+     * Plugins overriding {@link #onRequestPermissionsResult(int, String[], int[])} should call
+     * this method to save permission states correctly.
```
dwieeb, line 789: I don't think this is accurate any longer

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -614,6 +627,38 @@ public void releaseCall(PluginCall call) {
...
+
+        return getSavedCall(savedCallId);
+    }
+
+    /**
+     * Save a call to be retrieved after requesting permissions. Calls are saved in order.
+     *
+     * @param call The plugin call to save.
+     */
+    public void savePermissionCall(PluginCall call) {
```
dwieeb, line 651: Does this need to be public?

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -713,13 +758,136 @@ public void onRequestPermissionsResult(int requestCode, String[] permissions, in
...
+        return true;
+    }
+
+    /**
+     * Helper to check all permissions and see the current states of each permission.
+     *
+     * @since 3.0.0
+     * @return A mapping of permissions to the associated granted status.
+     */
+    public JSObject getPermissionStates(Plugin plugin) {
```
dwieeb, line 847: We should make this protected. If the API needs to be exposed to the plugin developer, we can make a helper in the Plugin class that just does `this.bridge.getPermissionStates(this)` (so the usage would be `this.getPermissionStates()`, much cleaner).

carlpoole: I will do that. We have the new code in permission states that indicates if its a `prompt-with-rational` or hard deny, and I think that could be useful for plugin authors to know from this helper.

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
carlpoole: Not any more, I can also refine the permission requesting helpers on Plugin to save the call so the plugin author won't need to call this at all

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -255,105 +267,87 @@ public boolean hasRequiredPermissions() {
...
+     *
      * @param permission the permission to request
      * @param requestCode the requestCode to use to associate the result with the plugin
      */
+    @Deprecated
     public void pluginRequestPermission(String permission, int requestCode) {
         ActivityCompat.requestPermissions(getActivity(), new String[] { permission }, requestCode);
     }
 
+    public JSObject getPermissionStates() {
```
dwieeb, line 347: This may need to be documented for plugin devs

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -255,105 +267,93 @@ public boolean hasRequiredPermissions() {
...
      * @since 3.0.0
+     * @param call the plugin call
+     * @param permission the permission to request
+     * @param requestCode the requestCode to use to associate the result with the plugin
      */
-    @PluginMethod
-    public void checkPermissions(PluginCall pluginCall) {
-        JSObject permissionsResult = getPermissionStates();
-        pluginCall.resolve(permissionsResult);
+    public void requestPermission(PluginCall call, String permission, int requestCode) {
```
dwieeb, line 277: Is this intended for plugins devs to use?

carlpoole: Yes, its a refactor of the existing `pluginRequestPermission` that includes the savePermissionCall

dwieeb (approved):

jcesarmobile (approved):

Issue #3266, opened 2020-07-14
Automatically import Android plugins


dwieeb:
It is possible to automatically import Android plugins, instead of requiring the user to manually add them to `MainActivity.java`. This will achieve closer platform parity (it is strange manual work is required for one platform but not the other, and personally I forget this all the time). Let's do this for Capacitor 3.

See https://github.com/ionic-team/capacitor/pull/3054

PR #3788, opened 2020-11-10
feat: automatically import Android plugins

dwieeb:
This is an implementation of the **Android Plugin Auto-registration Proposal**.

TODO:
- [x] Support adding additional classes during initialization (e.g. plugins from the app itself)
- [x] Make it backwards compatible with `@NativePlugin` (?)
- [x] Think about error handling when reflection fails

Internally, the CLI creates a `capacitor.plugins.json` file inside `android/app/src/main/assets` (alongside `capacitor.config.json`) which is meant to be shipped in the final Android bundled app.

This file contains a list of classpaths to the Capacitor plugins it finds during `npx cap update android`. It looks like this:

```json5
[
        {
                "pkg": "@capacitor/action-sheet",
                "classpath": "com.capacitorjs.plugins.actionsheet.ActionSheetPlugin"
        },

        ...
]
```

Then, in Java, we use Reflection to load classes via `Class.forName()`.

This simplifies the app developer workflow because they no longer need to use `add()` to add the class names to the list. `MainActivity.java` in most apps can look like this:

```java
public class MainActivity extends BridgeActivity { }
```

If they want to add app plugins, they need to do something like this:

```java
public class MainActivity extends BridgeActivity {
  @Override
    registerPlugin(CustomPlugin.class);
  }
}
```

~depends on https://github.com/ionic-team/capacitor/pull/3866~
~depends on https://github.com/ionic-team/capacitor/pull/3762~
resolves https://github.com/ionic-team/capacitor/issues/3266


android/capacitor/src/main/java/com/getcapacitor/PluginManager.java
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    private JSONArray parsePluginsJSON() throws PluginLoadException {
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(assetManager.open("capacitor.plugins.json")))) {
+            StringBuilder b = new StringBuilder();
+            String line;
+            while ((line = br.readLine()) != null) {
+                b.append(line);
+            }
+            String jsonString = b.toString();
```
carlpoole, line 48: :)

```suggestion
        try (BufferedReader reader = new BufferedReader(new InputStreamReader(assetManager.open("capacitor.plugins.json")))) {
            StringBuilder builder = new StringBuilder();
            String line;
            while ((line = reader.readLine()) != null) {
                builder.append(line);
            }
            String jsonString = builder.toString();
```

carlpoole (approved):

jcesarmobile (approved):

Issue #3789, opened 2020-11-11
bug: Geolocation.getCurrentPosition throws error on new permission dialog on iOS 14.2


abarax:
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 2.4.2
  @capacitor/core: 2.4.2
  @capacitor/android: 2.4.2
  @capacitor/electron: 2.4.2
  @capacitor/ios: 2.4.2

Installed Dependencies:

  @capacitor/android not installed
  @capacitor/cli 2.4.2
  @capacitor/core 2.4.2
  @capacitor/ios 2.4.2
  @capacitor/electron not installed

  Found 0 Capacitor plugins for ios:
[success] iOS looking great! 👌
```

### Platform(s)

iOS 14.2

### Current Behavior

Calling the getCurrentPosition API results in failure during the permission check on iOS 14.2 only.  Seemingly the call times out after 2 seconds of waiting on the permission dialog. Console Output from XCode via emulator:

To Native ->  Geolocation getCurrentPosition 80463459
APP INACTIVE
ERROR MESSAGE:  {"message":"The operation couldn’t be completed. (kCLErrorDomain error 1.)","errorMessage":"The operation couldn’t be completed. (kCLErrorDomain error 1.)"}
⚡️  [error] - ERROR Error: Uncaught (in promise): Error: The operation couldn’t be completed. (kCLErrorDomain error 1.)

<img width="1116" alt="CleanShot 2020-11-11 at 17 47 37@2x" src="https://user-images.githubusercontent.com/124785/98783892-01b6cd00-2446-11eb-8d50-96afcfed6ad8.png">

### Expected Behavior

getCurrentPosition should wait until an option has been selected on the permission dialog.

### Code Reproduction

Basic reproduction. Simply hit the 'test' button on tab 1 and you will see the above error in the xcode console.
https://github.com/abarax/capacitor-bug

Only Code added is here:

https://github.com/abarax/capacitor-bug/blob/main/src/app/explore-container/explore-container.component.ts

https://github.com/abarax/capacitor-bug/blob/main/src/app/explore-container/explore-container.component.html

To run:

> npm run build
> npx cap sync
> npx cap open ios

Run via XCode using iOS 14.2 simulator

zsyphon:
@abarax Have you made any progress on this issue? I am getting the same error.

abarax:
[...]

PR #3802, opened 2020-11-13
fix(ios): Don't get location if permission is not determined

jcesarmobile:
Don't try to get the location when the permission status is `.notDetermined`, otherwise the failure delegate gets called before the user reacts to the permissions prompt.

closes https://github.com/ionic-team/capacitor/issues/3789


zsyphon:
@ikeith Is it likely this will be merged soon?

ikeith (approved):

PR #3797, opened 2020-11-12
chore(cli): bundle CLI

dwieeb:
previously: https://github.com/ionic-team/capacitor/pull/3786

Issue #3290, opened 2020-07-17
Archive CLI assets


dwieeb:
npm can do weird things when publishing files (see https://github.com/ionic-team/capacitor/pull/3285). This isn't the only place where we rename/move files when applying the templates. If we ship tar.gz archives of the assets, it'll be cleaner and more reliable (probably).

IT-MikeS:
So the idea would be to take a template, let's say the android one, and before release to npm you would archive it, then ship the archive while the template folder as it is now gets added to the .npmignore file?

dwieeb:
The CLI assets (`android-template/`, `ios-template/`, `app-template/`, `capacitor-cordova-android-plugins`, `capacitor-cordova-ios-plugins`, and `plugin-template`) are copied into `cli/assets/` before it's published. This change would ship tar.gz archives instead.

IT-MikeS:
Okay, and then on the users system it would just do an extract on `add` which will use an npm package for the function, or roll one into the cli common functions lib?

dwieeb:
The tar.gz archives would still be in the `@capacitor/cli` package.

IT-MikeS:
Yeah I meant what is the idea for extraction, a built in extract lib or an npm package?

dwieeb:
tar: https://www.npmjs.com/package/tar

top1st:
no such file or directory, stat node_modules\@capacitor\cli\assets\capacitor-cordova-android-plugins.tar.gz

Why I am getting this problem?

PR #3803, opened 2020-11-15
refactor(cli): shuffle imports to reduce startup time

dwieeb:
Using the following code after config is loaded in the `run()` function of `cli/src/index.ts`:

```js
console.log(Object.keys(require.cache).length);
```

as well as `time npx cap --version` to test startup time.

**Benchmarks**

These are the numbers I'm getting on my machine (MacBook Pro (16-inch, 2019) / 2.4 GHz 8-Core Intel Core i9 / 32 GB 2667 MHz DDR4):

`main`
- 280 modules in cache
- 0.42s startup
- package: 141 files (500.7 kB, 829.7 kB unpacked)

`cli-bundle` (https://github.com/ionic-team/capacitor/pull/3797)
- 241 modules in cache
- 0.40s startup
- package: 102 files (497.3 kB, 807.7 kB unpacked)

`import-refactor` (this PR)
- 100 modules in cache
- 0.32s startup 
- package: 145 files (501.1 kB, 832.9 kB unpacked)

**Results**

This has ~100ms faster startup than `cli-bundle` while maintaining the simple build process and useful dist stack traces and while adding little to no maintenance overhead. To reduce the number of files published we can implement https://github.com/ionic-team/capacitor/issues/3290 (wouldn't reduce the packaged size much, but would offer additional benefits).

alternative to https://github.com/ionic-team/capacitor/pull/3797

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -1,21 +1,8 @@
 import program from 'commander';
 
 import c from './colors';
-import { logFatal } from './common';
 import { loadConfig } from './config';
-import { output } from './log';
-import { addCommand } from './tasks/add';
-import { copyCommand } from './tasks/copy';
-import { createCommand } from './tasks/create';
-import { doctorCommand } from './tasks/doctor';
-import { initCommand } from './tasks/init';
-import { listCommand } from './tasks/list';
-import { newPluginCommand } from './tasks/new-plugin';
-import { openCommand } from './tasks/open';
-import { runCommand } from './tasks/run';
-import { serveCommand } from './tasks/serve';
-import { syncCommand } from './tasks/sync';
-import { updateCommand } from './tasks/update';
+import { output, logFatal } from './log';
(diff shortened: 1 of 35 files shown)


cli/src/index.ts
```diff
@@ -1,21 +1,8 @@
...
-import { doctorCommand } from './tasks/doctor';
-import { initCommand } from './tasks/init';
-import { listCommand } from './tasks/list';
-import { newPluginCommand } from './tasks/new-plugin';
-import { openCommand } from './tasks/open';
-import { runCommand } from './tasks/run';
-import { serveCommand } from './tasks/serve';
-import { syncCommand } from './tasks/sync';
-import { updateCommand } from './tasks/update';
+import { output, logFatal } from './log';
```
adamdbradley, line 5: These changes are awesome 😎

adamdbradley (approved):

PR #3818, opened 2020-11-17
refactor(cli): consolidate paths to config

dwieeb:
We want to rely on the config for file paths as much as possible to reduce hard-coding and open up the CLI to opportunity for configuring these paths in the future.

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -154,9 +154,12 @@ async function loadAndroidConfig(
   const name = 'android';
   const platformDir = extConfig.android?.path ?? 'android';
   const platformDirAbs = resolve(rootDir, platformDir);
-  const webDir = 'app/src/main/assets/public';
-  const resDir = 'app/src/main/res';
-  const buildOutputDir = 'app/build/outputs/apk/debug';
+  const appDir = 'app';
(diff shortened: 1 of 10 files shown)


ikeith (approved):

cli/src/config.ts
```diff
@@ -154,9 +154,12 @@ async function loadAndroidConfig(
   const name = 'android';
   const platformDir = extConfig.android?.path ?? 'android';
   const platformDirAbs = resolve(rootDir, platformDir);
-  const webDir = 'app/src/main/assets/public';
-  const resDir = 'app/src/main/res';
-  const buildOutputDir = 'app/build/outputs/apk/debug';
+  const appDir = 'app';
+  const srcDir = `${appDir}/src/main`;
```
jcesarmobile, line 158: Why is it called `srcDir ` but points to `main` dir instead of to `src`?
We should probably have both, with appropriate names.

dwieeb: I can do that, I just saw that the `main` source set appears to be where everything is (besides tests).

jcesarmobile: in regular apps yes, but there can also be `debug`, `release` and other folders inside `src`, so might need to have a src path and a main path

PR #3822, opened 2020-11-19
feat(android): modified plugin annotation format for multi-permissions and empty (auto-grant)

carlpoole:
Allows for the following annotation formats for permissions:

Multiple permissions can be grouped within the same `@Permission` sub-annotation and alias.

```
@CapacitorPlugin(
    name = "Geolocation",
    permissions = {
        @Permission(permission = { Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION }, alias = "location")
    },
    permissionRequestCode = GeolocationPlugin.GEOLOCATION_REQUEST_PERMISSIONS
)
```

Permission element can be omitted to define a permission that should be auto-granted. Useful when another platform has the permission but Android does not. 

```
@CapacitorPlugin(
    name = "Notifications",
    permissions = {
        @Permission(permission = {}, alias = "notifications")
    }
)
```

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
@@ -277,6 +277,8 @@ public class Plugin {
         JSObject permissionsResults = new JSObject();
         CapacitorPlugin annotation = handle.getPluginAnnotation();
         for (Permission perm : annotation.permissions()) {
+            // If a permission is defined with no permission constants, return "granted" for it.
+            // Otherwise, get its true state.
             if (perm.permission().length == 0 || (perm.permission().length == 1 && perm.permission()[0].isEmpty())) {
                 String key = perm.alias();
                 if (!key.isEmpty()) {
@@ -511,6 +513,8 @@ public class Plugin {
             if (providedPermsList == null || providedPermsList.isEmpty()) {
                 HashSet<String> permsSet = new HashSet<>();
                 for (Permission perm : annotation.permissions()) {
+                    // If a permission is defined with no permission constants, separate it for auto-granting.
+                    // Otherwise, it is added to the list to be requested.
                     if (perm.permission().length == 0 || (perm.permission().length == 1 && perm.permission()[0].isEmpty())) {
                         if (!perm.alias().isEmpty()) {
                             autoGrantPerms.add(perm.alias());
--- a/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
@@ -13,6 +13,7 @@ import com.getcapacitor.annotation.CapacitorPlugin;
 import com.getcapacitor.annotation.Permission;
 import com.getcapacitor.util.PermissionHelper;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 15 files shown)


android/capacitor/src/main/java/com/getcapacitor/annotation/Permission.java
```diff
@@ -9,11 +9,11 @@
...
     /**
-     * The Android string for the permission.
-     * Eg: Manifest.permission.ACCESS_COARSE_LOCATION
-     *     or "android.permission.ACCESS_COARSE_LOCATION"
+     * An array of Android permission strings.
+     * Eg: {Manifest.permission.ACCESS_COARSE_LOCATION}
+     *     or {"android.permission.ACCESS_COARSE_LOCATION"}
      */
-    String permission() default "";
+    String[] permission() default {};
```
dwieeb, line 16: Thoughts on pluralizing it now that it's an array?

```suggestion
    String[] permissions() default {};
```

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -489,15 +509,25 @@ public void requestPermissions(PluginCall call) {
         } else {
             // If call was made without any custom permissions, request all from plugin annotation
             if (providedPermsList == null || providedPermsList.isEmpty()) {
-                perms = new String[annotation.permissions().length];
-                for (int i = 0; i < perms.length; i++) {
-                    perms[i] = annotation.permissions()[i].permission();
+                HashSet<String> permsSet = new HashSet<>();
+                for (Permission perm : annotation.permissions()) {
+                    if (perm.permission().length == 0 || (perm.permission().length == 1 && perm.permission()[0].isEmpty())) {
```
dwieeb, line 514: Can you make a code comment on why this if-test is necessary?

android/capacitor/src/main/java/com/getcapacitor/annotation/Permission.java
carlpoole: It makes sense because it can be one or many but I hesitate because the parent element is called "permissions" which is made up of one or many `@Permission` annotations, which then can contain one or many "permission" manifest constants. Im thinking of renaming it to something like "constants" or "strings" but that sounds awfully generic

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -489,15 +511,27 @@ public void requestPermissions(PluginCall call) {
         } else {
             // If call was made without any custom permissions, request all from plugin annotation
             if (providedPermsList == null || providedPermsList.isEmpty()) {
-                perms = new String[annotation.permissions().length];
-                for (int i = 0; i < perms.length; i++) {
-                    perms[i] = annotation.permissions()[i].permission();
+                HashSet<String> permsSet = new HashSet<>();
+                for (Permission perm : annotation.permissions()) {
+                    // If a permission is defined with no permission constants, separate it for auto-granting.
```
dwieeb, line 516: I'm still confused as to why this part is necessary: `(perm.permission().length == 1 && perm.permission()[0].isEmpty())`

It is handling this case?
```
@Permission(permission = {""}, alias = ...
```

carlpoole: yes or this valid definition

```
@Permission(permission = "", alias = ...
```

which on the back end translates to an array containing one empty element

android/capacitor/src/main/java/com/getcapacitor/annotation/Permission.java
dwieeb: I don't mind `strings`, because that's what they're called in Android and we already have the context of permissions given the `@Permission` annotation. 

```java
@Permission(strings = { Manifest.permission.WRITE_EXTERNAL_STORAGE, Manifest.permission.READ_EXTERNAL_STORAGE }, alias = "publicStorage")
```

^ This looks good to me. 😁

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
dwieeb: Let's switch to Kotlin.

jcesarmobile (approved):

dwieeb (approved):

PR #3832, opened 2020-11-20
chore(ios): mark modulePrint and alert as public

dwieeb:
otherwise we get "x is inaccessible due to 'internal' protection level"

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeProtocol.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeProtocol.swift
@@ -70,13 +70,13 @@ import WebKit
  */
 extension CAPBridgeProtocol {
     // variadic parameters cannot be exposed to Obj-C
-    func modulePrint(_ plugin: CAPPlugin, _ items: Any...) {
+    public func modulePrint(_ plugin: CAPPlugin, _ items: Any...) {
         let output = items.map { String(describing: $0) }.joined(separator: " ")
         print(message: output, for: plugin)
     }
 
     // default arguments are not permitted in protocol declarations
-    func alert(_ title: String, _ message: String, _ buttonTitle: String = "OK") {
+    public func alert(_ title: String, _ message: String, _ buttonTitle: String = "OK") {
         showAlertWith(title: title, message: message, buttonTitle: buttonTitle)
     }
 


ikeith:
This is a subset of #3831

jcesarmobile:
I think we can remove modulePrint, I don’t see the difference between it and using CAPLog and only Splash class uses it

ikeith:
More than just splash screen uses `modulePrint`; I encountered this same problem with the camera plugin which is why I made #3831. Yes, it's a small wrapper that just calls through to CAPLog, but it does enforce some consistent formatting so it seems reasonable to leave it in.

jcesarmobile:
Well, it doesn't matter if it's just Splash or if it's Splash + Camera, all the rest of plugins just use CAPLog, which is the class that prints logs.

Also, modulePrint used to internally call CAPLog, but not anymore, which means those messages can't be disabled with hideLogs preference.
If it's about the formatting we could have a similar function on CAPLog, but having something that prints messages outside CAPLog makes no sense to me

ikeith:
`modulePrint` exists for backwards compatibility. [It calls ` print(message:, for:)` which calls `CAPLog`](https://github.com/ionic-team/capacitor/blob/264a96d5c68a026263482553138a8d624535b5fe/ios/Capacitor/Capacitor/CapacitorBridge.swift#L542) so it is subject to the logging flag.

jcesarmobile:
Ah, I saw it was calling print and though it was regular print, not bridge print.

Anyway, I don't see "Backward compatibility" a compelling reason for leaving it there when there are better alternatives. We have CAPLog that is static and can be used anywhere without depending on the bridge.

`modulePrint` is only available on iOS, depends on having access to the bridge, and in the end it just calls CAPLog.

dwieeb:
If `modulePrint` wasn't public before, we should be able to do whatever we want, right? FWIW I'd rather not make this public.

ikeith:
[It was `public` before](https://github.com/ionic-team/capacitor/blob/de1eac8b9abba59eacacfc6c4e1818fe0adf8b88/ios/Capacitor/Capacitor/CAPBridge.swift#L373); it was made `internal` by accident when I refactored the interface. That said, I have no idea if any other plugin ever used it. We could just make it deprecated now, though, and encourage people to use `CAPLog` directly.

ikeith:
Since the consensus is to remove `modulePrint, I have deprecated the method in #3831. Since it was public, that allows a smoother migration in case it was being used.

PR #3842, opened 2020-11-21
feat(ios): add path utilities to bridge

dwieeb:
Usable by plugins that deal with file paths, such as Camera and Notifications.


ios/Capacitor/Capacitor/CAPFile.swift
```diff
@@ -11,7 +11,7 @@ public class CAPFile {
  * disk, in a bundle, or in another location.
  */
 @objc public class CAPFileManager: NSObject {
-    static func get(path: String) -> CAPFile? {
+    public static func get(path: String) -> CAPFile? {
```
ikeith, line 14: Can this just return a `URL`? CAPFile is a pretty useless wrapper since it only has one property. In addition, using that object means that this method can't be exported to Obj-C because `CAPFile` is Swift-only.

ikeith, line 14: I think this would benefit from a more descriptive name. This isn't getting "getting" anything; CAPFileManager doesn't store data. This is transforming the parameter that gets passed in but it's impossible to tell what it is expected do to from the name.

ikeith (changes requested):
Since it's an internal implementation detail, I don't think it's critical. But the resolvers would be much cleaner if they were made into an enum instead of a collection of classes.

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -537,6 +537,57 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+            url = URL(fileURLWithPath: resourcePath + "/public" + inputURL.path)
+        case "file":
+            url = inputURL
+        default:
+            return nil
+        }
+
+        guard manager.fileExists(atPath: url.absoluteString.replacingOccurrences(of: "file://", with: "")) else {
+            return nil
+        }
```
dwieeb, line 572: @ikeith This is questionable. I kept the old behavior, but I'm thinking we should remove this check and let it fail whenever the file is actually used, vs when we transform the path.

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -537,6 +537,60 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+
+        return url
+    }
+
+    /**
+     * Translate a file URL for native iOS into a URL to load in the web view.
+     */
+    public func portablePath(fromLocalURL localURL: URL?) -> String? {
```
dwieeb, line 580: @ikeith What do you think about this also returning a proper `URL`, and then documenting that the plugin can use `url.absoluteString`?

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -537,6 +537,60 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+        }
+
+        let manager = FileManager.default
+        let url: URL
+
+        switch inputURL.scheme {
+        case "res":
+            let bundle = Bundle.main
+            let resourcePath = bundle.resourcePath!
+            url = URL(fileURLWithPath: resourcePath + "/public" + inputURL.path)
```
ikeith, line 563: URL's path manipulation methods are more robust than string modifications. Also, force-unwraps are bad.
```suggestion
            guard let resourcePath = Bundle.main.resourcePath else {
                return nil
            }
            url = URL(fileURLWithPath: resourcePath).appendingPathComponent("public").appendingPathComponent(inputURL.path)
```

ios/Capacitor/Capacitor/CapacitorBridge.swift
ikeith: I would be in favor of removing this. If we did keep it, though, it should simply pass `url.path` to `fileExist(atPath:)` instead of getting the absolute string and modifying it.

ios/Capacitor/Capacitor/CapacitorBridge.swift
ikeith: I like that better.

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -537,6 +537,60 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+            return nil
+        }
+
+        guard let portableSchemeAndHost = self.localUrl else {
+            return nil
+        }
+
+        let url = URL(string: portableSchemeAndHost + CapacitorBridge.fileStartIdentifier + inputURL.path)
+
+        return url?.absoluteString
```
ikeith, line 591: Same reasoning.
```suggestion
        guard let inputURL = localURL else {
            return nil
        }
        guard let urlString = self.localUrl, let portableURL = URL(string: urlString) else {
            return nil
        }
        return portableURL.appendingPathComponent(CapacitorBridge.fileStartIdentifier).appendingPathComponent(inputURL.path).absoluteString
```

ikeith (changes requested):

ikeith (approved):

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -533,6 +533,51 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+
+        let url: URL
+
+        switch inputURL.scheme {
+        case "res":
+            guard let resourcePath = Bundle.main.resourcePath else {
+                return nil
+            }
+
+            url = URL(fileURLWithPath: resourcePath).appendingPathComponent("public").appendingPathComponent(inputURL.path)
```
ikeith, line 560: To support the embedded case, this should use the configuration's web app location instead of the native app bundle/public path.
```suggestion
            url = config.appLocation.appendingPathComponent(inputURL.path)
```

ikeith (approved):

PR #3847, opened 2020-11-23
chore: Remove app schemes from the templates

jcesarmobile:
Users should be adding this themselves if using App plugin and want to launch the app from a custom url scheme.


dwieeb (commented):
Should we also remove `custom_url_scheme` from `android-template/app/src/main/res/values/strings.xml`?

jcesarmobile:
I didn't because the value is written by the CLI on `npx cap add android` and it's easier to document that they need to add the lines I've removed in the `AndroidManifest.xml` than adding those and also creating a new entry in `strings.xml`.

But I can remove it and make the CLI not write it.

dwieeb:
I think it's good, just checking

dwieeb (approved):

PR #3858, opened 2020-11-25
feat(cli): extendable plugin configuration types

imhoffd:
See https://github.com/ionic-team/capacitor-plugins/pull/115 for examples of plugin changes necessary for adding to the plugin config types.

App developers will need to add `<reference>` lines at the beginning of their config files to pull in the config types from each plugin.

Example `capacitor.config.ts`:

```ts
/// <reference types="@capacitor/keyboard" />

import type { CapacitorConfig } from '@capacitor/cli';

    appId: "com.capacitorjs.testapp",
    appName: "capacitor-testapp",
    bundledWebRuntime: false,
    webDir: "build",
    plugins: {
        Keyboard: {
            // TODO
        }
    },
    cordova: {}
}

export = config;
```

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -61,17 +61,17 @@ public class CAPBridgeViewController: UIViewController, CAPBridgeDelegate, WKUID
         // get the web view
         let assetHandler = CAPAssetHandler()
         assetHandler.setAssetPath(startPath)
-        webView = prepareWebView(with: configuration, assetHandler: assetHandler)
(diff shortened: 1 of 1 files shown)


jcesarmobile (commented):
If plugins are a separate thing, does it make sense to have types for their configurations here?
What about 3rd party plugins?
What if a plugin adds a new config option and can’t be used until a new CLI is released?

dwieeb:
> What about 3rd party plugins?

I was thinking about this, and maybe we can do something like `declare module '@capacitor/cli'` with the types in each plugin. I can play around with this idea. I mostly made this PR to split out these types from the allowlist PR.

jcesarmobile (approved):

PR #3863, opened 2020-11-25
fix(cli/cordova): replace AndroidManifest.xml variables with default value

jcesarmobile:
cordova-plugin-gelocation added a install variable on latest release that adds a variable in AndroidManifest.xml and it prevents Capacitor apps to build because it's not recognized.
This PR replaces the variable with the default value.


dwieeb (approved):

DanielOrtegaConesa:
It is not ideal, but as a temporary "solution", I have added the following script to my npm commands
`    
"android": "npx cap sync android && sed -i 's/$GPS_REQUIRED/true/' android/capacitor-cordova-android--lugins/src/main/AndroidManifest.xml"
`
So instead of `npx cap sync android` im usin npm run android

PR #3875, opened 2020-11-29
fix(ios): share message handler between webview and bridge

dwieeb:
This fixes the “App is not implemented on ios” error in main.


ikeith (approved):
Whoops 😅

PR #3880, opened 2020-11-30
chore: update husky to v5

### LINKED ISSUE #3839: bug: git hooks not setup after npm install when using npm 7.x
### Capacitor Version
Current `main`

### Platform(s)
Cli

### Current Behavior
The husky hooks are not set up

### Expected Behavior
The husky hooks should be set up

### Code Reproduction
- Clone `main`
- Run `npm install` when an npm version >= 7.0
- `ls -al ./.git/hooks`

Only the sample hooks will be there

### Other Technical Details

This seems to be related to https://github.com/npm/cli/issues/1905 and https://github.com/typicode/husky/issues/776

The fix is probably updating to Husky 5.0 which was released 4 days ago. npm 6 and below are unaffected so this is not a blocking issue.

thomasvidas:
# CHANGES
- Updated to husky 5.0.4 which fixes npm 7 hook issues
- Added `.husky` folder which contains the hooks
- Added `husky install` to the postinstall so it will be setup 
- Added `pinst` package which doesn't run `husky install` for devs that are using capacitor (see [here](https://typicode.github.io/husky/#/?id=install) for details)

My bad rebase brought in the previous commit from main 😅  but that will be fixed on merge. Also if lint is running twice, run `rm -rf .git/hooks` to delete the v4 husky hooks and then run `npx husky install` to setup the new ones

# LICENSE CHANGE IN DEPENDENCY
Husky v5 also has a license change from MIT to [The Parity Public License 7.0.0](https://raw.githubusercontent.com/typicode/husky/next/LICENSE-PARITY). Because Capacitor is open source and MIT Licensed, I don't believe this is an issue, but we should double check before merging this in

Fixes #3839


package.json
```diff
@@ -7,7 +7,9 @@
     "prettier": "prettier \"**/*.{css,html,java,js,ts}\"",
     "eslint": "eslint . --ext ts",
     "swiftlint": "node-swiftlint",
-    "postinstall": "lerna bootstrap"
+    "postinstall": "husky install && lerna bootstrap",
+    "prepublish": "pinst --disable",
+    "postpublish": "pinst --enable"
   },
   "husky": {
```
dwieeb, line 14: Can this block be removed, then?

thomasvidas: Yes! You are correct. When I was testing, my `npx husky` was v4 but not that I cleared the cache I can confirm we don't need this anymore

package.json
```diff
@@ -8,14 +8,11 @@
     "prettier": "prettier \"**/*.{css,html,java,js,mjs,ts}\"",
     "eslint": "eslint . --ext ts",
     "swiftlint": "node-swiftlint",
-    "postinstall": "lerna bootstrap",
+    "postinstall": "husky install && lerna bootstrap",
+    "prepublish": "pinst --disable",
+    "postpublish": "pinst --enable"
     "version": "npm run sync-peer-dependencies"
```
dwieeb, line 14:
```suggestion
    "postpublish": "pinst --enable",
    "version": "npm run sync-peer-dependencies"
```

.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname $0)/_/husky.sh"
+
+npm run lint
```
dwieeb, line 4:
```suggestion
npm run lint

```

dwieeb: This adds newline at EOF

package.json
thomasvidas: nice catch. Bad merge 👍

.husky/pre-commit
thomasvidas: I wonder if there's a way we can automate this if we need to add new hooks? 🤔 This is an auto generated file

package.json
```diff
@@ -8,14 +8,11 @@
     "prettier": "prettier \"**/*.{css,html,java,js,mjs,ts}\"",
     "eslint": "eslint . --ext ts",
     "swiftlint": "node-swiftlint",
-    "postinstall": "lerna bootstrap",
+    "postinstall": "husky install && lerna bootstrap",
+    "prepublish": "pinst --disable",
+    "postpublish": "pinst --enable",
```
dwieeb, line 13: We actually don't need `pinst` because our root package is private.

dwieeb (approved):

PR #3906, opened 2020-12-05
chore(android): setup linter

dwieeb:
Using a `lint-baseline.xml` file, we can enable linting on `@capacitor/android` today. The `lint` task will also now run in CI.

I ignored `GradleDependency` using a `lint.xml` file because it seems like it would break our build whenever new versions come out. The other rules are static and related to the code itself.

Lint fixes can be a part of other PR(s).

Resources:
- https://developer.android.com/studio/write/lint
- https://developer.android.com/reference/tools/gradle-api/4.2/com/android/build/api/dsl/LintOptions
- https://sites.google.com/a/android.com/tools/tips/lint-checks


carlpoole (approved):
Nice

PR #3911, opened 2020-12-08
feat(Android): Provide override hook for permissions

ikeith:
This branch changes the automatic permission handling on android to let Plugins provide an override state at runtime. This is needed in the case of the Camera plugin to conditionally skip the camera permissions depending on the contents of the manifest.

Draft pending testing.


dwieeb:
I have to ask the question now before we enter beta: Are we trying to do too much on Android for permissions? Should we roll back the implicit behavior and instead just define a contract: the Plugin class would have `requestPermissions` and `checkPermissions`, but they'd run `call.unimplemented()` unless overridden. We could define helpers to mitigate the implementation details for the plugin author.

dwieeb:
To get Camera functional for beta, LGTM

I think we should circle back and reevaluate, though.

dwieeb (approved):

ikeith:
Closing in preference of https://github.com/ionic-team/capacitor/pull/3939

PR #3914, opened 2020-12-08
chore(android): Move Splash code out of Bridge

jcesarmobile:


android/capacitor/src/main/java/com/getcapacitor/plugin/SplashScreen.java
```diff
@@ -9,6 +10,12 @@
 @NativePlugin
 public class SplashScreen extends Plugin {
 
+    public void load() {
+        if (getContext() instanceof BridgeActivity) {
+            Splash.showOnLaunch((BridgeActivity) getContext(), bridge.getConfig());
```
dwieeb, line 15: `Splash.showOnLaunch()` can just take a normal `Activity`, then this cast/import isn't needed

jcesarmobile: yeah, just moving things out from the Bridge class to the SplashScreen class, but at the moment Splash.java requires a BridgeActivity, so didn't want to change that on this PR, will change it on the new plugin. In fact, it would use `AppCompatActivity` instead of `Activity`/`BridgeActivity`

jcesarmobile: In fact, this PR doesn't really need to get merged as next PR will be to remove the Splash plugin entirely, including Splash.java, but wanted you to take a look first and see if you could still see the "blink" between screens.

carlpoole: This reverts the flicker fix because the splash screen needs to be visible while the web view is initialized

dwieeb: Sure, I can try!

jcesarmobile: > This reverts the flicker fix because the splash screen needs to be visible while the web view is initialized

Have you tried it? I don't see the flicker with this change.

I don't think it reverts the fix, because before it was called on BridgeActivity after the whole bridge was created, now it gets called during the bridge creation, on registerAllPlugins method on the bridge, that is still called before the `loadWebView`, which is the code that initializes and loads the webview.

https://github.com/ionic-team/capacitor/blob/1d3dd0ae1677d1f56a9ce0efa847426a2b43cca5/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L170-L172

carlpoole: it happens in `this.initWebView();` and it was only reproducible on certain devices where it was more visible. I will check.

jcesarmobile: I put the call in BridgeActivity to make sure I could reproduce on one of my devices (the slower one) and then moved it to SplashPlugin class to verify it didn't happen anymore. But if you have a slower device, please, test.

`initWebView` just add a few settings, the method that took more time is `loadWebView`. `registerAllPlugins` can also take some time if there are a lot of plugins. At the moment, since SplashScreen is the only core plugin, it gets registered the first one, so there is no delay, once we move it out, things might slow down as it would be one of the last ones because the automatic register is alphabetical.
Maybe we should make the CLI add some sort of priority for some plugins on the `capacitor.plugins.json` based on some value in the plugin's package.json?
cordova had an `onload` attribute that gives plugins priority.

carlpoole: Ok my slow phone finally booted, and you are right. I put it back to before the fix to reproduce and then the new method isn't flickering 👍🏼👍🏼

The priority idea sounds good, or some way to specify or hook in your plugin to be loaded prior to web view setup or something.

jcesarmobile: I've moved the `this.registerPlugin(SplashScreen.class);` so it's the last one instead of the first one and I still don't see a flicker, so I think we are good and don't need the priority for now, but something to have in mind in case some user complains in the future if they have a lot of plugins that do a lot of work on load method.

PR #3915, opened 2020-12-08
feat(android): support plugin init

dwieeb:
To support an alternative to https://github.com/ionic-team/capacitor/pull/3914

Plugins can add the following static method which will get called during plugin initialization:

```java
public static void init(Bridge bridge) {
    // do things
}
```

The splash screen plugin's would look like this:

```java
public static void init(Bridge bridge) {
    Splash.showOnLaunch(bridge.getActivity(), bridge.getConfig());
}
```

This is a "breaking change" in that it restricts `PluginHandle` and `PluginMethodHandle` to package-only visibility, not `public`.


dwieeb:
Since this is not currently necessary for Splash Screen, we are going to sideline it for now.

PR #3918, opened 2020-12-08
refactor(cli): pack/extract assets

dwieeb:
This removes the gitignore renaming hack (no longer necessary) and ensures the CLI assets are no longer subject to weird npm publishing rules.

~Before~ After:

```
npm notice package size:  505.2 kB
npm notice unpacked size: 649.5 kB
npm notice total files:   55
```

~After~ Before:

```
npm notice package size:  504.2 kB
npm notice unpacked size: 843.7 kB
npm notice total files:   161
```

resolves https://github.com/ionic-team/capacitor/issues/3290


package.json
```diff
@@ -19,11 +19,13 @@
...
     "@ionic/swiftlint-config": "^1.1.2",
+    "@types/tar": "^4.0.4",
     "eslint": "^7.11.0",
     "husky": "^5.0.4",
     "lerna": "^3.22.1",
     "prettier": "~2.2.0",
     "prettier-plugin-java": "~1.0.0",
-    "swiftlint": "^1.0.1"
+    "swiftlint": "^1.0.1",
+    "tar": "^6.0.5"
```
jcesarmobile, line 29: should this and the types/tar packages be here on the root? they are already on the cli's package.json

jcesarmobile (commented):
Added a comment.

Also, you are saying that with the tar we now have more files unpacked (the triple) that were missed when doing a copy?

package.json
dwieeb: Technically yes because `scripts/pack-cli-assets.mjs` also imports tar.

dwieeb:
> Also, you are saying that with the tar we now have more files unpacked (the triple) that were missed when doing a copy?

No, that's the final number of files in the published npm package. 

I checked that the number of assets that were being copied is identical to the number now being packed into the tar.gz files.

package.json
jcesarmobile: ah, feels weird to have the script outside the cli folder, but considering all the other code shared it's probably ok

jcesarmobile:
Not sure if you got confused, but the "before" results are after doing this change, and the "after" results are before this change, so makes sense that there are less files now after the change since the templates are packed.

dwieeb:
Oh lol

jcesarmobile (approved):

Issue #3088, opened 2020-06-11
bug: Keyboard overlaps ion-input when used with StatusBar.overlay as true


manishoswal:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 2.2.0

  @capacitor/core: 2.2.0

  @capacitor/android: 2.2.0

  @capacitor/electron: 2.2.0

  @capacitor/ios: 2.2.0

Installed Dependencies:

  @capacitor/ios not installed

  @capacitor/electron not installed


  @capacitor/cli 2.1.2

  @capacitor/core 2.1.2

  @capacitor/android 2.2.0

[success] Android looking great! 👌```

### Platform(s)
<!--
List the platforms that this bug affects.
-->

Android

### Current Behavior
<!--
Describe how the bug manifests. Be specific.
-->

The android keyboard overlay the input on the screen when the StatusBar.overlay is set to true

### Expected Behavior
<!--
Describe what the behavior should be.
-->

The screen should scroll to the focused input. 

### Example Code or Repo
<!--
If you are able to illustrate the bug with an example, please provide sample code snippets or a sample application using a public repo (preferred).
-->

Go to any screen with ion-inputs
Click on the input sometimes
See error

![image](https://user-images.githubusercontent.com/2525704/84349544-2b3ca580-abd5-11ea-91d9-40f25803ebd9.png)

### Code Reproduction
<!--
To isolate the cause of the problem, we ask you to provide a minimal sample application that demonstrates the issue.
For full instructions, see: https://github.com/ionic-team/capacitor/blob/master/CONTRIBUTING.md#creating-a-code-reproduction
-->



### Other Technical Details
<!--
Please provide the following information with your request and any other relevant technical details (versions of IDEs, local environment info, plugin information or links, etc).
-->

`npm --version` output:

`node --version` output:

`pod --version` output (iOS issues only):

### Additional Context
<!--
List any other information that is relevant to your issue. Stack traces, related issues, suggestions on how to fix, Stack Overflow links, forum links, etc.
-->

This has been resolved in the ionic cordova keyboard plugin by adding a resizeOnFullscreen preference. Do we have anything like this in capacitor keyboard.

jcesarmobile:
can you provide a sample app?
I can't reproduce on the one I used to implement the `resizeOnFullscreen` preference for the Cordova plugin, that's why I didn't implement it for Capacitor.

manishoswal:
[...]

PR #3929, opened 2020-12-10
refactor(iOS): move webview keyboard overrides

ikeith:
This branch moves the web content view method swizzling (to override keyboard focus behavior) out of the bridge view controller and into a class extension. This resolves several issues with the previous implementation (in relative order of severity):
- The functionality didn't need to be in the view controller, which complicated the VCs implementation.
- The override flag could only be set once as its value was baked into the replacement implementations.
- The override applied to all webviews inside the application.
- There were no protections against swizzling multiple times, leading to very unpredictable results if more than one view controller was ever instantiated.

All of which made the implementation unsuitable and potentially dangerous in embedded use. This branch fixes the swizzling to ensure it only happens once and changes the override to be _per instance_ of webview allowing for more granular control and to toggle the behavior on and off as needed.


BerkeAras:
@ikeith Would this fix #3088?

ikeith:
@BerkeAras No, it's unrelated. I'm not familiar with the specifics of that issue but this PR is about changing the _location of the code_ within the iOS codebase (and fixing a couple of potential problems in it). It should not change the visible behavior of the keyboard on any platform.

dwieeb (approved):

jcesarmobile (approved):

PR #3947, opened 2020-12-14
feat(android): Add WebColor utility for parsing color

### LINKED ISSUE #3903: iOS RGBA / Android ARGB color parsing disparity
On iOS, colors are parsed as RGBA, but on Android, colors are parsed as ARGB. This makes it difficult to use the same color input string cross-platform.

We should include a utility in `@capacitor/android` that offers RGBA color parsing to plugins so that we can consistently parse color input strings to Android values and the same color input string can be used in iOS and Android.

Don't forget to add this as a change that affects users in the [Upgrade Guide](https://github.com/ionic-team/capacitor/issues/3320). 😁 

ref: https://github.com/ionic-team/capacitor/pull/3719

thomasvidas:
Under `package main.java.com.getcapacitor.util;`

Uses `Color.parseColor` under the hood but allows the user to specify an ColorFormat. Right now only RGB, RGBA and ARGB are supported. But it would be cool to add something like HSV or HSL in the future.

Fixes #3903 (but will require changes to a bunch of the plugins)


jcesarmobile:
On iOS users can use RGB or RGBA indistinctly, so this should behave the same way.
Instead of having the param for the format, it should only accept the HEX color string, check the length and consider it RGB or RGBA based on it.
And I think iOS also accepts with and without `#`

dwieeb:
Summary of slack conversation: don't extend `Color` (just a simple `parseColorRGBA(String) -> int`), ignore Android color names (hopefully we didn't document support for those anywhere... and if we do accept color names, we should use [this list](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value), not Android's)

android/capacitor/src/main/java/com/getcapacitor/util/WebColor.java
```diff
@@ -0,0 +1,27 @@
+package main.java.com.getcapacitor.util;
```
dwieeb, line 1: Shouldn't this be `com.getcapacitor.util`?

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -43,6 +43,8 @@
 import org.json.JSONException;
 import org.json.JSONObject;
 
+import main.java.com.getcapacitor.util.WebColor;
```
carlpoole, line 46:
```suggestion
import com.getcapacitor.util.WebColor;
```

carlpoole (changes requested):
Looks good

carlpoole (approved):

thomasvidas:
I will merge this when the plugin PR goes up later today

PR #3951, opened 2020-12-15
chore(cli): remove default SplashScreen launchShowDuration value

jcesarmobile:
Max wanted launchShowDuration value to be 0, but we couldn't do it in the middle of v2 as it was a breaking change, so we added this setting for new apps.

For the new SplashScreen plugin I'm making the default value 0, so the CLI don't need to add the value anymore.

### DIFF
--- a/cli/src/tasks/init.ts
+++ b/cli/src/tasks/init.ts
@@ -144,8 +144,7 @@ async function mergeConfig(
       ...oldConfig,
       ...extConfig,
       ...{
-        plugins: extConfig.plugins ??
-          oldConfig.plugins,
+        plugins: extConfig.plugins ?? oldConfig.plugins,
       },
     },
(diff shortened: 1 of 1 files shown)


cli/src/tasks/init.ts
```diff
@@ -144,8 +144,7 @@ async function mergeConfig(
       ...oldConfig,
       ...extConfig,
       ...{
-        plugins: extConfig.plugins ??
-          oldConfig.plugins ?? { SplashScreen: { launchShowDuration: 0 } },
+        plugins: extConfig.plugins ?? oldConfig.plugins,
       },
```
dwieeb, line 149: I think you can delete this entire thing now 

```suggestion
```

dwieeb (approved):

PR #3953, opened 2020-12-15
refactor(iOS): Move webview delegation out of CAPBridgeViewController

ikeith:
This branch refactors `CAPBridgeViewController` to move the WebKit delegation into a new object. None of the functionality needed to be in the view controller and was split between a protocol extension (adopted by the view controller) and the view controller class itself. This was a poor separation of concerns and made `CAPBridgeViewController` quite large.

- `CAPMessageHandlerWrapper` is removed. It only existed as a temporary refactor to break retain cycles in the object graph.
- `WebViewDelegationHandler` has now inherited the functionality that was split between `CAPMessageHandlerWrapper` and `CAPBridgeDelegate`.
- The `WKNavigationDelegate` and `WKUIDelegate` protocol conformance has been moved out of `CAPBridgeViewController` and into `WebViewDelegationHandler`.
- `CAPAssetHandler` has been renamed to `WebViewAssetHandler` to match `WebViewDelegationHandler`.

The result is a cleaner separation between **configuration/initialization** of the environment, which `CAPBridgeViewController` handles, and **runtime management** of the environment, which is split between the handlers and owned by the bridge.


ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -1,7 +1,8 @@
 import Foundation
 import MobileCoreServices
 
-internal class CAPAssetHandler: NSObject, WKURLSchemeHandler {
+@objc(CAPWebViewAssetHandler)
+internal class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
```
dwieeb, line 5: Just curious, how come internal classes need an Obj-C name?

dwieeb (approved):

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
ikeith: Because inheriting from `NSObject`, which is necessary when conforming to certain protocols from UIKit, means that the class is bridged into the Obj-C runtime regardless of any access controls. Being internal, it's interface is not exposed outside of the framework but at runtime the class name would still exist in the flat namespace of the application.

jcesarmobile (approved):

PR #3968, opened 2020-12-18
feat(core): WebPlugin native call

thomasvidas:
Add `WebPlugin.callNative()` to the `WebPlugin` class, allowing users to call native functions from the web side.


core/src/web-plugin.ts
```diff
@@ -129,6 +129,11 @@ export class WebPlugin implements Plugin {
   load(): void {
     this.loaded = true;
   }
+
+  callNative(methodName: string, parameters?: any): Promise<any> {
+    const pluginName = this.constructor.name;
```
dwieeb, line 134: We will need to come up with a more reliable way to retrieve the "plugin class".

I looked into this a little bit and this may involve some refactoring and a change in the API.

thomasvidas: Yeah, I've been trying to see if we could use the proxies in `runtime.ts` to add in the pluginName from there but I'm not making much progress. We could make another constructor that takes in the name optionally? Or we could just use `callNative` to mask the `Capacitor.nativePromise` call

thomasvidas:
When #3983 is merged, I might be able to grab the pluginName in the proxy and inject it into the native function 🤞. I'd really like to avoid having to pass in the name directly

thomasvidas:
Now that #3983 is merged, we can use `Reflect.apply` to include `pluginName` to the scope of the Plugin class without using the config

core/src/web-plugin.ts
```diff
@@ -129,6 +132,10 @@ export class WebPlugin implements Plugin {
   load(): void {
     this.loaded = true;
   }
+
+  callNative(methodName: string, parameters?: any): Promise<any> {
```
dwieeb, line 136:
```suggestion
  protected callNative(methodName: string, parameters?: any): Promise<any> {
```

core/src/web-plugin.ts
```diff
@@ -17,8 +22,9 @@ export class WebPlugin implements Plugin {
   listeners: { [eventName: string]: ListenerCallback[] } = {};
   windowListeners: { [eventName: string]: WindowListenerHandle } = {};
 
-  // This gets injected into the scope via the Proxy in runtime.ts
-  pluginName: string;
+  // These gets injected into the scope via the Proxy in runtime.ts
+  private pluginName: string;
+  private cap: CapacitorInstance;
```
dwieeb, line 27: Maybe prefix with underscores to further convey that these are private

core/src/runtime.ts
```diff
@@ -203,7 +211,11 @@ export const createCapacitor = (win: WindowCapacitor): CapacitorInstance => {
 
                   const func = Reflect.get(loadedImpl, prop);
                   if (typeof func === 'function') {
-                    return Reflect.apply(func, loadedImpl, args);
+                    return Reflect.apply(
+                      func,
+                      { ...loadedImpl, pluginName, cap },
```
dwieeb, line 216: Won't you need to change each of these to include the underscore?

dwieeb:
Can you run prettier?

thomasvidas:
Closing after our convo off Github. We'll figure out a better API for this rather than rushing in to get it in 3.x

Issue #3913, opened 2020-12-08
feat: Ability to modify config during bridge initialization


dwieeb:
We can save ourselves from an overwhelming number of `Bridge.setThisValue` and `Bridge.setThatValue` if we load the configuration from `capacitor.config.json` into a Configuration object and then allow the app developer to modify values as needed.

Issue #3912, opened 2020-12-08
feat: Support loading from 


dwieeb:

rraihansaputra:
I'd like to share another usecase, currently I'm building a Capacitor app with Svelte and Routify, and for SSR purposes the default entrypoint is `__app.html`. Currently my workaround for building the Capacitor app both locally and in CI is to have a separate build step for the web app which copies `dist/__app.html` to `dist/index.html` prior to executing `npx cap copy`. If this can be configured, it would resolve the separate jobs needed to make this work.

sabahallah:
Dear @imhoffd 

Any luck when this could be resolved.

We now architecting an application that will use Angular built-in i18n for multi-lingual app. and with using Angular built-in i18n there would be a separate build for each language. We will use Capacitor as a hybrid mobile solution, but Capacitor copy the application from www directory, in my case there will be folder for each language. So how to handle this in Capacitor, and is it possible to change the URL in runtime as the user change the language.

meriturva:
Any news here? we would like to use angular i18n and we have to decide how to proceed. Thanks!

UltraCakeBakery:
There is currently no way to do native translations that allow a user to only install for example the english version of your app in your app bundle. 

You do however have access in your web app to https://developer.mozilla.org/en-US/docs/Web/API/Navigator/languages, which allows you to determine the devices preferred language.

meriturva:
Thanks @UltraCakeBakery for your feedback.

My idea is to pack capacitor with all language folders produced by ng build and add an index.html file on the root folder.
The root index file has just to check the device's preferred language and redirects to the correct folder.

What do you think?

UltraCakeBakery:
> Thanks @UltraCakeBakery for your feedback.
> 
> My idea is to pack capacitor with all language folders produced by ng build and add an index.html file on the root folder. The root index file has just to check the device's preferred language and redirects to the correct folder.
> 
> What do you think?

That'll work fine I think. It is all local so redirects won't hurt performance too much.

rdlabo:
> Thanks @UltraCakeBakery for your feedback.
> 
> My idea is to pack capacitor with all language folders produced by ng build and add an index.html file on the root folder. The root index file has just to check the device's preferred language and redirects to the correct folder.
> 
> What do you think?

This work well. I prepared this html:

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <script type="text/javascript">
[...]

PR #3973, opened 2020-12-21
feat(iOS): Open CAPBridgeViewController for subclassing

ikeith:
This branch refactors CAPBridgeViewController to make it available for subclassing. This is necessary for many embedded uses and allows it address several outstanding issues.

- Subclasses can override `webView(with:configuration:)` to return a custom subclass of WKWebView if desired. This addresses https://github.com/ionic-team/capacitor/pull/3095

- Subclasses can override `instanceDescriptor` to modify the capacitor configuration at runtime after the config file is loaded, or to completely customize the environment. This addresses https://github.com/ionic-team/capacitor/issues/3912 & https://github.com/ionic-team/capacitor/issues/3913 as well as allow for embedded uses where the web app will not exist in the expected locations in the application bundle.

- Subclasses can override `webViewConfiguration` to make any changes to the configuration before it is used to instantiate the web view. This addresses https://github.com/ionic-team/capacitor/issues/1097

- Subclasses can override `capacitorDidLoad` or `viewDidLoad` to make any desired changes to the webview or related views. This provides a stable point for operations like https://github.com/ionic-team/capacitor/pull/3546 without requiring that the code be merged in upstream or added via a plugin.

In addition, this branch unifies much of the path handling to ensure that the config is the one true source for where the web app is located. It also eliminates some of the side effects of the previous implementation like `isNewBinary()` modifying UserDefaults such that it would return different results on subsequent invocations.


jcesarmobile (changes requested):
Can we instead of having the `webViewClass` thing have an open method where the WebView is created so they can override that instead?
That way if their custom `WKWebView` subclass has a custom init they can use it, with current implementation it's not possible.

jcesarmobile (approved):

dwieeb (approved):

tossaro:
hello everyone, first of all thanks for opening CAPBridgeViewController so we can create subclasses, but I am facing blockers when adding hooks to web view lifecycle like didFinish, didFail etc, due to existing loadView is final function that have:
`let delegateHandler = WebViewDelegationHandler()`
I have created a subclass of WebViewDelegationHandler then use at capacitorDidLoad(), but it makes the bridge not work.
So could you please make loadView not final function? or maybe any additional open function that refer to delegation like instanceDescriptor function. thank you

existing capacitor 6.1.0 file CAPBridgeViewController:
`override public final func loadView() {
        // load the configuration and set the logging flag
        let configDescriptor = instanceDescriptor()
        let configuration = InstanceConfiguration(with: configDescriptor, isDebug: CapacitorBridge.isDevEnvironment)
        CAPLog.enableLogging = configuration.loggingEnabled
        logWarnings(for: configDescriptor)

        setStatusBarDefaults()
        setScreenOrientationDefaults()

        // get the web view
        let assetHandler = WebViewAssetHandler(router: router())
        assetHandler.setAssetPath(configuration.appLocation.path)
        assetHandler.setServerUrl(configuration.serverURL)
        let delegationHandler = WebViewDelegationHandler()
        prepareWebView(with: configuration, assetHandler: assetHandler, delegationHandler: delegationHandler)
        view = webView
        // create the bridge
        capacitorBridge = CapacitorBridge(with: configuration,
                                          delegate: self,
                                          cordovaConfiguration: configDescriptor.cordovaConfiguration,
                                          assetHandler: assetHandler,
                                          delegationHandler: delegationHandler)
        capacitorDidLoad()

        if configDescriptor.instanceType == .fixed {
            updateBinaryVersion()
        }
    }`

PR #3981, opened 2020-12-22
chore: bump targetSdkVersion and compileSdkVersion to 30

jcesarmobile:
in variables.gradle, capacitor and capacitor-cordova-android-plugins

Added QueryPermissionsNeeded lint suppress because that it used for input file and getUserMedia, but most users won't use it, so adding intent queries on Capacitor's manifest wouldn't be a good idea.

But we would need to document somewhere that for Android 11, users need to add those queries if using input file and/or getUserMedia (not sure where would be the better place for that)

### DIFF
--- a/android-template/variables.gradle
+++ b/android-template/variables.gradle
@@ -1,7 +1,7 @@
 ext {
     minSdkVersion = 21
-    compileSdkVersion = 29
-    targetSdkVersion = 29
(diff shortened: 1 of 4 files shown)


dwieeb (approved):

dwieeb:
Is this more of a feat?

PR #3999, opened 2020-12-29
feat(cli): create TS configuration files in `init`

dwieeb:


cli/src/tasks/init.ts
```diff
@@ -132,18 +124,68 @@ async function getWebDir(config: Config, webDir?: string) {
...
-    { spaces: 2 },
+  await writeConfig(newConfig, newConfigPath);
+}
+
+function printNextSteps(newConfigName: string) {
+  logSuccess(`${c.strong(newConfigName)} created!`);
+  output.write(
+    `\nAdd platforms using ${c.input('npx cap add')}:\n` +
+      `  ${c.input('npx cap add android')}\n` +
+      `  ${c.input('npx cap add ios')}\n\n` +
```
jcesarmobile, line 186: Not related to this change since you just moved it, but, since platforms are not automatically installed anymore, this commands will end up in errors, so maybe we should remove it or reword?

dwieeb: Sure, I think we should remove it and keep the link to the workflow docs below. We can add a section about adding platforms there.

jcesarmobile (changes requested):
The generated `.ts` file in an existing angular app makes vs code to complain with

> Export assignment cannot be used when targeting ECMAScript modules. Consider using 'export default' or another module format instead.

so probably not a good idea to default to the .ts file if it can cause problems or the .ts file should be changed to support angular apps.

cli/src/tasks/init.ts
dwieeb: https://github.com/ionic-team/capacitor/issues/4030

dwieeb:
> or the .ts file should be changed to support angular apps.

Agreed, see https://github.com/ionic-team/capacitor/pull/4031

dwieeb:
https://github.com/ionic-team/capacitor/pull/4031 is now merged. I think we should document `export default` as the preferred strategy

jcesarmobile:
not just document it, this PR should also use it on the initial .ts file creation

jcesarmobile (approved):

PR #4002, opened 2020-12-31
fix: move `public` into iOS target directory

dwieeb:
The `public` directory, which contains generated web assets, should exist alongside the generated `capacitor.config.json` within the iOS target directory, not the Xcode workspace directory.

TODO:
- [ ] Add migration steps to upgrade guide.
- [x] Add a check for old pbxproj and link to upgrade guide.
- [ ] Update [these docs](https://capacitorjs.com/docs/v3/ios/configuration#renaming-your-app)

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -1,4 +1,4 @@
-import { pathExists, readJSON } from '@ionic/utils-fs';
+import { pathExists, readFile, readJSON } from '@ionic/utils-fs';
 import Debug from 'debug';
 import { dirname, join, relative, resolve } from 'path';
 
@@ -14,6 +14,7 @@ import type {
 } from './definitions';
 import { OS } from './definitions';
 import { fatal, isFatal } from './errors';
+import { logger } from './log';
 import { tryFn } from './util/fn';
 import { resolveNode, requireTS } from './util/node';
 import { lazy } from './util/promise';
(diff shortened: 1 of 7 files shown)


jcesarmobile (approved):
looks good

the .gitignore change for existing apps should be documented

also the native iOS project change, I find this one more problematic and easier to miss and can lead to a lot of users not knowing why their file changes don't get updated after the migration

should the CLI remove the existing `ios/app/public` folder? or should also be documented that users need to remove it?

dwieeb:
The CLI should probably continue to support `ios/app/public`, warning the user during every `sync` to migrate with a link to the migration guide. When they delete the folder we can assume they made the project changes and start writing to the new location. Will that work?

jcesarmobile:
since the public folder is gitignored, there can be the case that an user downloads the project and the folder is not there so the CLI thinks the migration was done, but it wasn't

maybe make the CLI check the `project.pbxproj` to see if the `path = public; sourceTree = SOURCE_ROOT;` was updated to `path = public; sourceTree = "<group>";` and if not, copy to the old location?

dwieeb:
Good point.

> maybe make the CLI check the `project.pbxproj`

Yeah, should be possible. 👍

ikeith (approved):

PR #4019, opened 2021-01-05
fix(android): always inject js script

adamdbradley:
While unlikely, it's still possible to have an html file without a head element. Instead of not working at all, fallback to adding the script before all of the html.


dwieeb (approved):

jcesarmobile:
Closing the PR.
Injecting the javascript before the html document start caused problems in the past, like adding a big white gap on top the app's ion-header, which was hard to debug.
Docs mention that a head is needed, so I think it's better to keep warning that Capacitor was not injected than cause weird display glitches.

PR #4032, opened 2021-01-07
refactor(cli): move xcodeproj/xcworkspace to config

dwieeb:
Additional config unification for xcodeproj and xcworkspace directories.

precursor to: https://github.com/ionic-team/capacitor/pull/4002

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -1,6 +1,6 @@
 import { pathExists, readJSON } from '@ionic/utils-fs';
 import Debug from 'debug';
-import { dirname, join, resolve } from 'path';
+import { dirname, join, relative, resolve } from 'path';
 
 import c from './colors';
 import type {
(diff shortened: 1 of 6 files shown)


cli/src/config.ts
```diff
@@ -237,6 +237,8 @@ async function loadIOSConfig(
   const platformDirAbs = resolve(rootDir, platformDir);
   const nativeProjectDir = 'App';
   const nativeTargetDir = `${nativeProjectDir}/App`;
+  const nativeXcodeProjDir = `${nativeProjectDir}/App.xcodeproj`;
+  const nativeXcodeWorkspaceDir = `${nativeProjectDir}/App.xcworkspace`;
```
dwieeb, line 241: Note: with this PR, we now blindly assume `App.xcworkspace` (instead of what was happening in `findXcodePath()`).

What was the reason for `findXcodePath()`? Are there apps with workspace directories named something other than `App.xcworkspace`?

jcesarmobile: It's not possible at the moment since we hardcode `App` in a few places, but users have tried to rename their projects and it broke because our hardcoded values.

jcesarmobile, line 241:
```suggestion
  const nativeTargetDir = `${nativeProjectDir}/${nativeProjectDir}`;
  const nativeXcodeProjDir = `${nativeProjectDir}/${nativeProjectDir}.xcodeproj`;
  const nativeXcodeWorkspaceDir = `${nativeProjectDir}/${nativeProjectDir}.xcworkspace`;
```
if we do it like this, would be easier to allow users to rename their project (if we decide to make the `nativeProjectDir` value configurable)

cli/src/ios/open.ts
```diff
@@ -1,22 +1,9 @@
...
-
-  if (!xcodeProject) {
-    fatal(
-      'Xcode workspace does not exist.\n' +
-        `Run ${c.input('npx cap add ios')} to bootstrap a new iOS project.`,
-    );
-  }
-
-  await open(xcodeProject, { wait: false });
+  await open(config.ios.nativeXcodeWorkspaceDirAbs, { wait: false });
```
jcesarmobile, line 7: I think we should check if it exists before trying to open, as open doesn't provide any feedback if Xcode was opened or not

jcesarmobile (changes requested):

cli/src/ios/open.ts
dwieeb: I agree, but this should happen automatically when we try to load the value from the config.

jcesarmobile (approved):

PR #4033, opened 2021-01-07
feat(android): switch to new callback-style permission requests

carlpoole:
**Changes**
- PluginState enum replaces raw strings representing the state of a permission
- Permission response codes are no longer necessary and have been removed
- Behavior run after permission grant/deny is defined in a callback method.
- Permission callback methods are defined on the plugin class with the signature format `private void myPermissionCallback(PluginCall, Map<String, PluginState>)`
- PluginMethods that result in permission request calls must specify the name of a callback method in the `@PluginMethod` annotation, for example: `@PluginMethod(permissionCallback = "myPermissionCallback")`
- The `permissionCallback` annotation element is important in registering the permission callback in the Activity lifecycle and ensuring the code is run after permission is granted/denied.

**2.x**
- For backwards compatibility with 2.x, any plugins defined with a permission response code in a @NativePlugin annotation will be handled first before delegating the permission handling to Android for the new callback method.


android/capacitor/src/main/java/com/getcapacitor/PermissionResponse.java
```diff
@@ -0,0 +1,19 @@
+package com.getcapacitor;
```
dwieeb, line 1: This file can be deleted, I believe.

android/capacitor/src/main/java/com/getcapacitor/cordova/MockCordovaInterfaceImpl.java
```diff
@@ -14,4 +16,23 @@ public MockCordovaInterfaceImpl(Activity activity) {
     public CordovaPlugin getActivityResultCallback() {
         return this.activityResultCallback;
     }
+
+    /**
+     * Use in lieu of {@link CordovaInterfaceImpl#onRequestPermissionResult(int, String[], int[])} to return
```
dwieeb, line 21: Is `CordovaInterfaceImpl#onRequestPermissionResult()` still used? Pretty sure we can remove it.

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -511,7 +509,7 @@ public PluginHandle getPluginWithRequestCode(int requestCode) {
                 }
             }
 
-            if (permissionRequestCode == requestCode) {
+            if (permissionRequestCode != null && permissionRequestCode == requestCode) {
                 return handle;
             }
```
dwieeb, line 514: I would move this up into the legacy annotation block and then define `permissionRequestCode` as a local variable

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -765,42 +763,29 @@ public void startActivityForPluginWithResult(PluginCall call, Intent intent, int
      * @param requestCode the code that was requested
      * @param permissions the permissions requested
      * @param grantResults the set of granted/denied permissions
+     * @return true if permission code was handled by a plugin explicitly, false if not
```
dwieeb, line 766: I think we should somehow indicate in the javadoc that this method is for Cordova plugins and legacy plugins only. It may be a good place to document how all the new permission handling works, because it's complex. Either here or `BridgeActivity.java`, what do you think?

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -58,6 +64,16 @@
     // Stored event listeners
     private final Map<String, List<PluginCall>> eventListeners;
 
+    /**
+     * Base activity result launcher used by the {@link #requestPermissions(PluginCall)} plugin call
+     */
+    protected ActivityResultLauncher<String[]> basePermissionLauncher = null;
```
dwieeb, line 70: I think it's possible to make this private now.

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -283,47 +375,116 @@ public boolean hasRequiredPermissions() {
...
-     * @param permission the permission to request
-     * @param requestCode the requestCode to use to associate the result with the plugin
      */
-    public void requestPermission(PluginCall call, String permission, int requestCode) {
-        requestPermissions(call, new String[] { permission }, requestCode);
+    protected void requestAllPermissions(@NonNull PluginCall call) {
+        String callingMethodName = call.getMethodName();
+        ActivityResultLauncher<String[]> activityResultLauncher = permissionLaunchers.get(callingMethodName);
+        if (activityResultLauncher == null && !callingMethodName.equals("requestPermissions")) {
+            Logger.error(
```
dwieeb, line 387: Have you thought about rejecting the call here instead?

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -283,47 +375,116 @@ public boolean hasRequiredPermissions() {
...
+                    "method that has the correct signature: (PluginCall, Map<String, PermissionState>)",
+                    callingMethodName
+                )
+            );
+        }
+
+        // default to base permission launcher that returns permission states via the plugin call
+        if (activityResultLauncher == null) {
+            activityResultLauncher = basePermissionLauncher;
+        }
```
dwieeb, line 447: Can this logic be put into a private method for reuse? (it looks the same as in `requestAllPermissions`)

android/capacitor/src/main/java/com/getcapacitor/cordova/MockCordovaInterfaceImpl.java
carlpoole: That's defined on the parent class in Apache's library, and I don't see it referenced anywhere else now so I can just clean up this comment to avoid confusion.

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -758,8 +758,9 @@ public void startActivityForPluginWithResult(PluginCall call, Intent intent, int
     }
 
     /**
-     * Handle a request permission result by finding the that requested
-     * the permission and calling their permission handler
+     * Check for legacy Capacitor or Cordova plugins that may have registered to handle a permission
```
carlpoole, line 761: @dwieeb How is this verbiage? See javadoc added in BridgeActivity line 167 also

dwieeb: Perfect 👌

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
carlpoole: We could. Would it be more desirable to have a hard fail in this instance instead of falling back to a functional permission request and result (the basePermissionLauncher)? I see benefits to the fallback, but this proposal would be a more explicit indicator to the developer that something was incorrectly implemented.

dwieeb: Yeah I think it might be better to fail the request so the plugin developer knows something was incorrectly implemented. (Assuming I'm understanding all the pieces at play here)

dwieeb (approved):

PR #4043, opened 2021-01-11
fix(iOS): properly handle date types during JSValue coercion

theproducer:
This PR fixes a missing case in `coerceToJSValue` in which NSDate values were not being returned, resulting in nil values.

Example:
```
const mydate = new Date();
const result = await ExamplePlugin.testMethod({ date: mydate });
```

```
@objc func testMethod(_ call: CAPPluginCall) {
         // date will be nil
        let date = call.getDate("date")
        
       
        if let date = date {
            print("Date: \(date)")
        }
    }
```

### DIFF
--- a/ios/Capacitor/Capacitor/JSTypes.swift
+++ b/ios/Capacitor/Capacitor/JSTypes.swift
@@ -9,7 +9,7 @@ extension Float: JSValue {}
 extension Double: JSValue {}
 extension NSNumber: JSValue {}
 extension Array: JSValue {}
-extension NSDate: JSValue {}
+extension Date: JSValue {}
 extension Dictionary: JSValue where Key == String, Value == JSValue {}
 
 // convenience aliases
@@ -206,7 +206,7 @@ private func coerceToJSValue(_ value: Any?) -> JSValue? {
         return floatValue
     case let doubleValue as Double:
         return doubleValue
-    case let dateValue as NSDate:
+    case let dateValue as Date:
         return dateValue
     case let arrayValue as NSArray:
         return arrayValue.compactMap { coerceToJSValue($0) }
--- a/ios/Capacitor/Capacitor/JSTypes.swift
+++ b/ios/Capacitor/Capacitor/JSTypes.swift
@@ -9,6 +9,7 @@ extension Float: JSValue {}
 extension Double: JSValue {}
 extension NSNumber: JSValue {}
 extension Array: JSValue {}
+extension NSDate: JSValue {}
 extension Dictionary: JSValue where Key == String, Value == JSValue {}
 
 // convenience aliases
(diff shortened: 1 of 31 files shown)


ios/Capacitor/Capacitor/JSTypes.swift
```diff
@@ -9,6 +9,7 @@ extension Float: JSValue {}
 extension Double: JSValue {}
 extension NSNumber: JSValue {}
 extension Array: JSValue {}
+extension NSDate: JSValue {}
```
ikeith, line 12: This should be `Date`. Since these are Swift convenience wrappers, we should prefer the Swift data types wherever possible.
```suggestion
extension Date: JSValue {}
```

ios/Capacitor/Capacitor/JSTypes.swift
```diff
@@ -151,6 +152,10 @@ extension JSValueContainer {
         if let isoString = jsObjectRepresentation[key] as? String {
             return Self.jsDateFormatter.date(from: isoString)
         }
+
```
ikeith, line 155: Since the string coercion has already failed by this point, this whole block could be reduced to a single line:
`return jsObjectRepresentation[key] as? Date`

ios/Capacitor/Capacitor/JSTypes.swift
```diff
@@ -201,6 +206,8 @@ private func coerceToJSValue(_ value: Any?) -> JSValue? {
         return floatValue
     case let doubleValue as Double:
         return doubleValue
+    case let dateValue as NSDate:
```
ikeith, line 209:
```suggestion
    case let dateValue as Date:
```

ikeith (changes requested):
Looks good; just some minor changes. And please remove the unrelated whitespace diffs, if possible.

dwieeb (approved):

ios/Capacitor/Capacitor/JSTypes.swift
```diff
@@ -148,10 +149,7 @@ extension JSValueContainer {
     }
 
     public func getDate(_ key: String) -> Date? {
-        if let isoString = jsObjectRepresentation[key] as? String {
-            return Self.jsDateFormatter.date(from: isoString)
-        }
-        return nil
+        return jsObjectRepresentation[key] as? Date
```
ikeith, line 152: Sorry if it wasn't clear. I meant that the multiple lines with the date checking could be coalesced into a single line. I think we need to keep the string interpretation intact so we can work with JSON strings or date objects.
```suggestion
        if let isoString = jsObjectRepresentation[key] as? String {
            return Self.jsDateFormatter.date(from: isoString)
        }
        return jsObjectRepresentation[key] as? Date        
```

ikeith (changes requested):

ios/Capacitor/Capacitor/JSTypes.swift
theproducer: Ah, its ok, makes sense now

ikeith (approved):
Looks good. The type bridging actually has a lot of test coverage (the exception in the project) so it'd be good to update the unit tests to verify the additional date handling. But that can be addressed in a separate branch.

PR #4045, opened 2021-01-12
docs(cli): add a hint for installing the platform

dwieeb:


cli/src/common.ts
```diff
@@ -64,14 +64,17 @@ export async function checkPackage(): Promise<string | null> {
...
 ): Promise<string | null> {
   const pkg = await getCapacitorPackage(config, platform);
 
   if (!pkg) {
-    return `Could not find the ${c.input(
-      platform,
-    )} platform. Does it need to be installed?\n`;
+    return (
+      `Could not find the ${c.input(platform)} platform.\n` +
+      `Does it need to be installed first? e.g. w/ ${c.input(
```
jcesarmobile, line 74: I don't think this should be a question. It needs to be installed first.

cli/src/common.ts
```diff
@@ -64,14 +64,17 @@ export async function checkPackage(): Promise<string | null> {
 
 export async function checkCapacitorPlatform(
   config: Config,
-  platform: string,
+  platform: 'android' | 'ios',
```
jcesarmobile, line 67: on every other method we use string, so for consistency I think we should keep string

maybe later on we can do a refactor to replace platform with `'android' | 'ios'`

dwieeb: I agree, I was just worried somewhere it was being called for custom platforms. I'll change it back

jcesarmobile (approved):

PR #4067, opened 2021-01-14
chore(android): remove back button handling and listeners

jcesarmobile:


carlpoole (approved):
Is this worth deprecating in case there are some plugins using it?

jcesarmobile:
it was added in a previous alpha, so no plugin should be using it

dwieeb (approved):

PR #4072, opened 2021-01-15
fix(iOS): preserve null values in bridged types

ikeith:
This branch extends JSValue to include NSNull in order to preserve null values sent from JS. It also adds some small extensions to Array to improve the ergonomics when working in Swift, and adds unit tests to document and verify the behavior.

This branch also deprecates `hasOption` on `CAPPluginCall` as the presence of a key in the options isn't significant and we shouldn't encourage that pattern. Typing the represented value if it is not nil is more meaningful.

This addresses the underlying cause that https://github.com/ionic-team/capacitor/pull/4048 worked around.


dwieeb (commented):
Would adding a test for [JSValue?] and heterogeneous arrays make sense (in `BridgedTypesTests.swift`)?

ikeith:
> Would adding a test for [JSValue?] and heterogeneous arrays make sense (in `BridgedTypesTests.swift`)?

More tests are good, but I'm not clear on what use-case you're suggesting covering?

jcesarmobile (approved):

dwieeb:
@ikeith Does something like this make sense?

```swift
let array: [Any] = ["test string 1", 1, NSNull()]
let sparseArray = JSTypes.coerceArrayToJSArray(array)?.capacitor.replacingNullValues() as? [JSValue?]
// sparseArray should cast successfully and not be nil
```

ikeith:
@dwieeb That kind of test was already added to the test suite: `testSparseArrayCastSuccess` in BridgedTypesTests.swift

PR #4073, opened 2021-01-16
feat(cli): configurable iOS build scheme

dwieeb:
This allows custom scheme names to be used when running `npx cap run`.

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -262,6 +262,7 @@ async function loadIOSConfig(
   const podPath = determineCocoapodPath();
   const platformDir = extConfig.ios?.path ?? 'ios';
   const platformDirAbs = resolve(rootDir, platformDir);
+  const scheme = extConfig.ios?.scheme ?? 'App';
   const nativeProjectDir = 'App';
   const nativeProjectDirAbs = resolve(platformDirAbs, nativeProjectDir);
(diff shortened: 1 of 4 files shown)


jcesarmobile (approved):
looks good and works fine

but wouldn't it better to make it a param on the run command?
i.e: `capacitor run ios --scheme="MyScheme"`

Also the `scheme` name might be a little confusing since we have `iosScheme` on `server` object that configures the scheme of the url used to load the app and also the app url scheme for opening the app from another app.

dwieeb:
> but wouldn't it better to make it a param on the run command?

Maybe, but later on we may want to use the build scheme to derive more information about the iOS project for `sync`, etc. By using `xcodebuild -showBuildSettings` with the configured scheme, we can get the folder structure of the project and know exactly where everything is, which means developers can rename folders. I also think it will be useful if we ever add the `npx cap build` command.

> `scheme` name might be a little confusing since we have `iosScheme`

I would rather deprecate `iosScheme` in favor of `server.ios.httpScheme` (and `server.android.httpScheme` for Android) to clear up any potential confusion. What do you think?

jcesarmobile:
Then `httpScheme` name is confusing because you can't really use `http` value.

Anyway, I approved the changes, just sharing my concerns.

dwieeb:
Wow, my brain said `urlScheme` but I wrote `httpScheme`.

PR #4084, opened 2021-01-19
feat(cli): send config, rootDir, and webDir to custom platform hooks

IT-MikeS:
This solves the issue of finding the users app/project directory, cleaning up the currently hard coded "up 3 levels" way of doing it. (which sometimes failed)


cli/src/common.ts
```diff
@@ -5,6 +5,7 @@ import c from './colors';
 import type { Config, PackageJson } from './definitions';
 import { fatal } from './errors';
 import { output, logger } from './log';
+import { configCommand } from './tasks/config';
```
dwieeb, line 8:
```suggestion
```

cli/src/common.ts
```diff
@@ -168,6 +170,7 @@ export async function runPlatformHook(
       cwd: platformDir,
       env: {
         INIT_CWD: platformDir,
+        CAP_CONFIG: config,
```
dwieeb, line 173: There are a couple things we will need to address here:
- I'm not sure how `config` is being serialized into an environment variable here, we'll need to check.
- The CLI has "lazy values" now (https://github.com/ionic-team/capacitor/pull/4007) and any place we want the **full** config we'll need to use that `deepAwait` helper.
- But... I'm not sure we want to expose the full config here. Maybe just `config.app.extConfig` and a few key directories? `config.app.rootDir`?

dwieeb (approved):

PR #4093, opened 2021-01-20
chore(android): Update dependencies to latest on variables.gradle

jcesarmobile:
Updated the dependencies in `variables.gradle` file to use latest stable dependencies available (except for android material that was already using a RC, so continued doing so)

Removed `androidxLocalbroadcastmanagerVersion` since it was used for the BackgroundTask plugin which is no longer present.

Tested the plugins and they seemed ok and showed no warnings.

But maybe we should remove all that are not used by capacitor/android and just document them in the plugins?

I created an issue for documenting them https://github.com/ionic-team/capacitor-plugins/issues/123


dwieeb:
> But maybe we should remove all that are not used by capacitor/android and just document them in the plugins?

It's questionable because plugins can share these values

jcesarmobile:
This change will only affect new apps, not existing apps, if the variable doesn't exist on a plugin, then the plugin uses the default configured value, so wouldn't break them
Only if the default values are different can cause issues, that's why plugins should document their variables.

But there is no harm in having the variables defined here just in case.

dwieeb (approved):

PR #4100, opened 2021-01-20
fix(android): get PermissionState enum by state value

carlpoole:
Matching the lower-case and kebab case state values to the enum names was not working as intended

The enum method `valueOf` must match the enum name exactly. 
Eg. "denied" did not match to DENIED


jcesarmobile (approved):

dwieeb (approved):

leo6104:
I'm waiting for this commit to be released 👍

PR #4114, opened 2021-01-25
fix: fix isPluginAvailable() for web and native

dwieeb:
This PR introduces the idea of "plugin headers" that are written to JS alongside the plugin skeleton implementations.

This is required because the [core refactor](https://github.com/ionic-team/capacitor/pull/3748) changed how the `Capacitor.Plugins` object is populated and broke `isPluginAvailable()`. The new headers are necessary to know if a plugin has an implementation on native. `isPluginAvailable()` has been updated to check JS and native implementations.

I want to do a refactor later to remove the plugin skeleton implementations because they're unnecessary with our proxy and with bundling the core together.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/JSExport.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/JSExport.java
@@ -4,6 +4,11 @@ import static com.getcapacitor.FileUtils.readFile;
 
 import android.content.Context;
 import android.text.TextUtils;
+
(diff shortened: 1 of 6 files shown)


mburger81:
Looks good, can you please release it in beta as soon as possibile, thx.

dwieeb:
@mburger81 Unfortunately this PR does not yet address all the issues with `isPluginAvailable()`.

Issue #4050, opened 2021-01-12
AndroidX Activity Result API Update


carlpoole:
- Update the Activity intent and result mechanism to use the AndroidX Activity Result API changes in v1.2

PR #4127, opened 2021-01-28
feat(android): activity result use new API and updated permission result callbacks to match

carlpoole:
- Removes the need to define request codes on plugins. `requestCodes` element removed from `@CapacitorPlugin` annotation (breaking change for plugins with activity results on Capacitor 3 beta)
- Activity result callbacks are registered by adding the `@ActivityCallback` annotation to a method with the parameters (PluginCall, ActivityResult).
- Permission result callbacks are registered by adding the `@PermissionCallback` annotation to a method with the parameter (PluginCall)
- The `permissionCallback` element is removed from the `@PluginMethod` annotation (breaking change for plugins with permission requests on Capacitor 3 beta)
- Activities are launched using the new method `startActivityForResult(PluginCal call, Intent intent, String callbackName);` and link to the registered callback by providing the name of the callback method.
- Existing permission request methods were modified to add the callback method name parameter, for example `requestPermissionForAlias(String alias, PluginCall call, String callbackName);`

Resolves https://github.com/ionic-team/capacitor/issues/4050


android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -91,87 +95,123 @@ public Plugin() {
...
-                            if (!savedPermissionCall.isReleased() && !savedPermissionCall.isSaved()) {
-                                savedPermissionCall.release(bridge);
-                            }
-                        }
-                    }
-                );
+    void initializeActivityLaunchers() {
+        try {
+            // load the default checkPermission callback from the plugin parent class
+            Method method = getClass().getSuperclass().getMethod("checkPermissions", PluginCall.class);
```
dwieeb, line 104: Did we decide not to get this dynamically via the annotations in the `Plugin` superclass?

carlpoole: Yes, thank you. Fixed.

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -900,36 +905,36 @@ protected boolean validatePermissions(Plugin plugin, PluginCall savedCall, Map<S
      * @param resultCode
      * @param data
      */
-    public void onActivityResult(int requestCode, int resultCode, Intent data) {
+    boolean onActivityResult(int requestCode, int resultCode, Intent data) {
```
jcesarmobile, line 908: shouldn't this still be public?

android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -91,85 +95,111 @@ public Plugin() {
...
-                    new ActivityResultContracts.RequestMultiplePermissions(),
-                    permissions -> {
-                        PluginCall savedPermissionCall = bridge.getPermissionCall(handle.getId());
+     * Registers activity result launchers defined on plugins, used for permission requests and
+     * activities started for result.
+     */
+    void initializeActivityLaunchers() {
+        List<Method> pluginClassMethods = new ArrayList<>();
+        pluginClassMethods.addAll(Arrays.asList(getClass().getSuperclass().getDeclaredMethods()));
+        pluginClassMethods.addAll(Arrays.asList(getClass().getDeclaredMethods()));
```
dwieeb, line 104: Something I just realized with this approach: we're assuming the super class is `Plugin`, which is _probably_ fine, but if someone extends another subclass, this may become difficult to debug. What do you think?

carlpoole: I think to be safe I could rework this to scan up through the class hierarchy and stop at Plugin.class

jcesarmobile (commented):
I've been doing some testing on the sample app and on camera page the app crash on the second time I click the "Request all permissions" button (or with the photo permissions too) because of null `savedCall` on `triggerPermissionCallback` method.
I tried to put a null check and that only avoids the crash, then one of each call works and the following doesn't because it's null, then the following works again and so on, so loos like the `bridge.getPermissionCall(handle.getId());` is not working properly or it's not getting stored properly

carlpoole:
> I've been doing some testing on the sample app and on camera page the app crash on the second time I click the "Request all permissions" button (or with the photo permissions too) because of null `savedCall` on `triggerPermissionCallback` method.

Nice catch. Thanks for testing, the call was being saved twice so subsequent permissions checks on the same plugin were trying to find an old call and fetching `null` instead.

jcesarmobile (approved):

dwieeb (approved):

PR #4133, opened 2021-01-29
fix(iOS): Handle NSString in JSValue type casting

theproducer:
This PR adds support for type casting NSString to JSValue

See below for example problem (taken from https://github.com/ionic-team/capacitor-plugins/pull/211):
```swift
var notificationExtras = request.content.userInfo["cap_extra"] as? [String: Any?]
if let notificationExtras = notificationExtras {
    var extra: JSObject = [:]
    for(key, value) in notificationExtras {
        extra[key] = value as? JSValue // will be nil if value is NSString
    }    
}
```

### DIFF
--- a/ios/Capacitor/Capacitor/JSTypes.swift
+++ b/ios/Capacitor/Capacitor/JSTypes.swift
@@ -2,6 +2,7 @@ import Foundation
 
 // declare our empty protocol, and conformance, for typing
 public protocol JSValue {}
+extension NSString: JSValue {}
 extension String: JSValue {}
 extension Bool: JSValue {}
 extension Int: JSValue {}


ikeith:
I don't think this is the right approach. `JSValue` is a Swift-only abstract protocol while most foundation types are actually [Obj-C class clusters](https://developer.apple.com/library/archive/documentation/General/Conceptual/CocoaEncyclopedia/ClassClusters/ClassClusters.html) which is why we get these weird corner cases where the runtimes intersect. If we go down this route, we'd have to add protocol conformance for all the possible types (potentially including the private classes hidden inside the clusters).

As a general rule, you shouldn't be casting to JSValue unless you're casting _up_ from one of the concrete subtypes that conform to the protocol.

However, this problem is already solved via `JSTypes.coerceDictionaryToJSObject`! Rather than casting the individual values, you should coerce the whole collection and then you can pull out `JSValue` objects as needed.

theproducer:
Thank you!

Issue #4125, opened 2021-01-28
bug: live-reload not working when using adb reverse


czabaj:
## Bug Report

App set for live-reload with

```
	"android": {
		"webContentsDebuggingEnabled": true
	},
	"server": {
		"cleartext": true,
		"url": "http://localhost:8080"
	}
```

The app runs in the simulator, displays the "Using app server http://localhost:8080" toast, but it serves static assets from the build folder and the live-reload really does not work.

I start the Android studio with 

```
yarn cap sync android && yarn cap copy android && yarn cap open android
```

### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 2.4.6
  @capacitor/core: 2.4.6
  @capacitor/android: 2.4.6
  @capacitor/electron: 2.4.6
  @capacitor/ios: 2.4.6

Installed Dependencies:

  @capacitor/cli 2.4.5
  @capacitor/core 2.4.5
  @capacitor/ios 2.4.5
  @capacitor/android 2.4.5
  @capacitor/electron not installed

[success] Android looking great! 👌
  Found 9 Capacitor plugins for ios:
    capacitor-dark-mode (1.0.5)
    capacitor-plugin-ios-swipe-back (1.0.3)
    capacitor-rate-app (0.1.1)
    capacitor-secure-storage-plugin (0.5.0)
    cordova-plugin-appcenter-crashes (0.5.1)
    cordova-plugin-appcenter-shared (0.5.1)
    cordova-plugin-fingerprint-aio (4.0.2)
    cordova-plugin-inappbrowser (5.0.0-dev)
    onesignal-cordova-plugin (2.11.2)
[success] iOS looking great! 👌
```

### Platform(s)

Android

### Other Technical Details

`yarn --version` output: `1.22.10`

`node --version` output: `v12.19.0`

jcesarmobile:
see https://github.com/ionic-team/capacitor/discussions/4123#discussioncomment-315565

czabaj:
@jcesarmobile oh, I see. Thank you!

Previously, setting `localhost` worked because we set

```
adb reverse tcp:8080 tcp:8080
```

for the simulator or device.

But recently, this stopped working and the fact, that the simulator uses static files from `public/` dir instead, cost me a lot of time since it pretends it is working even showing the "Serving from ..." toast.

I set the server to `0.0.0.0` instead of the `localhost` and now it works again (after the `adb reverse ...` command).

[...]

PR #4141, opened 2021-02-02
fix(android): live reload not working when using adb reverse

jcesarmobile:
When using adb reverse live reload is not working because of the `loadingUrl.getHost().equalsIgnoreCase(bridge.getHost())` check, so add an extra check to see if live reload is enabled, and if so, use the proxy instead of loading local assets.
Also split each check in different functions to make it more readable.

closes https://github.com/ionic-team/capacitor/issues/4125


dwieeb (approved):

carlpoole (approved):
Looks great

PR #4172, opened 2021-02-08
refactor(android): reduce visibility for methods not intended to be used in Capacitor API

carlpoole:


carlpoole:
Closing for now will revisit

PR #4200, opened 2021-02-12
feat(cli): do not require webDir when server.url is set

dwieeb:
The Capacitor CLI developer experience can be improved when using livereload by not requiring the web asset directory to exist if `server.url` is being used. 

BEFORE:

image

AFTER:

image

### DIFF
--- a/cli/src/cordova.ts
+++ b/cli/src/cordova.ts
@@ -1,6 +1,7 @@
 import {
   copy,
   ensureDir,
+  mkdirp,
   pathExists,
   readFile,
   remove,
@@ -29,7 +30,6 @@ import {
   getPlugins,
   printPlugins,
 } from './plugin';
-import { copy as copyTask } from './tasks/copy';
 import { resolveNode } from './util/node';
 import { buildXmlElement, parseXML, readXML, writeXML } from './util/xml';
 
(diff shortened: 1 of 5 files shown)


cli/src/cordova.ts
```diff
@@ -291,9 +291,9 @@ export async function handleCordovaPluginsJS(
   config: Config,
   platform: string,
 ): Promise<void> {
-  if (!(await pathExists(await getWebDir(config, platform)))) {
-    await copyTask(config, platform);
```
dwieeb, line 295: I believe we were running `copyTask` here simply so that the directory existed before we started writing the Cordova JS files, but I don't think we need to run a full copy, do we?

jcesarmobile: It was sort of a fail safe in case the user cloned a repository and ran update without running copy first, not just to prevent the error.

But in this PR you have added the copyTask execution to both ios and android update commands if the folder doesn't exists, so, do we want that to happen?

dwieeb: I think we do when running `update`. Eventually I'd like to see both `copy` and `update` deprecated and combined into `sync`. They're confusing and not very useful as separated commands IMO. Also, they do a bunch of stuff with Cordova that doesn't really fit into either.

jcesarmobile (approved):

Issue #4198, opened 2021-02-12
bug: Capacitor 3 adds the 'cordova' and 'hybrid' platforms to web


lincolnthree:
### Capacitor Version

```
sharktop:example lincoln$ npx cap doctor
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 2.4.6
  @capacitor/core: 2.4.6
  @capacitor/android: 2.4.6
  @capacitor/ios: 2.4.6

Installed Dependencies:

  @capacitor/cli: 3.0.0-beta.2
  @capacitor/core: 3.0.0-beta.2
  @capacitor/android: 3.0.0-beta.2
  @capacitor/ios: 3.0.0-beta.2


```

### Platform(s)

Web, PWA

### Current Behavior

Fairly simple issue. After upgrading to Capacitor 3.x, we noticed that Native plugin wrappers were attempting (and failing) to activate in our Web & PWA platform builds, across the board.

After a night of investigation, we've determined that Capacitor is incorrectly adding the 'cordova' and 'hybrid' platforms when running in the 'web' platform.

See screenshot of printed platforms running with `ionic serve --port 8101` (port 8100 was taken):

![image](https://user-images.githubusercontent.com/362329/107789462-9f5cb980-6d1f-11eb-8bcc-2a114fb638b5.png)

### Expected Behavior

As with Capacitor 2.4.6, we'd expect the additional platforms to be added only when running in the actual environments to which they represent. Like so: 

![image](https://user-images.githubusercontent.com/362329/107789722-f5c9f800-6d1f-11eb-9f97-2c9417868407.png)

### Code Reproduction

##Demonstrates the breaking behavior:
```
git clone git@github.com:lincolnthree/capacitor-3-platforms-incorrect.git
npm i
ionic serve
# Open browser dev tools and view printed platforms in console
```
https://github.com/lincolnthree/capacitor-3-platforms-incorrect

##Demonstrates the expected behavior:
```
git clone git@github.com:lincolnthree/capacitor-3-platforms-incorrect.git
git checkout 2.4.6
npm i
ionic serve
# Open browser dev tools and view printed platforms in console
```
https://github.com/lincolnthree/capacitor-3-platforms-incorrect/tree/2.4.6

### Other Technical Details

`npm --version` output: 6.14.9

`node --version` output: v14.15.0

`pod --version` output (iOS issues only):

### Additional Context

[...]

PR #4214, opened 2021-02-16
fix(core): do not add window.cordova on web apps

elylucas:
I wrapped the logic in the legacy handlers to only add the empty Cordova object and the native event handlers to only be executed when it is running on a device. This fixes the issue with Ionic Core wrongly identifying web apps as being identified as Cordova and hybrid apps as brought up in issue #4198.

### DIFF
--- a/core/src/tests/legacy.spec.ts
+++ b/core/src/tests/legacy.spec.ts
@@ -7,7 +7,7 @@ import { WebPlugin } from '../web-plugin';
 describe('legacy', () => {
   let win: WindowCapacitor;
   let cap: CapacitorGlobal;
-  const LegacyWebPlugin = class extends WebPlugin {};
+  const LegacyWebPlugin = class extends WebPlugin { };
   const orgConsoleWarn = console.warn;
   const noop = () => {
     // do nothing
@@ -102,8 +102,8 @@ describe('legacy', () => {
 
   it('doc.addEventListener backbutton', done => {
     const AppWeb = class {
-      async addListener(eventName: string) {
-        expect(eventName).toBe('backButton');
+      async addListener(event: any) {
+        expect(event.eventName).toBe("backButton");
         done();
       }
     };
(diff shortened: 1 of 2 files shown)


core/src/tests/legacy.spec.ts
```diff
@@ -102,8 +102,8 @@ describe('legacy', () => {
 
   it('doc.addEventListener backbutton', done => {
     const AppWeb = class {
-      async addListener(eventName: string) {
-        expect(eventName).toBe('backButton');
+      async addListener(event: any) {
+        expect(event.eventName).toBe('backButton');
```
dwieeb, line 106: Why was this change necessary? `addListener()` methods within plugin classes shouldn't have this signature.

elylucas: when I added the androidBridge to the window object (line119), it changed the result that was coming back from the method from just a string to the object.

lincolnthree:
Glad I'm not crazy. I thought I was losing my mind for a while, trying to track down what I'd done that had caused this to start happening.

dwieeb (approved):

dwieeb:
@lincolnthree Thank you for reporting it!

lincolnthree:
Hey @dwieeb, did this make it into the beta.3 build? Tried it out and it looks like the compiled sources do not include the fix:

image

```
sharktop:topdecked-unified lincoln$ npx cap doctor
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 2.4.6
  @capacitor/core: 2.4.6
  @capacitor/android: 2.4.6
  @capacitor/ios: 2.4.6

Installed Dependencies:

  @capacitor/cli: 3.0.0-beta.3
  @capacitor/core: 3.0.0-beta.3
  @capacitor/android: 3.0.0-beta.3
  @capacitor/ios: 3.0.0-beta.3

sharktop:topdecked-unified lincoln$ 
```

lincolnthree:
Actually hang on. I think there's some weird caching going on here. I see the updated dist/ in node_modules, but the `ionic serve` code does not seem to reflect the change, even with cache disabled in browser.

lincolnthree:
Okay, confirmed! This is working :) Was a weird NPM multi-version precedence thing... Thanks again!

Issue #4030, opened 2021-01-07
Clean up `npx cap add` hints in the CLI


dwieeb:
We suggest running `npx cap add` in 10+ places in the CLI, but we should probably just link to the docs.

PR #4216, opened 2021-02-16
chore: clean up npx cap add hints

jcesarmobile:
Removed the `platforms.length === 0` checks as we always return at least web as platform, so they will never be 0.
Removed the `checkIOSProject` checks as the ios folder was already being checked by `selectPlatforms` function, so that code was not reachable.
Replaced the npx cap add instructions on `init` with a link to `https://capacitorjs.com/docs/v3/getting-started#where-to-go-next`

closes https://github.com/ionic-team/capacitor/issues/4030

### DIFF
--- a/cli/src/common.ts
+++ b/cli/src/common.ts
@@ -299,9 +299,7 @@ export async function selectPlatforms(
       }
       fatal(
         `${c.strong(platformName)} platform has not been added yet.\n` +
-          `Use ${c.input(
-            `npx cap add ${platformName}`,
-          )} to add the platform to your project.`,
+          `https://capacitorjs.com/docs/v3/${platformName}#adding-the-${platformName}-platform`,
       );
     }
 
--- a/cli/src/common.ts
+++ b/cli/src/common.ts
@@ -304,7 +304,11 @@ export async function selectPlatforms(
       }
       fatal(
         `${c.strong(platformName)} platform has not been added yet.\n` +
-          `https://capacitorjs.com/docs/v3/${platformName}#adding-the-${platformName}-platform`,
+          `See the docs for adding the ${c.strong(
+            platformName,
+          )} platform: ${c.strong(
+            `https://capacitorjs.com/docs/v3/${platformName}#adding-the-${platformName}-platform`,
+          )}`,
       );
     }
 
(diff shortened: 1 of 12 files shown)


cli/src/tasks/init.ts
```diff
@@ -166,10 +166,10 @@ async function mergeConfig(
 function printNextSteps(newConfigName: string) {
   logSuccess(`${c.strong(newConfigName)} created!`);
   output.write(
-    `\nAdd platforms using ${c.input('npx cap add')}:\n` +
-      `  ${c.input('npx cap add android')}\n` +
-      `  ${c.input('npx cap add ios')}\n\n` +
-      `Follow the Developer Workflow guide to get building:\n${c.strong(
+    `\nCheck next steps: \n${c.strong(
+      `https://capacitorjs.com/docs/v3/getting-started#where-to-go-next`,
```
dwieeb, line 170: I wonder if we should just link to the iOS and Android platform guides directly.

jcesarmobile: the "where to go next" has links to the iOS guide, the Android guide and the workflow, so didn't want to add that many links. In fact, I could probably remove the workflow link since it's included in the "where to go next"

dwieeb: 👍

cli/src/tasks/init.ts
```diff
@@ -166,11 +166,8 @@ async function mergeConfig(
 function printNextSteps(newConfigName: string) {
   logSuccess(`${c.strong(newConfigName)} created!`);
   output.write(
-    `\nAdd platforms using ${c.input('npx cap add')}:\n` +
-      `  ${c.input('npx cap add android')}\n` +
-      `  ${c.input('npx cap add ios')}\n\n` +
-      `Follow the Developer Workflow guide to get building:\n${c.strong(
-        `https://capacitorjs.com/docs/v3/basics/workflow`,
-      )}\n`,
+    `\nCheck next steps: \n${c.strong(
```
dwieeb, line 169:
```suggestion
    `\nNext steps: \n${c.strong(
```

cli/src/common.ts
```diff
@@ -304,9 +304,7 @@ export async function selectPlatforms(
       }
       fatal(
         `${c.strong(platformName)} platform has not been added yet.\n` +
-          `Use ${c.input(
-            `npx cap add ${platformName}`,
-          )} to add the platform to your project.`,
+          `https://capacitorjs.com/docs/v3/${platformName}#adding-the-${platformName}-platform`,
```
dwieeb, line 307: Links should use `c.strong()` and have some context like "See these docs: <link>"

PR #4219, opened 2021-02-18
feat(core): dont overwrite Capacitor props when using custom platform

IT-MikeS:
For custom platform support, if the custom platform is injecting overrides for the runtime, don't let the Web runtime overwrite those changes at runtime via a custom flag on the Capacitor object.


IT-MikeS:
Going to close this, after discussing with @adamdbradley in favor of the functionality outlined in this issue (#4243)

Issue #4223, opened 2021-02-19
bug: [Cap 3 beta] cap init fails in Cordova project with config.xml file


dotNetkow:
### Capacitor Version
```
3.0.0-beta.3
```

### Current Behavior
Add Capacitor 3 to an existing Cordova project ([this demo app](https://github.com/ionic-team/demo-authconnect-auth0)) with a valid config.xml file then run `npx cap init`.

After answering all prompts:
- name = demo-authconnect-auth0
- package id = io.ionic.demo.ac.auth0
- web asset dir = www

it fails with: **[error] Unable to read: [local path]/demo-authconnect-auth0/config.xml**
and capacitor.config.json is not created.

### Expected Behavior
The `init` command should complete successfully.

### Code Reproduction
Clone this repo: https://github.com/ionic-team/demo-authconnect-auth0

### Other Technical Details

`npm --version` output: 6.14.6
`node --version` output: v14.5.5

jcesarmobile:
Interesting, I can reproduce with beta 3, but not with the local copy that should be exactly the same. @dwieeb thoughts?
In fact, I can reproduce with all the betas.

PR #4224, opened 2021-02-19
fix(cli): init failure if config.xml is present

jcesarmobile:
For some unknown reason, `parseStringPromise` is failing only when you install from npm or from npm pack, but not if installed locally or linked

The PR uses the callback based `parseString` that seems to work fine.

closes https://github.com/ionic-team/capacitor/issues/4223


dwieeb:
Alternative: https://github.com/ionic-team/capacitor/pull/4227

jcesarmobile:
closing in favor of https://github.com/ionic-team/capacitor/pull/4227

Issue #4159, opened 2021-02-04
android: update internal permission/activity results to use new Activity Result API


carlpoole:
Some internal permission requests in Capacitor are still using request codes. Update for AndroidX Activity Result API.

PR #4238, opened 2021-02-22
refactor: Use the new Activity Result API in BridgeWebChromeClient

jcesarmobile:
Use the new Activity Result API in BridgeWebChromeClient instead of using Cordova plugins.

closes https://github.com/ionic-team/capacitor/issues/4159


carlpoole (approved):
Nice!

PR #4239, opened 2021-02-22
chore: Update androidxActivityVersion and androidxFragmentVersion

jcesarmobile:
both got final versions, so better ship them instead of the rc ones


carlpoole (approved):
Great

PR #4250, opened 2021-02-25
ci: Update Xcode versions

jcesarmobile:
Capacitor 3 is supposed to only support Xcode 12, so I've removed Xcode 11.7 and added 12.4

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -84,7 +84,6 @@ jobs:
       matrix:
         xcode:
           - /Applications/Xcode_12.4.app
-          - /Applications/Xcode_12.2.app
     steps:
       - run: sudo xcode-select --switch ${{ matrix.xcode }}
       - uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


ikeith:
Is there a reason to keep 12.2 in the list? This was originally setup to build with the latest version of 11 and 12 so dropping 11 makes sense (as does updating the version of 12). But I'm not aware of a reason to keep two versions of 12 running. We probably want to stay with just the latest until Xcode 13 comes out.

jcesarmobile:
we used to have different versions, not just latest, but usually one was a beta and another one a final.
I can remove 12.2, not sure if 14.5 (beta) is available, will give it a try and if not present I'll remove it.

jcesarmobile:
macos-latest is still using macOS 11.15, so Xcode 12.5 can't be used, will just remove 12.2 then
https://github.com/actions/virtual-environments#available-environments

ikeith:
12.5 requiring Big Sur (once we can run it in CI) could be an argument to keep 12.4 around since there might be users who can't/won't upgrade their OS. But I don't see why we need more than one version at this time.

ikeith (approved):

Issue #4243, opened 2021-02-24
feat: Create a runtime Capacitor addPlatform() api


adamdbradley:
### Description

Right now ios and android are baked into the runtime, which is fine and will probably stay that way. However, it should be easier for other platforms to add themselves to the Capacitor runtime, such as Electron.

### Preferred Solution

Something along the lines of:

```
import { addPlatform } from '@capacitor/core';

addPlatform({
  name: 'electron',
  isNativePlatform: () => !!window.electron,
  ...
});
```

And this same API can be what's used for ios and android internally. I think we'll want to keep a map of all the platforms available, and when the top level API, such as `Capacitor.getPlatform()` is called, it loops through each of the platforms to see if it provided its own API or not, otherwise falling back to what it's doing today.

IT-MikeS:
This API would also allow for the override of other platform functions such as `isPluginAvailable`. 

For example in electron, we would want electron Plugins first but all web ones also work just fine, so the above function should return electron first then web if possible.

thomasvidas:
Merged in #4255

PR #4255, opened 2021-02-27
feat(core): platforms api

IT-MikeS:
Start of a more clean platforms api, for adding and setting platforms. Addresses #4243


core/src/runtime.ts
```diff
@@ -22,17 +23,20 @@ export interface RegisteredPlugin {
...
-  const getPlatform = () => getPlatformId(win);
+  const defaultGetPlatform = () => getPlatformId(win);
+  const getPlatform = capPlatforms.currentPlatform.getPlatform || defaultGetPlatform;
 
-  const isNativePlatform = () => getPlatformId(win) !== 'web';
+  const defaultIsNativePlatform = () => capPlatforms.currentPlatform.name !== 'web';
+  const isNativePlatform = capPlatforms.currentPlatform.isNativePlatform || defaultIsNativePlatform;
 
-  const isPluginAvailable = (pluginName: string): boolean => {
+  const defaultIsPluginAvailable = (pluginName: string): boolean => {
```
adamdbradley, line 39: Why does this have to get renamed?

core/src/runtime.ts
```diff
@@ -22,17 +23,20 @@ export interface RegisteredPlugin {
...
   const cap: CapacitorInstance = win.Capacitor || ({} as any);
+  const capPlatforms: CapacitorPlatformsInstance = win.CapacitorPlatforms;
 
   const Plugins = (cap.Plugins = cap.Plugins || ({} as any));
 
   const webviewServerUrl =
     typeof win.WEBVIEW_SERVER_URL === 'string' ? win.WEBVIEW_SERVER_URL : '';
 
-  const getPlatform = () => getPlatformId(win);
+  const defaultGetPlatform = () => getPlatformId(win);
```
adamdbradley, line 33: Not sure I follow why we need to have a `currentPlatform` and rename some of the apis to `default`

adamdbradley:
Since this PR there has been a significant internal update to how the native bridge works. Overall I think it's a good update, but would rather not rename any of the public APIs we already have established. Would you be able to rebase from master and not rename any of the public APIs? Thanks

core/src/runtime.ts
IT-MikeS: The defaults are not renaming the consts that are exposed, they are just holding the default version of that function so in the case of a custom platform not overriding a piece then the runtime will have the default to fall back on

IT-MikeS:
Did a rebase and a quick clean up

thomasvidas (approved):

thomasvidas:
Looks great! If you need any help with getting https://github.com/capacitor-community/electron/ updated I'm happy to help. Just DM me on the community slack 😄

IT-MikeS:
> Looks great! If you need any help with getting https://github.com/capacitor-community/electron/ updated I'm happy to help. Just DM me on the community slack 😄

Thanks! I'm pretty well done with the update but if anything comes up I'll be sure to take you up on that 😁

Issue #4157, opened 2021-02-04
feat: Add configurable app path urls to support reuse of web app in Embedded Capacitor


carlpoole:
### Description
Capacitor currently supports setting the app url to load a web app into the web view (1 web app to 1 web view). This allows multiple different web apps to be bundled in the app and load in different web views the native app.

Capacitor should also allow the developer to set a path to load a section of the provided web app. This would allow multiple different web views in the app to use the same web app assets, and enter into different areas of the same app.

### Platform(s)
Android, iOS

### Preferred Solution
Add a configurable app path url

### Additional Context
https://github.com/ionic-team/capacitor/pull/3405

dwieeb:
Duplicate of https://github.com/ionic-team/capacitor/issues/3912

PR #4264, opened 2021-03-01
feat(android): add configurable app path for embedded capacitor

carlpoole:
This PR adds a configurable path string to load a specific initial path for a Capacitor app.

See #3405 and #4157

Example usage:

```java
BridgeFragment embeddedFragment = BridgeFragment.newInstance("app1");
CapConfig config = new CapConfig.Builder(this)
            .setPath("/feature_1")
            .create();
        
embeddedFragment.setConfig(config);
```


android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -135,6 +137,7 @@ private void deserializeConfig(@Nullable Context context) {
         html5mode = JSONUtils.getBoolean(configJSON, "server.html5mode", html5mode);
         serverUrl = JSONUtils.getString(configJSON, "server.url", null);
         hostname = JSONUtils.getString(configJSON, "server.hostname", hostname);
+        path = JSONUtils.getString(configJSON, "server.path", path);
```
ikeith, line 140: My understanding is that this was going to be a run-time only config, not supported in the config json?

carlpoole: Yeah I don't think it makes sense for now. Removed

ikeith (approved):

jcesarmobile (approved):

fkirc:
Thanks for implementing this much-needed feature!
I will migrate to V3 as soon as those runtime-paths are working for both Android and iOS.
Until now, I was using a path-redirect-workaround via a custom plugin, but this path-redirect-workaround was only working after an initial rendering of the app.

PR #4268, opened 2021-03-02
feat(android): set WebChromeClient geolocation permission strings in Activity or Plugin

carlpoole:
On Android, Google Chrome calls into BridgeWebChromeClient when the web app tries to use the web geolocation API to get permission information (not native mobile code).

This PR removes the geolocation permission strings from BridgeWebChromeClient and requires to use `registerGeolocationPermissions` from their Activity, or plugin authors to use `registerGeolocationPermissions` from their Plugin class if they wish to use the web geolocation API from the web app used in Capacitor.

Google has started scanning apps for their permission usage and identifies all Capacitor apps as using geolocation services, even if they are not explicitly using it in the web app. This can result in Play Store submission rejections if Google does not think the location permission usage is appropriate, such as in apps targeted towards children.

Usage examples:

From an Activity
```java
@Override
public void onStart() {
    super.onStart();
    registerGeolocationPermissions(() -> new String[]{ 
        Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION 
    });
}
```

From a Plugin
```java
@Override
public void load() {
    implementation = new Geolocation(getContext());
    registerGeolocationPermissions(() -> new String[] { 
        Manifest.permission.ACCESS_COARSE_LOCATION, Manifest.permission.ACCESS_FINE_LOCATION 
    });
}
```

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -92,6 +92,7 @@ public class Bridge {
     private CordovaWebView cordovaWebView;
     private CordovaPreferences preferences;
     private BridgeWebViewClient webViewClient;
+    private BridgeWebChromeClient webChromeClient;
     private App app;
 
     // Our MessageHandler for sending and receiving data to the WebView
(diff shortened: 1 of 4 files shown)


jcesarmobile:
Looks like it's safe to have the old code as long as the geolocation permissions are not in the apps `AndroidManifest.xml`, so going to close.

PR #4281, opened 2021-03-03
fix(android): calls re-saved during permission/activity result callbacks were being released

### LINKED ISSUE #4277: bug: Attempt to invoke virtual method 'boolean com.getcapacitor.PluginCall.isReleased()' on a null object reference
### Capacitor Version
beta 6

```
PASTE OUTPUT HERE
```

### Platform(s)
android

### Current Behavior
test app crash on after first install if clicking the "take picture and save" button, after the permission prompt, with error

`java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.getcapacitor.PluginCall.isReleased()' on a null object reference`

### Expected Behavior

to not crash

### Code Reproduction

https://github.com/ionic-team/capacitor-testapp

### Other Technical Details

`npm --version` output:

`node --version` output:

carlpoole:
A bug was observed (resolves #4277) where if permission was requested and the callback contained another permission/activity launch with a result, the app would crash claiming the savedCall was null.

What was happening was the call was originally saved for the first permission request, then during the callback the call would save again for the following activity result. The initial permission request would then try to cleanup after the callback was called and remove the saved call that was still being used by the waiting callback, resulting in it being null when the second callback returned. 

Fix: I moved the cleanup of the savedCall before the callback is triggered so that if the same PluginCall is saved again, it won't be accidentally removed.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Plugin.java
@@ -140,15 +140,15 @@ public class Plugin {
 
         // validate permissions and invoke the permission result callback
         if (bridge.validatePermissions(this, savedCall, permissionResultMap)) {
+            if (!savedCall.isKeptAlive()) {
+                savedCall.release(bridge);
+            }
+
             try {
                 method.setAccessible(true);
                 method.invoke(this, savedCall);
             } catch (IllegalAccessException | InvocationTargetException e) {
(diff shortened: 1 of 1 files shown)


theproducer (approved):

jcesarmobile (approved):
works fine, I hope calling release before the actual method invoke doesn't bring weird edge cases in the future

PR #4299, opened 2021-03-05
feat(cli): allow async in capacitor config file

buschtoens:
This allows users to generate their config asynchronously.

```ts
// capacitor.config.ts
import type { CapacitorConfig } from '@capacitor/cli';

async function generateConfig(): CapacitorConfig {
  // ...
}

export default generateConfig();
```


alfredosalzillo:
Would be better if instead of exporting a Promise as default it's allowed to export function and async function ?

```
export default async function () {
  ....
  return yourconfig
}
```

buschtoens:
Another, possibly more standard / intuitive, option is using [top-level `await`](https://v8.dev/features/top-level-await).

```ts
// capacitor.config.ts
import type { CapacitorConfig } from '@capacitor/cli';

const appId = await determineAppId();

  appId,
  appName: 'My Capacitor App',
  // ...
};

export default config;
```

ZaLiTHkA:
while working on a "whitelabel" project using Ionic 6 with Angular 14 and Capacitor 4, I ran into an issue with trying to retrieve the current dev machine's primary network interface.. the only NPM packages that I found with enough "business logic" to sufficiently retrieve and inspect the current host's network interfaces all work with Promises, whereas those with pure synchronous logic simply don't get the correct information. **this** behaviour would have solved that issue for me.

I see this PR has been open and untouched for a long time now, so just to check: is there any update on this idea? or do we simply need to find (or make) another plan for cases like mine?

jcesarmobile:
@giralte-ionic @theproducer @markemer @dallastjames 
thoughts on this?
Having a .ts/.js config file but not being able to use async code is very limiting.
I tried in the past to add top level await support, which would be the ideal feature to have, but I was not able to make it work, so this would be at least a big improvement and it's simple enough.

markemer:
> @giralte-ionic @theproducer @markemer @dallastjames thoughts on this? Having a .ts/.js config file but not being able to use async code is very limiting. I tried in the past to add top level await support, which would be the ideal feature to have, but I was not able to make it work, so this would be at least a big improvement and it's simple enough.

I mean, it seems harmless enough to me with a useful feature addition.

jcesarmobile (approved):

PR #4303, opened 2021-03-06
Support retrieval of NSNumber values from CAPPluginCall options

lincolnthree:
There is no current shortcut to retrieve `NSNumber` values from CAPPluginCall options (only NSString and BOOL are currently supported). 

This PR adds the missing convenience method which is required for retrieving numeric and numeric enum values passed through native plugins via the IOS capacitor bridge.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPPlugin.m
+++ b/ios/Capacitor/Capacitor/CAPPlugin.m
@@ -36,6 +36,19 @@
   return TRUE;
 }
 
+-(NSNumber *) getNumber:(CAPPluginCall *)call field:(NSString *)field defaultValue:(NSNumber *)defaultValue 
+{
+  id idVal = [call.options objectForKey:field];
+  if(![idVal isKindOfClass:[NSNumber class]]) {
+    return defaultValue;
+  }
+  NSNumber *value = (NSNumber *)idVal;
+  if(value == nil) {
+    return defaultValue;
+  }
+  return value;
+}
+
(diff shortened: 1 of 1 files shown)


lincolnthree:
Forgive me if this PR does not conform to process. It was submitted via Github UI. Happy to update but I am not an iOS developer so this is about as good as it gets from me so far. Happy to learn :)

ikeith:
Thank you for the PR!

This PR highlighted the fact that these accessors on `CAPPlugin` had been overlooked so we appreciate you bringing that to our attention. They are deprecated going forward and this will be reflected in the next release. Our preferred pattern is to have the accessors on `CAPPluginCall` instead. That is a simpler and more ergonomic API than these methods where you have to pass the call as an argument since the call is the object that actually holds the data.

If you are using Swift, [the accessors can be found here](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/JSTypes.swift).

If you are using Obj-C, [the accessors can be found here](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/CAPBridgedJSTypes.h). These now contain the `getNumber` method you need. (these do require an extra import statement since the methods are not exposed by default to avoid conflicts with the Swift implementations)

So instead of:
 `[self getNumber:call field:@"foo" defaultValue:@1];`
you would use:
`[call getNumber:@"foo" defaultValue:@1];`

lincolnthree:
Thanks, @ikeith - Sounds good. We will update :) Keep up the good work! Love the project.

PR #4309, opened 2021-03-09
feat(iOS): Obj-C convenience accessors on CAPPluginCall

ikeith:
This branch revisits the Obj-C versions of the CAPPluginCall convenience accessors, deprecates the redundant methods on CAPPlugin, adds test coverage, and fixes a bug with dates.

- Adds a `getNumber` method for NSNumber values.
- Switches the redundant `CAPPlugin` accessors to use the call's implementation.
- Marks the redundant `CAPPlugin` accessors as deprecated.
- Fixes the incorrect handling of NSDate objects in `getDate`.
- Adds unit tests to verify behavior.


jcesarmobile (commented):
Shouldn't the bridged types have a getBool like the swift plugin call has?
Having to use NSNumber every time you want a boolean might make users use the deprecated getBool from CAPPlugin class instead as it's easier to get despite it's deprecated.

ikeith:
@jcesarmobile
Swift cares about types more which is why it has more accessors. But we can add a `getBool`, although I'm not sure how helpful it is since you can use `boolValue` and NSNumber:
`[call getBool:@"foo" defaultValue:TRUE];`
vs
`[[call getNumber:@"foo" defaultValue:@TRUE] boolValue];

ikeith:
That's a fair point. It probably makes sense to ease the migration from the deprecated methods.

Issue #4307, opened 2021-03-08
bug: Attempt to invoke virtual method 'boolean com.getcapacitor.PluginCall.isKeptAlive()' on a null object reference


jcesarmobile:
### Capacitor Version

main

```
PASTE OUTPUT HERE
```

### Platform(s)

android

### Current Behavior

java.lang.NullPointerException: Attempt to invoke virtual method 'boolean com.getcapacitor.PluginCall.isKeptAlive()' on a null object reference

### Code Reproduction

in the sample app, on Camera page change `allowEditing` to `true` and use the "Choose Picture" button, app will crash after you edit the picture

### Other Technical Details

`npm --version` output:

`node --version` output:

`pod --version` output (iOS issues only):

PR #4318, opened 2021-03-10
fix(android): Release the call after reject/resolve

jcesarmobile:
Capacitor calls are being saved when an intent is being launched and then released after the intent returns a result, but before the plugin returns the result, and in cases where there are multiple intents launched by the same plugin call it ends up creating null pointer exceptions in some cases where the call was released sooner than expected and needed later.

This PR removes the early releases of the calls and put it where the call actually rejects or resolves to the JS side.

closes https://github.com/ionic-team/capacitor/issues/4307

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/MessageHandler.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/MessageHandler.java
@@ -109,6 +109,9 @@ public class MessageHandler {
         } catch (Exception ex) {
             Logger.error("sendResponseMessage: error: " + ex);
         }
+        if (!call.isKeptAlive()) {
+            call.release(bridge);
+        }
     }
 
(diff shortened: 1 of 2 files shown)


carlpoole:
Will this not regress this fix if the call is saved again before it is resolved? https://github.com/ionic-team/capacitor/pull/4281

jcesarmobile:
I tried and doesn't look like it, if the call is saved again, since it's the same call, with same id, and savedCalls is a `Map`, it should just save the same call on the `Map` with no side effects as Maps don't allow duplicate keys.

carlpoole:
I understand this now. I do like this a lot better. So basically the life of the call lasts until it is resolved/rejected (aka finished) and if you want the call to last past that, use the keepAlive setting. And it works because the call isn't resolved until the native work has been completed - which can last through multiple intents.

carlpoole (approved):
Tested and seems to work great

PR #4344, opened 2021-03-15
feat(ios): add credentials to webview for auth challenge

VictorKachalov:
If the hybrid approach is required when Capacitor is an addition to the native app, sometimes the Web App may request basic auth in order to allow the connection. In order not to inject the credentials inside the capacitor.config.json file, there should be an opportunity to provide credentials in a native way when the callback 

`webView(_ webView: WKWebView, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void)`

is called (for iOS).

### DIFF
--- a/ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
+++ b/ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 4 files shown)


jcesarmobile:
Thanks for the Pull Request.
We should expose a way for users to implement/overwrite some of the WKNavigationDelegate methods themselves or with plugins, but your changes won't be added to Capacitor, sorry.

Maybe you can create a plugin similar to this one I created for ignoring SSL errors and use method swizzling to provide your own implementation of `didReceiveAuthenticationChallenge`
https://github.com/jcesarmobile/ssl-skip/blob/main/ios/Plugin/SslSkipPlugin.m

PR #4346, opened 2021-03-16
refactor: getPlatform to return string literal

mesqueeb:


mesqueeb:
@jcesarmobile any feedback on my PR? I think it would be a great value to add before release v3

jcesarmobile:
Since it's possible to add custom platforms, this would prevent those platforms from getting a correct name, in example if you add @capacitor-community/electron, it will return electron and with this changes it won't be a valid value, so I think we will keep the string

PR #4363, opened 2021-03-18
feat(ios): swift package manager support

VictorKachalov:
If the hybrid approach is required when Capacitor is an addition to the native app, sometimes not all native apps have a cocoapods support. Since the Capacitor and CapacitorCordova have their own Podspecs and are published on cocoapods.org in order to add the dependency without npm and directly to the iOS project, it will be great to have such support for Swift Package Manager


ikeith:
Thank you for the PR!

Do you have an example of this working? The Package files appear invalid because the paths should be relative to the `Package.swift` file and Capacitor's dependency on CapacitorCordova is not described within them.

However, the core problem is that projects like Capacitor will quickly hit errors of the type `target at <path> contains mixed language source files; feature not supported`. We've looked into adding SPM support but each `target` in the package must be a single language, either Swift or a C language. That means that the Swift files and the Obj-C files need to be in separate targets but the circular dependencies between them prevent them from compiling independently. It would require a significant refactoring to address it, if it is even possible.

VictorKachalov:
> Thank you for the PR!
> 
> Do you have an example of this working? The Package files appear invalid because the paths should be relative to the `Package.swift` file and Capacitor's dependency on CapacitorCordova is not described within them.
> 
> However, the core problem is that projects like Capacitor will quickly hit errors of the type `target at <path> contains mixed language source files; feature not supported`. We've looked into adding SPM support but each `target` in the package must be a single language, either Swift or a C language. That means that the Swift files and the Obj-C files need to be in separate targets but the circular dependencies between them prevent them from compiling independently. It would require a significant refactoring to address it, if it is even possible.

Hi @ikeith 

thank you for your response. I made some adjustment and moved the package directly to the folder where .git folder is presented. Howerver I receive the error you mention `target at <path> contains mixed language source files; feature not supported`. 

I have a question, is it possible to install Capacitor via Carthage?

ikeith:
@kvs As far as I know, there's no reason Carthage couldn't work but I've not tried it. Adding support is not something that's been a concern as CocoaPods works for the expected use case.

Neil3079:
Just want to add that there is very good reason to move away from Cocoapods and towards Swift Package Manager;

1. SPM is native tooling and is the desired approach moving forwards not just of Apple and the Swift Foundation but also native iOS Developers everywhere.

2. New third party dependancies are already dropping Cocoapods support and just supporting SPM directly

3. Creating modular apps such would benefit Ionic style applications is better defined and much simpler with SPM

4. If Ionic wants to get a foothold in existing applications switching the dependancy management framework from SPM to cocoa pods will be a huge point of resistance.

Issue #4396, opened 2021-03-24
bug: v3 run ios fails with wireless device


alexcroox:
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 2.4.7
  @capacitor/core: 2.4.7
  @capacitor/android: 2.4.7
  @capacitor/ios: 2.4.7

Installed Dependencies:

  @capacitor/cli: 3.0.0-rc.0
  @capacitor/android: 3.0.0-rc.0
  @capacitor/core: 3.0.0-rc.0
  @capacitor/ios: 3.0.0-rc.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Current Behavior

`npx cap run ios` returns list of devices, if the device is connected wirelessly then the device ID is undefined and triggers an error when selecting.

![image](https://user-images.githubusercontent.com/1065835/112303983-d87c3800-8c94-11eb-920b-c5fbcecedf43.png)

The first `Alex iPhone 12` only appears when I plug the phone in via a cable. The undefined one I believe is the wifi version.

![image](https://user-images.githubusercontent.com/1065835/112303998-e0d47300-8c94-11eb-9829-5e0335b0f1c4.png)

### Expected Behavior

No error or at least filter the wireless device from the list if wifi installing of the app is not supported

PR #4397, opened 2021-03-24
fix(cli): filter targets without id from run device list

jcesarmobile:
If the device doesn't have an id it fails to run, so filter out the devices without id from the run list

closes https://github.com/ionic-team/capacitor/issues/4396


thomasvidas (approved):

tonycesar:
Now by the console show the error message:  Invalid target ID: ?.
How can I connect to device connected by wifi

PR #4402, opened 2021-03-28
fix(android): resolve issue with activity result API registration for fragments

carlpoole:
Prerequisite: #4405

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,9 +117,9 @@ jobs:
       - run: npm install
       - run: npm run verify
         working-directory: ./android
-  deploy:
+  deploy-next:
     runs-on: macos-latest
-    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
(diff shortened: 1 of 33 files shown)


jcesarmobile (commented):
I think we could have a helper on Bridge class that help us get the activity or fragment, something like this:

```
    public ActivityResultCaller getResultCaller() {
        if (this.fragment != null) {
            return this.fragment;
        } else {
            return this.context;
        }
    }
```

This way we can avoid having the fragment null check and having to duplicate the code that is called once with `getFragment` and another one with `getActivity when `registerForActivityResult` is used

carlpoole:
> I think we could have a helper on Bridge class that help us get the activity or fragment that is going to register for result

Thanks for the idea, this does make it cleaner. Take a look what I added

jcesarmobile:
Looks good, can you use it in BridgeWebChromeClient too?

thomasvidas (approved):

jcesarmobile (approved):
looks good!

PR #4405, opened 2021-03-29
feat(android): ability to add listeners to the Capacitor WebView

carlpoole:
Adds the ability to register a callback to execute when certain events occur in the WebView in Capacitor. This PR includes a method to add an event when the page is finished loading. In the future this can also be used to add page error event callbacks, for instance.

Usage example: masking the view and fading it out when the web app is loaded, to hide ugly gaps in app loading

```
addWebViewListener(new WebViewListener() {
            @Override
            public void onPageLoaded(WebView webView) {
                fadeView.animate()
                    .alpha(0f)
                    .setDuration(duration)
                    .setListener(new AnimatorListenerAdapter() {
                        @Override
                        public void onAnimationEnd(Animator animation) {
                            fadeView.setVisibility(View.GONE);
                        }
                    });
            }});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,27 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [3.0.0-rc.0](https://github.com/ionic-team/capacitor/compare/3.0.0-beta.6...3.0.0-rc.0) (2021-03-10)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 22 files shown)


thomasvidas (approved):

thomasvidas (approved):

jcesarmobile (changes requested):
This implementation only allows to add listeners from the bridge builder. I think it should allow to add (and maybe remove) listeners after the bridge has been created too, so, in example, a plugin or some custom user code in the activity class can add new listeners

jcesarmobile (approved):
looks good

PR #4415, opened 2021-03-31
WIP: inject native-bridge.js

jcesarmobile:
Still work in progress, it needs that the native-bridge.js is generated from typescript, at the moment it's a sample file to demonstrate how it works.

the build command in core copies native-bridge.js to ios and android

on android it's distributed as a regular file on assets folder that gets merged into the final app assets folder

on iOS, CocoaPods generates a bundle with the file and the code reads the file from the bundle.


jcesarmobile:
superseded by https://github.com/ionic-team/capacitor/pull/4446

PR #4416, opened 2021-03-31
feat: Unify logging behavior across environments 

ikeith:
This branch updates the rules for when/how statements should be logged at runtime to reconcile differences between JS and native implementations. Previously, in the default case of `hideLogs` in the config file being either false or missing, the native code would log statements in production builds while the JS code would not. The previous boolean flag was insufficient for describing the state(s).

- `loggingBehavior` is a new config option with possible values of `none`, `debug`, and `production`. Both JS and native code follow the same rules:
  1. If `none`, no log statements are generated.
  2. If `debug` (the default), log statements are generated in debug builds but not production builds.
  3. If `production`, log statements are generated in all builds.
- `hideLogs` is now deprecated in the config file with a warning in the CLI. If `loggingBehavior` is absent but `hideLogs` exists, `false` is treated as `debug` and `true` is treated as `none`.
- The JS bridge now has a `isLoggingEnabled` flag to avoid the conflation between debug and logging states.

Companion documentation change: https://github.com/ionic-team/capacitor-site/pull/258

Closes https://github.com/ionic-team/capacitor/issues/3723


ios/Capacitor/Capacitor/CAPInstanceDescriptor.h
```diff
@@ -63,15 +69,15 @@ NS_SWIFT_NAME(InstanceDescriptor)
...
+ @discussion Defaults to @c debug. Set by @c loggingBehavior in the configuration file but will inherit the deprecated @c hideLogs flag if @c loggingBehavior is absent.
  */
-@property (nonatomic, assign) BOOL enableLogging;
+@property (nonatomic, assign) CAPInstanceLoggingBehavior loggingBehavior;
 /**
  @brief Whether or not the web view can scroll.
  @discussion Set by @c ios.scrollEnabled in the configuration file. Corresponds to @c isScrollEnabled on WKWebView.
  */
-@property (nonatomic, assign) BOOL enableScrolling;
+@property (nonatomic, assign) BOOL scrollingEnabled;
```
ikeith, line 80: This is an unrelated change but was made for the sake of consistency. It not only matches `loggingEnabled`, it better aligns with the config file option and the actual setting on WKWebView so the original name was probably a bad choice.

jcesarmobile: Even if it's for consistency, changes like this should be made in separate PRs, then link with this one and mention that it's for consistency.
If not we can enter into pointless discussions about if we should use `scrollingEnabled` or `scrollEnabled` as the config value name or even another name, instead of focusing on the important changes.

thomasvidas:
Is there ever a situation where we'd native native logs but _not_ webview logs? Would that be an option worth exploring?

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -207,6 +213,7 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
         do {
             try JSExport.exportCapacitorGlobalJS(userContentController: webViewDelegationHandler.contentController,
                                                  isDebug: isDevMode(),
```
jcesarmobile, line 215: Unrelated to the change, but since `isDevMode` is deprecated, we should use `isDevEnvironment`

cli/src/declarations.ts
```diff
@@ -268,11 +289,22 @@ export interface CapacitorConfig {
      *
      * Overrides global `hideLogs` option.
      *
-     * @since 1.1.0
+     * @since 2.1.0
```
jcesarmobile, line 292:
```suggestion
     * @since 1.1.0
```
For iOS it was implemented on 1.1.0, since Android allows to hide logs natively, but some people requested the same configuration option for Android and was added in 2.1.0

carlpoole (approved):

jcesarmobile (changes requested):
Added a few minor comments.

I personally don't like the "production" name. There is already a lot of confusion between the web production build and the native production build. Having a "production" value looks like it's the value you should use for the production build, so can lead to users enabling logs in production builds while thinking they are disabling it.

cli/src/declarations.ts
ikeith: Ah, I didn't know the history and assumed it was a typo.

ios/Capacitor/Capacitor/CapacitorBridge.swift
ikeith: Oh, good catch.

ikeith:
> I personally don't like the "production" name. 

I'm open to ideas. "release", perhaps? I was trying to keep the values short but we could use longer, more descriptive strings like "allowInProduction" or "disabledForRelease" or something. But those could get awkward.

ikeith:
> Is there ever a situation where we'd native native logs but _not_ webview logs? Would that be an option worth exploring?

@thomasvidas I can't think of one. We could offer it but, logically, then we should also offer the inverse, i.e. web logs but not native logs. But that feels like a lot of complication for a relatively simple feature. You'll need to parse a bunch of noise to find the info you need no matter what. I think the biggest concern is leaking sensitive information by logging in release builds without realizing it (which can happen now).

jcesarmobile (approved):

thomasvidas (approved):
LGTM.

bridge.ts moved over to native-bridge.js in my PR, but I think everything else should merge in cleanly

thomasvidas (approved):

cli/src/declarations.ts
```diff
@@ -152,10 +170,21 @@ export interface CapacitorConfig {
      * Overrides global `hideLogs` option.
      *
      * @since 2.1.0
+     * @deprecated 3.0.0
      * @default false
      */
     hideLogs?: boolean;
 
+    /**
+     * The build configuration under which Capacitor will generate logs.
```
jcesarmobile, line 179: Mention that it's for Android only

cli/src/declarations.ts
```diff
@@ -269,10 +298,21 @@ export interface CapacitorConfig {
      * Overrides global `hideLogs` option.
      *
      * @since 1.1.0
+     * @deprecated 3.0.0
      * @default false
      */
     hideLogs?: boolean;
 
+    /**
+     * The build configuration under which Capacitor will generate logs.
```
jcesarmobile, line 307: mention that is for iOS only

jcesarmobile (changes requested):
two minor things

thomasvidas (approved):

jcesarmobile (approved):

carlpoole (approved):

PR #4437, opened 2021-04-06
fix(android): Avoid handling redirection response in WebViewLocalServer (#4240)

### LINKED ISSUE #4240: bug: Android web view encapsulate 302 redirection response
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 2.4.6
  @capacitor/core: 2.4.6
  @capacitor/android: 2.4.6
  @capacitor/electron: 2.4.6
  @capacitor/ios: 2.4.6

Installed Dependencies:

  @capacitor/cli 2.4.2
  @capacitor/android 2.4.2
  @capacitor/core 2.4.2
  @capacitor/ios 2.4.2
```

### Platform(s)

Only on android

### Current Behavior

When the android web view navigates to a URL that supposed to returns a 302 redirection response to a different domain, a 200 response was received with the HTML content. So that the web view does not perform any URL redirection. Instead the window URL remain unchanged. Assets (eg. images, CSS, static JS files) with relative URLs in the HTML reference the original window URL rather than the redirected URL. A broken HTML page is shown as a result of missing assets required by the HTML.

### Expected Behavior

When navigating to URL which returns a 302 response, the web view should perform the redirection instead of it being resolved by Capacitor. The web view window location should be updated to the redirected location and the assets with relative URL in the HTML should be resolved with the correct URL path

### Code Reproduction

Navigating to any URL which returns a 302 redirection response to a different domain produce this issue.

I made a repo to reproduce the issue here: https://github.com/mrkelvinli/capacitor-android-issue-4240-reproduction

### Other Technical Details

`npm --version` output: `6.14.10`

`node --version` output: `v14.15.4`

#### I did some investigation on this in `@capacitor/android`

When the web view made a request to the serverUrl domain, Capacitor will proxy the network request from the web view into WebViewLocalServer. If the request in `shouldInterceptRequest` is pass into `handleProxyRequest()`, that means it has a PathHandler associated to the hostname of the request. If the request also has `text/html` in the HTTP accept header, then the `HttpURLConnection` object which performs the network request will **follow redirects** by default (as `HttpURLConnection::setFollowRedirects()` is true by default). Thus, all successfully request will result in a 200 (status code is from the PathHandler associated to the request hostname) response and redirection will be resolved within `HttpURLConnection` instead of the web view.

### Solution I propose
```patch
diff --git a/node_modules/@capacitor/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java b/node_modules/@capacitor/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
index d632065..bc10831 100755
--- a/node_modules/@capacitor/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/node_modules/@capacitor/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -311,6 +311,28 @@ public class WebViewLocalServer {
           conn.setRequestMethod(method);
           conn.setReadTimeout(30 * 1000);
           conn.setConnectTimeout(30 * 1000);
+
+          /**
+           * This should be the final step in the setup process of the HttpURLConnection object.
+           *
+           * Initiate the connection with follow redirects disable. It allows us
+           * identifying any redirection response and avoid intercepting such request.
+           * This is because returning a 200 response when a redirection response (ie. 301 or 302)
+           * is expected could result in issues with resolving relative URLs for static assets
+           * in the HTML.
+           */
+          conn.setInstanceFollowRedirects(false);
+          conn.connect();
+          int status = conn.getResponseCode();
+          if (
+            status == HttpURLConnection.HTTP_MOVED_TEMP ||
+              status == HttpURLConnection.HTTP_MOVED_PERM ||
+              status == HttpURLConnection.HTTP_SEE_OTHER
+          ) {
+            // Return null so that this request will not be intercepted.
+            return null;
+          }
+
           String cookie = conn.getHeaderField("Set-Cookie");
           if (cookie != null) {
             CookieManager.getInstance().setCookie(url, cookie);
```

mrkelvinli:
Fixes #4240

In `WebViewLocalServer.handleProxyRequest()`, avoid intercepting network request which will result in a redirection response. This is because intercepting redirection response in `handleProxyRequest()` will caused the redirection resolving in HttpURLConnection, rather than in web view. This in turn causes the window URL to be unchanged when we expected it to change after a redirect response.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -362,20 +362,23 @@ public class WebViewLocalServer {
                      */
                     conn.setInstanceFollowRedirects(false);
                     conn.connect();
-                    int status = conn.getResponseCode();
-                    if (
-                        status == HttpURLConnection.HTTP_MOVED_TEMP ||
-                        status == HttpURLConnection.HTTP_MOVED_PERM ||
-                        status == HttpURLConnection.HTTP_SEE_OTHER
-                    ) {
-                        // Return null so that this request will not be intercepted.
-                        return null;
-                    }
 
                     String cookie = conn.getHeaderField("Set-Cookie");
                     if (cookie != null) {
                         CookieManager.getInstance().setCookie(url, cookie);
                     }
+
+                    int status = conn.getResponseCode();
+                    if (
+                        status == HttpURLConnection.HTTP_MOVED_TEMP ||
+                        status == HttpURLConnection.HTTP_MOVED_PERM ||
+                        status == HttpURLConnection.HTTP_SEE_OTHER ||
+                        status == 307 // Temporary Redirect
+                    ) {
+                        // Return null so that this request will not be intercepted.
+                        return null;
+                    }
+
                     InputStream responseStream = conn.getInputStream();
                     responseStream = jsInjector.getInjectedStream(responseStream);
                     return new WebResourceResponse(
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -350,6 +350,28 @@ public class WebViewLocalServer {
                         String base64 = Base64.encodeToString(userInfoBytes, Base64.NO_WRAP);
                         conn.setRequestProperty("Authorization", "Basic " + base64);
                     }
+
+                    /**
+                     * This should be the final step in the setup process of the HttpURLConnection object.
+                     *
+                     * Initiate the connection with follow redirects disable. It allows us
+                     * identifying any redirection response and avoid intercepting such request.
+                     * This is because returning a 200 response when a redirection response (ie. 301 or 302)
+                     * is expected could result in issues with resolving relative URLs for static assets
+                     * in the HTML.
+                     */
+                    conn.setInstanceFollowRedirects(false);
+                    conn.connect();
+                    int status = conn.getResponseCode();
+                    if (
+                        status == HttpURLConnection.HTTP_MOVED_TEMP ||
+                        status == HttpURLConnection.HTTP_MOVED_PERM ||
+                        status == HttpURLConnection.HTTP_SEE_OTHER
+                    ) {
+                        // Return null so that this request will not be intercepted.
+                        return null;
+                    }
+
                     String cookie = conn.getHeaderField("Set-Cookie");
                     if (cookie != null) {
                         CookieManager.getInstance().setCookie(url, cookie);


Moarcoo:
This indeed fixes my redirect problems! I do however notice that after a redirect, the capacitor and plugin packages have lost all knowledge of the native-bridge (no more device info). It restarts properly after a webview refresh. I'm not sure if this is a side effect of this fix though.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -350,6 +350,28 @@ private WebResourceResponse handleProxyRequest(WebResourceRequest request, PathH
...
+                     * is expected could result in issues with resolving relative URLs for static assets
+                     * in the HTML.
+                     */
+                    conn.setInstanceFollowRedirects(false);
+                    conn.connect();
+                    int status = conn.getResponseCode();
+                    if (
+                        status == HttpURLConnection.HTTP_MOVED_TEMP ||
+                        status == HttpURLConnection.HTTP_MOVED_PERM ||
+                        status == HttpURLConnection.HTTP_SEE_OTHER
```
tony-scio, line 369: This PR worked great for me! I'm depending on a fork now. The one thing is that I also had to add `|| status == 307` here.

mrkelvinli: Yeah, that make sense. Though 307 is not a very common status code isn't it?

mrkelvinli:
Glad that this works for you both @Moarcoo and @tony-scio 

If I recall correctly, one of the side effect to the redirected page is that it will not have the capacitor runtime (ie. the native bridge) as it skipped the `jsInjector.getInjectedStream` part which is located right after the if statement.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
tony-scio: It is not, that's why there's no constant for it in `HttpURLConnection`. But a backend we had to support happens to use it and it is standard: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307

Thanks for adding it!

tony-scio (approved):
I'm not a project member so not sure what the approval process is. But we have this change deployed to prod and it's working well for our particular usage.

tony-scio:
> If I recall correctly, one of the side effect to the redirected page is that it will not have the capacitor runtime (ie. the native bridge) as it skipped the `jsInjector.getInjectedStream` part which is located right after the if statement.

We didn't try changing the order of that. Worth a go.

agclark27:
This is an issue that's been affecting us on Android, too, and while I think this may fix this issue with regards to redirects, there still exists a broader and more substantive difference between how the Capacitor code is injected on iOS and how it's injected on Android. The iOS approach is pretty clean and ensures that the code is always loaded, redirects are handled within the webview, etc. The Android approach, on the other hand, relies on proxying the content and delicately searching for <head> tags to inject the Capacitor code. We've been experimenting with a fork that we've created that appears to eliminate the need for the handleProxyRequest function altogether.

Specifically, we (a) eliminate handleProxyRequest, (b) set the jsInjector property to public on WebViewLocalServer, (c) remove references to jsInjector.getInjectedStream in handleLocalRequest, and (d) in Bridge.java, add to the reset() function a call to webView.evaluateJavascript that executes the script aggregate, which appears to have a pretty equivalent effect as the current behavior. With this new approach, however, no proxying is required, everything can happen in the webview, and we achieve a pretty equivalent behavior between the iOS and Android implementations.

We're happy to submit this as a PR, if folks may think this would be helpful, but I thought I'd chime in here first to see if we're missing some special context as to why this approach hasn't been taken already. The proxy code just seems like it's always going to be a fragile way to handle this remotely-loaded content on Android.

avallete:
> This is an issue that's been affecting us on Android, too, and while I think this may fix this issue with regards to redirects, there still exists a broader and more substantive difference between how the Capacitor code is injected on iOS and how it's injected on Android. The iOS approach is pretty clean and ensures that the code is always loaded, redirects are handled within the webview, etc. The Android approach, on the other hand, relies on proxying the content and delicately searching for tags to inject the Capacitor code. We've been experimenting with a fork that we've created that appears to eliminate the need for the handleProxyRequest function altogether.
> 
> Specifically, we (a) eliminate handleProxyRequest, (b) set the jsInjector property to public on WebViewLocalServer, (c) remove references to jsInjector.getInjectedStream in handleLocalRequest, and (d) in Bridge.java, add to the reset() function a call to webView.evaluateJavascript that executes the script aggregate, which appears to have a pretty equivalent effect as the current behavior. With this new approach, however, no proxying is required, everything can happen in the webview, and we achieve a pretty equivalent behavior between the iOS and Android implementations.
> 
> We're happy to submit this as a PR, if folks may think this would be helpful, but I thought I'd chime in here first to see if we're missing some special context as to why this approach hasn't been taken already. The proxy code just seems like it's always going to be a fragile way to handle this remotely-loaded content on Android.

We have similar issues with some webapp using Auth0 for login and do redirects for tracking tools.
The fix provided in this PR is good enough for our use case, but I'll be very interested to see what solution you came with because as you say this fix doesn't really "fix" the underlying issue with the way webview handle requests.

Zacherl:
Is there any information when this might be integrated? And is there a fix that also prevents the problem with the missing native bridge?

rengert:
Hi Capacitor team, this can be fix by any chance? We want to use keycloak and this is the only blocking issue at the moment. Thank you for your work.

PR #4440, opened 2021-04-06
feat(cli): Add signup prompt on first init

jcesarmobile:
prompt users to create an Ionic account
prompt only once, the first time they run init command
prompt only if init is executed in interactive mode (so doesn't prompt if init is called from Ionic CLI or another tooling)


perrygovier (approved):

dotNetkow (approved):

dotNetkow:
thanks julio!

jthoms1 (approved):

thomasvidas (approved):

PR #4441, opened 2021-04-06
feat(iOS): Renaming scrolling flag

ikeith:
As per the [discussion on 4416](https://github.com/ionic-team/capacitor/pull/4416), this change was unrelated and has been moved to a separate PR.

This branch renames the programmatic config property from `enableScrolling` to `scrollingEnabled`. It now better matches the loggingEnabled property as well as the config file option and the actual setting on WKWebView.


jcesarmobile (approved):
looks good, but let's try to not merge it until right before we are about to do a new release so it doesn't break keyboard plugin

PR #4446, opened 2021-04-07
refactor: Inject native-bridge as first script

thomasvidas:
Kind of a revert of #3748. There were several issues with plugins not being initialized correctly as well as issues with the new `PluginHeaders`

This refactor allows us to initialize the necessary native parts `Capacitor` that are required for plugins like `toNative`, `nativeCallback`, `nativePromise`, etc.


core/src/global.ts
```diff
@@ -1,18 +1,8 @@
...
-    ? globalThis
-    : typeof self !== 'undefined'
-    ? self
-    : typeof window !== 'undefined'
-    ? window
-    : typeof global !== 'undefined'
-    ? global
-    : {}) as any,
-);
+export const Capacitor = (window as WindowCapacitor).Capacitor;
```
adamdbradley, line 5: Avoid using `window` directly like this because it'll likely fail in many SSR and node testing framework implementations. In today's landscape, and the minimum of what we require for Capacitor, it might be safe to start using just `globalThis` instead. Otherwise you can do this check:

```
(typeof globalThis !== 'undefined'
    ? globalThis
    : typeof self !== 'undefined'
    ? self
    : typeof window !== 'undefined'
    ? window
    : typeof global !== 'undefined'
    ? global
    : {}) as any,
```

thomasvidas: Fixed it and also added a fallback

core/src/global.ts
```diff
@@ -1,18 +1,43 @@
...
+} from './definitions-internal';
 import { legacyRegisterWebPlugin } from './legacy/legacy-web-plugin-merge';
-import { initCapacitorGlobal } from './runtime';
 import type { WebPlugin } from './web-plugin';
 
-export const Capacitor = /*#__PURE__*/ initCapacitorGlobal(
-  (typeof globalThis !== 'undefined'
-    ? globalThis
+// eslint-disable-next-line
+const createCapacitor = require('../native-bridge');
```
jcesarmobile, line 9: I think this is importing the whole native-bridge into core, which makes the native code to ignore the native-bridge.js file and still use the core code

jcesarmobile:
Added a comment, it's not working as expected.

Also, I don't think we need the `.gitkeep` files since we are committing the native-bridge.js on both iOS and Android

core/src/global.ts
thomasvidas: Yeah, I don't think its working as expected...I mostly did that so the tests would pass. I will keep tinkering with the tests so we don't have to require the bridge like here

thomasvidas:
After a quick chat with @adamdbradley , I've changed the following

- Revert back to using the `.ts` files for the following files
  - `events.ts`
  - `runtime.ts`
  - `vendor.ts`
- Remove everything besides `initBridge()` (and `initLogger()`, which is a part of `initBridge()`) from `native-bridge.js` so that it is truly only the bridge. Then it calls `initBridge` which creates an Capacitor object that kinda looks like this...
```typescript
{
  "logToNative": (console) => void,
  "logFromNative": (console) => void,
  "handleWindowError": (msg, url, lineNo, columnNo, err) => void,
  "toNative": (pluginName, methodName, options, storedCallback) => int | null,
  "fromNative": (result) => void,
  "nativeCallback": (pluginName, methodName, options, callback) => int | null,
  "nativePromise": (pluginName, methodName, options) => Promise<void>,
}
```

- Revert back to the original `initCapacitorGlobal` which reads the above object and augments it with the rest of the Capacitor goodies.

- `initBridge` is exported as CJS only for use in tests. Its not used in the actual bridge/runtime.

The tests pass after I updated to use `initBridge`, and the testapp seems to work on web, iOS, and Android again.

jcesarmobile:
If we want users to no need to import the whole capacitor/core for things they don't use themselves but that Capacitor uses internally, the native-bridge would need:

`Capacitor.createEvent`, used from `cordova.js` ❌
`Capacitor.fromNative`, used from the native bridge ✅
`Capacitor.toNative`, used from the js bridge ✅
`Capacitor.withPlugin`, used from iOS native bridge ❌
`Capacitor.triggerEvent`, used from the native bridge ❌
`Capacitor.logJs`, used from the native bridge ❌
`Capacitor.logToNative`, used for Console/logs  ✅
`Capacitor.logFromNative`, used for Console/logs  ✅
`Capacitor.handleWindowError`, used for Console/logs  ✅
`Capacitor.nativeCallback`, used by plugins  ✅
`Capacitor.nativePromise`, used by plugins  ✅
`Capacitor.addListener`, used by plugins  ❌
`Capacitor.removeListener`, used by plugins  ❌

Expose `Ionic` object with the WebView wrapper functions ❌
https://github.com/ionic-team/capacitor/blob/2.x/core/native-bridge.js#L444-L463

Expose `navigator.app.exitApp` and fire `deviceready` ❌

https://github.com/ionic-team/capacitor/blob/2.x/core/native-bridge.js#L107-L128

✅  is what's already done
❌  is what's missing

core/native-bridge.js
```diff
@@ -0,0 +1,563 @@
...
+  const callbacks = new Map();
+
+  // Counter of callback ids, randomized to avoid
+  // any issues during reloads if a call comes back with
+  // an existing callback id from an old session
+  let callbackIdCount = Math.floor(Math.random() * 134217728);
+
+  let postToNative = null;
+
+  const isNativePlatform = () => getPlatformId(win) !== 'web';
```
jcesarmobile, line 323:
```suggestion
  const isNativePlatform = () => true;
```

native-bridge.js is executed in native apps only, so no need to check the platform, will always be true

core/native-bridge.js
```diff
@@ -0,0 +1,563 @@
...
+      }
+    }
+    return false;
+  };
+
+  win.Capacitor = cap;
+};
+
+const initLegacyHandlers = (win, cap) => {
+  if (cap.isNativePlatform()) {
```
jcesarmobile, line 101: native-brige.js is always used in native platform, no need to check

core/native-bridge.js
```diff
@@ -0,0 +1,543 @@
...
+  // define cordova if it's not there already
+  win.cordova = win.cordova || {};
+
+  const doc = win.document;
+  const nav = win.navigator;
+
+  if (nav) {
+    nav.app = nav.app || {};
+    nav.app.exitApp = () => {
+      cap.nativeCallback('App', 'exitApp', {});
```
jcesarmobile, line 110: This uses App plugin, we should check if it's installed and warn if it isn't

core/native-bridge.js
```diff
@@ -0,0 +1,543 @@
...
+      const handler = args[1];
+      if (eventName === 'deviceready' && handler) {
+        Promise.resolve().then(handler);
+      } else if (eventName === 'backbutton' && cap.Plugins.App) {
+        // Add a dummy listener so Capacitor doesn't do the default
+        // back button action
+        cap.Plugins.App.addListener('backButton', () => {
+          // ignore
+        });
+      }
```
jcesarmobile, line 127: should also warn here if no app plugin is present

core/native-bridge.js
```diff
@@ -0,0 +1,543 @@
...
+  cap.platform = cap.getPlatform();
+  cap.isNative = cap.isNativePlatform();
+
+  win.Capacitor = cap;
+};
+
+const initVendor = (win, cap) => {
+  const Ionic = (win.Ionic = win.Ionic || {});
+  const IonicWebView = (Ionic.WebView = Ionic.WebView || {});
+  const Plugins = cap.Plugin;
```
jcesarmobile, line 142:
```suggestion
  const Plugins = cap.Plugins;
```

It's Plugins

jcesarmobile (approved):
Looks good, but would be good if @adamdbradley could also do a review since is a major change

adamdbradley:
All of the contents of native-bridge.js should be put inside of a closure, otherwise each of those functions be end up on `window` as a const, and accessible by anything. It would have the potential for breaking an app that'd also be using one of the functions with the same name. 

Next, all of this should continue to be in TypeScript, then use rollup to bundle it all together into an `iife`. This would be apart of the build step.

You could also return `initBridge` from the closure which may be helpful for tests, rather than adding `module.exports = initBridge;` only for the tests.

thomasvidas:
Made the following changes in that last big commit
- Migrated `native-bridge.js` to `native-bridge.ts`
- Added `rollup.bridge.config.js` for the bridge
  - The build config removes the export and module variables and assigns them to a dummy variable so we can use it in Android/iOS fine as well as import like normal for the tests. Maybe a little hacky, but works fine.
- Modified build commands in package.json to use `build.runtime` and `build.bridge` commands
- Updated the tests to use `import` instead of `require`, otherwise the tests all still pass and it works on the test app

adamdbradley:
Looks like the typescript target for native-bridge is ES3, so there's a lot of extra async/await polyfills that aren't needed and make it harder to debug. Could you set the target to es2017 instead?

Next, since the output of native-bridge.js is part of a build step now, could we gitignore them from their platform asset directories, then probably create a .npmignore to ensure they publish? This way we won't have two copies of the build artifacts committed to the repo, which may cause issues when contributors are trying to make updates. Could also remove the "Do not edit" warning.

Overall I think it's looking good and these last changes should be it. Thanks

thomasvidas:
I added the `--target es2017` and `--moduleResolution node` flags so that the emitted output is `es2017`. I also added the files to the `android/package.json` and `ios/package.json` and testing with some combos of `npm pack` that seems to work. And the `// Do not edit` is no more!

thomasvidas:
Added the builds back to their respective folders after our chat on slack. Think this is good now

Cc @adamdbradley @jcesarmobile

core/native-bridge.ts
```diff
@@ -0,0 +1,566 @@
...
+  PluginResult,
+  WindowCapacitor,
+} from './src/definitions-internal';
+
+// For removing exports for iOS/Android, keep let for reassignment
+// eslint-disable-line
+const dummy = {};
+
+const initBridge = (w: any): void => {
+  const getPlatformId = (win: WindowCapacitor): 'android' | 'ios' | 'web' => {
```
jcesarmobile, line 15: since it's a .ts file now, we can use the getPlatformId from util instead of duplicating

android/package.json
```diff
@@ -16,7 +16,8 @@
     "capacitor/build.gradle",
     "capacitor/lint-baseline.xml",
     "capacitor/lint.xml",
-    "capacitor/src/main/"
+    "capacitor/src/main/",
+    "capacitor/src/main/assets/native-bridge.js"
```
jcesarmobile, line 20: Since it's committed, no need to add the line as everything inside main is already published

ios/package.json
```diff
@@ -16,7 +16,8 @@
     "Capacitor/Capacitor/",
     "CapacitorCordova/CapacitorCordova/",
     "Capacitor.podspec",
-    "CapacitorCordova.podspec"
+    "CapacitorCordova.podspec",
+    "Capacitor/Capacitor/assets/native-bridge.js"
```
jcesarmobile, line 20: same, as it's committed there is no need to add the line since everything on Capacitor/Capacitor/ is already published

jcesarmobile (changes requested):
Added a few comments, also, since we commit the files, there is no need for the .gitkeep files

And since we are committing the duplicated files, should we add the "Do not edit" or similar messaging on the copies to avoid confusions?

jcesarmobile (approved):

adamdbradley (approved):

PR #4454, opened 2021-04-11
Ensure capacitor fires pause and resume events for Ionic/Angular

toxaq:
Capacitor does not fire pause or resume events on the document. This means the Ionic/Angular Platform service pause and resume events never fire. They do on the Android release.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/App.swift
+++ b/ios/Capacitor/Capacitor/Plugins/App.swift
@@ -45,6 +45,7 @@ public class CAPAppPlugin : CAPPlugin {
   }
   
   public func fireChange(isActive: Bool) {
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the PR, but this will duplicate the events when users have Cordova plugins installed.

Capacitor already fires the events, but only if Cordova plugins are present.
We will fix it in a way that it fires them even when no Cordova plugins are present, but that doesn't duplicate the events if Cordova plugins are present, you can subscribe to https://github.com/ionic-team/capacitor/issues/4456

toxaq:
Very good. That explains why my functionality stopped working on iOs. I migrated all of my plugins to Capacitor and got stuck in a spin for 2 days. Have followed #4456. I have a repro if required but it's trivial to replicate as I assume you have.

Issue #4456, opened 2021-04-12
bug: pause/resume document events only fired if cordova plugins are installed


jcesarmobile:
### Capacitor Version

2.x+

```
PASTE OUTPUT HERE
```

### Platform(s)

iOS

### Current Behavior

For compatibility purposes, Capacitor fires pause and resume events, but on iOS they only fire if some cordova plugin is installed, while in Android they are always fired

### Expected Behavior

both iOS and Android fire the events even if there are no cordova plugins installed

### Code Reproduction
```
function onPause() {
    console.log('paused');
}

function onResume() {
    console.log('resume');
}
document.addEventListener("pause", onPause, false);
document.addEventListener("resume", onResume, false);
```

### Other Technical Details

`npm --version` output:

`node --version` output:

`pod --version` output (iOS issues only):

### Additional Context

It affects Ionic apps since they rely on those events instead of in Capacitor ones

PR #4467, opened 2021-04-15
fix(ios): fire resume/pause events if no cordova plugins installed

jcesarmobile:
Cordova fires pause and resume events, but if there are no cordova plugins installed, cordova doesn't get initialized, so the events are not fired.
This PR fires pause/resume when there are no cordova plugins installed for compatibility

closes https://github.com/ionic-team/capacitor/issues/4456


thomasvidas (approved):

jcesarmobile:
I try to avoid making changes in cordova classes because they are mostly copies of the original cordova ones, so we can reintroduce it in future updates without noticing and cause duplicate events

ikeith (approved):

PR #4502, opened 2021-04-27
Hot push code suport

tamanhquyen:
Change class to objc & objc method support capacitor code push
Plugin link : https://github.com/tamanhquyen/capacitor-codepush


tamanhquyen:
Update again

PR #4504, opened 2021-04-27
(ios) change method setServerBasePath can access method from Objective-C file

tamanhquyen:
All change support for hot push code plugins
App example : https://github.com/tamanhquyen/capacitor-code-push-example
Plugins Example : https://github.com/tamanhquyen/capacitor-codepush


ikeith:
Thank you for the PR. However, we are not supporting feature work on 2.x at this time; bug fixes only. The 3.0 version is currently in the release candidate stage and this PR is obviated by its changes.

PR #4505, opened 2021-04-27
feat(iOS): Include native-bridge.js as a resource in both Cocoapods and direct build

ikeith:
This branch updates the podspec and direct build process of the Capacitor.framework to ensure that they build functionally identical frameworks. Specifically, the `native-bridge.js` file is added as a resource for the framework in both scenarios. It also adds a test case to verify that the injection method can locate the javascript file as expected.


jcesarmobile:
If we are going to have the bundle file, then we should use it instead of having two (this one and the generated by CocoaPods)
Can be done with the resource option
https://guides.cocoapods.org/syntax/podspec.html#resources

ikeith:
I'm not sure I understand. This changes has Xcode generate a `.bundle` file as a build artifact and includes it in the framework. So there isn't a file that can be referenced from the podspec (although the final products match). Are you suggesting that we create the bundle as a static resource and check into git?

ikeith:
The other option is that we just use the Cocoapods `resource` option for the `native-bridge.js` file. A resource bundle is only needed to avoid conflicts if the binary will be a static library, but a framework is our default configuration for capacitor apps.

I don't have a strong opinion either way, I just want to ensure that the final product is the same structure regardless of how it gets built (from the capacitor .xcworkspace or via Cocoapods)

jcesarmobile:
I tried that and wasn’t able to read the file, but if you can make it work I’m ok with it

ikeith:
Updated. This uses the podspec `resources` key to package the native bridge file and updates the direct build process to match. The direct build version is verified through the new unit test and the pod version passes integration testing via the test app.

jcesarmobile (approved):
works fine and reduces the code, nice!

PR #4524, opened 2021-05-03
fix(bridge): Fix type errors with new Platforms API

thomasvidas:
Merging in #4255 seemed to make tests and types not work anymore even though they worked fine on their own individual branch...This PR makes some changes such as

- Making the `CapacitorPlatforms` optional like all of the other types on `WindowCapacitor`
- optional chaining to allow web platform to default correctly to the bridge functions if no platform was initialized (mostly for tests)

### DIFF
--- a/core/native-bridge.ts
+++ b/core/native-bridge.ts
@@ -559,7 +559,7 @@ initBridge(
     ? (window as any)
     : typeof global !== 'undefined'
     ? (global as any)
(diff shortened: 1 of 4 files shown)


thomasvidas:
Lint/build failing on `pull_request_target` is because they are failing on `main`. You should be able to ignore that

elylucas (approved):

PR #4526, opened 2021-05-03
Append user agent properly on iOS when "appendUserAgent" is used

### LINKED ISSUE #4517: bug: iOS user-agent is missing Mobile/15E148 when appendUserAgent is used
### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 2.4.7
  @capacitor/core: 2.4.7
  @capacitor/android: 2.4.7
  @capacitor/ios: 2.4.7

Installed Dependencies:

  @capacitor/cli: 3.0.0-rc.0
  @capacitor/android: 3.0.0-rc.0
  @capacitor/core: 3.0.0-rc.0
  @capacitor/ios: 3.0.0-rc.0

```

### Platform(s)

iOS

### Current Behavior

After adding `appendUserAgent` (with `MY_APP` as value) as described in https://capacitorjs.com/docs/v3/config I get following user-agent
```
Mozilla/5.0 (iPhone; CPU iPhone OS 14_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) MY_APP
```

### Expected Behavior

I expect to see following user-agent, which contains `Mobile/15E148` as part of user-agent
```
Mozilla/5.0 (iPhone; CPU iPhone OS 14_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 MY_APP
```

### Other Technical Details

`npm --version` output:

`node --version` output:

thomasvidas:
Fixes #4517

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -105,12 +105,11 @@ import Cordova
         webViewConfiguration.allowsAirPlayForMediaPlayback = true
         webViewConfiguration.mediaTypesRequiringUserActionForPlayback = []
         if let appendUserAgent = instanceConfiguration.appendedUserAgentString {
-            guard webViewConfiguration.applicationNameForUserAgent != nil else {
+            if let appName = webViewConfiguration.applicationNameForUserAgent {
+                webViewConfiguration.applicationNameForUserAgent = "\(appName)  \(appendUserAgent)"
+            } else {
                 webViewConfiguration.applicationNameForUserAgent = appendUserAgent
-                return webViewConfiguration;
-            };
-            
-            webViewConfiguration.applicationNameForUserAgent! += " " + appendUserAgent;
+            }
         }
         return webViewConfiguration
     }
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/CAPBridgeViewController.swift
```diff
@@ -105,7 +105,12 @@ import Cordova
         webViewConfiguration.allowsAirPlayForMediaPlayback = true
         webViewConfiguration.mediaTypesRequiringUserActionForPlayback = []
         if let appendUserAgent = instanceConfiguration.appendedUserAgentString {
-            webViewConfiguration.applicationNameForUserAgent = appendUserAgent
+            guard webViewConfiguration.applicationNameForUserAgent != nil else {
+                webViewConfiguration.applicationNameForUserAgent = appendUserAgent
+                return webViewConfiguration;
+            };
+            
+            webViewConfiguration.applicationNameForUserAgent! += " " + appendUserAgent;
```
ikeith, line 113: While it is true that force unwrapping `applicationNameForUserAgent` should be safe here because it has already been checked for nil, it will still generate a warning from the linter because it doesn't know that. And the string interpolation could be easier to mentally parse.
```suggestion
            if let appName = webViewConfiguration.applicationNameForUserAgent {
                webViewConfiguration.applicationNameForUserAgent = "\(appName)  \(appendUserAgent)"
            } else {
                webViewConfiguration.applicationNameForUserAgent = appendUserAgent
            }
```

ikeith (approved):

PR #4530, opened 2021-05-05
feat(android): allow to add listeners for receivedError and receivedHttpError events

crixx:
Since [PR 4405](https://github.com/ionic-team/capacitor/pull/4405) we can listen for the pageLoad event. Actually, I want also to listen to some errors. This PR allows to register listeners for receivedError and receivedHttpError events.


crixx:
What I'm actually not sure about is, if it is really a good idea to further follow this path. IMHO in the end, we just forward all event listeners from the WebViewClient to the listeners. Would it be wiser, to allow to register a Custom WebViewContr or if it were better to allow to register a custom WebViewClient instead/in addition of/in the BridgeWebViewClient?

carlpoole:
I think we could allow the registration of a custom WebViewClient. At the moment we have some logic in the BridgeWebViewClient that we want to keep beside allowing the overriding of these listeners.

crixx:
@carlpoole So, how do we proceed? What's missing to get this merged?

crixx:
Any news on this? What's missing?

elylucas:
Thanks for the PR @crixx!

PR #4550, opened 2021-05-07
feat(cli): auto detect framework's webDir

### LINKED ISSUE #4546: feat: Detect framework within cap init, set webDir from framework default
### Description

When running `npx cap init`, detect which framework is being used, then depending on the framework, set the `webDir` property to the framework's default build directory. Similar to how vercel detects frameworks and knows its default outputDirectory: https://github.com/vercel/vercel/blob/abf159d5e08164b88caa31cdb25d6843faade91a/packages/frameworks/src/frameworks.ts#L1217

Also, this is just a helper to get started faster and take a good guess at what they'll want to set the `webDir` to. But webDir is still a config that can be changed after cap init.

### Preferred Solution

Create an interface for each framework config with just a `detector` property and `outputDirectory` (like how Vercel is doing it). Then on `cap init`, loop through the configs to see if there's a match. If there's a match, use the framework setting, otherwise fallback to the current default.

adamdbradley:
Closes #4546


adamdbradley:
Takes a best guess, if it doesn’t figure it out then falls back to what it does today

thomasvidas (approved):

PR #4586, opened 2021-05-19
chore(ios): Use AnyObject instead of class for protocol inheritance

MortenGregersen:
When compiling with @capacitor/ios 3.0.0-rc.3 we get the following warning. This fixes this warning.

/Users/morten/MyProject/node_modules/@capacitor/ios/Capacitor/Capacitor/CAPBridgeDelegate.swift:3:38: Using 'class' keyword for protocol inheritance is deprecated; use 'AnyObject' instead


jcesarmobile (approved):
nice, thanks!

PR #4617, opened 2021-05-24
fix(cli): add priority to framework detection

adamdbradley:


elylucas (approved):

jcesarmobile (approved):

jcesarmobile:
changing title to fix since it's not really a feature for users and fix a problem where some apps were getting the wrong webDir value set

Issue #4717, opened 2021-06-13
bug: `npx cap sync` fails to complete its task! Because it can't simply update my iOS native deps and `xcodebuild` is not recognized on Windows!


imalitavakoli:
## Bug Report
**I know!** I'm not running on a Mac OS! But that's one of the reasons I'm working on a hybrid app and have chosen Capacitor and Appflow for my project! As you already said that we don't need to have a Mac! Right?

So, because I don't have a Mac, I can simply use [OpenSSL to build my certification keys](https://nintendoengineer.medium.com/creating-an-apple-provisioning-profile-and-a-p12-certificate-using-openssl-ebd443d3b8bb). And then I like to use Capacitor and Appflow to deploy the `.ipa` file for me!

But because `npx cap sync` command fails, I cannot run `ionic deploy add --app-id="APP-ID" --channel-name="Production" --update-method="background"` command successfully as [Appflow suggests in its DOC](https://ionic.io/docs/appflow/quickstart/installation) as well!

### Capacitor Version
```
   Capacitor Doctor

Latest Dependencies:

  @capacitor/cli: 3.0.1
  @capacitor/core: 3.0.1
  @capacitor/android: 3.0.1
  @capacitor/ios: 3.0.1

Installed Dependencies:

  @capacitor/android: 3.0.0
  @capacitor/core: 3.0.0
  @capacitor/cli: 3.0.0
  @capacitor/ios: 3.0.0

[success] Android looking great! 👌
[error] Xcode is not installed
```

### Platform(s)
This kills the Ionic CLI command processing! So it affects both Android and iOS platforms!

### Current Behavior
Because I'm not on a Mac, when I run `npx cap sync` command, it cannot simply update my iOS native deps and `xcodebuild` is not recognized:
```
√ Copying web assets from web to android\app\src\main\assets\public in 57.28ms
√ Creating capacitor.config.json in android\app\src\main\assets in 3.16ms
| copy android [info] Found 2 Cordova plugins for android:
       cordova-plugin-ionic@5.4.7
       cordova-plugin-whitelist@1.3.4
√ copy android in 191.74ms
√ Updating Android plugins in 24.24ms
\ update android [info] Found 2 Cordova plugins for android:
       cordova-plugin-ionic@5.4.7
       cordova-plugin-whitelist@1.3.4
√ update android in 176.05ms
√ Copying web assets from web to ios\App\App\public in 60.78ms
[...]

PR #4723, opened 2021-06-15
fix(cli): Don't error on ios sync on non macOS

jcesarmobile:
If sync is run on windows or linux, just skip the pod install step instead of erroring

closes https://github.com/ionic-team/capacitor/issues/4717


thomasvidas (approved):

imalitavakoli:
In what version of Capacitor will this change be available?

jcesarmobile:
3.0.2 or 3.1.0 depending on the changes we make from now until we release, probably next week

PR #4536, opened 2021-05-06
feat(android): Enable Kotlin for Cordova Android Plugins

EinfachHans:
closes #4426

If `kotlin_version` is set (because Kotlin is configured in module) the **capacitor-cordova-android-plugins** module now also respects that and configures itself.

PR #4739, opened 2021-06-18
feat(cli): Add hooks to call into npm scripts when capacitor commands run

### LINKED ISSUE #3593: feat: ability to hook into cap commands
### Description

Enable users to hook into any `cap` command to run before and after hooks. This is useful, e.g. to allow users to run custom scripts when running `cap sync`.

#### `.ts` file
Allow users to add a `.ts` file in the project root that contains hooks.

#### `capacitor.config.ts`
As the `capacitor.config.json` will be generated on-the-fly in Capacitor 3, we could also export the hooks here to avoid adding another file.

#### `package.json` scripts
Check the `scripts` defined in the `package.json`. Scripts named `cap:${command}:(before|after)` are executed, e.g. `cap:copy:before`.

EinfachHans:
As my other PR #4536 is still pending, this would be another way for me to make my android Build work in Appflow.

## Changes in this PR:

- The existing method `runPlatformHook` was extended with a new Environment Variable: **CAPACITOR_PLATFORM_NAME**
- The same Method is used to trigger the following hooks from `package.json:`
    - `capacitor:sync:before`
    - `capacitor:sync:after`
    - `capacitor:copy:before`
    - `capacitor:copy:after`
    - `capacitor:update:before`
    - `capacitor:update:after`

closes #3593

### DIFF
--- a/cli/src/common.ts
+++ b/cli/src/common.ts
@@ -158,6 +158,7 @@ export async function wait(time: number): Promise<void> {
 
 export async function runPlatformHook(
   config: Config,
+  platformName: string,
   platformDir: string,
   hook: string,
 ): Promise<void> {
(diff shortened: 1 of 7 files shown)


jcesarmobile (approved):
looks good, thanks!

PR #4748, opened 2021-06-21
fix(android): Set WEBVIEW_SERVER_URL before injecting native-bridge

WIStudent:
According to the [docs](https://capacitorjs.com/docs/basics/utilities#convertfilesrc) `Capacitor.convertFileSrc` should return an URL starting with `http://localhost` on android, but instead this part is missing and something like `/_capacitor_file_/path/to/device/file` is returned. `convertFileSrc` gets the host part from `webviewServerUrl`, which is set when native-bridge is loaded:

https://github.com/ionic-team/capacitor/blob/ca4e3b62dba6a40e593a1404ba2fe2b416a4ac14/core/native-bridge.ts#L356-L357

Because of the injection order, `win.WEBVIEW_SERVER_URL` has not been set yet when this line is executed. This PR fixes that by injecting the part that sets `WEBVIEW_SERVER_URL` before native-bridge.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/JSInjector.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/JSInjector.java
@@ -48,6 +48,8 @@ class JSInjector {
         return (
             globalJS +
             "\n\n" +
+            localUrlJS +
+            "\n\n" +
             bridgeJS +
             "\n\n" +
             pluginJS +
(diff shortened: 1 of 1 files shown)


jcesarmobile (changes requested):
good catch and thanks for the PR!

Can you only change the `getScriptString` function and not all function params?

Since they are public it would be a breaking change if somebody is using them internally.

WIStudent:
Good point, I changed it.

jcesarmobile (approved):
looks good, thanks!

PR #4751, opened 2021-06-21
fix(core): convertFileSrc returning wrong url

jcesarmobile:
`WEBVIEW_SERVER_URL` value was being read before being injected, so it was empty and convertFileSrc was returning a wrong url, the PR restores old behavior where it's read when it's needed.


jcesarmobile:
looks like this change was done on purpose so the server URL doesn't change if somebody changes `WEBVIEW_SERVER_URL` as it's global, should be fixed in native

PR #4769, opened 2021-06-28
fix(camera): properly append exif data into the returned images

theproducer:
Backporting android camera exif fix from Capacitor 3.x
https://github.com/ionic-team/capacitor-plugins/pull/480


jcesarmobile:
changes look good, but on Capacitor 2 the exif was not properly returned on iOS, so that should be fixed too.

theproducer:
> changes look good, but on Capacitor 2 the exif was not properly returned on iOS, so that should be fixed too.

Can you show me where this fix was made in Cap 3?

jcesarmobile (approved):
I've added the needed iOS changes and another android fix we made on https://github.com/ionic-team/capacitor-plugins/pull/510

PR #4779, opened 2021-06-30
fix(ios): isNewBinary is true if defaults have no stored versions

### LINKED ISSUE #4778: bug: isNewBinary is always false
### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 3.0.2
  @capacitor/core: 3.0.2
  @capacitor/android: 3.0.2
  @capacitor/ios: 3.0.2

Installed Dependencies:

  @capacitor/cli: 3.0.1
  @capacitor/core: 3.0.1
  @capacitor/ios: 3.0.1
  @capacitor/android: 3.0.1
```

### Platform(s)
iOS

### Current Behavior

`CAPBridgeViewController` field `isNewBinary` is always false

### Expected Behavior

`isNewBinary` should be true for new binary versions (and first launch)

### Code Reproduction

To experience that, we need to have Ionic Deploy set up OR install new binary. Will create a code reproduction if needed, but static analysis of android and ios can point to that:

Compare Android version:
https://github.com/ionic-team/capacitor/blob/b893a57aaaf3a16e13db9c33037a12f1a5ac92e0/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L259-L284

And iOS version:
https://github.com/ionic-team/capacitor/blob/779b650b9d213bc49f1e2e42000ba736105c8eb6/ios/Capacitor/Capacitor/CAPBridgeViewController.swift#L18-L27

Notice UserDefault can be only changed in that method (which will never happened because of guard):
https://github.com/ionic-team/capacitor/blob/779b650b9d213bc49f1e2e42000ba736105c8eb6/ios/Capacitor/Capacitor/CAPBridgeViewController.swift#L303-L316

### Other Technical Details

Bug first appeared in https://github.com/ionic-team/capacitor/commit/a601705f8116ac10d1a0b5942511952c07cf474e

`npm --version` output: 7.17.0

`node --version` output: v16.3.0

`pod --version` output (iOS issues only): 1.10.0

### Additional Context

This bug was found during searching for solution for keeping old update in app (fetched from Ionic AppFlow) after binary update (ignoring app delivered with new binary).

Zapominacz:
closes #4778

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -22,6 +22,7 @@ import Cordova
                let lastVersionName = UserDefaults.standard.string(forKey: "lastBinaryVersionName") {
                 return (curVersionCode.isEqual(lastVersionCode) == false || curVersionName.isEqual(lastVersionName) == false)
             }
+            return true
         }
         return false
     }()


jcesarmobile (approved):
good catch, thanks!

PR #4780, opened 2021-06-30
fix(android): add proguard rules for minifyEnabled

### LINKED ISSUE #739: Issues in release build when minifyEnabled is set to true
## Capacitor version

1.0.0-beta.4

## Platform

Android

## Issue

Strangely, many Capacitor elements behave differently in `release` build.

For instance, with `debug` we don't have any issues, while `release` builds getting few crashes during testing all the time:

image

For instance, my friend's Galaxy S6 Android 6.0 (debloated, not stock) always crashing on launch with `release` version, but not with `debug`.

Another issue with `navigator.geolocation` described in #738

And we noticed that for some reason `Plugins.App.openUrl` stops to work in `release` version. It just doing nothing.

I'm sure there also other issues specific to the `release` build. It makes me feel there some kind of global issue with `release` build in Capacitor.

What can I do to help debug those issues?

Zapominacz:
closes #739 . It will only apply for the new application

### DIFF
--- a/android-template/app/proguard-rules.pro
+++ b/android-template/app/proguard-rules.pro
@@ -1,9 +1,30 @@
 # Add project specific ProGuard rules here.
+# This rules will only works when `minifyEnabled true` in app/build.gradle
 # You can control the set of applied configuration files using the
 # proguardFiles setting in build.gradle.
 #
 # For more details, see
 #   http://developer.android.com/guide/developing/tools/proguard.html
+
+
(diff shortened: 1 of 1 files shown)


carlpoole:
Thank you for your PR! I am doing some testing and looking at adding these rules directly into the Capacitor Android library package `proguard-rules.pro` file

carlpoole:
Applied changes into library instead of app template in #5048

PR #4788, opened 2021-07-02
fix(android): proper targetSdkVersion configuration on cordova plugins

jcesarmobile:
on cordova plugins project, it was not properly formed


captainjim1:
for anyone wondering, you need to update @capacitor/cli to 2.4.9 to get this fix.

PR #4832, opened 2021-07-14
Give option to skip sync on add 

SmileyJames:
The same as #4384 but without merge conflicts.

Here at the BBC, we've used Capacitor 2 to released both Cbeebies Playtime Island and Cbeebies Storytime to the Android Play store, iOS app store, Amazon Appstore and Amazon Kids+ store! We are really pleased with the resulting technology stack and it's help us deliver more for less.

We are really keen to help contribute to open source.

This small change means we can move from Capacitor 2 to Capacitor 3 - so much good stuff we want to use: a bug fix for Android deeplinks; lovely new plugin API! Really want to keep up to date with your awesome changes.

## Change

This adds a new option to Capacitor's config called skipSyncOnAdd. If set to true, it will not do a cap sync during the cap add stage. false by default to retain original behaviour.

## Use Case

When doing a cap add, Capacitor also does a cap sync internally. Configuring add to skip sync allows for manipulation of the files between adding and syncing. This allows for increased automation and flexibility for Capacitor projects.

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -59,6 +59,7 @@ export async function loadConfig(): Promise<Config> {
       },
       ...conf,
       bundledWebRuntime: conf.extConfig.bundledWebRuntime ?? false,
+      skipSyncOnAdd: conf.extConfig.skipSyncOnAdd ?? false,
     },
   };
 
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -42,6 +42,17 @@ export interface CapacitorConfig {
   bundledWebRuntime?: boolean;
 
   /**
+   * Whether to skip doing a `cap sync` during `cap add`
+   *
+   * Sometimes you may not wish to perform a `cap sync` when adding a platform. e.g. if you're using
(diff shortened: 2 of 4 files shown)


consolenaut:
+1 from me on this, this is really useful for the way we use capacitor, and means we'd be able to move off our Capacitor 3 fork to the real thing :+1: 

@jcesarmobile @carlpoole is there anything we need to do to get this ready to merge? Give us a shout if anything needs changing!

jcesarmobile:
Thanks for the PR.
We talked in the past about having a CLI option that could be passed to add command to skip sync (like --no-sync), but we discarded it, having it as a config option is even less appealing because add should be run only once per platform and app.

There is a workaround that you can use if you don't want sync to run, if you change the `webDir` value to point to a non existent folder, sync will be skipped.
But note that is only a workaround, if you want a real solution, I would recommend using `capacitor:sync:before` [hook](https://capacitorjs.com/docs/cli/sync#hooks). Hooks allow to  run scripts, so you can make `capacitor:sync:before` run some code that edits the Podfile and adds the needed sources to sync doesn't fail when run.

PR #4856, opened 2021-07-20
(feat) android: adds support in CapacitorPlugin for handleOnActivityResult

ItsChaceD:
These proposed changes will allow support for using request codes in the `@CapacitorPlugin` annotation while also calling the `handleOnActivityResult` method in the Plugin as in previous versions of Capacitor.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -993,29 +993,23 @@ public class Bridge {
             return cordovaInterface.onActivityResult(requestCode, resultCode, data);
         }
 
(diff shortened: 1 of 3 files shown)


carlpoole:
@ItsChaceD I ran a couple tests with this and looks good. I experimented with an alternate solution that used a registration method instead of adding requestCodes back to the annotation but I think this is cleaner in the end. Added some verbiage to the comment to make it more clear this is a legacy support option.

carlpoole (approved):

ItsChaceD:
@carlpoole Awesome, looks good! Thanks for getting back to me! When can we expect a new version to be deployed with these changes?

carlpoole:
Release should be around Weds next week

PR #4861, opened 2021-07-23
fix(android): make openUrl create a new activity

Matheus-de-Souza:
When openUrl opens other webview app it opens the other webview app inside itself. My PR fix this behaviour.


carlpoole:
Thanks for submitting the PR. This changes the behavior of openUrl and while we would be open to adding this as an option via feature request in the [Capacitor 3.x app-launcher plugin](https://github.com/ionic-team/capacitor-plugins/tree/main/app-launcher) we are only making bug fixes to Capacitor 2.x at this time.

Issue #4904, opened 2021-08-09
feat: set targetSdkVersion to 30 for Capacitor 2.x


antoniogenaro:
## Feature Request
Set default `targetSdkVersion` and `compileSdkVersion` to 30 in Capacitor 2.x

### Description
Google doesn't accept apps with `targetSdkVersion` setted to 29 for new Google Play apps, so version 30 is needed for apps generated with Capacitor 2.x

### Platform(s)
Android

### Preferred Solution
Default `targetSdkVersion` and `compileSdkVersion` should be 30

### Additional Context
In Capacitor 3.x this variable is setted to 30, but in Capacitor 2.x is still to 29

jcesarmobile:
that's a breaking change, so can't be done in Capacitor 2

anyway, users can configure their app's `targetSdkVersion` and `compileSdkVersion` in their `variables.gradle` file (as their own risk as some things might break)

PR #4905, opened 2021-08-09
chore: set targetSdkVersion and compileSdkVersion to 30

antoniogenaro:
Related to #4904


jcesarmobile:
as explained on the linked issue, this is a breaking change and can't be done in Capacitor 2

PR #4992, opened 2021-09-01
Backported fix for Geolocation in V2

KoenLav:
Based on: https://github.com/ionic-team/capacitor-plugins/pull/589


jcesarmobile:
Thanks for the PR, but when porting changes from the plugins back to capacitor 2, we ask the original author to do it to maintain authorship, in this case it was me, so I've sent another PR and already released capacitor/android 2.5.0 with the fix.

Issue #5003, opened 2021-08-27
Android: Camera Plugin returns null data when Activity not kept


mfs409:
### Plugin(s)
Camera

### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 3.2.0
  @capacitor/core: 3.2.0
  @capacitor/android: 3.2.0
  @capacitor/ios: 3.2.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 3.2.0
  @capacitor/android: 3.2.0
  @capacitor/core: 3.2.0

[success] Android looking great! 👌
[error] Xcode is not installed
```

### Platform(s)
Android

### Current Behavior
When "Don't keep Activities" is selected in Android's Developer Options, the Camera Plugin returns an error code.

### Expected Behavior

Camera Plugin should return a `resultCode` that is not `-1`, and the `data` object should not be `null`.

### Code Reproduction

```
import { Camera, CameraResultType, Photo, CameraSource } from '@capacitor/camera';
import { CameraHelper } from "./CameraHelper"; // This hooks into a custom webpack build rule for bundling PWA on web, but not including on Android/iOS

export function takePicture(success: (image: string) => void, failure: (reason: any) => void) {
    let f = () => {
        Camera.getPhoto({
            quality: 90,
            allowEditing: false,
            direction: CameraDirection.Front, // Commenting this does not change the behavior
            resultType: CameraResultType.Uri,
            source: CameraSource.Camera,
            saveToGallery: true  // Commenting this does not change the behavior
        }).then((image: Photo) => {
            success(image.webPath!);
        }).catch((reason: any) => {
            failure(reason);
        });
    }
    // NB: on web/electron, need to load PWA support first
    if (platform === "web") {
        new CameraHelper().injectPwa(f);
    }
    else {
        f();
    }
}
```

### Other Technical Details

* The above code works for PWA, iOS, and Android 7/10 when Activities are kept.  When Activities are not kept, the code breaks on Android 7/10.

* I have a handler in place for `appRestoredResult`, but it never fires:
```
    public void processCameraImage(PluginCall call, ActivityResult result) {
        if (imageFileSavePath == null) {
            // processCameraImage() is always returning here, because imageFileSavePath is null
            call.reject(IMAGE_PROCESS_NO_FILE_ERROR);
            return;
        }
        // ...
    }
```

### Additional Context

[...]

PR #5004, opened 2021-09-02
fix(android): save activity result launcher calls

jcesarmobile:
When using activity result launchers, the call is not being saved, so if the app dies, the `appRestoredResult` event won't be fired

Also `getSavedCall()` is deprecated and not returning the proper value, switched to `bridge.getSavedCall(lastPluginCallId);`

closes https://github.com/ionic-team/capacitor/issues/5003

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -690,6 +690,10 @@ public class Bridge {
         return pluginCallForLastActivity;
     }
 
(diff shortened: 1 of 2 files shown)


carlpoole (approved):

jcesarmobile:
the fix will be available on next @capacitor/android release, so once it's released you can do `npm install @capacitor/android@latest` and you'll get the fix, but you'll have to wait for the release

mfs409:
FWIW, I replicated the changes to `Bridge.java` and `Plugin.java` in the corresponding files in my `node_modules` tree, and then re-built the minimal failing code.  The behavior changed, and now the failing code is passing.  Thanks so much!

PR #5096, opened 2021-10-04
Fix multiple wildcard processing

macdja38:
At the moment allowed hosts with multiple wildcards are processed in-correctly
For example
Pattern: `*.*.example.com`
Host: `test1.test2.example.com`

Iteration 1 wildcard.offset is 0
Pattern becomes: `*.example.com`
Host becomes: `test2.example.com`

Iteration 2 wildcard.offset is 1
Pattern becomes: `*.com`
Host becomes:  `test2.com`

The correct outcomes would be
Pattern: `example.com`
Host: `example.com`

Reversing the direction of iteration fixes this, as removing the item at index 1 first is safe and doesn't effect the next step of removing the item at index 0.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
+++ b/ios/Capacitor/Capacitor/CAPInstanceConfiguration.swift
@@ -52,7 +52,7 @@ extension InstanceConfiguration {
             return false
         }
         // remove any wildcard segments
-        for wildcard in patternComponents.enumerated().filter({ $0.element == "*" }) {
+        for wildcard in patternComponents.enumerated().reversed().filter({ $0.element == "*" }) {
             hostComponents.remove(at: wildcard.offset)
             patternComponents.remove(at: wildcard.offset)
         }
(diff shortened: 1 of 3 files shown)


macdja38:
This is actually the old behaviour, the following code was removed ~10 months ago.

```swift
let wildcards = pattern.enumerated().filter { $0.element == "*" }
         for wildcard in wildcards.reversed() {
             host.remove(at: wildcard.offset)
             pattern.remove(at: wildcard.offset)
         }

         return host == pattern
```

Should I add a comment to ensure it's not removed again?

macdja38:
Added a test instead of a comment.

StareInTheAir:
I found this exact bug on Friday and wanted to create a PR just now, with the almost exact same fix. You beat me to it :)

jcesarmobile:
We have a setting that doesn't allow us to merge if the branch is not in sync with main branch, I usually see a button to update the branch, but not in your PR, did you check the checkbox to allow changes from maintainers when you created the PR?

Anyway, you probably see the button to update the branch, can you click it?

jcesarmobile (approved):

jcesarmobile:
Thanks!

PR #5134, opened 2021-10-14
feat(cli): Allow to configure access origin on cordova config.xml

Dante1349:
Hey,

i reecognized that the access origins where hard coded here, but i wanted to restrict the access origins because we are using `cordova-plugin-inappbrowser` which therefore can load any url.

### DIFF
--- a/cli/src/cordova.ts
+++ b/cli/src/cordova.ts
@@ -249,15 +249,18 @@ export async function autoGenerateConfig(
     }
   });
 
-  let allowedHostsString: string[] = [];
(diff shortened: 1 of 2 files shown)


jcesarmobile:
According to the own plugin docs, cordova-plugin-inappbrowser is not subject to the whitelist restrictions:
https://github.com/apache/cordova-plugin-inappbrowser/blob/master/README.md#windowopen

Can you provide a sample app where the cordova-plugin-inappbrowser respects the whitelist?

PR #5146, opened 2021-10-16
refactor(android): use StringBuilder in place of StringBuffer

larrydiamond:
… to eliminate synchronization on file reads 
this method used in 20 places across Capacitor


larrydiamond:
Happy to help, thanks for creating and managing such good tools for the community.

I agree the performance difference wont be big in most cases, but for large files on slower devices it might be noticeable.   For reading a 20 byte config file on a brand new top tier phone it's probably noise.

PR #5176, opened 2021-10-27
Android: Update AGP to 4.2.2 and limit jCenter to Cordova 7

saschpe:
Apply Gradle repository order, Maven Central should always come before
the Google repository when resolving dependencies and jCenter should be
last. Limit jCenter to only Cordova 7.x.x as it's going to be shut down
soon.

Update Android Gradle Plugin (AGP) to version 4.2.2, see upstream
release notes [0].

Remove ExampleUnitTest as it does not test anything.

[0]: https://androidstudio.googleblog.com/2021/06/android-studio-422-available.html


jcesarmobile:
If I create a new Android project from Android Studio, google is the first repository, not sure why do you want to put MavenCentral first. Also MavenCentral has had some reachability problems in the past, while google hasn't (as far as I know), and specially for capacitor android project, most of the dependencies are google dependencies and google recommends using google in their docs.

Also, not sure why you removed `androidx.activity:activity` dependency, that's needed for the plugins activity callbacks and it's not mentioned in the description.

As for the jcenter dependency, they announced that they weren't going to shut down after all, they will leave the repository as read only indefinitely. And restricting to only cordova is not a good solution as some plugins (specially cordova ones) can still be using old versions of dependencies not hosted in MavenCentral or google, so doing this now will potentially break a lot of user apps. We will probably remove jcenter in a future major release (maybe Capacitor 4) and restrict to using cordova-android 10+ as it's hosted on MavenCentral.

saschpe:
> If I create a new Android project from Android Studio, google is the first repository, not sure why do you want to put MavenCentral first.

The project template from Android Studio is outdated and quirky. It's pretty natural that Google puts their Maven repository first. Their project template still uses Groovy, `buildscript` and the `ext` Plugin. All of which are no longer recommended to use.

The general reason is that the repository order in Gradle matters a lot. Gradle tries to resolve *all* artifacts from repositories in that order. Let's assume you're working on a proprietary project that has private or even confidential dependencies. Gradle will ask Google's repo first, then go on. This isn't an issue here as the `repositories` declaration in that Gradle build file is only relevant for that particular Gradle project. 

> Also MavenCentral has had some reachability problems in the past, while google hasn't (as far as I know), and specially for capacitor android project, most of the dependencies are google dependencies and google recommends using google in their docs.

I've been doing Android development for a long time now and can't really share that experience. However, Google recommends their repositories simply because they can better track what OSS dependencies project uses. Ideally, Google's repository is only used to resolve Android-specific artifacts that are not published on Maven Central.

> Also, not sure why you removed `androidx.activity:activity` dependency, that's needed for the plugins activity callbacks and it's not mentioned in the description.

That's an error, I'll update the MR.

> As for the jcenter dependency, they announced that they weren't going to shut down after all, they will leave the repository as read only indefinitely. And restricting to only cordova is not a good solution as some plugins (specially cordova ones) can still be using old versions of dependencies not hosted in MavenCentral or google, so doing this now will potentially break a lot of user apps. We will probably remove jcenter in a future major release (maybe Capacitor 4) and restrict to using cordova-android 10+ as it's hosted on MavenCentral.

This is probably a misunderstand. The `repositories` clause modified here only applies for that particular Gradle project. It is not propagating to other Gradle projects unless put in a `allprojects` or `subprojects` clause. Other capacitor plugins or user apps can specify repositories as they see fit.

Similarly, the line `implementation fileTree(dir: 'libs', include: ['*.jar'])` could have been removed as well as there are no local libraries included here but I didn't want touch too much at once. 

While jCenter stays around for longer, it is still recommend to limit dependency resolution to the what is absolutely needed. An advantage of being explicit about it is to raise awareness on what needs to be ported / upgraded still.

saschpe:
Just for your info, jCenter's TLS certificate just expired:

```
  project :buildSrc > org.gradle.kotlin.kotlin-dsl:org.gradle.kotlin.kotlin-dsl.gradle.plugin:2.1.6 > org.gradle.kotlin:gradle-kotlin-dsl-plugins:2.1.6 > org.jetbrains.kotlin:kotlin-gradle-plugin:1.5.21
215      > Could not resolve org.jetbrains.kotlin:kotlin-scripting-compiler-embeddable:1.5.21.
216         > Could not get resource 'https://plugins.gradle.org/m2/org/jetbrains/kotlin/kotlin-scripting-compiler-embeddable/1.5.21/kotlin-scripting-compiler-embeddable-1.5.21.pom'.
217            > Could not HEAD 'https://jcenter.bintray.com/org/jetbrains/kotlin/kotlin-scripting-compiler-embeddable/1.5.21/kotlin-scripting-compiler-embeddable-1.5.21.pom'.
218               > PKIX path validation failed: java.security.cert.CertPathValidatorException: validity check failed
```

saschpe:
One final point about the stability of Google's Maven Repositories. It's not working at this moment:

Screenshot 2021-11-12 at 10 54 25

I think this can happen to all online services every now and then but Maven Central is usually the most reliable source. Let alone it's the canonical one for almost all Java / Kotlin projects (of which many Google jars depend on as well).

PR #5195, opened 2021-11-02
fix(android): Integer conversion of compileSdk/minSdk/targetSdk

### LINKED ISSUE #5194: bug: Unable to customize `compileSdkVersion` with Gradle Kotlin DSL
### Capacitor Version
3.1.5

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not determine the dependencies of task ':capacitor-android:compileReleaseAidl'.
> Failed to find target with hash string '31' in: EXAMPLE
```

Screenshot 2021-11-02 at 09 44 31

### Platform(s)
Android

### Current Behavior
The default project template and capacitor-cordovan-android-plugins try to read rootProject.ext.compileSdkVersion as integer directly, which only works for Gradle Groovy projects. Projects using Gradle Kotlin DSL won't have access to the `ext` plugin. The new mechanism is to put properties into `gradle.properties`. Those however, are always strings.

### Expected Behavior
`compileSDK`, `minSdk` and `targetSdk` can be customized in Gradle Kotlin DSL projects.

saschpe:
Gradle properties are always parsed as strings from `gradle.properties`. While Gradle Groovy can utilize the `ext` Gradle plugin directly, Gradle Kotlin DSL projects have to specify properties in `gradle.properties`. Convert those to integers as required by Android Gradle Plugin (AGP). This won't affect projects that still use Gradle Groovy.

Resolves #5194

### DIFF
--- a/android/capacitor/build.gradle
+++ b/android/capacitor/build.gradle
@@ -42,10 +42,10 @@ if (System.getenv("CAP_PUBLISH") == "true") {
 
 android {
     namespace "com.getcapacitor.android"
-    compileSdkVersion project.hasProperty('compileSdkVersion') ? rootProject.ext.compileSdkVersion : 33
+    compileSdkVersion project.hasProperty('compileSdkVersion') ? rootProject.ext.compileSdkVersion as Integer : 33
     defaultConfig {
-        minSdkVersion project.hasProperty('minSdkVersion') ? rootProject.ext.minSdkVersion : 22
-        targetSdkVersion project.hasProperty('targetSdkVersion') ? rootProject.ext.targetSdkVersion : 33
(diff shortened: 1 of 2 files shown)


saschpe:
I would be nice if someone could provide some feedback for this fix. Thanks.

saschpe:
I would appreciate if this could be looked at. At the moment it is not possible to put this into one's **gradle.properties**, simply because those are interpreted as strings:

```
compileSdkVersion=33
minSdkVersion=23
targetSdkVersion=33
```

[The same change for **capacitor-plugins**](https://github.com/ionic-team/capacitor-plugins/pull/677)

saschpe:
Final ping, I think this is still relevant. Worth rebasing once more?

PR #5196, opened 2021-11-02
feat(ios): Add new iOS15 media capture permission delegate

samydoesit:
Delegates media capture permissions to the app and remembers the user's decision for iOS15.

[https://developer.apple.com/documentation/webkit/wkuidelegate/3763087-webview#discussion](https://developer.apple.com/documentation/webkit/wkuidelegate/3763087-webview#discussion)

Thanks and credits to @7freaks-otte who found the solution. :)
[https://forum.ionicframework.com/t/prevent-for-reprompting-webrtc-webview-permissions-ios15/215240/2?u=samydoesit](https://forum.ionicframework.com/t/prevent-for-reprompting-webrtc-webview-permissions-ios15/215240/2?u=samydoesit)

You can test these changes using my getusermedia demo case when build on Xcode13 -> https://github.com/samydoesit/capacitor-getusermedia-test

### DIFF
--- a/ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
+++ b/ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
@@ -47,7 +47,8 @@ internal class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDel
         bridge?.reset()
     }
     
-    // This delegates media capture permission requests to the app and also remembers the user’s decision.
+    // TODO: remove once Xcode 12 support is dropped
+    #if compiler(>=5.5)
     @available(iOS 15, *)
     func webView(
         _ webView: WKWebView,
--- a/ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
+++ b/ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
@@ -46,6 +46,18 @@ internal class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDel
         // Reset the bridge on each navigation
         bridge?.reset()
     }
+    
+    // This delegates media capture permission requests to the app and also remembers the user’s decision.
+    @available(iOS 15, *)
+    func webView(
+        _ webView: WKWebView,
+        requestMediaCapturePermissionFor origin: WKSecurityOrigin,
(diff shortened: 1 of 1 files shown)


samydoesit:
Looks like that the CI should use latest Xcode13 on macos-11. https://github.com/actions/virtual-environments/issues/4060

ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
```diff
@@ -46,6 +46,18 @@ internal class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDel
         // Reset the bridge on each navigation
         bridge?.reset()
     }
+    
+    // This delegates media capture permission requests to the app and also remembers the user’s decision.
```
jcesarmobile, line 50:
```suggestion
    // TODO: remove once Xcode 12 support is dropped
    #if compiler(>=5.5)
```

ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
```diff
@@ -46,6 +46,18 @@ internal class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDel
...
+    @available(iOS 15, *)
+    func webView(
+        _ webView: WKWebView,
+        requestMediaCapturePermissionFor origin: WKSecurityOrigin,
+        initiatedByFrame frame: WKFrameInfo,
+        type: WKMediaCaptureType,
+        decisionHandler: @escaping (WKPermissionDecision) -> Void
+    ) {
+        decisionHandler(.grant)
+    }
```
jcesarmobile, line 60:
```suggestion
    }
    #endif
```

jcesarmobile (changes requested):
Hello.
Thanks for the PR!

Since we still need to support Xcode 12 and the method is not available there, the function needs to be wrapped inside a `#if compiler(>=5.5)` as proposed since that's the swift version shipped on Xcode 13, so that code won't run on Xcode 12 but the build won't fail as it does now.

jcesarmobile (approved):
Thanks!
It looks good, we will merge once we start working on next minor release, might take a few weeks, but will be included.

Issue #5177, opened 2021-10-27
feat: iOS Bundler support


saschpe:
### Description
The current implementation relies on the CocoaPods 'pod' binary to be installed as a system-wide Ruby gem. Similarly, all other gems needed during build (and thus specified in `Podfile`) to be available there too. [Bundler](https://bundler.io) provides a way to declare per-project RubyGems dependencies similar to how `node_modules` works.  It is used by many iOS projects to isolate from the host OS (different developers using different CocoaPods versions). Sometimes, a specific CocoaPods version has to be pinned, i.e. when a newer version had a regression (happened more than once in the past).

### Platform(s)
* iOS

### Preferred Solution

Thus, `checkCocoaPods` should run `bundle` first and see if

1) Bundler is installed 
    * Could be done with `which bundler` but the version has to match as well. Simply running `bundle` and checking for errors catches both cases.
2) The project has a `Gemfile`

If so, invoke `bundle exec pod install`. Otherwise, continue as before by looking for CocoaPods binary installed elsewhere (e.g. system-wide or on a custom path).

### Alternatives
Bundler is [recommended by CocoaPods.org](https://guides.cocoapods.org/using/a-gemfile.html) in such situations. 

### Additional Context

A typical `Gemfile` is situated at the repository root and might look like this:

```ruby
source 'https://rubygems.org'

## iOS Development
# DEVELOPER: You now need to prepend calls to rubygems with "bundle exec "
# e.g. instead of calling "pod install" you now call "bundle exec pod install"
gem 'cocoapods', '1.11.2'
gem 'cocoapods-art', '1.1.0'  # Cocoapods Artifactory plugin used to access private Artifactory server
gem 'xcpretty', '0.3.0'       # For prettier `xcodebuild` invocations on the console

## Documentation
gem 'asciidoctor', '2.0.16'
gem 'asciidoctor-diagram', '2.2.1'
```

This project fetches iOS dependencies from a private Artifactory instance (NPM and CocoaPods mirror plus proprietary stuff). It's `Podfile` looks like this (stripped for brevity):

```ruby
platform :ios, '12.0'
use_frameworks!

plugin 'cocoapods-art', :sources => [ 'cocoapods.github.com.proxy', ]

def capacitor_pods
  pod 'Capacitor', :path => '../../node_modules/@capacitor/ios'
  pod 'CapacitorCordova', :path => '../../node_modules/@capacitor/iOS'
  ....
```

[...]

PR #5205, opened 2021-11-04
feat(ios): Add Bundler support

saschpe:
Bundler isolates RubGems dependencies from the host operating system and creates a consistent environment. Gem dependencies can be pinned to exact versions by using a `Gemfile`. CocoaPods is modeled after Bundler.

This ensures all developers and CI machines use the same CocoaPods version. It's also conveniently used to install CocoaPods plugins, e.g. `cocoapods-art`, or additional iOS development tools, such as `xcpretty`.

Multiple cases are now handled correctly. If a `Gemfile` is found:

 - At the Git repository root, in case of a multi-app project
 - At or below the app directory, in case of a single-app project

Then:

 1. Use Bundler to install RubyGems, including CocoaPods:
   1. If Bundler is outdated, update first by executing
      `gem install bundler`
   2. If RubyGems bundle is outdated, update first by executing
      `bundle install`
   3. Install CocoaPods and other gems by executing
      `bundle exec pod install`

Or, just like before, if:

 4. CocoaPods is available on PATH, execute `pod install`
 5. Neither global CocoaPods nor Gemfile are found
    Skip CocoaPods install entirely

Implements #5177

# Remarks
The following screenshots demonstrate the CLI output and actions that now occur (in reverse order but with increasing coolness):

### Case 5
Screenshot 2021-11-04 at 10 25 04

### Case 4
Screenshot 2021-11-04 at 10 25 53

### Case 1.2 and 1.3
Screenshot 2021-11-04 at 10 24 32

### Case 1.1 and 1.3
Screenshot 2021-11-04 at 09 40 26

### DIFF
--- a/cli/src/ios/common.ts
+++ b/cli/src/ios/common.ts
@@ -1,4 +1,5 @@
 import { readFile, writeFile } from '@ionic/utils-fs';
+import { execSync } from 'child_process';
 import { resolve } from 'path';
 
 import c from '../colors';
@@ -6,16 +7,46 @@ import { checkCapacitorPlatform } from '../common';
 import { getIncompatibleCordovaPlugins } from '../cordova';
 import { OS } from '../definitions';
 import type { Config } from '../definitions';
+import { logger } from '../log';
 import { PluginType, getPluginPlatform } from '../plugin';
 import type { Plugin } from '../plugin';
-import { isInstalled } from '../util/subprocess';
+import { isInstalled, runCommand } from '../util/subprocess';
 
 export async function checkIOSPackage(config: Config): Promise<string | null> {
   return checkCapacitorPlatform(config, 'ios');
 }
 
+function execBundler() {
+  try {
+    const bundleOutput = execSync('bundle &> /dev/null ; echo $?');
+    return parseInt(bundleOutput.toString());
+  } catch (e: any) {
+    return -1;
+  }
+}
+
+export async function checkBundler(config: Config): Promise<string | null> {
+  if (config.cli.os === OS.Mac) {
+    let bundlerResult = execBundler();
+    if (bundlerResult === 1) {
+      // Bundler version is outdated
+      logger.info(`Using ${c.strong('Gemfile')}: Bundler update needed...`);
+      await runCommand('gem', ['install', 'bundler']);
+      bundlerResult = execBundler();
+    }
+    if (bundlerResult === 0) {
+      // Bundler in use, all gems current
+      logger.info(`Using ${c.strong('Gemfile')}: RubyGems bundle installed`);
+    }
+  }
+  return null;
+}
+
 export async function checkCocoaPods(config: Config): Promise<string | null> {
-  if (!(await isInstalled(config.ios.podPath)) && config.cli.os === OS.Mac) {
+  if (
+    !(await isInstalled(await config.ios.podPath)) &&
+    config.cli.os === OS.Mac
+  ) {
     return (
       `CocoaPods is not installed.\n` +
       `See this install guide: ${c.strong(
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -9,7 +9,6 @@ import Debug from 'debug';
 import { dirname, extname, join, relative, resolve } from 'path';
 
 import c from './colors';
-import { OS } from './definitions';
 import type {
   AndroidConfig,
   AppConfig,
@@ -19,12 +18,14 @@ import type {
   IOSConfig,
   WebConfig,
 } from './definitions';
+import { OS } from './definitions';
 import { fatal, isFatal } from './errors';
 import { logger } from './log';
 import { tryFn } from './util/fn';
 import { formatJSObject } from './util/js';
-import { resolveNode, requireTS } from './util/node';
+import { requireTS, resolveNode } from './util/node';
 import { lazy } from './util/promise';
+import { getCommandOutput } from './util/subprocess';
 
 const debug = Debug('capacitor:config');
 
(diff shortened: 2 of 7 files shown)


saschpe:
I didn't dare to touch the iOS project template yet. Once this change is accepted, it might make sense to do that as well to better isolate iOS builds from the host machine used.

saschpe:
I addressed the eslint issues but now need a workflow approval / re-run. TIA.

saschpe:
Any remarks or comments?

saschpe:
Once again, any news, decisions or comments on the matter?

saschpe:
It would be nice if this could get some feedback, I just rebased onto most recent `main` changes.

fxsalazar:
This is awesome 👍🏽  hope this can be merged as fast as possible.

ffluk3:
Hey, bump on this one, does the Ionic team see any value in bringing this into the project?

markemer:
Hey, update this with the latest from main as well. As best as you can anyway - I know we're coming fast and furious with commits - but I don't like to update other peoples forks unless I'm about to merge any minute.

saschpe:
And so I keep it current even though I already gave up hope that contributing to CapacitorJS is a thing at all.

saschpe:
Now that some attention is on this one, may I shamelessly point you guys to https://github.com/ionic-team/capacitor/pull/5195 as well? It's equally important when trying to use latest Android SDK versions

cli/src/ios/common.ts
```diff
@@ -1,21 +1,60 @@
...
+    }
+    if (bundlerResult === 0) {
+      // Bundler in use, all gems current
+      logger.info(`Using ${c.strong('Gemfile')}: RubyGems bundle installed`);
+    } else {
+      // Bundler not in use
+      logger.info(`No ${c.strong(
+        'Gemfile',
+      )} found, trying system-wide CocoaPods.
+Consider adopting Bundler, see: ${c.strong(
```
markemer, line 45: I'm all for this, but I'll have to ask if we might want to remove the Consider adopting Bundler bit, since we don't have a Gemfile in the template.

markemer (approved):
Looks good, only thing we might want to pull is the CTA about installing Bundler - but I'll have to ask the team.

saschpe:
Let's see wait they say. I'd be happy to adjust the default project template afterwards.

markemer:
@saschpe You have a conflict here since I removed the deletion of the `Podfile.lock` as that was done a while back to work around a bug. Should be an easy fix. I can even fix it if you like.

saschpe:
I'm currently rebasing all my open pull requests...

cli/src/ios/common.ts
saschpe: I'd say this is merely a recommendation. People are free to add their own `Gemfile`.

markemer:
I haven't forgotten about this - I'm going to see if we can get this into the 5-rc or if not that, the next point release.

cli/src/ios/common.ts
IT-MikeS: I'm up for the changes to allow support for Bundler but I don't believe we should have this else statement. This is to avoid confusion for current users who don't use/won't need, Bundler.

IT-MikeS:
Please see my new comment:

https://github.com/ionic-team/capacitor/pull/5205#discussion_r1156179404

cli/src/ios/common.ts
markemer: Yeah - I think that was the consensus. @saschpe can you pull that?

IT-MikeS (approved):

jskrepnek:
Is there a way to tell Capacitor 5 to ignore or opt-out of the bundler integration?  We have a Gemfile for automation purposes, but don't want it used for development at all (at least not yet).

jskrepnek:
@markemer @IT-MikeS 

I'm trying to figure out if I'm seeing an issue with this change.

Running `capacitor sync ios` I see the logs:

```
[capacitor] [info] Using Gemfile: RubyGems bundle installed
...
[capacitor] [warn] Skipping pod install because CocoaPods is not installed
[capacitor] ✔ Updating iOS native dependencies with pod install in 1.27s
```

The update [code](https://github.com/ionic-team/capacitor/blob/4099969f70e9b995182bacecc16e160d89bbc746/cli/src/ios/update.ts#L139) appears to confusingly think that it's not using bundler, despite the earlier message.

Tracing down further, I think the issue is that it's using single quotes in the [test](https://github.com/ionic-team/capacitor/blob/4099969f70e9b995182bacecc16e160d89bbc746/cli/src/config.ts#L503C53-L503C53) to determine if Cocoapods is referenced in the Gemfile, but it's valid to use double quotes here and our file does.

saschpe:
Double quoted strings allow interpolation which is a bit unusual in the case of gem names. They're rather fixed. Then again, Ruby has another string literal variant, percent strings like `%q{hello}`. 

While it's straightforward to expand the test to check for double-quoted strings, I am not sure we'd want to support percent strings as well. I can already see someone who is using regex strings :-)

toxaq:
Yes, how do I opt out of this? We have a Gemfile for fastlane but this is trying to install bundler on xcode cloud build?

saschpe:
> Yes, how do I opt out of this? We have a Gemfile for fastlane but this is trying to install bundler on xcode cloud build?

You're using Fastlane and Xcode Cloud? There aren't many reasons left these days to still use Fastlane. Almost everything can be achieved with `xcodebuild`. But if you want to avoid Bundler, you could simply `rm Gemfile` in the Xcode Cloud build phase.

toxaq:
You’re using bundled to install cocoa pods?

PR #5209, opened 2021-11-04
chore: Carthage Bundling for Ionic Portals

thomasvidas:
- Adds `build-carthage-zip.sh` file for creating Carthage binaries for Ionic Portals
  - Will emit `Capacitor.xcframework.zip` which can be attached to a release and used with ionic-portals to include Capacitor and CapacitorCordova as dependencies


jcesarmobile (changes requested):
The `build-carthage-zip.sh` would need executable permissions, or maybe we could move it inside ios folder (modifying paths as needed) and add a new script in package.json that runs `sh build-carthage-zip.sh`.

thomasvidas:
Closed in favor of #5415

Issue #5252, opened 2021-11-18
feat: add productFlavor support to --livereload --external


sandor-biro:
### Description
Currently `assembleDebug` is hardcoded so there's no possibility to `--livereload --external` a given `productFlavor`.

```
productFlavors {
        versionA {
            applicationId "com.company.versionA"
        }

        versionB {
            applicationId "com.company.versionB"
        }

        versionC {
            applicationId "com.company.versionC"
        }
    }
```

### Platform(s)
Android

### Preferred Solution
Change the hardcoded `const gradleArgs = ['assembleDebug'];` (https://github.com/ionic-team/capacitor/blob/main/cli/src/android/run.ts#L22)

to a dynamic one, so various `productFlavors` can be generated correctly when using `--livereload --external` build args.

App should generate `gradleArgs` based on the given productFlavor i.e. `assembleVersionADebug` etc.

```
const arg = `assemble${flavour}Debug`;
const gradleArgs=[arg];
```

### Alternatives
... remove extra product flavours from `build.gradle` while working, then remember copy them back.

### Additional Context
I'm happy to submit a PR for this.

Something similar was reported here also: https://github.com/ionic-team/capacitor/issues/5104

jcesarmobile:
assembleDebug is supposed to build a debug apk for all the configured flavors, isn’t it working? Maybe you have some setting that prevents it from working?

sandor-biro:
I understand all flavors are supposed to be built at every save. Is this what you meant?

My question is: what if there are more than 10 flavors? All of them will (i.e. should) be built into the .apk at every save?

jcesarmobile:
No at every save, as you are using live reload the native app is not rebuilt on every save, just the web part and it reloads the WebView without rebuilding the native part.

adrianpunkt:
Our use case here is to build one flavor at a time since we do some additional configuration for that flavor. The build process fails since it cannot build all of them at the same time. Since the `capacitor.config.ts` includes one flavor, it would make sense to build only one, hence the need to call `assembleFlavorDebug` instead of `assembleDebug` which builds all:

```
const config: CapacitorConfig = {
 ...
  },
  ios: {
    scheme: 'XCODE_SCHEME',
  },
  android: {
    flavor: 'ANDROID_FLAVOR'
  }
};
```

adrianpunkt:
For livereload, it also doesn't make sense to build multiple flavors at once.

carlpoole:
[...]

PR #5256, opened 2021-11-19
fix(android): restrict android run command to configured flavour

sandor-biro:
Solves request: https://github.com/ionic-team/capacitor/issues/5252

Live reload should now work with any `productFlavor`.

### DIFF
--- a/cli/src/android/run.ts
+++ b/cli/src/android/run.ts
@@ -19,7 +19,8 @@ export async function runAndroid(
     selectedTarget,
   );
 
-  const gradleArgs = ['assembleDebug'];
+  const arg = `assemble${config.android?.flavor || ''}Debug`;
+  const gradleArgs = [arg];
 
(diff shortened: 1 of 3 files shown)


cli/src/android/run.ts
```diff
@@ -19,7 +19,8 @@ export async function runAndroid(
     selectedTarget,
   );
 
-  const gradleArgs = ['assembleDebug'];
+  const arg = `assemble${config.android?.flavor || ''}Debug`;
```
adrianpunkt, line 22: @sandor-biro You still need the word `Debug`, the correct argument is actually `assembleFlavorDebug` where `Flavor` is the value from the config

jcesarmobile: Debug word is at the end of the template string, so if there is a flavor it will be assembleFlavorNameDebug and if not assembleDebug

carlpoole:
Can you run `npm run fmt` and commit the changes please

Issue #5266, opened 2021-11-22
feat: Update the GitHub action setup-java from adopt to zulu of the OpenJDK.


carldea:
## Feature Request
Not a feature but an update to the GitHub Actions setup-java to update the discontinued `adopt` build distribution of the OpenJDK.

### Description

The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

Please see more details at Foojay.io:
https://foojay.io/today/github-actions-with-java-part-2/

### Platform(s)

N/A

### Preferred Solution

```yaml
        uses: actions/setup-java@v2
         with:
           java-version: '11'
           distribution: 'zulu'
```

### Alternatives

distribution: 'temurin'

### Additional Context

https://adoptopenjdk.net/
https://foojay.io/today/github-actions-with-java-part-2/

carlpoole:
Adopted this, thank you for the recommendation

carldea:
> Adopted this, thank you for the recommendation

Glad to have helped contribute to this excellent project!

PR #5267, opened 2021-11-22
#5266 Using TCK Tested JDK builds of OpenJDK

carldea:
#5266 Update the GitHub action setup-java from adopt to zulu of the OpenJDK.

The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK. 

Please see more details at Foojay.io:
https://foojay.io/today/github-actions-with-java-part-2/

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/setup-java@v2
         with:
           java-version: '11'
-          distribution: 'adopt'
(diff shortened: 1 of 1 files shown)


carlpoole:
Thanks for the contribution!

carldea:
> Thanks for the contribution!

Glad to have helped!

PR #5286, opened 2021-12-01
WIP: resolve to accept other data types

### LINKED ISSUE #1132: FR: support non JSObject method result and call arguments
Currently Capacitor allows to return from native method only JSObject (a map, with string key and any json serializable value). The same goes to method arguments - only one arguments of JSObject is supported (of course, it is a map, where multiple parameters can be passed). That restriction requires additional steps when interacting with native methods and is boilerplate (e.g. ```StatusBar#setStyle(options: {style: StatusBarStyleOptions})``` or ```Geolocation#getCurrentPosition() => Promise<{coords: any}>```).

# Calling native method
Arguments of js method call are accessible in js as array (arguments variable) and that array should be passed to PluginCall (instead of a map) as call data/options. Native PluginCall instance should have methods for reading the arguments:
- int getLength() -> returns the size of arguments array
- String getString(int , String defaultValue) -> returns the argument at given index as string. 
- ... other getters for json types

Now, it maybe looks like a problem to support array arguments and JSObject argument in the same time (e.g. what if there is only one argument in a js call and it is JSObject), but it is not. PluginCall data/options should always be (underneath) an array of arguments, that comes from js but in the same time getString(String key), getDouble(String key), get(String key) should work, but will refer to first argument, which should be JSObject (if argument is not a JSObject error should be thrown). That way, we can provide backward compatibility and additional feature, which imho is clean and flexible ;-)

# Returning from native method
PluginCall should have a methods for returing from method invocation for each json type, e.g.:
- void resolve(String value)
- void resolve(int value)
- void success(String value)
- void success(String[] value)

romaia:
This is a first attempt to allow `PluginCall.resolve` to accept other data types than `JSObject`.

From what I can see in the native-bridge, it should already handle any value coming "fromNative":

https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/assets/native-bridge.js#L420

So, simply overriding `sendReponseMessage` to accept other values than `PluginResult` seems like it should be enough.

This of course needs further development, but I would like some feedback from the maintainers on whether this approach is reasonable (and so I would proceed) or if this is just a bad idea.#5267 

This should partially fix #1132

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/MessageHandler.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/MessageHandler.java
@@ -79,6 +79,21 @@ public class MessageHandler {
     }
 
     public void sendResponseMessage(PluginCall call, PluginResult successResult, PluginResult errorResult) {
+        _sendResponseMessage(call, successResult, errorResult);
+    }
+
+    public void sendResponseMessage(PluginCall call, JSObject successResult, PluginResult errorResult) {
+        _sendResponseMessage(call, successResult, errorResult);
+    }
+
+    public void sendResponseMessage(PluginCall call, boolean successResult, PluginResult errorResult) {
+        _sendResponseMessage(call, successResult, errorResult);
+    }
+
+    public void sendResponseMessage(PluginCall call, String successResult, PluginResult errorResult) {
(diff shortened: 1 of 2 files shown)


eric-horodyski:

More context:

Issue #658, opened 2018-07-14
TypeError: cameraModal.present is not a function


dotnetdreamer:
I am using with following dependencies with ionic:
```
 "dependencies": {
    "@angular/common": "6.0.2",
    "@angular/core": "6.0.2",
    "@angular/forms": "6.0.2",
    "@angular/http": "6.0.2",
    "@angular/platform-browser": "6.0.2",
    "@angular/platform-browser-dynamic": "6.0.2",
    "@angular/pwa": "^0.6.8",
    "@angular/router": "6.0.2",
    "@angular/service-worker": "6.0.2",
    "@capacitor/android": "^1.0.0-alpha.39",
    "@capacitor/cli": "^1.0.0-beta.1",
    "@capacitor/core": "^1.0.0-beta.1",
    "@ionic-native/core": "5.0.0-beta.10",
    "@ionic-native/splash-screen": "5.0.0-beta.10",
    "@ionic-native/status-bar": "5.0.0-beta.10",
    "@ionic/angular": "4.0.0-alpha.7",
    "@ionic/schematics-angular": "latest",
    "chart.js": "^2.7.2",
    "core-js": "^2.5.3",
    "moment": "^2.22.2",
    "rxjs": "6.1.0",
    "ydn.db": "^1.3.0",
    "zone.js": "^0.8.26"
  }
```
When i try to use its camera plugin. i am getting error:
```
        const image = await Camera.getPhoto({
            quality: 90,
            allowEditing: true,
            source: CameraSource.Camera,
            resultType: CameraResultType.Base64
        });
        console.log(image.base64Data);
```
```
core.js:1598 ERROR Error: Uncaught (in promise): TypeError: cameraModal.present is not a function
TypeError: cameraModal.present is not a function
    at CameraPluginWeb.<anonymous> (camera.js:49)
    at step (tslib.es6.js:91)
    at Object.next (tslib.es6.js:72)
    at fulfilled (tslib.es6.js:62)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:388)
    at Object.onInvoke (core.js:4062)
    at ZoneDelegate.push../node_modules/zone.js/dist/zone.js.ZoneDelegate.invoke (zone.js:387)
    at Zone.push../node_modules/zone.js/dist/zone.js.Zone.run (zone.js:138)
    at zone.js:872
[...]

PR #5289, opened 2021-12-02
fix(android): Prevent crash in restoreInstanceState if bundleData is null

hermitdemschoenenleben:
this PR fixes this (closed) issue: https://github.com/ionic-team/capacitor-plugins/issues/658
closes: https://github.com/ionic-team/capacitor/issues/5290

Without the fix, capacitor calls the camera plugin's `restoreState()` method even though `bundleData` is `null`

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -807,8 +807,10 @@ public class Bridge {
             // Let the plugin restore any state it needs
             Bundle bundleData = savedInstanceState.getBundle(BUNDLE_PLUGIN_CALL_BUNDLE_KEY);
             PluginHandle lastPlugin = getPlugin(lastPluginId);
-            if ((bundleData != null) && (lastPlugin != null)) {
+            if (bundleData != null && lastPlugin != null) {
                 lastPlugin.getInstance().restoreState(bundleData);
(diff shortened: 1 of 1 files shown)


hermitdemschoenenleben:
corresponding issue: https://github.com/ionic-team/capacitor/issues/5290

jcesarmobile (approved):
Thanks for the PR.

I've made a few additions to make sure we don't save other data if we couldn't save the bundle and also to log error if the saving/restoring failed, since this will only prevent the crash, but the messages can maybe figure out why it's failing to save in the future.

sebsoftpak:
Hi @hermitdemschoenenleben,

Thank you for making this fix. Do you know if this fix has been released?
(Haven't seen #658 in the release notes yet, that's why i'm asking.)

We are experiencing the same NullPointerException with @capacitor/camera 1.3.0.

Kind regards,
Sebastian Sørensen

hermitdemschoenenleben:
@sebsoftpak yes, the fix has been released (at least I don't see the error in production anymore)

sebsoftpak:
@hermitdemschoenenleben  ok, good to hear! will upgrade from 1.3.0 to 1.3.1, maybe that makes the difference :)

jcesarmobile:
The fix is in capacitor/android 3.3.3, not in the camera plugin

PR #5350, opened 2022-01-06
fix(android): Add android:exported parameter for android 12 support

Seungwoo-Yu:
References:
https://developer.android.com/about/versions/12/behavior-changes-12#exported


jcesarmobile:
Thanks for the PR, but this is not a fix, it's a feature as it's only needed if you want to target SDK 31 and capacitor 2 targets SDK 29.
This is something we will add in Capacitor 4 as feature.

We are not adding any new features to Capacitor 2, sorry.

Issue #5237, opened 2021-11-12
activity needs android:exported value when targeting SDK 31


jcesarmobile:
### Capacitor Version
4

### Platform(s)
android

### Current Behavior
android template doesn't have a value for android:exported and it's required if targeting SDK 31

### Expected Behavior
we should update the template to make the MainActivity entry to include `android:exported="true"`

beruic:
Will this be backported to Capacitor 2 and 3?

jcesarmobile:
no, but this is the starting template, any user that wants to target SDK 31 can just do this on their own app, for users migrating from 3 to 4 will be documented on the upgrade guide

productdevbook:
some problem Capacitor 3

piotr-cz:
@jcesarmobile
In that case Capacitor 3 docs should be upgraded to clarify that SDK 31+ target isn't supported

jcesarmobile:
this has been addressed [here](https://github.com/ionic-team/capacitor/pull/5351) and available on capacitor 4.0.0 beta 0

PR #5351, opened 2022-01-06
feat(cli): export android-template activity for SDK 31 support

Seungwoo-Yu:
References:
https://developer.android.com/about/versions/12/behavior-changes-12#exported

closes https://github.com/ionic-team/capacitor/issues/5237


android-template/app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,8 @@
             android:name="com.getcapacitor.myapp.MainActivity"
             android:label="@string/title_activity_main"
             android:theme="@style/AppTheme.NoActionBarLaunch"
-            android:launchMode="singleTask">
+            android:launchMode="singleTask"
+			android:exported="true">
```
jcesarmobile, line 19:
```suggestion
android:exported="true">
```
Can you align the android word with the other lines?

jcesarmobile (approved):
Thanks!

We will merge once we start working on Capacitor 4.

android-template/app/src/main/AndroidManifest.xml
```diff
@@ -15,7 +15,8 @@
             android:name="com.getcapacitor.myapp.MainActivity"
             android:label="@string/title_activity_main"
             android:theme="@style/AppTheme.NoActionBarLaunch"
-            android:launchMode="singleTask">
+            android:launchMode="singleTask"
+            android:exported="true">
```
beruic, line 19: Why is this set to `"true"`?

According to developer.android.com/about/versions/12/behavior-changes-12#exported it should be set to `"false"` in most cases when the app is not a launcher. What is the case for it being `"true"`for Capacitor?

I should mention that my context is making this work for a Capacitor 2 project.

PR #5352, opened 2022-01-06
feat(android): Allow overriding web chrome client

SimonBackx:
Currently it is not possible to override the WebChromeClient of the Bridge. With this change, it becomes possible to override the web chrome client in plugins (the same way this currently is possible for the web view client). This is required if you want to add more custom behaviour in a Capacitor app and need to have a reference to the WebChromeClient.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -107,6 +107,7 @@ public class Bridge {
     private CordovaWebView cordovaWebView;
     private CordovaPreferences preferences;
     private BridgeWebViewClient webViewClient;
+    private BridgeWebChromeClient webChromeClient;
     private App app;
 
     // Our MessageHandler for sending and receiving data to the WebView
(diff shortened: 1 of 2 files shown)


SimonBackx:
Any update on this? This would help me implement some changes on WebChromeClient via our own plugins. Let me know if you have questions! :) Thanks!

riderx (approved):

jcesarmobile:
Thanks for the pull request, but it has always been possible to override the `WebChromeClient` by calling `bridge.getWebView().setWebChromeClient()` from the `MainActivity.java`'s `onCreate` method or from a plugin's `handleOnStart()` as the `WebView` has the setters for the `WebViewClient` and `WebChromeClient`.

For `WebViewClient` we added the `setWebViewClient` helper in the bridge as we wanted people to keep using a `BridgeWebViewClient` since it's essential for Capacitor's inner working, but the `BridgeWebChromeClient` is not essential and users overriding the `WebChromeClient` are free to use a `WebChromeClient` subclass or a `BridgeWebChromeClient` subclass if they want to take advantage of the methods we already implement and just change some others or implement new ones.

PR #5355, opened 2022-01-07
null-check before removing a listener

Monchi:
`remove` would be undefined on calling `PluginListenerHandle#remove` before capacitor initialized. This causes `remove is not a function.` error.

This PR adds null-check before calling `PluginListenerHandle#remove` to avoid the error.

### DIFF
--- a/core/src/runtime.ts
+++ b/core/src/runtime.ts
@@ -166,7 +166,9 @@ export const createCapacitor = (win: WindowCapacitor): CapacitorInstance => {
         });
 
         if (prop === 'addListener') {
-          (p as any).remove = async () => remove();
+          (p as any).remove = async () => {
+            remove && remove();
+          };
         }
 
         return p;


jcesarmobile:
Can you provide a sample app where the listener get removed before capacitor being initialized?
In which platform does it happen?

davidisaaclee:
This is an issue for me (in that it causes noise in my application's logs).

I'm using Capacitor with a React app running on an iOS iPhone. I use a `useEffect` hook to add listeners to a Capacitor plugin – something like:

```ts
useEffect(() => {
  const subscription = CapacitorAppPlugin.addEventListener('resume', onResume);
  return () => {
    subscription.remove();
  };
}, [onResume]);
```

(Happy to explain more if you're not familiar with React.) With [React's strict mode](https://react.dev/reference/react/StrictMode) enabled, this effect fires twice at launch, triggering the cleanup function, which hits this null `remove` issue.

I don't see any downside of adding this check – did you have something in mind?

jcesarmobile:
can you provide a sample app?

davidisaaclee:
I could not reproduce using https://github.com/ionic-team/capacitor-testapp as a base; if I find time, I'll try to reproduce again. (Is there a more barebones Capacitor + iOS sample project? For my purposes, something that already has Capacitor, iOS project, and ideally React.)

PR #5356, opened 2022-01-07
fix(android): Support for Android 12 / API 31

saschpe:
Activities, providers and services have to set the `exported` flag
explicitly to either 'true' or 'false' when the component has an intent
filter.

See https://developer.androi.com/guide/topics/manifest/activity-element#exported


saschpe:
I didn't see https://github.com/ionic-team/capacitor/pull/5351, sorry for the noise.

PR #5377, opened 2022-01-16
Remove all references to jcenter()

cromefire:
All references to should be removed, because they will be non functional in 2 weeks (on 01.02.2022) so I guess better break it now and then fix it an get a release out, than wait it out and in 2 weeks all builds are failing.

https://blog.gradle.org/jcenter-shutdown

Yes, this will probably break the build, but it will break in 2 weeks on it's own anyway, so probably better break the dev branch now and fix it, than break every production build in 2 weeks without a fix.

### DIFF
--- a/android-template/build.gradle
+++ b/android-template/build.gradle
@@ -1,10 +1,9 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    
+
     repositories {
         google()
-        jcenter()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:4.2.1'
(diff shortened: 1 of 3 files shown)


cromefire:
Timeline for the shutdown has been canceled, but the latest problems are still concerning, so this should be killed anyway

mlynch:
As mentioned, JCenter has moved to a read-only repo and will continue to work just fine. Of course, new packages can't be published to it but all old ones will continue to work: https://developer.android.com/studio/build/jcenter-migration

cromefire:
Yeah but it seems to have some dropout over that last days, so while not that critical it's still very disruptive if CI builds just randomly fail.

cromefire:
Also all dependencies seem to be available on mavenCentral at least in an updated form.

PR #5405, opened 2022-02-02
fix(cli): Better error on gradlew permission problems

jcesarmobile:


nphyatt:
do we have easy color highlighting/bolding in the CLI? would be nice to make `chmod +x /path/to/gradlew` standout so they know exactly what to copy paste.

jcesarmobile:
yeah, added it

IT-MikeS (approved):

PR #5427, opened 2022-02-10
feat(android): Use addWebMessageListener where available

jcesarmobile:
This PR will use the `addWebMessageListener` API if available, if not it will default to the old bridge.

I made that the bridge is injected to the capacitor default url (i.e. http://localhost), to `server.url` if configured, and also to all the urls in `server.allowNavigation`.
Also only passes the message to the js side if it's the main frame.


IT-MikeS (approved):

giralte-ionic (approved):

jcesarmobile:
did a minor change to use a lambda that simplifies the code

carlpoole (approved):

giralte-ionic (approved):

IT-MikeS (approved):

Issue #5278, opened 2021-11-26
bug: service worker server.url not injecting capacitor


dylanvdmerwe:
### Capacitor Version
```
Latest Dependencies:

  @capacitor/cli: 3.3.2
  @capacitor/core: 3.3.2
  @capacitor/android: 3.3.2
  @capacitor/ios: 3.3.2

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 3.3.2
  @capacitor/android: 3.3.2
  @capacitor/core: 3.3.2
```

### Current Behavior
Calling any native plugins when an app is served from an Angular service worker causes the plugins to fail. Using `server.url` to load the site on Android.
![image](https://user-images.githubusercontent.com/2852839/143595172-71102760-69e3-403f-afbc-058dce62e1cc.png)

Platform ready does not help for this. 

### Expected Behavior
Capacitor needs to be injected when a site is served through a service worker.

### Code Reproduction

app.module.ts
```
export function init_app(appLoadService: AppLoadService) {
  return () => appLoadService.initApp();
}

@NgModule({
  declarations: [AppComponent],
  entryComponents: [],
  imports: [
    BrowserModule,
    IonicModule.forRoot(),
    AppRoutingModule,
    ServiceWorkerModule.register('ngsw-worker.js', {
      enabled: environment.production,
      // Register the ServiceWorker as soon as the app is stable
      // or after 30 seconds (whichever comes first).
      registrationStrategy: 'registerWhenStable:30000',
    }),
  ],
  providers: [
    { provide: RouteReuseStrategy, useClass: IonicRouteStrategy },
    {
      provide: APP_INITIALIZER,
      useFactory: init_app,
      deps: [AppLoadService],
      multi: true,
    },
  ],
  bootstrap: [AppComponent],
})
export class AppModule {}
```

app-load-service.ts
```
@Injectable({
  providedIn: 'root',
})
export class AppLoadService {
  constructor(
    private device: DeviceInfo,
    private platform: Platform
  ) {}

  async initApp(): Promise<any> {
    await this.platform.ready();

    SplashScreen.hide(); // breaks
  }
}
```

dylanvdmerwe:
OR it seems that the capacitor script is not being injected after the site is served by a service worker. 

I have included this on Android but still getting plugin errors:
![image](https://user-images.githubusercontent.com/2852839/143600123-4ffb5f95-0386-4c63-814d-cd535243858d.png)

[...]

PR #5428, opened 2022-02-11
Service worker injection workaround

chrisharris77:
Requests from service workers use "Accept: */*" so checking for "text/html" in Accept header can't work. As a workaround, add check if proxy request URL contains ".html" and treat it as HTML if it does. This is the same as what handleLocalRequest does to identify HTML files.

See https://github.com/ionic-team/capacitor/issues/5278


riderx:
@chrisharris77 we don't need the fix in ios ?

PR #5438, opened 2022-02-15
fix(android): compatibility with cordova kotlin plugins

IT-MikeS:
This PR allows supported Cordova plugins written in Kotlin to be used by `capacitor-cordova-android-plugins`.


LifeIsStrange:
How does it compare to https://github.com/ionic-team/capacitor/pull/4536 ?
Anyways it's about time for capacitor to get support for Kotlin cordova plugins..

jcesarmobile (changes requested):
this should be done in a way that the CLI adds this properties/dependencies if there is a plugin installed that uses kotlin 

also the versions should be configurable from the variables.gradle instead of using hardcoded values

android-template/variables.gradle
```diff
@@ -11,4 +11,5 @@ ext {
     androidxJunitVersion = '1.1.2'
     androidxEspressoCoreVersion = '3.3.0'
     cordovaAndroidVersion = '7.0.0'
+    kotlin_version = '1.4.32'
```
jcesarmobile, line 14:
```suggestion
```
we only add variables to the template that are used by default

ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
```diff
@@ -62,9 +62,9 @@ internal class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDel
 
     @available(iOS 15, *)
     func webView(_ webView: WKWebView,
-    requestDeviceOrientationAndMotionPermissionFor origin: WKSecurityOrigin,
-         initiatedByFrame frame: WKFrameInfo,
-          decisionHandler: @escaping (WKPermissionDecision) -> Void) {
+                 requestDeviceOrientationAndMotionPermissionFor origin: WKSecurityOrigin,
```
jcesarmobile, line 65: looks like there is some problem with swiftlint or it's not happy with this lines anymore, but can you remove it from this PR?

cli/src/android/update.ts
```diff
@@ -350,6 +350,65 @@ ext {
...
+  config: Config,
+  cordovaPlugins: Plugin[],
+  pluginsGradlePath: string,
+) {
+  let kotlinNeeded =
+    config.app.extConfig.cordova?.preferences?.GradlePluginKotlinEnabled ===
+    'true';
+  if (!kotlinNeeded) {
+    for (const plugin of cordovaPlugins) {
+      let platforms: any[] = plugin.xml.platform;
```
jcesarmobile, line 366: we have a helper function to get the platform, better use that, see an example of how to get the source-files, it's similar, but for consistency
https://github.com/ionic-team/capacitor/blob/main/cli/src/android/update.ts#L360-L365

cli/src/android/update.ts
```diff
@@ -350,6 +350,65 @@ ext {
     join(config.android.cordovaPluginsDirAbs, 'cordova.variables.gradle'),
     cordovaVariables,
   );
+  await kotlinNeededCheck(config, cordovaPlugins, pluginsGradlePath);
+}
+
+async function kotlinNeededCheck(
```
jcesarmobile, line 356: since it's called kotlinNeededCheck, I think it should just return true or false and keep the gradle writing in the handleCordovaPluginsGradle function, you can move the check before the current write on line 341 and avoid to read and rewrite
Also, you can add the `'androidx.core:core-ktx:1.6.0'` and `"org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"` dependencies to the `frameworksArray` when needed so you can avoid a second replace of `SUB-PROJECT\sDEPENDENCIES\sSTART`
The core-ktx version should be configurable too.

jcesarmobile (changes requested):
added a few comments

ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
IT-MikeS: Yeah not sure what's up there. Removed it 👍

cli/src/android/update.ts
IT-MikeS: Restructured 👍

giralte-ionic (approved):

jcesarmobile (approved):

LifeIsStrange:
Thanks @IT-MikeS !!

PR #5495, opened 2022-03-10
feat(ios): add getConfig to CAPPlugin

### LINKED ISSUE #5476: feat(ios): add getConfig and deprecate getConfigValue
### Description
android has a getConfig method for easier access to plugin configurations
it also has a getConfigValue, but it's deprecated

iOS only has the getConfigValue

We should deprecate getConfigValue and add a getConfig for consistency between platforms

### Platform(s)
iOS

IT-MikeS:
closes #5476

### DIFF
--- a/ios/Capacitor/Capacitor/CAPPlugin.h
+++ b/ios/Capacitor/Capacitor/CAPPlugin.h
@@ -4,6 +4,8 @@
 @protocol CAPBridgeProtocol;
 @class CAPPluginCall;
 
+@class CAPInstanceConfiguration;
+
 @interface CAPPlugin : NSObject
 
 @property (nonatomic, weak, nullable) WKWebView *webView;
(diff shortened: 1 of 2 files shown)


jcesarmobile (changes requested):
The idea is to do something like what was done for Android here https://github.com/ionic-team/capacitor/commit/9820a30688f0a774eced1676f1927cacde53301f

To have a `PluginConfig` class with helper methods to get strings, ints, etc. and the new getConfig should return that class with the config for that plugin only, not the whole configuration, as that's already accesible from the bridge.

ios/Capacitor/Capacitor/CAPPlugin.h
```diff
@@ -44,7 +46,8 @@
 -(NSString* _Nonnull)getId;
 -(BOOL)getBool:(CAPPluginCall* _Nonnull) call field:(NSString* _Nonnull)field defaultValue:(BOOL)defaultValue DEPRECATED_MSG_ATTRIBUTE("Use accessors on CAPPluginCall instead. See CAPBridgedJSTypes.h for Obj-C implementations.");
 -(NSString* _Nullable)getString:(CAPPluginCall* _Nonnull)call field:(NSString* _Nonnull)field defaultValue:(NSString* _Nonnull)defaultValue DEPRECATED_MSG_ATTRIBUTE("Use accessors on CAPPluginCall instead. See CAPBridgedJSTypes.h for Obj-C implementations.");
--(id _Nullable)getConfigValue:(NSString* _Nonnull)key;
+-(id _Nullable)getConfigValue:(NSString* _Nonnull)key __deprecated;
```
jcesarmobile, line 49:
```suggestion
-(id _Nullable)getConfigValue:(NSString* _Nonnull)key __deprecated_msg("use getConfig() and access config values using the methods available depending on the type.");
```
Using the message users can know what to use instead

ios/Capacitor/Capacitor/CAPPlugin.m
```diff
@@ -1,5 +1,6 @@
 #import "CAPPlugin.h"
 #import "CAPBridgedJSTypes.h"
+#import "CAPInstanceConfiguration.h"
```
jcesarmobile, line 3:
```suggestion
```

Not needed, I think you imported it on the first commit when the function returned CAPInstanceConfiguration

jcesarmobile (changes requested):
getArray on Android has 2 functions, one with default and another without it, here it's missing the without default option

run `npm run fmt` as swift files have a bunch of  `;` at the end on swift files that not sure why the lint is not complaining about them, but fmt removes them.

IT-MikeS:
adjusted in latest commit

jcesarmobile (approved):

PR #5503, opened 2022-03-15
chore(android): bump core min sdk

IT-MikeS:
addresses: WN-345


IT-MikeS:
Makes sense, I've adjust the PR

jcesarmobile (approved):

a-poliusov:
Guys, why do you manually update the minSdkVersion? What is the problem to give the user some option to set the version by themself?

Issue #5408, opened 2022-02-03
prevent app from loading if the url field is incorrect


jcesarmobile:
It has been reported a few times an issue where the plugins don't work on android and it's due to the url being incorrect, just the IP address and port, without http or https, the app still loads, but not the plugins
the url field should be validated and show some log about it being incorrect and prevent from loading the app content

jcesarmobile:
addressed here https://github.com/ionic-team/capacitor/pull/5544

PR #5544, opened 2022-03-28
fix(android): prevent app from loading if server.url is invalid 

theproducer:
closes: https://github.com/ionic-team/capacitor/issues/5408


jcesarmobile (changes requested):
When the url is created, we have a try-catch that is doing nothing at the moment.
I think it should log the error there instead of validating the url after creation, if it's invalid the catch will be run with a MalformedURLException

https://github.com/ionic-team/capacitor/blob/622d62fc0d7cd79558bf6f11331bd7d6690aa4f9/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L215-L218

jcesarmobile (approved):

PR #5546, opened 2022-03-29
feat(ios): Add overrideable routing for CAPBridgeViewController subclasses

Steven0351:


ios/Capacitor/Capacitor/Router.swift
```diff
@@ -0,0 +1,27 @@
...
+//
+
+import Foundation
+
+public protocol Router {
+    func route(for path: String) -> String
+}
+
+// swiftlint:disable:next type_name
+internal struct _Router: Router {
```
Steven0351, line 16: I disabled the swiftlint violation here because this is [common practice](https://github.com/apple/swift-corelibs-foundation/blob/759d9ebcd00914d0ed5c4f63b56a594f37983a3a/Sources/Foundation/JSONDecoder.swift#L21) in the Swift standard library to use an _ prefix for non-public types

Steven0351: I'm fine with naming it something different if that's what we want to do as well.

jcesarmobile (approved):

Issue #3854, opened 2020-11-24
bug: file scheme doesn't work on Android


jcesarmobile:
from https://forum.ionicframework.com/t/cannot-change-server-androidscheme-with-angular-capacitor-app/200109

if setting the android scheme to file, the app doesn't correctly load

```
"server": {
    "androidScheme": "file"
}
```

I think we should not allow to use file as value on `androidScheme`, instead show an error and default to http

jcesarmobile:
addressed here https://github.com/ionic-team/capacitor/pull/5550

PR #5550, opened 2022-03-30
feat(android):  don't allow server.androidScheme to be set to schemes handled by WebView

theproducer:
closes: https://github.com/ionic-team/capacitor/issues/3854

### DIFF
--- a/android/capacitor/build.gradle
+++ b/android/capacitor/build.gradle
@@ -47,8 +47,8 @@ android {
         warningsAsErrors true
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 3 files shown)


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -32,6 +32,7 @@
 import com.getcapacitor.util.PermissionHelper;
 import com.getcapacitor.util.WebColor;
 import java.io.File;
+import java.lang.reflect.Array;
```
jcesarmobile, line 35:
```suggestion
```

it's unused now that the code has been moved

jcesarmobile (approved):

PR #5558, opened 2022-04-08
feat(cli): support capacitor portals plugin changes needed

carlpoole:


cli/src/tasks/copy.ts
```diff
@@ -70,13 +72,38 @@ export async function copy(
...
+          config.app.webDirAbs,
+        );
+      }
       await copyCapacitorConfig(config, config.ios.nativeTargetDirAbs);
       const cordovaPlugins = await getCordovaPlugins(config, platformName);
       await handleCordovaPluginsJS(cordovaPlugins, config, platformName);
     } else if (platformName === config.android.name) {
-      await copyWebDir(config, config.android.webDirAbs);
+      if (isFederated) {
+        await copyFederatedWebDirs(config, await config.android.webDirAbs);
```
Steven0351, line 99: I noticed here that we're awaiting here and not on line 103, did `config.android.webDirAbs` become async?

carlpoole: Nice catch, thanks

Steven0351 (approved):

cli/src/tasks/copy.ts
```diff
@@ -70,23 +72,55 @@ export async function copy(
       'capacitor:copy:before',
     );
 
+    const allPlugins = await getPlugins(config, platformName);
+    let usesCapacitorPortals = false;
+    if (
+      allPlugins.filter(plugin => plugin.id === '@capacitor/federation')
```
carlpoole, line 78: This needs to be changed to reflect the actual package name we use for Capacitor Portals Plugin

IT-MikeS (approved):
This looks good to me, nice and clean 👍

jthoms1:
@jcesarmobile Brought up the point that we might use hooks from plugins rather than having the copy code in CLI core. We discussed this as a group and feel like that might cause more issues moving forward.

Hooks for platforms makes more sense because the code mods are only controlled by a single codebase. Electron/iOS/Android.

But in the case of allowing for code mods or hooks from Plugins we run into issues. These hooks can introduce unpredictable side effects. Where the order of plugin loading can yield different results and individual plugins can cause major issues with projects without devs realizing this in the usage.

For this case we believe it is best to continue with this simplistic approach of having the code mods in the CLI core codebase.

IT-MikeS (approved):

Steven0351 (approved):

PR #5577, opened 2022-04-22
fix(ios): Add _Router() to WebViewAssetHandler initializer in CAPWebView

Steven0351:


jcesarmobile:
I've cherry-picked into https://github.com/ionic-team/capacitor/pull/5576

Issue #5564, opened 2022-04-17
feat: load plugin classes in `BridgeFragment` 


wokkaflokka:
### Description

`BridgeActivity` calls `loadPluginClasses()` in `onStart` and registers/loads the plugins it [finds](https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java#L100-L110).

`BridgeFragment` does not provide this behavior. BridgeFragment can be subclassed to add this functionality, but it would be preferable if this were in the base implementation.

### Platform(s)
Android

### Preferred Solution
Update the `load` method in `BaseFragment` to load plugins using the same technique as `BridgeActivity`.

### Alternatives
Sub-classing.

jcesarmobile:
This is on purpose, while a regular Capacitor app, which only has a `BridgeActivity` should make all plugins available to the app, when using `BridgeFragment` you might want to make available only certain plugins depending on the fragment, since you can have several `BridgeFragment`s with different code that would be using different plugins.

PR #5578, opened 2022-04-22
feat(android): load JSON-defined plugins in `BridgeFragment` (#5564)

wokkaflokka:
This PR proposes a solution for #5564.

### Description

`BridgeActivity` calls `loadPluginClasses()` in `onStart` and registers/loads the plugins it [finds](https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/java/com/getcapacitor/BridgeActivity.java#L100-L110).

`BridgeFragment` does not provide this behavior. BridgeFragment can be subclassed to add this functionality, but it would be preferable if this were in the base implementation.

### Platform(s)
Android

### Preferred Solution
Update the `onCreateView` method in `BaseFragment` to load plugins using the same technique as `BridgeActivity`.

### Alternatives
Sub-classing.


jcesarmobile:
This is on purpose, while a regular Capacitor app, which only has a `BridgeActivity` should make all plugins available to the app, when using `BridgeFragment` you might want to make available only certain plugins depending on the fragment, since you can have several `BridgeFragment`s with different code that would be using different plugins.

wokkaflokka:
@jcesarmobile thanks for following up and for the explanation. That's reasonable when explained in that way.

I could see a case for manifesting this information in documentation somewhere, but I could also see how it may be low value given that using `BridgeFragment` directly may be a less-common use-case.

Thanks again!

PR #5579, opened 2022-04-22
fix(android): move initialFocus on webview into config

carlpoole:
Moves the initial focus action of the webview on Android into a config flag which is `true` by default for Capacitor proper, and `false` by default for programmatic setup (i.e: Portals) where we may not want the webview to have the initial focus.

This line causes some problems on Portals apps by causing the webview to be selected when a page loads, and this may not be desired. With it being a config property, a developer could still enable it at any time if preferred. Portals apps will be configured to set this property to false by default.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -447,7 +447,10 @@ public class Bridge {
             Logger.debug("WebView background color not applied");
         }
 
-        webView.requestFocusFromTouch();
+        if (config.isInitialFocus()) {
+            webView.requestFocusFromTouch();
+        }
+
         WebView.setWebContentsDebuggingEnabled(this.config.isWebContentsDebuggingEnabled());
     }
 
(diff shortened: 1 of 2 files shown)


jcesarmobile (changes requested):
Despite the main purpose of this configuration is to be used in Portals, I think it still should be included in the [CLI declarations ](https://github.com/ionic-team/capacitor/blob/main/cli/src/declarations.ts)so it can be configured from typescript too

carlpoole:
> Despite the main purpose of this configuration is to be used in Portals, I think it still should be included in the [CLI declarations ](https://github.com/ionic-team/capacitor/blob/main/cli/src/declarations.ts)so it can be configured from typescript too

Thank you, good catch. This was the intention but I missed it. I added it now

cli/src/declarations.ts
```diff
@@ -203,6 +203,14 @@ export interface CapacitorConfig {
      * @since 3.1.0
      */
     flavor?: string;
+
+    /**
+     * Wether to give the webview initial focus.
```
Steven0351, line 208: slight typo here: Wether -> Whether

Steven0351: Screen Shot 2022-04-29 at 1 43 58 PM
😄

Steven0351 (approved):

jcesarmobile (approved):

Issue #4444, opened 2021-04-07
bug: capacitor fails to load wasm (Incorrect response MIME type. Expected 'application/wasm)


JoshuaBThompson:
### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 2.4.7
  @capacitor/core: 2.4.7
  @capacitor/android: 2.4.7
  @capacitor/electron: 2.4.7
  @capacitor/ios: 2.4.7

Installed Dependencies:

  @capacitor/android not installed
  @capacitor/cli 2.4.6
  @capacitor/core 2.4.6
  @capacitor/ios 2.4.6
  @capacitor/electron not installed


```

### Platform(s)

mac os (10.15.7 catalina)

### Current Behavior
installed micropython.js (npm install micropython) which uses wasm
vue js app with capacitor

in app.js
require("micropython");

when loading browser after running npm run serve I get:
```
micropython.js:1779 wasm streaming compile failed: TypeError: Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'.
(anonymous) @ micropython.js:1779
Promise.then (async)
createWasm @ micropython.js:1776
push../node_modules/micropython/lib/micropython.js.Module.asm @ micropython.js:1809
(anonymous) @ micropython.js:3320
./node_modules/micropython/lib/micropython.js @ micropython.js:3985
__webpack_require__ @ bootstrap:848
fn @ bootstrap:150
(anonymous) @ index.js:19
./node_modules/micropython/index.js @ index.js:132
__webpack_require__ @ bootstrap:848
fn @ bootstrap:150
./node_modules/cache-loader/dist/cjs.js?!./node_modules/babel-loader/lib/index.js!./node_modules/cache-loader/dist/cjs.js?!./node_modules/vue-loader/lib/index.js?!./src/App.vue?vue&type=script&lang=js& @ App.vue:63
__webpack_require__ @ bootstrap:848
fn @ bootstrap:150
./src/App.vue?vue&type=script&lang=js& @ App.vue?9241:1
__webpack_require__ @ bootstrap:848
fn @ bootstrap:150
./src/App.vue @ App.vue:1
__webpack_require__ @ bootstrap:848
fn @ bootstrap:150
./src/main.js @ FileExplorer.vue?7552:1
__webpack_require__ @ bootstrap:848
fn @ bootstrap:150
1 @ app.js:3604
__webpack_require__ @ bootstrap:848
[...]

Issue #1864, opened 2019-08-11
WebAssembly Mime-type problem


XamiKami:
Hey!

Im currently trying to set up an Ionic/React based App with Capacitor, that in part is based on WebAssembly.

Trying to start that App on the Emulator i get a blank screen. After looking into the chrome://inspect thing, I'm able to see the problem:
```
Uncaught (in promise) TypeError: Failed to execute 'compile' on 'WebAssembly': 
Incorrect response MIME type. Expected 'application/wasm'.
```
When using it in the browser (ionic serve etc) it does work, so this seems to be a Capacitor-problem.

Any help? 
if you believe [#774](https://github.com/ionic-team/capacitor/issues/774), Wasm is supposed to work...

Thanks in advance for any help ;D

jcesarmobile:
sample app https://github.com/XamiKami/Capacitor-Wasm-problem-simple-case-

rokit:
@jcesarmobile That link is now broken.

Issue #774, opened 2018-08-25
WebAssembly support


HannanAkbar:
It would be great to add WebAssembly support. Is it planned?

bmsantos:
@HannanAkbar, yes, Capacitor works just fine with WASM and I don't think that it requires any official support. You can try it with Blazor using Bionic, an experimental framework for Blazor that tries to have the amazing feeling of Ionic CLI. Its all experimental at this point.
https://bmsantos.github.io/bionic/

jcesarmobile:
As @bmsantos said, it already works as it’s implemented by the browser, going to close the issue. 

If you face specific wasm issues that only apply to Capacitor, please, report them providing more details and we will try to fix it

XamiKami:
It in fact does not work, [see my issue here](https://github.com/ionic-team/capacitor/issues/1864)
because it doesn't seem to accept the `application/wasm` Mime-type

PR #5585, opened 2022-04-29
Fixes WASM mimetype for static file server in a production-bundled iOS app

ggaabe:
WebAssembly imports don't work correctly on iOS production builds that must serve files within the app instead of over a dev server. This happens because iOS production builds use the `file://` protocol to retrieve build files from a static file server, which can strip the MIME type if the file extension isn't recognized by the static file server.

This issue doesn't happen on Android, or in dev iOS builds because both retrieve their build files from a localhost server, which provides the correct MIME type.

This fix defines the WASM file extension for the static file server within the iOS app.

This fix may resolve the WASM MIME type issues listed here:
https://github.com/ionic-team/capacitor/issues/774
https://github.com/ionic-team/capacitor/issues/1864
https://github.com/ionic-team/capacitor/issues/4444

**Minimum reproduction:**

1. Install npm package `@emurgo/cardano-serialization-lib-browser`
2. Attempt to import it in a packaged production iOS build (not hosted from a dev server): 
```ts 
await (
      await import('@emurgo/cardano-serialization-lib-browser/cardano_serialization_lib')
).default
````
3. The app will throw an error: `Failed to execute 'compile' on 'WebAssembly': Incorrect response MIME type. Expected 'application/wasm'`

### CODE REVIEW

**Comment by iamcco** [comment]
Any update ?

**Comment by ggaabe** [comment]
@iamcco Not sure how long it'll take before / if this gets reviewed but if you add the changes I made here to your own info.plist the issue will be resolved.

**Comment by iamcco** [comment]
> @iamcco Not sure how long it'll take before / if this gets reviewed but if you add the changes I made here to your own info.plist the issue will be resolved.

Thanks, it works!

**Comment by Janaka-Steph** [comment]
Thank you so much, it works!

**Comment by mudlee** [comment]
I can confirm that these added lines solve my wasm issue as well.

**Review by markemer — APPROVED** [review]
This seems reasonable - I don't have a lot of WASM to test with but now that we have the Cap5 WASM fix, this seems like a good idea.

**Comment by jcesarmobile** [comment]
This is a change in the iOS app template, so if this gets merged it would be included in every capacitor app created after the PR gets merged, even if not using WASM. Most users don’t use WASM, so I don’t think this should be part of every app, in any case we should document that this might be needed for using WASM, not even sure if it’s needed for most WASM users as I’ve used WASM in the past without doing this change. 
If something broke since last time I tested WASM, then we should try to fix it from the code that returns the mime type, not from the template.

**Comment by ggaabe** [comment]
I provided minimum conditions needed to reproduce with a specific WASM NPM package installation and import to test with that would trigger the issue.

This issue only affects iOS users on production-packaged builds.

I no longer work at the job where this issue had occurred and my fix was needed so I probably will not also spend time building out a template app to demonstrate reproduction. But I know this is an issue that has affected other WASM users with capacitor and that the fix works.

**Comment by markemer** [comment]
I don't know how we handle `file://` urls in the app - but this is a pretty easy way to support wasm from the filesystem or remote file servers. I'm not sure of the harm of adding it for people that don't need it, but if we wanted to add it using the CLI or further refine the WASM code that could work too. @ItsChaceD had some WASM fixes, but I think they were for HTTP delivered assets.

**Comment by arunrdd** [comment]
Sharing my experience in setting up [sql.js](https://sql.js.org/) with capacitor, which worked nicely when the changes proposed above are made to the `Info.plist` file.

I used `vite-plugin-static-copy` to copy the wasm file to the build folder using following configuration:
```
import { defineConfig } from 'vite';
import { viteStaticCopy } from 'vite-plugin-static-copy';

export default defineConfig({
	plugins: [
		viteStaticCopy({
			targets: [{
				src: 'node_modules/sql.js/dist/sql-wasm.wasm',
				dest: 'wasm-files'
			}]
		})
	]
});
```

When the app tried to access `${window.location.origin}/wasm-files/sql-wasm.wasm`, it received a wrong response MIME type `application/octet-stream`. I can confirm now that this works in production iOS builds. Thank you @ggaabe for the fix!

PR #5590, opened 2022-05-03
chore(ci): change ci to use xcode 13

IT-MikeS:


jcesarmobile (changes requested):
I think it's better to use macos-12 instead of macos-latest as latest will eventually point to 12 (it's still 11 at the moment) and things might break/change.
Also, I think we should use latest Xcode (13.3.1) instead of 13.0

jcesarmobile (approved):

PR #5610, opened 2022-05-11
chore: remove labels from PluginConfig accessors

jcesarmobile:
To be consistent with PluginCall accessors
Also add @objc so they can be used from Objective-C too


giralte-ionic (approved):

theproducer (commented):
Looks good, but is there a reason why you didn't expose getArray and getObject to ObjC?

jcesarmobile:
Because JSArray and JSObject can’t be converted to Objective-C, the compiler was complaining about it

theproducer (approved):

IT-MikeS (approved):

PR #5613, opened 2022-05-12
fix: window error msg is null in native-bridge

iGroovyboy:


Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

PR #5622, opened 2022-05-16
feat: promise with callback plugin method

### LINKED ISSUE #5618: feat: plugin method with callback and custom return type
### Description

I want to be able to have a custom return type with a plugin method that has a callback. According to the documentation about [plugin method types](https://capacitorjs.com/docs/plugins/method-types) the return type is a `string` which represents the callback id, but i want to be able to change this type when i don't need the callback id. For my use case i want to call a method which runs a task and gives status updates about it via the callback, after some time the task is expected to finish with a result which i want to receive via the resolved promise.

### Platform(s)

Android

### Preferred Solution

```java
@CapacitorPlugin(name = "Example")
public class ExamplePlugin extends Plugin {
    @PluginMethod(returnType = PluginMethod.RETURN_CALLBACK)
    public void example(PluginCall call) {
        String v = call.getString("v");
        call.setKeepAlive(true);
        call.resolve(new JSObject().put("n", 1));
        call.resolve(new JSObject().put("n", 2));
        call.resolve(new JSObject().put("n", 3));
        call.setKeepAlive(false);
        call.resolve(new JSObject().put("i", 42).put("r", v));
    }
}
```
I would expect when i set keep alive to true to return values via the callback and when i set it again to false the promise is resolved with my custom type. 

### Alternatives

```java
@PluginMethod(returnType = PluginMethod.RETURN_CALLBACK_WITH_PROMISE)
public void example(PluginCall call) {/* ... */}
```
Introduce a new method type which could be called `RETURN_CALLBACK_WITH_PROMISE` to allow a custom return type in the promise.

#### Web
On the web this is already possible:
```typescript
import { WebPlugin } from "@capacitor/core";

interface ExampleOptions {
  v: string
}

interface ExampleCallbackValue {
  n: number
}

type ExampleCallback = (value: ExampleCallbackValue) => void;

interface ExampleReturnValue {
  i: number,
  r: string
}

interface ExamplePlugin {
  example(options: ExampleOptions, callback: ExampleCallback): Promise<ExampleReturnValue>;
}

export class ExampleWeb
  extends WebPlugin
  implements ExamplePlugin
{
  async example(options: ExampleOptions, callback: ExampleCallback): Promise<ExampleReturnValue> {
    return new Promise(resolve => {
      callback({ n: 1 });
      callback({ n: 2 });
      callback({ n: 3 });
      resolve({ i: 42, r: options.v })
    });
  }
}
```

Calling the example method with the following code:

```typescript
Example.example({ v: "foo" }, (value: ExampleCallbackValue) => console.log("callback", value))
          .then((value: ExampleReturnValue) => console.log("resolved", value));
```

Prints the expected behaviour into the console:

web console output

#### Android
The prefered solution currently passes the object in the last `call.resolve()` after setting keep alive to false to the callback.

android console output

matthiasschwarz:
### Description
Fixes #5618
Allows to add a callback function to a plugin method which returns a promise. Calling `call.resolve()` on keep alive passes the object to the callback - when a callback was provided. Setting keep alive to false passes the object to the promise and resolves it. `call.reject()` rejects the promise and deletes the saved callback.

This should not be a breaking change to the [value return method type](https://capacitorjs.com/docs/plugins/method-types#value-return) when no callback function is provided to the method or keep alive is never set to true.

#### Android example usage
```java
@CapacitorPlugin(name = "Example")
public class ExamplePlugin extends Plugin {
    @PluginMethod()
    public void example(PluginCall call) {
        String v = call.getString("v");
        call.setKeepAlive(true);
        call.resolve(new JSObject().put("n", 1));
        call.resolve(new JSObject().put("n", 2));
        call.resolve(new JSObject().put("n", 3));
        call.setKeepAlive(false);
        call.resolve(new JSObject().put("i", 42).put("r", v));
    }
}
```

### Platform(s)
- Android
- iOS (untested)

### DIFF
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -405,7 +405,23 @@ const nativeBridge = (function (exports) {
                                 return err;
                             }, new cap.Exception(''));
                         }
-                        if (typeof storedCall.callback === 'function') {
+                        if (typeof storedCall.callback === 'function' &&
+                            typeof storedCall.resolve === 'function') {
+                            // promise with callback
+                            if (result.success) {
+                                if (result.save) {
+                                    storedCall.callback(result.data);
+                                }
+                                else {
+                                    storedCall.resolve(result.data);
+                                }
+                            }
+                            else {
+                                storedCall.reject(null, result.error);
+                                callbacks.delete(result.callbackId);
+                            }
+                        }
+                        else if (typeof storedCall.callback === 'function') {
                             // callback
                             if (result.success) {
                                 storedCall.callback(result.data);
@@ -451,9 +467,10 @@ const nativeBridge = (function (exports) {
                 }
                 return cap.toNative(pluginName, methodName, options, { callback });
             };
-            cap.nativePromise = (pluginName, methodName, options) => {
+            cap.nativePromise = (pluginName, methodName, options, callback) => {
                 return new Promise((resolve, reject) => {
                     cap.toNative(pluginName, methodName, options, {
+                        callback: callback,
                         resolve: resolve,
                         reject: reject,
                     });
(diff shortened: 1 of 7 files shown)


eric-horodyski:

More context:

PR #5629, opened 2022-05-18
chore(android): silencing / fixing warnings in bridge

theproducer:


giralte-ionic (approved):

jcesarmobile (changes requested):

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -474,19 +472,32 @@ public void registerPlugins(Class<? extends Plugin>[] pluginClasses) {
...
+        CapacitorPlugin pluginAnnotation = pluginClass.getAnnotation(CapacitorPlugin.class);
+
+        if (pluginAnnotation == null) {
+            NativePlugin legacyPluginAnnotation = pluginClass.getAnnotation(NativePlugin.class);
+            if (legacyPluginAnnotation == null) {
+                Logger.error("Plugin doesn't have the @CapacitorPlugin annotation. Please add it");
+                return null;
+            }
+
+            return legacyPluginAnnotation.name();
```
jcesarmobile, line 486: I meant to put in a new function just the deprecated code, with a name like "getLegacyName" so in the future when it's no longer used we just remove the legacy functions.

theproducer: Ah, makes sense!  I just pushed an update.

PR #5636, opened 2022-05-23
chore: cap4 deployment automations

IT-MikeS:


giralte-ionic (approved):

theproducer (approved):

.github/workflows/ci.yml
```diff
@@ -127,58 +127,58 @@ jobs:
...
-        with:
-          fetch-depth: 0
-      - name: Restore Dependency Cache
-        uses: actions/cache@v1
-        with:
-          path: ~/.npm
-          key: ${{ runner.OS }}-dependency-cache-${{ hashFiles('**/package.json') }}
-      - run: npm install
-      - run: npm run lerna:publish:latest || true
+  # deploy-next:
```
jcesarmobile, line 130: delete instead of comment

jcesarmobile (changes requested):
Added a minor comment.
Also I've realized we are missing the RC release, I think we should have that one too (there was a `release:rc` that has been removed). We need that because we do alpha, then beta and finally rc before the final is released.
I'm not sure about using the alpha/beta tags, at the moment we are using next tag for all alpha, beta and rc releases, I think a single tag for those would be better since we do them sequentially.

jcesarmobile (approved):

jcesarmobile (approved):

giralte-ionic (approved):

PR #5645, opened 2022-05-26
feat(cli): Add --forwardPorts option to Capacitor CLI

thomasvidas:
Exposes the `--forwardPorts` option from [native-run](https://github.com/ionic-team/native-run) to the Capacitor CLI. When an `npx cap run android --forwardPorts 8000:8000` is run, the following happens.

- A Capacitor Sync
- Device Boot & Connection
- `adb reverse tcp:${port} tcp:${port}`
- Install and run a debug APK

Opening as a Draft PR because
- changes may be needed in native-run to work properly
- maybe a `--live-reload` flag with the port in the config makes more sense?


bastienlemaitre:
@thomasvidas  any new about this feature ?

peitschie:
I'd be keen on this one landing too! Hit an issue today where manual port forwards are cleared by `native-run` under some circumstances, so having a flag like this would make the whole thing easier to deal with. I ended up needing to use the `wait-on` package to help work around things.

markemer:
I've updated this to be merged to `main` and am looking at making sure it works.

theproducer (approved):

IT-MikeS (approved):

jcesarmobile (changes requested):
It's not working, the forwardPorts option is not being passed to the runCommand, should look like this:

```
async (platform, { scheme, flavor, list, target, sync, forwardPorts }) => {
            const { runCommand } = await import('./tasks/run');
            await runCommand(config, platform, {
              scheme,
              flavor,
              list,
              target,
              sync,
              forwardPorts
            });
          },
```

markemer:
Huh, it worked for @theproducer - I'll take a look

markemer:
> await runCommand

It does seem like that is missing.

IT-MikeS (approved):

giralte-ionic (approved):

jcesarmobile (approved):

PR #5647, opened 2022-05-27
fix: Scheme in config file if User rename the Project

oneclickwebstudio:
I have came across to this problem after renaming the Project in Xcode.

```
const nativeXcodeProjDir = `${nativeProjectDir}/App.xcodeproj`;`
```

Shouldn't it be like that?
```
const nativeXcodeProjDir = `${nativeProjectDir}/${scheme}.xcodeproj`;
```


oneclickwebstudio:
By the way after editing this part of code.

Build was successfully finished

jcesarmobile:
Thanks for the PR.
We don't support changing the project name at the moment.
We support changing the scheme name or having multiple schemes and this PR will break that functionality since the scheme name doesn't have to match the project name.

PR #5648, opened 2022-05-27
feat(android): Redirect to Google Play Store if WebView is below the minimum

### LINKED ISSUE #4884: feat: Minimum WebView version on Android
### Description
It would be useful to be able to require a specific minimum WebView version on Android to avoid the "white screen of death" if the application won't work due to lack of newer JS feature support.

### Platform(s)
Android

### Preferred Solution
If the WebView is too old, Capacitor could, for example, display a (localized) error dialog, advise the user to update WebView and abort.

thomasvidas:
Update WebView Version

Fixes #4884

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -1,8 +1,10 @@
 package com.getcapacitor;
 
 import android.app.Activity;
+import android.app.Application;
 import android.content.ActivityNotFoundException;
 import android.content.Context;
+import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.SharedPreferences;
 import android.content.pm.ApplicationInfo;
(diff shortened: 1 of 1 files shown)


thomasvidas:
Future improvements to be made:
- Customizable text/icon. Should we add this to the Capacitor config?
- Localization. We should check the locale and use the localized version of this message for non-English speaking countries.
- Better support for Android R and above. The R minimum WebView version is 61, but the `TYPE_TOAST` was deprecated in O, and the replacement requires a system permission. We should modify the splash screen plugin to allow this alert on top or add steps on how to add the permission.

jcesarmobile:
closing in favor of https://github.com/ionic-team/capacitor/pull/5723

francolondra:
@jcesarmobile  Do you need implement some code ? because I have 
    "@capacitor/android": "4.4.0",
    "@capacitor/app": "4.1.0",
    "@capacitor/core": "4.4.0",
    
and I don't see this popup .

francolondra:
I was investigating and the problem comes from chrome, which has a version lower than 60, so what I implemented was to force it to be updated, I created that script and put it in the angular index.html. It worked very well for me.
`  <script type="text/javascript">
    var userAgentArray = window.navigator.userAgent.split(" ");
    var chromeVersion = a.find(c => c.includes('Chrome/'));
    var version = b.split("/")[1];
    var numberVersion = parseInt(c.toString().split(".")[0]);
    if(numberVersion <=60)
    {
        alert("Debe actualizar su navegador chrome.")
        var url = "https://play.google.com/store/apps/details?id=com.android.chrome";
        window.open(url);
    }
</script>`

PR #5671, opened 2022-06-09
chore(ios): Remove unneeded AppDelegate and CDVViewController from

Steven0351:
Remove unneeded AppDelegate and CDVViewController from CapacitorCordova. This is conflicting with Portals users who are using Objective-C and the CapacitorCordova AppDelegate is clashing with the users AppDelegate in the Objective-C runtime. Since CapacitorCordova is meant for use with Capacitor, there should be no issues removing this as anyone using a pure Cordova app should be using Apache Cordova, not CapacitorCordova.


Steven0351:
> Those classes are needed for the Cordova plugin compatibility, can’t be removed without making apps using plugins that reference AppDelegate and CDVViewController fail to build.
> 
> We could remove this for Capacitor 4 if really really really needed, but will break a few Cordova plugins.

The `AppDelegate` should still exist at runtime for a Capacitor application since it's generated as part of the template. Every iOS application requires one, so I don't think removing `AppDelegate` would be a problem. Do these plugins create an instance of `CDVViewController`? I would imagine the only way this would work generally would be to create from using `NSClassFromString` because that would need to be imported from CapacitorCordova to work to create one directly. This should mean that `CDVViewController` is not created by the Capacitor runtime and would always be `nil`. I would expect if these were being used in a capacitor application that the builds would have failed to compile. 

I of course could be completely wrong, but the fact that Capacitor will compile fine without any of these leads me to believe that the CapacitorCordova `AppDelegate` and the `CDVViewController` are never initialized. That all being said, I'm not opposed to this being a Capacitor 4 change.

jcesarmobile:
A few of Cordova plugins use Objective-C categories on AppDelegate.m and/or CDVViewController.m.
Since Capacitor uses AppDelegate.swift, AppDelegate.m doesn’t exist and those plugin will fail to build, so, it’s not unused, it’s there for that reason and removing them will make a few Cordova plugins to fail to build.
You can remove most code in CapacitorCordova and Capacitor will still build, but that doesn’t mean is not needed, most Cordova plugins will not compile or will not work if you remove CapacitorCordova code that they rely on.

PR #5706, opened 2022-06-21
feat: Add referrerUrl

MateiNenciu:
Expose referrerUrl from userActivity so we can use it on "appUrlOpen" event from capacitor-plugins/app. Apple documentation: https://developer.apple.com/documentation/foundation/nsuseractivity/2875762-referrerurl/
  We got a request from data department and we need to process this referrer.


eric-horodyski:

More context:

PR #5715, opened 2022-06-23
feat: Add CapWebView 

jcesarmobile:


giralte-ionic (approved):

Steven0351 (approved):

Steven0351:
One thing here that slipped my attention was the merge of main back into `portals-dev`. The `native-publish` branch references `portals-dev` for publishing to Cocoapods, which means all the future releases to Cocoapods will follow the 4.0.0 releases. I think while `native-publish` exists for iOS we should probably have `portals-dev` track the 3.x branch and if any changes need to go into `portals-dev` pre 4.0.0 general release we should cherry pick them into main as well.

jcesarmobile:
The idea of merging portals-dev into main is to get rid of portals-dev branch and with that the required extra step on publishing.

native-publish branch will always exist for publishing, we will need to update it to use main instead of portals-dev branch as portals-dev branch would be deleted after the merge of this PR.

If we are going to keep publishing v3 versions, we will need a new native-publish-v3 branch that points to the 3.x branch and cherry pick this CapWebView changes into 3.x branch too.

Issue #5336, opened 2021-12-22
bug: On android the signature of the reject method is different than on ios


1024gs:
## Bug Report
On android the signature of the reject method is different than on ios.

On ios it takes a fourth parameter called data [(see here)](https://github.com/ionic-team/capacitor/blob/a20684180a9b6fd50547ae578f21531faa116da5/ios/Capacitor/Capacitor/CAPPluginCall.swift#L60) and then in javascript the promise will resolve(fail) with that data.

On android It only takes three parameters [(see here)](https://github.com/ionic-team/capacitor/blob/a20684180a9b6fd50547ae578f21531faa116da5/android/capacitor/src/main/java/com/getcapacitor/PluginCall.java#L120) and the promise in javascript will resolve(fail) with a payload such as {message: 'some-string', code: 'some-other-string'}

Can you please make the reject method on android to take the same parameters and to behave the same as on ios.

jcesarmobile:
Android signature is correct, it's iOS what shouldn't return data on error. Capacitor returns a CapacitorException, which should only include a message and a code
https://github.com/ionic-team/capacitor/blob/12c6294b9eb82976b1322f00da9ba5a6004f7977/core/src/util.ts#L22-L26

What's your use case for this?

1024gs:
Here at Masabi we develop a plugin that behaves rather like a proxy, it calls a couple of methods of a third-party SDK and then passes the result up to either resolve or reject methods.

So in my case the SDK fails with a payload, more than just a message and a code, we call it sdkError, and we need it in javascript.

The way ios works at the moment is what we're looking for and besides that, we have done half of the work, for ios, and just now realised that android behaves differently.

I hope it makes sense and that you can help me with this.

1024gs:
@jcesarmobile can I kindly ask you what is your final answer, will you consider to return data on error, would you accept a PR or Capacitor categorically rejects with a CapacitorException, which should only include a message and a code?

jcesarmobile:
I added the issue to the Capacitor 4.0.0 milestone, so we won't be doing any changes on this short term, we might add the data, in that case it should also changed in the CapacitorException types (in capacitor/core), or we might remove the data parameter from iOS. It's not really my decision, will be discussed with the team internally.

1024gs:
Thank you for your answer.

webuniverseio:
I run into a similar issue, for rejection we also wanted to use Capacitor to return specifically formatted rejection reason which would benefit from having just a code and message.

PR #5719, opened 2022-06-24
feat(android): Add Optional Data Param for Error Object

ItsChaceD:
This PR adds an optional data object to be passed to `call.reject()` in order to be consistent with the method signatures on iOS.

closes https://github.com/ionic-team/capacitor/issues/5336


jcesarmobile (changes requested):
The types should be updated to include the optional data
https://github.com/ionic-team/capacitor/blob/12c6294b9eb82976b1322f00da9ba5a6004f7977/core/src/util.ts#L22-L26

IT-MikeS (approved):

jcesarmobile (approved):

PR #5723, opened 2022-06-27
feat: Add option for custom error page

theproducer:
This PR adds an option for displaying a local, custom error page in the case of an outdated web view (on Android), or in cases of web view errors.

closes: https://github.com/ionic-team/capacitor/issues/4884

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -172,7 +172,7 @@ public class WebViewLocalServer {
             return null;
         }
 
-        if (isLocalFile(loadingUrl) || isMainUrl(loadingUrl) || !isAllowedUrl(loadingUrl)  ||isErrorUrl(loadingUrl)) {
+        if (isLocalFile(loadingUrl) || isMainUrl(loadingUrl) || !isAllowedUrl(loadingUrl) || isErrorUrl(loadingUrl)) {
             Logger.debug("Handling local request: " + request.getUrl().toString());
             return handleLocalRequest(request, handler);
         } else {


giralte-ionic (approved):

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -257,10 +261,56 @@ private void loadWebView() {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            PackageInfo info = WebView.getCurrentWebViewPackage();
+            String majorVersionStr = info.versionName.split("\\.")[0];
+            int majorVersion = Integer.parseInt(majorVersionStr);
+            return majorVersion >= MINIMUM_ANDROID_WEBVIEW_VERSION;
+        }
+
+        // Otherwise manually check WebView versions
+        try {
+            PackageInfo info = pm.getPackageInfo("com.google.android.webview", 0);
```
jcesarmobile, line 293:
```suggestion
            String webViewPackage = "com.google.android.webview";
            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
                webViewPackage = "com.android.chrome";
            }
            PackageInfo info = pm.getPackageInfo(webViewPackage, 0);
```

On Android N the device uses chrome as the system webview, current code returns 0 on my Android N device, but with this change it returns the correct version

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -257,10 +261,56 @@ private void loadWebView() {
...
+            String majorVersionStr = info.versionName.split("\\.")[0];
+            int majorVersion = Integer.parseInt(majorVersionStr);
+            return majorVersion >= MINIMUM_ANDROID_WEBVIEW_VERSION;
+        } catch (Exception ex) {
+            Logger.warn("Unable to get package info for 'com.google.android.webview'", ex.toString());
```
jcesarmobile, line 298:
```suggestion
            Logger.warn("Unable to get package info for 'com.google.android.webview'");
```
Logger.warn accepts tag + message or message, so you are passing the message as tag
Either remove the ex.toString() or concat it to the message instead of being an additional param

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -257,10 +261,56 @@ private void loadWebView() {
...
+        }
+
+        try {
+            PackageInfo info = pm.getPackageInfo("com.android.webview", 0);
+            String majorVersionStr = info.versionName.split("\\.")[0];
+            int majorVersion = Integer.parseInt(majorVersionStr);
+            return majorVersion >= MINIMUM_ANDROID_WEBVIEW_VERSION;
+        } catch (Exception ex) {
+            Logger.warn("Unable to get package info for 'com.android.webview'", ex.toString());
```
jcesarmobile, line 307:
```suggestion
            Logger.warn("Unable to get package info for 'com.android.webview'");
```
Logger.warn accepts tag + message or message, so you are passing the message as tag
Either remove the ex.toString() or concat it to the message instead of being an additional param

jcesarmobile (changes requested):

jcesarmobile (changes requested):
Usually when we have a config option that can be configured for ios as `ios.something` and for android as `android.something` we also have a common `something` config option on the root so it can be configured for both platforms on a single place.

Also, not sure if it would make more sense to have the config option inside the `server` object since the error path is related to the WebView navigation (I mean, instead of having three `errorPath`, `ios.errorPath` and `android.errorPath`, to have only `server.errorPath`.

theproducer:
> Usually when we have a config option that can be configured for ios as `ios.something` and for android as `android.something` we also have a common `something` config option on the root so it can be configured for both platforms on a single place.
> 
> Also, not sure if it would make more sense to have the config option inside the `server` object since the error path is related to the WebView navigation (I mean, instead of having three `errorPath`, `ios.errorPath` and `android.errorPath`, to have only `server.errorPath`.

Do you think there would be any benefit to allowing the error path to be set independently per platform?  If not, I agree with having one option in `server.errorPath`.

jcesarmobile:
Not sure, but now that we allow .ts config files people can still have different values if they need to, so not sure if it's worth having both options.

jcesarmobile (changes requested):
How should the `errorPath` look like?
On Android I had to use it like `"errorPath": "/error.html"`, while on iOS I had to use it like `"errorPath": "error.html"` (without the `/`).
I think it should be like on iOS (without the `/`).

If using a `server.url` and there is no internet connection, it will try reload form the `server.url`, but appending the error page. i.e. if `server.url` is `https://nonexistent.com` and `errorPath` is `error.html`, it will try to reload `https://nonexistent.com/error.html`, which will continue to fail because there is no internet. It should try to load from the local url, in example `capacitor://localhost/error.html` on iOS or `http://localhost/error.html` on Android. (there seem to be a problem on iOS that prevents the `capacitor://localhost/error.html` url at the moment, we should probably fix that first.

Also, Capacitor's javascript should not be injected on the error page, since the injected javascript is not compatible with WebView in version 55 or older. Should be handled on the WebViewLocalServer file to exclude the error path.

jcesarmobile (changes requested):

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -257,10 +261,60 @@ private void loadWebView() {
...
+                Logger.error(MINIMUM_ANDROID_WEBVIEW_ERROR);
+            }
+        }
+
         // Get to work
         webView.loadUrl(appUrl);
     }
 
+    @SuppressLint("WebViewApiAvailability")
+    private boolean isMinimumWebViewInstalled() {
```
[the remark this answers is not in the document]
theproducer: Good point, will do!

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
```diff
@@ -56,6 +56,15 @@ public void onReceivedError(WebView view, WebResourceRequest request, WebResourc
                 listener.onReceivedError(view);
             }
         }
+
+        String errorPath = bridge.getErrorUrl();
+        String appUrl = bridge.getAppUrl();
+        if (appUrl.charAt(appUrl.length() - 1) != '/') {
+            appUrl += "/";
+        }
+        if (errorPath != null && request.getUrl().toString().equals(appUrl)) {
```
jcesarmobile, line 65: Why does this compare if the url to load is different from the appUrl? on iOS it's not comparing that.
If it's because when loading the error page the app enters into a loop of failures because of favicon.ico not loading, I think it's best to check if the request was to the main frame
`if (errorPath != null && request.isForMainFrame()) {`.

jcesarmobile (changes requested):
Line 175 of WebViewLocalServer.java should also be changed to check if it is the error url, otherwise when using server.url, if the url fails to load it won't be able to load the local assets. Should be something like:
`if (isLocalFile(loadingUrl) || isMainUrl(loadingUrl) || !isAllowedUrl(loadingUrl) || isErrorUrl(loadingUrl)) {`

Also, error page redirect code is also needed in `onReceivedHttpError` for Android in case it's using server.url and the url trying to load returns 404 or similar error, in that case it goes to `onReceivedHttpError` instead of `onReceivedError`

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
theproducer: > If it's because when loading the error page the app enters into a loop of failures..
Yes, this was exactly why.  Thanks!

jcesarmobile (approved):
looks good

justinpathrose:
@theproducer We are trying to implement this in our app. I get that error page is shown based on minWebViewVersion, but can't figure out  how this works on iOS. Is there a config we can use to say the minimum supported iOS version?

Thanks.

PR #5731, opened 2022-06-30
chore(ci): Change script to publish from 3.x branch

jcesarmobile:
After [CapWebView is added to 3.x branch](https://github.com/ionic-team/capacitor/pull/5730), native-publish will need to pick the code from there instead of from portals-dev


jcesarmobile:
good point, the `fetch-depth: 0` is also no longer needed if using the tag

Steven0351 (approved):

IT-MikeS (approved):

Issue #3302, opened 2020-07-20
Upgrade Tool


imhoffd:
We could create a visual upgrade tool that shows the differences between Capacitor versions of the base project. React Native has this: https://react-native-community.github.io/upgrade-helper/?from=0.62.2&to=0.63.1

- [https://reactnative.dev/docs/upgrading](https://reactnative.dev/docs/upgrading)
- [https://react-native-community.github.io/upgrade-helper/](https://react-native-community.github.io/upgrade-helper/)
- [https://github.com/ionic-team/capacitor/issues/3302](https://github.com/ionic-team/capacitor/issues/3302)
- [https://github.com/ionic-team/capacitor/discussions/3298#discussioncomment-39895](https://github.com/ionic-team/capacitor/discussions/3298#discussioncomment-39895)

ref: https://github.com/ionic-team/capacitor/discussions/3298

PR #5762, opened 2022-07-12
feat(cli): add migrator for cap3 to cap4

IT-MikeS:
This ports in the Ionic vscode extensions migration functions.

closes https://github.com/ionic-team/capacitor/issues/3302


giralte-ionic (approved):

cli/src/tasks/migrate.ts
```diff
@@ -0,0 +1,590 @@
...
+}
+
+function updateAndroidManifest(filename: string) {
+  const txt = readFile(filename);
+  if (!txt) {
+    return;
+  }
+
+  // AndroidManifest.xml add attribute: <activity android:exported="true"
+  if (txt.includes('<activity android:exported="')) {
```
jcesarmobile, line 292: the exported can be on any line of the activity tag, I personally put it on the last line and this doesn't detect it and duplicates it.

cli/src/tasks/migrate.ts
```diff
@@ -0,0 +1,590 @@
...
+        replaced,
+        `classpath '${dep}:`,
+        `'`,
+        neededDeps[dep],
+      );
+      logger.info(`Migrated build.gradle set ${dep} = ${neededDeps[dep]}.`);
+    }
+  }
+
+  // Replace jcenter()
```
jcesarmobile, line 343: the upgrade guide has this step as optional since users might still have some plugin that depends on some library that is only in jcenter, so maybe it should prompt so user is aware of the change?

cli/src/tasks/migrate.ts
```diff
@@ -0,0 +1,590 @@
...
+  );
+}
+
+function updateGradleWrapper(filename: string) {
+  const txt = readFile(filename);
+  if (!txt) {
+    return;
+  }
+  let replaced = txt;
+  if (replaced.includes('gradle-7.0-all.zip')) {
```
jcesarmobile, line 417: Since Android Studio prompts users to update gradle from time to time, it's possible that they are in a newer version than 7.0, so I think this should just replace whatever is on the url with 7.4.2

jcesarmobile (changes requested):
When running the migrate command it looks like the command hanged, we have ways of showing progress by showing a "spinning wheel", check update command as example when it says "Updating iOS native dependencies with pod install".

Not sure if would be helpful to have "migrate ios" and "migrate android" options too, but would definitely help maintaining the code by having separate files for iOS and Android changes.

If no ios or no android folders are present it will show a ton of errors, the command should check the folder existence first and if it's not there just warn that no migration will be done for that platform and not try to do the migration to that platform to avoid all those errors.

Some variables are out of date, I've commented on the lines, but since the template is shipped with the CLI, not sure if we could get latest versions from the variables.gradle file on the template instead of hardcoding them in code?
Or if it's hard to get from the template since it's zipped, maybe get it from https://github.com/ionic-team/capacitor/blob/main/android-template/variables.gradle

In Capacitor 3 we removed all the code that ran npm install to avoid problems for people using different package managers, but this new migrate command will run npm install.
I think it should at least prompt about it to users so they can skip the install step if desired so they can run their package manager equivalent command.

The changes on the .gitignore files are not helpful as the files already exist.

The migrate command is also doing some Capacitor 2 -> 3 changes, not sure if that was expected or if it was a mistake, if it's supposed to update from 2 -> it should do more things.

Related to that, while this change was from 2 to 3, it was not mandatory and now the init method is gone in capacitor 4, so that should be removed if present
https://capacitorjs.com/docs/updating/3-0#switch-to-automatic-android-plugin-loading

The migrator uninstall @capacitor/storage and install @capacitor/preferences, not sure if it should also search the user code and change the import and usage?

jcesarmobile (changes requested):
If I choose not to install the packages are not updated in package.json, can the version "text"  be updated to "next" so users have just to run the install command, i.e `npm install`. 
This works for npm, not sure about other package managers.

The gitignore changes are still being done but as the files exist it has no effect, better remove that code, it's not really a crucial change to edit the user's gitignore file and to properly work they will need to remove the cache and commit the removal of the files, so I think it's not worth the effort.

If the users chooses to not replace jcenter, it should still add mavenCentral in a few places where it doesn't exist.

There are a lot of `join('ios', 'XXX',`)`, where XXX are different files or folders, we have constants for most of them, even for the ios and android folders, in fact it can be configured to not be ios nor android, it should use those constants instead of the hardcoded values. (i.e `platformDirAbs`)

jcesarmobile (changes requested):
`platformDirAbs` was just an example of one of the values, but most of the `join` calls should be replaced with all the already existing folder/file paths in https://github.com/ionic-team/capacitor/blob/main/cli/src/config.ts so in case we change that in the future we don't have to go file by file updating the `join` calls to use the new paths

IT-MikeS:
Oh I see, my bad. Fixing

cli/src/tasks/migrate.ts
```diff
@@ -0,0 +1,734 @@
...
+            );
+          },
+        );
+
+        // Remove NSAppTransportSecurity
+        await runTask(
+          `Migrating info.plist by removing NSAppTransportSecurity key.`,
+          () => {
+            return removeKey(
+              join(config.ios.nativeTargetDirAbs, 'info.plist'),
```
jcesarmobile, line 160:
```suggestion
              join(config.ios.nativeTargetDirAbs, 'Info.plist'),
```

Info.plist is uppercase, this might cause problems in case sensitive OS

cli/src/tasks/migrate.ts
```diff
@@ -0,0 +1,734 @@
...
+              join(config.ios.nativeTargetDirAbs, 'AppDelegate.swift'),
+              `override func touchesBegan`,
+              `}`,
+            );
+          },
+        );
+
+        // Remove NSAppTransportSecurity
+        await runTask(
+          `Migrating info.plist by removing NSAppTransportSecurity key.`,
```
jcesarmobile, line 157:
```suggestion
          `Migrating Info.plist by removing NSAppTransportSecurity key.`,
```

jcesarmobile (changes requested):
In Capacitor 4 we have removed the deprecated init method, some Capacitor 3 apps might be still using it, so should be removed from `MainActivity.java`.
Should look something like this but probably with multiple `add(Plugin.class);` entries depending on the number of plugins installed.

```
    // Initializes the Bridge
    this.init(savedInstanceState, new ArrayList<Class<? extends Plugin>>() {{
      // Additional plugins you've installed go here
      // Ex: add(TotallyAwesomePlugin.class);
    }});
```

The migrator should also enable the Android 12 Splash Screen API (see https://github.com/ionic-team/capacitor-docs/pull/17)

cli/src/tasks/migrate.ts
```diff
@@ -0,0 +1,894 @@
...
+          return updateBuildGradle(
+            join(config.android.platformDirAbs, 'build.gradle'),
+            typeof leaveJCenterPrompt === 'string' &&
+              leaveJCenterPrompt.toLowerCase() === 'y',
+            variablesAndClasspaths,
+          );
+        });
+
+        // Update app.gradle
+        await runTask(`Migrating app.gradle file.`, () => {
```
jcesarmobile, line 221:
```suggestion
        await runTask(`Migrating app/build.gradle file.`, () => {
```

PR #5768, opened 2022-07-18
feat(android): Add android.minWebviewVersion configuration option

theproducer:
Adds a `android.minWebviewVersion` config option for allowing developers to set the minimum web view version for their app.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
@@ -2,6 +2,8 @@ package com.getcapacitor;
 
 import static com.getcapacitor.Bridge.CAPACITOR_HTTP_SCHEME;
 import static com.getcapacitor.FileUtils.readFile;
+import static com.getcapacitor.Bridge.DEFAULT_ANDROID_WEBVIEW_VERSION;
+import static com.getcapacitor.Bridge.MINIMUM_ANDROID_WEBVIEW_VERSION;
 
 import android.content.Context;
 import android.content.pm.ApplicationInfo;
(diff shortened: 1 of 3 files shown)


android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -436,6 +449,7 @@ public static class Builder {
         private Boolean webContentsDebuggingEnabled = null;
         private boolean loggingEnabled = true;
         private boolean initialFocus = false;
+        private int minWebViewVersion = 60;
```
jcesarmobile, line 452:
```suggestion
        private int minWebViewVersion = DEFAULT_ANDROID_WEBVIEW_VERSION;
```
use the `DEFAULT_ANDROID_WEBVIEW_VERSION` constant instead of hardcoding the value in two places

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -279,6 +284,14 @@ public boolean isInitialFocus() {
         return initialFocus;
     }
 
+    public int getMinWebViewVersion() {
+        if (minWebViewVersion < MINIMUM_ANDROID_WEBVIEW_VERSION) {
+            return MINIMUM_ANDROID_WEBVIEW_VERSION;
```
jcesarmobile, line 289: add a logger warning about the version being smaller than the allowed one and defaulting to MINIMUM_ANDROID_WEBVIEW_VERSION

jcesarmobile (changes requested):

jcesarmobile (approved):

lincolnthree:
What happens if the version requirement is not met? Forgive me if I didn't see the docs for this but I couldn't find them.

PR #5769, opened 2022-07-19
Added different background colors for light/dark mode

nkalupahana:
This adds the ability to specify different background colors for the webview depending on whether the device is in light or dark mode. The app I'm developing has different splash screens for light vs. dark mode, so without this separation, the screen will flash a different color in between when the splash screen disappears and when the app appears. Based on what I've seen online, I think a lot of people could benefit from this.

This is a breaking change, so I'm guessing y'all probably don't want it in 3.x. I just put it here because my app is in prod and I need this change, but I can't use 4.x until it's out. However, let me know where you'd like this change or what changes you'd like to see to it. I'm happy to change anything if y'all need me to do that.

These changes are fully tested on physical iOS and Android devices.


nkalupahana:
Superceded by #5840.

PR #5843, opened 2022-08-08
feat(cli): Option to inline JS source maps during sync

theproducer:
Adds an option to the `sync` command , `--inline`, which will, after syncing, inline all JS source maps for easier debugging on mobile devices.

### DIFF
--- a/cli/src/tasks/sourcemaps.ts
+++ b/cli/src/tasks/sourcemaps.ts
@@ -9,6 +9,22 @@ import { join, extname } from 'path';
 import type { Config } from '../definitions';
 import { logger } from '../log';
 
+
+function findJSAssetsDir(buildDir: string): string {
+    const reactJSDir = "/static/js";
+    const vueJSDir = "/js";    
+
+    if (existsSync(buildDir + reactJSDir)) {
+        return reactJSDir;
+    }
+
+    if (existsSync(buildDir + vueJSDir)) {
+        return vueJSDir;
+    }
+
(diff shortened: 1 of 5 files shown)


IT-MikeS (approved):

cli/src/tasks/sourcemaps.ts
```diff
@@ -0,0 +1,64 @@
...
+  readFileSync,
+  writeFileSync,
+  unlinkSync,
+} from '@ionic/utils-fs';
+import { join, extname } from 'path';
+
+import type { Config } from '../definitions';
+import { logger } from '../log';
+
+function findJSAssetsDir(buildDir: string): string {
```
dtarnawsky, line 13: JS files could be any folder. Im not sure you would want to hard code this for React and Vue.

cli/src/tasks/sourcemaps.ts
```diff
@@ -0,0 +1,64 @@
...
+  const reactJSDir = '/static/js';
+  const vueJSDir = '/js';
```
dtarnawsky, line 15: Path.join is required for folder separators which will change based on platform (eg on windows its \ rather than /)

dtarnawsky (commented):
Im a little concerned you'll miss JS files if you hardcode a subfolder in the output directory.

dtarnawsky (approved):
Looks good to me.

giralte-ionic (approved):

Issue #5793, opened 2022-07-26
bug: MIME Types ignored/blocked in strict mode


CrystalLake87:
### Platform(s)
iOS (14.0+)

### Current Behavior
XCode / iOS Simulator is not loading local stylesheets when running on brand new MacBook with M1 Pro - page loads but local CSS files do not load. Safari Inspector shows error messages related to the MIME types of the specified stylesheets, even when `type="text/css"` is specified in the tags. XCode is NOT running in Rosetta mode as suggested in previous bug reports.

Safari Error Message:
"Did not parse stylesheet at 'capacitor://{app_name}/{file_name}.css' because non CSS MIME types are not allowed in strict mode."

### Expected Behavior
Specified stylesheets should load without error and be applied properly.

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

CrystalLake87:
Issue can be reproduced using the standard capacitor test app at:
https://github.com/ionic-team/capacitor-testapp.git

Steps to reproduce:

From MacOS running on M1 or M1 Pro, clone repo:

- Clone repo from https://github.com/ionic-team/capacitor-testapp.git
- Install dependencies
- build package
- npx cap sync iOS
- npx cap open iOS
- run app in simulator (iOS 15.5, any simulated device)
- open safari web inspector
- reload page 
- inspect console output:

Console gives following output:

[Error] Did not parse stylesheet at 'capacitor://localhost/static/css/7.532f36fa.chunk.css' because non CSS MIME types are not allowed in strict mode.
[Error] Did not parse stylesheet at 'capacitor://localhost/static/css/main.fc34ace5.chunk.css' because non CSS MIME types are not allowed in strict mode.

CrystalLake87:
Inspecting the Network requests for these files shows they were served with MIME Type "application/octet-stream":


Summary
URL: capacitor://localhost/static/css/7.532f36fa.chunk.css
Status: 200
Source: —
Initiator: 
home:1

Request
Accept: text/css,*/*;q=0.1
[...]

PR #5853, opened 2022-08-12
fix(ios): return proper mimeType on M1 x86_64 simulators

arnold-plakolli:
In one of the projects I'm working on we are having an issue of app getting stuck on white screen in iOS simulators 14.0+ that run on M1 machines. 

After much trial and error we found out that in the `ios/Capacitor/Capacitor/WebViewAssetHandler.swift` file, the method `mimeTypeForExtension` was returning default `application/octet-stream` value for many path extensions. In real devices or simulators with lower iOS versions, it was returning correct value.

The issue was with the object `UTTypeCreatePreferredIdentifierForTag`. This object was returning `nil` value and as a result the method was defaulting to the value `application/octet-stream`.

This happens for projects:
- With checked `Open using Rosetta` option for Xcode
- With unchecked `Open using Rosetta` option for Xcode
  - Forcing architecture `x86_64` for simulator builds

`UTTypeCreatePreferredIdentifierForTag` is marked as deprecated and Xcode suggested using `UTType` instead. I did that but with same result as before.

I ended up providing a dictionary of path extensions and mime types. I added the dictionary as a fallback in case `UTTypeCreatePreferredIdentifierForTag` returns `nil`.

**Issue environment:**
* Machine:  M1 Mac (Apple Silicon)
* iOS Simulator: 14.0+

Sample projects:
* Referencing version `4.0.1` of `@capacitor/ios` that shows blank screen: https://github.com/arnold-plakolli/BlankScreenIssueDemo
* Referencing the fixed version in this PR of `@capacitor/ios` that works: https://github.com/arnold-plakolli/BlankScreenFixedIssueDemo

closes https://github.com/ionic-team/capacitor/issues/5793


jcesarmobile:
have you tried turning off `Open using Rosetta` (uncheck the checkbox) and restarting the computer? 
for me just unchecking the checkbox had no effect, I had to restart the computer.
Also needed to do a Product -> Clean Build Folder from Xcode before running

arnold-plakolli:
> have you tried turning off `Open using Rosetta` (uncheck the checkbox) and restarting the computer? for me just unchecking the checkbox had no effect, I had to restart the computer. Also needed to do a Product -> Clean Build Folder from Xcode before running

@jcesarmobile I did that for the blank Ionic project and it did work. The simulator loaded the app and it did not get stuck on the white screen. 

However when I ran the project, in which I initially had this bug on, it still showed the white screen. 

When I looked at the project build settings, I noticed that it was forcing architecture `x86_64` for simulators and that's when `UTTypeCreatePreferredIdentifierForTag` returns `nil`. So it seems this fix is still valid for those projects that want to use the architecture `x86_64` for simulators.

jcesarmobile:
You mean the project you shared or a different project?

where is that setting for using x86_64 on simulators?

arnold-plakolli:
@jcesarmobile 

The project I originally encountered the bug on is a different project not the one I shared.

I updated the PR description that points to two sample projects which both force arch `x86_64` for simulator builds. One of the projects points to version `4.0.1` of `@capacitor/ios` and the other one points to the fixed version in this PR.  You can run them both either with `Open using Rosetta` unchecked and you can see that the one that points to version `4.0.1` shows a blank screen. At least that's what I encountered on my tests.

jcesarmobile (approved):

giralte-ionic (approved):

PR #5861, opened 2022-08-16
feat: Capacitor Core Cookies

ItsChaceD:
This PR adds Cookies support via a Capacitor Plugin embedded into core. It also patches `document.cookie` to work synonymously with native apps.


ios/Capacitor/Capacitor/Plugins/CapacitorCookieManager.swift
```diff
@@ -0,0 +1,66 @@
...
+        let field = ["Set-Cookie": "\(key)=\(value)"]
+        let cookies = HTTPCookie.cookies(withResponseHeaderFields: field, for: url)
+        jar.setCookies(cookies, for: url, mainDocumentURL: url)
+    }
+
+    public func getCookies() -> String {
+        let jar = HTTPCookieStorage.shared
+        guard let url = self.getServerUrl() else { return "" }
+        guard let cookies = jar.cookies(for: url) else { return "" }
+        return cookies.map({"\($0.name):\($0.value)"}).joined(separator: ";")
```
epeach1, line 45: Should be name=value;
Not name:value;

PR #5862, opened 2022-08-16
feat(cli): environment config overwrites

IT-MikeS:
This PR adds the property `environmentOverrides` to `CapacitorConfig` which contains the name of the environment you want to use different config values in, and the config objects you wish to overwrite.

For example if you had a QA specific config you want to differ from your default config it would look like this:
```typescript

  appId: 'com.capacitorjs.app.testapp',
  appName: 'capacitor-testapp',
  webDir: 'build',
  ios: {
    scheme: 'App',
  },
  android: {
    flavor: 'dev',
  },
  plugins: {
    SplashScreen: {
      launchAutoHide: true,
    },
  },
  environmentOverrides: {
    'qa': {
      plugins: {
        SplashScreen: {
          launchAutoHide: false,
        },
      },
      ios: {
        scheme: 'AppQA',
      },
      android: {
        flavor: 'qa',
      },
    }
  }
};

export default config;
```

Where your android `flavor`, iOS `scheme`, and `SplashScreen` plugin configuration is changed if running in the `qa` environment.

You can set this environment by setting the `CAP_CONFIG_ENVIRONMENT` variable via `dotenv` or `export` ect. for example: `export CAP_CONFIG_ENVIROMENT=qa && npx cap sync`

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -52,6 +52,7 @@
     "debug": "^4.3.4",
     "env-paths": "^2.2.0",
     "kleur": "^4.1.4",
+    "lodash.merge": "^4.6.2",
     "native-run": "^1.6.0",
     "open": "^8.4.0",
     "plist": "^3.0.5",
@@ -65,6 +66,7 @@
   "devDependencies": {
     "@types/debug": "^4.1.7",
     "@types/jest": "^26.0.4",
+    "@types/lodash.merge": "^4.6.7",
     "@types/plist": "^3.0.2",
     "@types/prompts": "^2.0.14",
     "@types/rimraf": "^3.0.2",
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -33,12 +33,12 @@ export const CONFIG_FILE_NAME_TS = 'capacitor.config.ts';
 export const CONFIG_FILE_NAME_JS = 'capacitor.config.js';
 export const CONFIG_FILE_NAME_JSON = 'capacitor.config.json';
 
-export async function loadConfig(enviromentName = ''): Promise<Config> {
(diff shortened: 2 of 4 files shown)


giralte-ionic (approved):

jcesarmobile:
I don't see any real improvement in doing this, this is already possible with custom code in the .ts/.js files as the ticket indicates, this might be a little simpler, but still relies in scripting in the config file.

What the ticket was requesting is a parameter users or other tooling can pass to the run command, i.e `npx cap run android --flavor qa` on android or `npx cap run ios --scheme AppQA` on iOS that would override the flavor/scheme values in the config file (if any).

Please @dtarnawsky, correct me if I'm wrong.

dtarnawsky:
@jcesarmobile, @IT-MikeS,

Customers don't know how to write custom code to configure flavors/schemes as there aren't any written examples (and most customers don't read articles either), so the goal of the feature request is to make Flavors and schemes easy to specify via the CLI (first and foremost).

If you have a CLI argument then it's easier for the user to configure and easy for CI/CD as well as the VS Code extension. It also means you can ignore the contents of the capacitor config file which most customers dont change or care about.

Having the ability to override settings in the capacitor config based on CAP_CONFIG_ENVIRONMENT sounds like a nice feature to have but I don't have a solid use case for needing this as a feature.

IT-MikeS:
So this ticket is just about adding `--flavor` or `--scheme` to the `cap run [platform]` command?

dtarnawsky:
@IT-MikeS yes: something like --flavor or --scheme for the cap run command is the primary feature request.

IT-MikeS:
Going to close this PR but not delete the branch incase it becomes wanted as a feature later.

PR #5863, opened 2022-08-16
fix(cli): prevent error on migrate when devDependencies is missing

zookatron:
Currently the "cap migrate" command will crash with the error "Failed to migrate: TypeError: Cannot convert undefined or null to object" if you run it in project that has a package.json file without a devDependencies property. This PR fixes this issue.


jcesarmobile (approved):
good catch, thanks!

PR #5872, opened 2022-08-17
fix(android): Avoid handling responses in WebViewLocalServer (#4240, #5394)

### LINKED ISSUE #4240: bug: Android web view encapsulate 302 redirection response

### LINKED ISSUE #5394: bug: [Android] If response in WebView contains more than one Set-Cookie, only last is used
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 3.4.0
  @capacitor/core: 3.4.0
  @capacitor/android: 3.4.0
  @capacitor/ios: 3.4.0

Installed Dependencies:

  @capacitor/cli: 3.2.4
  @capacitor/android: 3.4.0
  @capacitor/core: 3.4.0
  @capacitor/ios: 3.3.3

```

### Platform(s)

Android

### Current Behavior

My application redirects from https://localhost out to a 3rd party authentication provider. This opens in a WebView. On that 3rd party site, the first GET request returns multiple (3) Set-Cookie headers in the response. Only the last cookie gets set in the WebView.

I have tracked the problem down to android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java lines 353-356:

```
                    String cookie = conn.getHeaderField("Set-Cookie");
                    if (cookie != null) {
                        CookieManager.getInstance().setCookie(url, cookie);
                    }
```
This code assumes that there is only one Set-Cookie header on the response. Since there might be multiple, it should be using conn.getHeaderFields().get("Set-Cookie") and iterating over each.

### Expected Behavior

All 3 cookies should be set in the WebView.

### Other Technical Details

`npm --version` output: 8.1.2

`node --version` output: v16.13.2

Zacherl:
Fixes #4240
Fixes #5394

The PR should fix the Android redirect issue.
More importantly, it fixes this issue without destroying the native bridge on redirects.

It actually also fixes this issue for me:
https://github.com/ionic-team/capacitor/issues/5394

It is inspired by this comment:
https://github.com/ionic-team/capacitor/pull/4437#issuecomment-1081300621

I do not know a lot about the codebase, so I am not sure if the workaround I proposed is suitable, but it works without issues for me; I used it for implementing single sign on with Keycloak.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -267,7 +267,7 @@ public class Bridge {
         final boolean html5mode = this.config.isHTML5Mode();
 
         // Start the local web server
-        localServer = new WebViewLocalServer(context, this, getJSInjector(), authorities, html5mode);
+        localServer = new WebViewLocalServer(context, this, authorities, html5mode);
         localServer.hostAssets(DEFAULT_WEB_ASSET_DIR);
 
         Logger.debug("Loading app at " + appUrl);
@@ -509,6 +509,11 @@ public class Bridge {
 
     public void reset() {
         savedCalls = new HashMap<>();
+
+        JSInjector jsInjector = getJSInjector();
+        if (jsInjector != null) {
+            webView.evaluateJavascript(jsInjector.getScriptString(), result -> {});
+        }
     }
 
     /**
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -17,19 +17,13 @@ package com.getcapacitor;
 
 import android.content.Context;
 import android.net.Uri;
-import android.util.Base64;
-import android.webkit.CookieManager;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.HttpURLConnection;
-import java.net.URL;
 import java.net.URLConnection;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.List;
 import java.util.Map;
 
 /**
@@ -56,7 +50,6 @@ public class WebViewLocalServer {
     private boolean isAsset;
     // Whether to route all requests to paths without extensions back to `index.html`
     private final boolean html5mode;
-    private final JSInjector jsInjector;
     private final Bridge bridge;
 
     /**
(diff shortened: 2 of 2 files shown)


Michriko:
Having the Same Issue with keycloak. 
Any updates to this PR? How can one support this fix?

Zacherl:
Hi @jcesarmobile, as you responded to the original issue, could you tell me what the next steps would be? I would love to dig deeper in this issue if the PR is not suitable, but I'm not sure how to proceed.

gkielwasser:
Hi @Zacherl Thank you for this PR !

It has fixed my redirection issue in Android.
Also I confirm it seems it doesn't breaks the native bridge as my different capacitor plugins are still working well after the redirect. This was not true when using the suggested patch in https://github.com/ionic-team/capacitor/issues/4240

PR #5873, opened 2022-08-17
feat(cli): add scheme and flavor options to run command

IT-MikeS:
This PR allows the `scheme` in iOS and `flavor` in Android to be set by their respective command options on the `run` command. For example:

`npx cap run ios --scheme AppQA`

`npx cap run android --flavor qa`


cli/src/tasks/run.ts
```diff
@@ -19,16 +21,33 @@ import { getPlatformTargets } from '../util/native-run';
...
+      scheme: string;
+    };
+  } = {};
+  if (options.flavor) {
+    toMerge.android = { flavor: options.flavor };
+  }
+  if (options.scheme) {
+    toMerge.ios = { scheme: options.scheme };
+  }
+  const config: Config = merge(cloneDeep(cfg), toMerge);
```
dtarnawsky, line 50: Seems like a little overkill to introduce lodash to change the config. Im cool with it but it does affect a users node_modules folder everytime a dependency is added

IT-MikeS: Config is read-only so this lets us take a deep clone of the object then merge deep safely without changing the properties of Config type, this also only includes the bare minimum as not all of Lodash is installed

giralte-ionic (approved):

cli/src/tasks/run.ts
jcesarmobile: You don't really need to clone and merge the config, what you can do is on the ios/android run.ts files read the scheme/flavor from the options and set a new scheme/flavor variable, something like
```
{ target: selectedTarget, flavor: selectedFlavor }: RunCommandOptions,
...
const runFlavor = selectedFlavor || config.android?.flavor || '';
...
 const arg = `assemble${runFlavor}Debug`;
```

jcesarmobile (approved):
looks good

dtarnawsky (approved):
Looks good - I assume there will be a PR to the Cap docs to document the new arguments?

PR #5891, opened 2022-08-30
feat(cli): add build command for android

IT-MikeS:
This PR adds the `build` command to the Capacitor CLI which will build the native project then create either a signed AAB or APK.

This command requires you to set up a Keystore prior to use. This is done via: opening Android Studio and choose Build -> Generate Signed Bundle / APK

After which you can use the command similar to the below example.

Example Usage:
`npx cap build android --keystorepath /path/to/my/keystore --keystorepass KEYSTORE_PASS --keystorealias KEY_ALIAS --keystorealiaspass KEY_PASS --androidreleasetype AAB`

Example TS Config File Usage (with environment vars):
```typescript
import * as dotenv from 'dotenv'

dotenv.config();

  appId: 'com.capacitorjs.app.testapp',
  appName: 'capacitor-testapp',
  webDir: 'build',
  plugins: {
    SplashScreen: {
      launchAutoHide: false,
    },
  },
  android: {
    buildOptions: {
      keystorePath: process.env.KEYSTORE_PATH,
      keystorePassword: process.env.KEYSTORE_PASSWORD,
      keystoreAlias: process.env.KEYSTORE_ALIAS,
      keystoreAliasPassword: process.env.KEYSTORE_ALIAS_PASSWORD,
      releaseType: 'AAB'
    }
  }
};

export default config;
```


phyr0s:
i think it would be better to have a config file like the build.json which is used in cordova and capacitor CLI just read that file by default or adding params about where is that file.
 
Example file

`"android": {
    "debug": {
      "packageType": "apk",
      "keystore": "my-release-key.keystore",
      "storePassword": "mypassword",
      "alias": "alias_name",
      "password": "mypassword",
      "keystoreType": "JKS"
    },
    "release": {
    "packageType": "apk",
      "keystore": "my-release-key.keystore",
      "storePassword": "mypassword",
      "alias": "alias_name",
      "password": "mypassword",
      "keystoreType": "JKS"
    }
  }`
  
  And command will be just like this
  
  ionic capacitor build android --release (default build.json in the proyect)
  ionic capacitor build android --configBuild="pathToConfig"


Also the same file could be valid for all platforms

giralte-ionic (approved):

dtarnawsky (approved):

theproducer (approved):
Tested the android config build options, works well.

phyr0s:
Tested APK, AAB, signed and unsigned. Works fine. Good job @IT-MikeS it will help a lot to create easier IC.

riderx:
I tried on my side ios build with no issue but android i cannot build:
CleanShot 2022-10-25 at 16 24 51@2x
@IT-MikeS you have idea why i got that ? on android studio no issue

IT-MikeS:
@riderx only time I've seen that is if your java version your machines default terminal has access to is an incompatible version with the requirements of Gradle or another lib you have in the app requires a different java version.

riderx:
@IT-MikeS thanks for this, i have check and seems good to me:
```
 …/capgo   main ↓1 ⍟1  java --version                                                                ✔  11362  12:40:24
openjdk 19 2022-09-20
OpenJDK Runtime Environment Temurin-19+36 (build 19+36)
OpenJDK 64-Bit Server VM Temurin-19+36 (build 19+36, mixed mode, sharing)
```

git4wht:
I had seted `"releaseType": "APK"` in `capacitor.config.json`, but it is not work when run 'npx cap build android'.
generated file is `app-release-signed.aab`
My Setting is:
```
  "android": {
    "buildOptions": {
      "keystorePath": "path to keystore",
      "keystorePassword": "password",
      "keystoreAlias": "alias",
      "keystoreAliasPassword": "password",
      "releaseType": "APK"
    }
  }
```

riderx:
my issue was my java version it was too early version and different from android studio

trinitiwowka:
> I had seted `"releaseType": "APK"` in `capacitor.config.json`, but it is not work when run 'npx cap build android'. generated file is `app-release-signed.aab` My Setting is:
> 
> ```
>   "android": {
>     "buildOptions": {
>       "keystorePath": "path to keystore",
>       "keystorePassword": "password",
>       "keystoreAlias": "alias",
>       "keystoreAliasPassword": "password",
>       "releaseType": "APK"
>     }
>   }
> ```

Hello, were you able to solve this problem? I have the same problem

PR #5901, opened 2022-09-07
feat(cli): add inline option to copy command

### LINKED ISSUE #5900: feat: Add --inline option to "npx cap copy" command to include JS source maps
## Feature Request
Add --inline option to "npx cap copy" command, as was added to "[npx cap sync --inline](https://capacitorjs.com/docs/cli/commands/sync)", in Capacitor 4.1.0

### Description

In capacitor 4.1.0 there is a new option [--inline] which enables syncing of all JS source maps, allowing for debugging an Android Web View in Chromium based browsers.
This option was added only to 'sync' command which syncs both native plugins & web app files.

During Ionic development, the 'npx cap copy' command is being used more often, when updating the web app files only, but this command doesn't support (returns error 'error: unknown option '--inline''). 

I suggest adding [--inline] option to copy command as well, since this is fast command that being used a lot for syncing web files and the --inline command is relevant to web files (JS source maps)

Thanks!

### Platform(s)
Android & iOS

### Preferred Solution
Adding [--inline] option to copy command as well: "npx cap copy --inline"

IT-MikeS:
This PR adds the inline option to the `copy` command while still allowing it to be passed from `sync` as well.

Closes #5900


giralte-ionic (approved):

theproducer (approved):
Looks good

cli/src/tasks/copy.ts
```diff
@@ -58,6 +61,7 @@ export async function copyCommand(
 export async function copy(
   config: Config,
   platformName: string,
+  inline: boolean,
```
jcesarmobile, line 64:
```suggestion
  inline: boolean = false,
```
If you set here the default value to false you don't need to edit multiple copy calls just to pass false.
Also, since we export the functions, there are possibilities of people using them in their tooling, so making the parameter mandatory can be considered a breaking change.

Same for copyCommand and the two sync equivalent methods were inline was introduced

jcesarmobile (approved):

Issue #5907, opened 2022-09-13
bug: wkwebview bouncing issue on iOS 16


rpanadero:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.2.0
  @capacitor/core: 4.2.0
  @capacitor/android: 4.2.0
  @capacitor/ios: 4.2.0

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/core: 4.2.0
  @capacitor/ios: 4.2.0
  @capacitor/cli: 4.2.0

[success] iOS looking great! 👌
```

### Platform(s)

- iOS

### Current Behavior

The webview bounces when the user slides down the view. I have checked the webview config and bouncing effect is disabled, but it seems it doesn't work.

Only reproduced on a physical iOS device with iOS 16.

### Expected Behavior

The webview shouldn't bounce when the user slides down the app view.

### Code Reproduction

https://github.com/rpanadero/demo-bouncing

Only reproduced on a physical iOS device with iOS 16.

### Other Technical Details

`npm --version` output: 8.19.1

`node --version` output: v18.9.0

`pod --version` output (iOS issues only): 1.11.3

### Additional Context

This bug was already reported as an iOS 16 issue on Apple forum one month ago, but it seems not to be fixed yet. Maybe, Apple will fix it in some days or months.

https://developer.apple.com/forums/thread/711601

Until Apple fixes the issue, we suggest to apply the workaround commented in this thread because the change is centralized and can be easily reverted when Apple fixes the real issue, if Capacitor team desires. The workaround consist on setting 'alwaysBounceVertical' webview property to false. Also, we have tested that the bug is still present on iOS 16 final release and the workaround still works.

We think that this change should be applied on Capacitor core code because this behaviour is undesired for all Capacitor iOS apps. In fact, webview bouncing is already disabled here:

https://github.com/ionic-team/capacitor/blob/76f28e70a5c0a03e4c6b9a93a0c068666a2c38ff/ios/Capacitor/Capacitor/CAPBridgeViewController.swift

https://github.com/ionic-team/capacitor/blob/76f28e70a5c0a03e4c6b9a93a0c068666a2c38ff/ios/Capacitor/Capacitor/CAPBridgeViewController.swift#L286

philmmoore:
+1 on this. Seeing the same thing in iOS16 on physical devices

mcfarljw:
[...]

PR #5909, opened 2022-09-13
fix(5907): webview bouncing issue on ios 16

rpanadero:
Fixes issue https://github.com/ionic-team/capacitor/issues/5907

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -284,6 +284,7 @@ extension CAPBridgeViewController {
         // create the web view and set its properties
         let aWebView = webView(with: .zero, configuration: webConfig)
         aWebView.scrollView.bounces = false
+        aWebView.scrollView.alwaysBounceVertical = false
         aWebView.scrollView.contentInsetAdjustmentBehavior = configuration.contentInsetAdjustmentBehavior
         aWebView.allowsLinkPreview = configuration.allowLinkPreviews
         aWebView.configuration.preferences.setValue(true, forKey: "allowFileAccessFromFileURLs")
--- a/ios/Capacitor/Capacitor/CAPWebView.swift
+++ b/ios/Capacitor/Capacitor/CAPWebView.swift
@@ -150,6 +150,7 @@ extension CAPWebView {
         loadInitialContext(webViewConfig.userContentController)
         let webView = WKWebView(frame: .zero, configuration: webViewConfig)
         webView.scrollView.bounces = false
+        webView.scrollView.alwaysBounceVertical = false
         webView.scrollView.contentInsetAdjustmentBehavior = configuration.contentInsetAdjustmentBehavior
         webView.allowsLinkPreview = configuration.allowLinkPreviews
         webView.configuration.preferences.setValue(true, forKey: "allowFileAccessFromFileURLs")


jcesarmobile:
Can you provide a video or gif of the behavior before and after the change?
I have tested the code and the WebView is still bouncing with the proposed changes.

thoasty-dev:
I can confirm the fix works. Would be great if you got this merged.

kemalcany:
Tested on the simulator and the bouncing seems to be fixed (haven't tested on a physical device tho)

giralte-ionic:
Could we get some videos or gifs of the problem behavior? We are thinking this may be a 16.0 issue and could be fixed in 16.1, so I'd also recommend testing with the current iOS Beta

fabltd:
How do I install this branch?

Sharcoux:
@rpanadero Do you use your fork internally? How can we use it?

rpanadero:
> @rpanadero Do you use your fork internally? How can we use it?

@Sharcoux  no, I don't. As a temporal fix until this PR is merged/reviewed, I developed an ad-hoc plugin for my project that changes the Capacitor webview configuration. Just add these two files to your iOS project:

WebViewPlugin.m
`
#import <Capacitor/Capacitor.h>

CAP_PLUGIN(WebViewPlugin, "WebViewPlugin",)
`

`
import Capacitor

@objc(WebViewPlugin)
public class WebViewPlugin: CAPPlugin {
    @objc override public func load() {
        // Called when the plugin is first constructed in the bridge
        self.bridge?.webView?.scrollView.bounces = false
        self.bridge?.webView?.scrollView.alwaysBounceVertical = false
        self.bridge?.webView?.scrollView.alwaysBounceHorizontal = false
    }
}
`

After adding/creating them to your iOS project, be sure that the files are included in the bundle resources.

IJustDev:
@rpanadero thank you very much this works for me!

If anybody reading this and don't know how to put this in practice:
- open Xcode (VSCode won't track them in the build process)
- create both files respectively on the same level as AppDelegate

IJustDev (commented):
LGTM!

willgardner9:
Thanks for the instructions @rpanadero and @IJustDev! I can't seem to get web view bounce to stop. I have created both files in the same dir as the AppDelegate.swift - I'm sure this is my fault for missing something

Screenshot 2022-11-28 at 15 17 49

Do I need to create a bridging file or anything? I can see the Xcode project has changed and the files are present in my `./ios/App/App.xcodeproj/project.pbxproj` file

Appreciate all your efforts on this, thanks so much 🙌

IJustDev:
Do these files appear in XCode?

bsalesc:
The above code didn't work, so creating a object c file (NoBounce) with below content worked:

```obj-c
#import <Foundation/Foundation.h>
#import <UIKit/UIKit.h>

@implementation UIScrollView (NoBounce)
- (void)didMoveToWindow {
    [super didMoveToWindow];
    self.bounces = NO;
}
@end
```

piotrfonte:
Can’t get any of the above snippets to work to solve the problem. How can check if they get included in the bundle?

thoasty-dev:
This is not an issue anymore on iOS 16.2

piotrfonte:
Gotcha. Can confirm that this is fixed in iOS 16.2.

BenRacicot:
I couldn't get any of the native code to work but the actual [capacitor.config.ios ](https://capacitorjs.com/docs/config)values fixed this for me.

```
"ios": {
    "contentInset": "always", <-- the cause
}
```

BenjaminDish:
> This is not an issue anymore on iOS 16.2

@thoasty-dev You say it is fixed on ios 16.2 but i can't understand what is the expected behaviour.

Is bouncing automatically disabled in capacitor apps by default ? (i don't think so as bouncing is ON on my ios 14 iphone)

Or, does it need any configuration/code to disable it ? If yes, which one ?

brilliantinsane:
Will this ever be merged?

eric-horodyski:

More context:

PR #5925, opened 2022-09-15
feat(cli): add build command for ios

IT-MikeS:
This PR adds the build command to the Capacitor CLI which will build the native project then create an IPA.

Example usage:
`npx cap build ios` additionally a custom scheme can be used either via the config file or by passing the `--scheme MyScheme` flag to the build command.


phyr0s:
I think that adding to capacitor.config file the option like the android PR that you open before will be very good to to remove complexity in the build process.

There is few options that probably help to build IPA easier --> 
The development team 
Package type: can be development, ad-hoc, app-store (in appstore accounts) or enterprise build for enterprise accounts

Provisioning profile: this field i think that can be optional or not necessary because now xcode manage autosign so if you 
have the provisioning profile in the provisioning profile folder should pick auto.... but maybe is good to have if you want to select different provisioning profile serving from jenkins before build for example.

Custom build flags array for example --> -UseModernBuildSystem=YES or -xcconfig=0

@IT-MikeS  thanks for your good job.

dtarnawsky (approved):

giralte-ionic (approved):

PR #5928, opened 2022-09-17
Fix update command when using custom iOS project name

nathanstitt:
The steps for renaming the iOS app https://capacitorjs.com/docs/ios/configuration#renaming-your-app cause the `cap sync ios`  to fail with:
```
✖ update ios - failed!
[error] Command line invocation:
        /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -project App.xcodeproj clean

        User defaults from command line:
        IDEPackageSupportUseBuiltinSCM = YES

        2022-09-16 23:55:10.098 xcodebuild[44343:3003925] Writing error result bundle to
        /var/folders/7s/9cx4q_5d1sq_8x9nl557s1fm0000gn/T/ResultBundle_2022-16-09_23-55-0010.xcresult
        xcodebuild: error: 'App.xcodeproj' does not exist.
```

Looks like this spot still has the hardcoded `App` project name

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -277,7 +277,7 @@ async function loadIOSConfig(
   const nativeProjectDirAbs = resolve(platformDirAbs, nativeProjectDir);
   const nativeTargetDir = `${nativeProjectDir}/App`;
   const nativeTargetDirAbs = resolve(platformDirAbs, nativeTargetDir);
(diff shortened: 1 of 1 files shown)


jcesarmobile:
The scheme name and the project name are different things, it’s allowed to have multiple schemes but only one project.
Looks like you did something wrong when you renamed the scheme, it should not change the project name.

PR #5940, opened 2022-09-22
add - definitions.ts : PlatformType

rbalet:
Add type to the `getPlatform()` method


jcesarmobile:
Since it's possible to add custom platforms, this would prevent those platforms from getting a correct name, in example if you add @capacitor-community/electron, it will return electron and with this changes it won't be a valid value, so I think we will keep the string

rbalet:
@jcesarmobile  Ok, I see, what about adding `| string` at the end? would make it being even more clearer. 
Don't you think?

Issue #5947, opened 2022-09-22
bug: ion icons disappear when Http enabled


jongbonga:
## Bug Report

Ionic icons not showing when `CapacitorHttp: {  enabled: true }` in `capacitor.config.ts` on native 

### Capacitor Version
```

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.3.0
  @capacitor/core: 4.3.0
  @capacitor/android: 4.3.0
  @capacitor/ios: 4.3.0

Installed Dependencies:

  @capacitor/cli: 4.3.0
  @capacitor/core: 4.3.0
  @capacitor/android: 4.3.0
  @capacitor/ios: 4.3.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

iOS

### Code Reproduction

```
import { CapacitorConfig } from '@capacitor/cli';

const config: CapacitorConfig = {
  ...
  plugins: {
    CapacitorHttp: {
      enabled: true
    }
  }
};

export default config;
```

nicklasuhde:
I have the same issue!

rensjaspers:
Here's a minimal sample application that demonstrates the issue:
https://github.com/rensjaspers/capacitor-http-missing-ion-icons-sample

I get "unsupported URL" errors like this one: 
```
⚡️  To Native ->  App addListener 124667420
2022-09-23 11:31:22.573110+0200 App[22553:2273293] Task <14803BC4-3F93-4D63-9370-506DC2DA7DAD>.<1> finished with error [-1002] Error Domain=NSURLErrorDomain Code=-1002 "unsupported URL" UserInfo={NSLocalizedDescription=unsupported URL, NSErrorFailingURLStringKey=/svg/accessibility.svg, NSErrorFailingURLKey=/svg/accessibility.svg, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataTask <14803BC4-3F93-4D63-9370-506DC2DA7DAD>.<1>"
), _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <14803BC4-3F93-4D63-9370-506DC2DA7DAD>.<1>, NSUnderlyingError=0x600000341fe0 {Error Domain=kCFErrorDomainCFNetwork Code=-1002 "(null)"}}
```

isabo:
FWIW the Chrome remote debugging console shows a `MalformedURLException`:

[...]

PR #5953, opened 2022-09-23
fix(http): fix local http requests on native platforms

ItsChaceD:
Addresses: https://github.com/ionic-team/capacitor/issues/5947

This PR fixes an issue with local requests not being handled properly on native platforms (iOS/Android). As a result, fetching `IonIcons` or local assets could crash the app.


anttironty:
This PR also fixes problems with firebase js sdk. Connections with firestore fail without the changes introduced here

edy-ap:
It works with relative urls?
Ex: "i18n/es.json" should be "http://localhost/i18n/es.json" on android and "capacitor://localhost/i18n/es.json" on iOS

Issue #2978, opened 2020-05-25
bug: File assets that are downloaded making use of HTTP Range requests fail to download (sometimes)


llorenspujol:
# Bug Report
File assets that are downloaded making use of [HTTP Range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) fail to download (sometimes).

## Capacitor Version

`npx cap doctor` output:

Installed Dependencies:

  @capacitor/ios not installed

  @capacitor/electron not installed

  @capacitor/cli 2.1.0

  @capacitor/android 2.1.0

  @capacitor/core 2.1.0

## Affected Platform(s)

- [X] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

## Current Behavior
File assets that are downloaded making use of [HTTP Range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) fail to download (sometimes).

In my case this assets are sound assets in .wav format. The issue is that this audios, are sometimes just partially loaded, or even not loaded at all... This bug is only reproducible when running the app with capacitor.

Here are some captures of the errors:

First request:
![image](https://user-images.githubusercontent.com/18361241/82796155-602ac780-9e75-11ea-90e4-fd8709d84076.png)

Second request (fails:
![image](https://user-images.githubusercontent.com/18361241/82796101-45585300-9e75-11ea-9b58-182e65624e4b.png)

## Expected Behavior

I expect that files that are downloaded using HTTP Range Request are fully and correctly downloded.

## Reproduction Steps
Just put some .wav files on the assets folder, and try to reproduce that audio in a capacitor android app. You will see that the downloading sometimes fails.

## Last words
I can't dig any more on the issue for now since I am a new capacitor user:), but if necessary and with some hints or guidance I could dig more into it and help solve this issue.
Thanks!

llorenspujol:
This issue could be related: https://github.com/ionic-team/capacitor/issues/1343

albertgasset:
I can reproduce this bug too when playing several audio files concurrently. In the Chrome inspector I can see that response headers are mixed up among requests, so one request gets the response range values of another request.

Looking at the code of WebViewLocalServer I suspect of the PathHandler instances: they contain the response headers and are shared among requests, leading to a data race.

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

[...]

PR #5956, opened 2022-09-26
Android range requests

steinjak:
This PR addresses issue #2978, which seems to be a problem carried over from the cordova plugin: ionic-team/cordova-plugin-ionic-webview/issues/453

The issue is with [`PathHandler`](android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java), where the `responseHeaders` map is shared between threads, causing simultaneous requests to return the same `Content-Length` to the browser in case of range requests. See the illustration in ionic-team/cordova-plugin-ionic-webview/issues/453 (the same issue is present Capacitor).

This PR proposes to return a copy of the map on each invocation.

Furhermore, it introduces skipping of the initial part of the file in case of a range request not starting from zero. This brings the Android implementation better in line with its iOS counterpart, see  https://github.com/ionic-team/capacitor/blob/3d4433b505252d263784ec0fafc72a9b7cd8591e/ios/Capacitor/Capacitor/WebViewAssetHandler.swift#L51

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -124,7 +124,7 @@ public class WebViewLocalServer {
         }
 
         public Map<String, String> getResponseHeaders() {
-            return responseHeaders;
+            return new HashMap(responseHeaders);
         }
     }
 
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -213,6 +213,10 @@ public class WebViewLocalServer {
                 String[] parts = rangeString.split("=");
                 String[] streamParts = parts[1].split("-");
                 String fromRange = streamParts[0];
+                int bytesToSkip = Integer.parseInt(fromRange);
+                if (bytesToSkip > 0) {
+                    responseStream.skip(bytesToSkip);
+                }
                 int range = totalRange - 1;
                 if (streamParts.length > 1) {
                     range = Integer.parseInt(streamParts[1]);


android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -213,6 +213,10 @@ private WebResourceResponse handleLocalRequest(WebResourceRequest request, PathH
                 String[] parts = rangeString.split("=");
```
peitschie, line 213: Am just driving by while hunting out my own problem... I wonder if a better way to do this is to make the copy only when it's about to be modified here.

I.e., instead of creating a copy every access, do it when the `tempResponseHeaders` variable is assigned at line 208:
```
Map<String, String> tempResponseHeaders = new HashMap(handler.getResponseHeaders());
```

Other than that though, this code definitely makes good sense to me.

steinjak: I see what you mean (a getter should not create new objects), but this will re-introduce the problem, as there are other invocations of this method as well - all passing the returned instance into the constructor of `WebResourceResponse` (which is where the value of `tempResponseHeaders` also ends up). Note that it's not the headers set on lines 225/226 that cause trouble here, but the `Content-Length` header, set somewhere else downstream from this code.

We could of course move the cloning to all call points of the getter, but that leaves a trap for the next person coming around and wanting to use this method, probably not knowing about this issue, so we should probably not do it without making the map read-only (through [Collections.unmodifiableMap(...)](https://docs.oracle.com/javase/8/docs/api/java/util/Collections.html#unmodifiableMap-java.util.Map-), ie. `this.responseHeaders = Collections.unmodifiableMap(new LinkedHashMap<String, String>(tempResponseHeaders))` on line 102). The downside to this approach, though, is that it triggers runtime exceptions - and I can't find the automated tests that will trigger these when developing, leaving it to the developer to thoroughly test all cases before submitting a patch...

Maybe it would be better to rename `WebViewLocalServer.PathHandler.getResponseHeaders()` to `WebViewLocalServer.PathHandler.buildDefaultResponseHeaders()` to clarify? I don't think I can see a case where the set of response headers need to be shared between responses, only cases where there is a base set of headers to be included in every response.

What do you think, @peitschie ?

visuallization:
I am currently experiencing a similar issue: https://github.com/ionic-team/capacitor/issues/6021#issuecomment-1470121335

Are there any updates on this PR?

JaHollTV:
are there updates?

visuallization:
> @visuallization I think the skipping part is unrelated here, as it's the `Content-Length` header that ends up with the wrong value due to a race condition. If I remember correctly, capacitor tries to serve the whole file at once, so I suspect the skipping part is mostly redundant (a long time since I looked at the code).
> 
> Sadly, there doesn't seem to be much interest in this PR.
> 
> For the next person facing this: We circumvent the issue by downloading the media file without streaming (using Angular `HTTP` service, `fetch()` would probably also work) and map it to a resource URL (`URL.createObjectURL(blob)`) that we point our media elements to.

@steinjak I patched capacitor in our project with your fix and it turns out the skipping part is actually an issue as some audio will just stop un mid sentence. So I removed the skipping part and now everything works as intended! Thanks for the fix, it helped a lot!

biguphpc:
Having the same issue with Range requests. This bug is easy to reproduce by puttin multiple `<audio>` tags on a page. The concurrent queries create the threading issues and the range headers response corruption. Right now i'm, having to patch [the Java file during the build](https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/632#issuecomment-1810870746) to disable the Range requests completely which is a terrible solution.

I've tested the PR and it solves our issue. It's also a lot better than the current status quo. Let me know if I can help do something to have the PR merged. Thanks

WoodyWoodsta:
How can we help with this PR? This seems like a fairly critical issue when loading anything larger and in parallel (a reasonable requirement) from the filesystem in Android.

cfpwastaken:
I am in need of this fix. How can I patch it in to my app for the time being?

biguphpc:
@cfpwastaken See[ this for patching the file during build]( https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/632#issuecomment-1810870746)

bolivir:
Patching out range requests isn’t a viable solution for our use case. In our application, streaming local videos is a core feature, not just a side functionality. This long-standing issue (open for more than three years now) is severely impacting our ability to deliver a reliable experience to our users.

We’re going to try implementing a fix ourselves, but this really feels like something that should be properly addressed in the framework. At this point, we’re seriously considering moving away from Ionic if there’s no clear plan or timeline for resolving this problem. Proper support for range requests is essential for making local video playback work as intended.

bolivir:
The proposed solution appears to be working well so far — we’re running extensive tests, and none of the videos have bricked yet.
I’ve also opened a separate PR (https://github.com/ionic-team/capacitor/pull/8190) to return a new HashMap instance for responseHeaders. Hopefully this can be merged soon.
In the meantime, we’ve implemented a temporary Nx hook to modify this line of code during the build process.

eric-horodyski:
Thanks for opening this PR!

OS-pedrogustavobilro:
Hey everyone, an up-to-date version of this PR was opened at https://github.com/ionic-team/capacitor/pull/8357, and we merged it. So I'll be closing this one.

Apologies it took us a long time to look at this 🙇 

Feel free to ask any questions if you have them

PR #5973, opened 2022-10-05
fix (http) : use setValue instead of addValue on iOS setRequestHeaders

bastyen:
As explain [here](https://github.com/capacitor-community/http/issues/261), there is a problem with accept-language header on iOS because of [this](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift#L21).


jcesarmobile:
should be covered by https://github.com/ionic-team/capacitor/pull/6303, which was released in 4.7.0 release

if not, please, create an issue and provide a sample app where it can be reproduced

Issue #5971, opened 2022-10-04
bug: CapacitorHttp JSONException when no content (204)


JumBay:
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 4.3.0
  @capacitor/core: 4.3.0
  @capacitor/android: 4.3.0
  @capacitor/ios: 4.3.0

Installed Dependencies:

  @capacitor/cli: 4.3.0
  @capacitor/ios: 4.3.0
  @capacitor/android: 4.3.0
  @capacitor/core: 4.3.0

[success] iOS looking great! 👌
[error] Missing <manifest package=""> attribute in app/src/main
```

### Platform(s)

- Android
- Web

### Current Behavior

On Android/Web, if the request returns an empty content with a JSON content type, which happens on `204` response, it raises a JSON Exception cause it fails to parse `""`  here https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java#L297

### Expected Behavior

Shouldn't raise an exception

### Code Reproduction

```ts
 try {
      const response = await CapacitorHttp.request({
        method: 'POST',
        url: 'https://httpstat.us/204',
        data: {
          value: 'test',
        },
        responseType: 'json',
      });
      console.log('CapacitorHttp Response', response);
    } catch (e) {
      console.log('CapacitorHttp Failure', e);
    }
```

### Other Technical Details

Adding this 
```java
 else if ("".equals(input.trim())) {
                return "";
            }
```
https://github.com/ionic-team/capacitor/blob/main/android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java#L306  fixes it

nicklasuhde:
I have the same issue

PR #5974, opened 2022-10-06
fix: #5971 empty body

riderx:
This fix the empty body issue.
#5971


ItsChaceD:
This is failing some checks so I created a new PR: https://github.com/ionic-team/capacitor/pull/6011

riderx:
thanks a lot!
i would have appreciate the explaining, before creating you own PR.
that would have let me a chance to fix it.
More we learn and practice to help more that will get easier for all :)

PR #6011, opened 2022-10-20
fix(http): fix exception thrown on 204 responses

ItsChaceD:
Addresses: https://github.com/ionic-team/capacitor/issues/5971

This PR fixes the exception thrown on Android when an empty string is returned with `Content-Type: "application/json"` (204 response).


jcesarmobile (changes requested):
native-bridge.ts has not changed on this commit, so the native-bridge.js files shouldn't be changed neither. Can you remove them from the pull request?

jcesarmobile (approved):

PR #6022, opened 2022-10-24
chore(android): native artifact publish logic

carlpoole:
Adds native publication step for android into Capacitor v4. No separate publication branch required like in v3.

Triggering a build requires ENV variable set for `CAP_PUBLISH=true` when running the build and publish Gradle commands so that the appropriate plugins are imported with Gradle. The CI publish script will include this.

ENV variable requirement fixes a previous error in traditional Capacitor apps where the parent App project threw a Gradle sync error due to the presence of the publishing plugin in the entire project.
```A problem occurred evaluating project ':capacitor-android'.
> Failed to apply plugin 'io.github.gradle-nexus.publish-plugin'.
   > Plugin must be applied to the root project but was applied to :capacitor-android
```

Removed the manual review step from Sonatype. **Whenever a fresh build is run with this publish script, as long as it is not already published on MavenCentral, Capacitor Core will be fully published** with these changes. Test version was successful. https://central.sonatype.dev/artifact/com.capacitorjs/core/4.4.0-beta1

### DIFF
--- a/.github/workflows/publish-android.yml
+++ b/.github/workflows/publish-android.yml
@@ -9,8 +9,6 @@ jobs:
       packages: write
     steps:
       - uses: actions/checkout@v2
-        with:
-          ref: native-publish
       - name: set up JDK 11
         uses: actions/setup-java@v2
         with:
--- a/android/capacitor/build.gradle
+++ b/android/capacitor/build.gradle
@@ -16,15 +16,28 @@ buildscript {
     repositories {
         google()
         mavenCentral()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:7.2.1'
+
+        if (System.getenv("CAP_PUBLISH") == "true") {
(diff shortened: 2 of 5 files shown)


Steven0351 (approved):

jcesarmobile (approved):
Tested on a capacitor android app and doesn't affect running or building a signed apk, so looks good to me.


Doesn't changing the publish-android.yml to not use native-publish will remove the possibility of publishing capacitor 3 versions? or you don't plan to publish more v3 versions? or you plan to create a new publish v3 workflow if needed?

carlpoole:
> Doesn't changing the publish-android.yml to not use native-publish will remove the possibility of publishing capacitor 3 versions? or you don't plan to publish more v3 versions? or you plan to create a new publish v3 workflow if needed?

Yeah the plan going forward is to migrate Portals to Cap v4 from now on. If we need to do any v3 builds we should still be able to do that now that the branch isn't hardcoded in the CI yaml I think we can manually run it from any branch.

PR #6047, opened 2022-11-02
chore(ios): Use ruby features instead of patch files for publishing

Steven0351:
This provides a cleaner way to dynamically change the paths than using the patchfile method.


Steven0351:
@jcesarmobile I initially tried to just use the same method as capacitor-plugins uses, but Capacitor and CapacitorCordova set a module_map and Capacitor has the resource file, so I needed to come up with something else for the paths.

ios/CapacitorCordova.podspec
```diff
@@ -1,5 +1,11 @@
 require 'json'
 package = JSON.parse(File.read(File.join(__dir__, 'package.json')))
+prefix = if ENV['NATIVE_PUBLISH']
```
jcesarmobile, line 3:
```suggestion
prefix = if ENV['NATIVE_PUBLISH'] == "true"
```
Better check the actual value and not just if it's defined.
Same for the `Capacitor.podspec` file.

And maybe set it to false after publication or call `unset NATIVE_PUBLISH`

Since it's going to be run on the GH actions, the variable will probably not persist executions, but just in case.

Steven0351: Also, just to assuage any unease, the only time the podspec is ever evaluated as part of our CI processes is when we're linting and publishing for cocoapods itself. Even if `NATIVE_PUBLISH` is present when publishing to NPM, the podpsec file is not evaluated until a user performs a `pod install` (or in most capacitor users cases, `cap sync`).

jcesarmobile (approved):
ah, right, I was testing from the capacitor-testapp running sync like this
`export NATIVE_PUBLISH=true && npx cap sync ios` and it doesn't work (expected) and then `export NATIVE_PUBLISH=false && npx cap sync ios` or just `npx cap sync ios` and they didn't work neither (not expected).

With the latest changes it works as expected.

You were right, when the `export NATIVE_PUBLISH=true` is done from the .sh file it's not set outside the script execution, so no need to unset it or set it to false.

PR #6060, opened 2022-11-07
fix(cli): warns about config that is present

IT-MikeS:
This PR fixes a behaviour where the CLI will warn about missing configuration in the Plist and AndroidManifest files even when the required config of the plugins is already present.

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -60,7 +60,6 @@
     "semver": "^7.3.7",
     "tar": "^6.1.11",
     "tslib": "^2.4.0",
-    "xml-js": "1.6.11",
     "xml2js": "^0.4.23"
   },
   "devDependencies": {
--- a/cli/package.json
+++ b/cli/package.json
@@ -60,6 +60,7 @@
     "semver": "^7.3.7",
     "tar": "^6.1.11",
     "tslib": "^2.4.0",
+    "xml-js": "1.6.11",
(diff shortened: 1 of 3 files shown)


cli/package.json
```diff
@@ -60,6 +60,7 @@
     "semver": "^7.3.7",
     "tar": "^6.1.11",
     "tslib": "^2.4.0",
+    "xml-js": "1.6.11",
     "xml2js": "^0.4.23"
```
dtarnawsky, line 64: Looks like there are 2 libraries now `xml-js` and `xml2js`. Possible to refactor down to 1?

cli/package.json
```diff
@@ -60,6 +60,7 @@
     "semver": "^7.3.7",
     "tar": "^6.1.11",
     "tslib": "^2.4.0",
+    "xml-js": "1.6.11",
```
jcesarmobile, line 63: Isn't there a more up to date library that can accomplish the same?
We already have `xml2js` which has not been updated in 3 years, but `xml-js` has not been updated in 4 years, so having more unmaintained dependencies will probably bring us problems in the future.

IT-MikeS:
Now only uses libs that were already there.

dtarnawsky (approved):

giralte-ionic (approved):

PR #6072, opened 2022-11-11
feat(ios): Plugin Registration and Plugin Instance Support

Steven0351:
feat(ios): Adds support for manual registration for plugins and the ability to register instances of plugins so that plugins can be instantiated by the user instead of relying on the bridge.

### DIFF
--- a/ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
+++ b/ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
@@ -89,7 +89,7 @@
 		A38C3D7B2848BE6F004B3680 /* CapacitorCookieManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = A38C3D7A2848BE6F004B3680 /* CapacitorCookieManager.swift */; };
 		A71289E627F380A500DADDF3 /* Router.swift in Sources */ = {isa = PBXBuildFile; fileRef = A71289E527F380A500DADDF3 /* Router.swift */; };
 		A71289EB27F380FD00DADDF3 /* RouterTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A71289EA27F380FD00DADDF3 /* RouterTests.swift */; };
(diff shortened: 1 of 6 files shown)


Steven0351:
This is an effort to solve problems we see in Portals as we now have customers using Capacitor with new use cases:

1. Being able to isolate what plugins are available to the Capacitor runtime. As it exists today, all plugins visible to Objective-C are auto-registered, which may not be desirable if multiple Capacitor instances are running in an application. This provides users the ability to explicitly decide what native functionality a Capacitor application has (similar to Android today).
2. Being able to control when and how a plugin is initialized. Capacitor plugins are now being used in cases where they are participants in the native application. This is in contrast to the more traditional role they took in mostly providing access to native API's through the web. Plugins that are purpose built for cohabiting in a native application should have the ability to participate in the programming paradigms of the native application.

All of these features are additive and do not affect status quo Capacitor users and is explicitly opt-in.

giralte-ionic (approved):

IT-MikeS (approved):

co-dax:
Hey @Steven0351, does this change (and related android change) allow to load plugins dynamically at runtime as/when/if needed? I am aksing this we are having several plugins in our app and this is causing slow application start so deferring the plugin  loading would be great. Also, some of the users would never need some of the plugins so by loading them if needed would reduce app start time.

Thanks!

Steven0351:
> Hey @Steven0351, does this change (and related android change) allow to load plugins dynamically at runtime as/when/if needed? I am aksing this we are having several plugins in our app and this is causing slow application start so deferring the plugin loading would be great. Also, some of the users would never need some of the plugins so by loading them if needed would reduce app start time.
> 
> Thanks!

@co-dax, I can't say for sure since that wasn't the use case I created this for and I didn't test for it. However, you might give it a try and see what happens. I think the question on whether or not it would work depends on how eagerly the parts of your app that depend on those plugins are loaded. With code-splitting it may not be much of an issue, but I'm sure that can change on a case-by-case basis.

co-dax:
thanks @Steven0351 !
I will give it a try.

PR #6073, opened 2022-11-11
feat(android): Plugin Instance Support

Steven0351:
feat(android): Adds support for registering already initialized instances of plugins to the bridge.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -120,6 +120,8 @@ public class Bridge {
 
     private final List<Class<? extends Plugin>> initialPlugins;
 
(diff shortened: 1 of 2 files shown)


Steven0351:
This is an effort to solve problems we see in Portals as we now have customers using Capacitor with new use cases where a Plugin is expected to participate in the application and allows for more flexibility in how plugins are used and authored. This is in contrast to the more traditional role they take in providing access to native API's through a web interface. Plugins that are purpose built for cohabiting in a native application should have the ability to participate in the programming paradigms of the native application.

The changes introduced in the PR are additive and do not affect status quo Capacitor users and is explicitly opt-in.

giralte-ionic (approved):

IT-MikeS (approved):

PR #6157, opened 2022-12-12
Fix error can't build android 12

N-H-Quang:


jcesarmobile:
Thanks for the PR, but we only backport security related changes to Capacitor 2.

Issue #5996, opened 2022-10-14
bug: Firebase getIdToken(true) for refresh token - Refresh token for persistence of user on app is not returning token but error - jsonexception


shreyasPL:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.3.0
  @capacitor/core: 4.3.0
  @capacitor/android: 4.3.0
  @capacitor/ios: 4.3.0

Installed Dependencies:

  @capacitor/core: 4.3.0
  @capacitor/ios: 4.3.0
  @capacitor/cli: 4.3.0
  @capacitor/android: 4.3.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```
Capacitor.config.js
```
{
  "appId": "our-app.com",
  "appName": "appname",
  "bundledWebRuntime": false,
  "npmClient": "yarn",
  "webDir": "build",
  "plugins": {
    "SplashScreen": {
      "launchShowDuration": 0
    },
    "FirebaseAuthentication": {
      "skipNativeAuth": false,
      "providers": [
        "google.com",
        "facebook.com",
        "apple.com"
      ]
    },
    "CapacitorHttp": {
      "enabled": true
    }
  },
  "server": {
    "androidScheme": "https",
    "iosScheme": "ionic", 
    "cleartext": true,
    "allowNavigation": [
      "*"
    ]
  }
}
```

### Platform(s)

1. capacitor android only - We are using capacitor.https -> enabled true flag as part of new release 4.3.0 of capcitor.

### Current Behavior

1. We know firebase access token expires after an hour.
2. We use getIdToken(true) - forcerefresh param to true, for forceful token with extended one hour token.
3. This function is returning jsonexception error when being called.
4. FirebaseError/auth-network-failed is returned with no other information.
NOTE: getIdToken() without param after logging in, returns token perfectly. But with parameter as true it fails all the time.

### Expected Behavior

1. getIdToken(true) should return new token, not the error.

### Code Reproduction

```
const app = getApp();
const auth = getAuth(app);
const updatedUser = auth.currentUser;
updatedUser?.getIdToken(true).then(token => {
     console.log(token, "Forceful"); // doesn't return this
}).catch(err => {
     console.error(err)
});
```
### Other Technical Details

`npm --version` output: 8.11.0

`node --version` output: v16.16.0

`pod --version` output (iOS issues only): nil

### Additional Context

firebase: 9.8.2

[...]

PR #6165, opened 2022-12-14
fix(android): get application/x-www-form-urlencoded as string

jcesarmobile:
As far as I understand, when using `application/x-www-form-urlencoded` you send a string, so that's why the `body.toJSObject()` is crashing, because it's a string, not a JSObject.

The PR removes the whole `application/x-www-form-urlencoded` else and goes directly to the final else where it read the body as string.

If you think there might be bases where the `application/x-www-form-urlencoded` could be getting a `JSObject`, let me know and I'll change the code to catch the JSON exception and fallback to string in that case.

closes https://github.com/ionic-team/capacitor/issues/5996

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/util/CapacitorHttpUrlConnection.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/util/CapacitorHttpUrlConnection.java
@@ -188,8 +188,6 @@ public class CapacitorHttpUrlConnection implements ICapacitorHttpUrlConnection {
                 dataString = call.getString("data");
             }
             this.writeRequestBody(dataString != null ? dataString : "");
-        } else if (contentType.contains("application/x-www-form-urlencoded")) {
-            this.writeRequestBody(body.toString());
         } else {
             this.writeRequestBody(body.toString());
         }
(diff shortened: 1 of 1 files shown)


theproducer (approved):
Looks good.  Tested this with one of the supplied reproductions and it worked.

As far as I know, I don't think you have to worry about JSObjects from a `x-www-form-urlencoded` response.  if that does happen, it should be on the caller to take the body response and parse as JSON.

IT-MikeS (approved):
`x-www-form-urlencoded` requests should always be sent with URL Query String formatted bodies. Users can accomplish this via a couple options:

- [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) in JS

- A custom mapping function that will convert your JS Object into query params:
```javascript
const bodyRequest = Object.keys(myJSObject).map(key => encodeURIComponent(key) + '=' + encodeURIComponent(myJSObject[key])).join('&');
```

In this case the Cap HTTP Plugin was accepting users sending JSON strings then auto converting them for users.

giralte-ionic (approved):

emelampianakis:
Hello, it throws an exception for me:
JSONException
    org.json.JSONException: JSValue could not be coerced to JSObject.
        at com.getcapacitor.JSValue.toJSObject(JSValue.java: 39)
        at com.getcapacitor.plugin.util.CapacitorHttpUrlConnection.setRequestBody(CapacitorHttpUrlConnection.java: 194)
        at com.getcapacitor.plugin.util.HttpRequestHandler.request(HttpRequestHandler.java: 399)
        at com.getcapacitor.plugin.CapacitorHttp$1.run(CapacitorHttp.java: 34)
        at java.lang.Thread.run(Thread.java: 920)

Tried both URLSearchParams & custom mapping but still getting an exception.
What additional info can i give you to resolve the issue

dnietodev:
Hello, when can we expect to have this fix with the next release of capacitor?

Thanks!

jcesarmobile:
This was released in 4.6.2

Issue #6174, opened 2022-12-17
bug: CapacitorHTTP fetch with Request object fails


jn42lm1:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

Installed Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

- iOS
- Android

### Current Behavior

The [Capacitor HTTP plugin](https://capacitorjs.com/docs/apis/http) supports a mode that patches the browser's window.fetch API when enabled.

With this active you can instead call the fetch API directly which will proxy all requests through the native Capacitor HTTP plugin on native platforms, i.e. Android, iOS. This is particularly useful to:

1. not have to refactor standard fetch calls you might have in your codebase already
2. enables support for observability frameworks, e.g. Instana, that also patch fetch to intercept HTTP traffic and that would not otherwise be able to detect calls to the ```Http.request()``` directly.

However, the current implementation of this patch in the Capacitor core/native-bridge.ts does not take into account that the [browser's fetch API](https://developer.mozilla.org/en-US/docs/Web/API/fetch) can be called with multiple argument signatures.

The patch expects all request to come in the form of ```fetch(resource: string, options: object)``` where **resource** is the URL string of the request, while **options** is the request options like a body, headers, etc.

The browser fetch API additionally allows for fetch requests to be made using a Request object instead, e.g. ```fetch(resource: Request)```. In this case the patch will fallback to using the browser fetch API instead of the Capacitor native HTTP plugin due to this limited implementation.

This is not the obvious behavior that a developer would expect. For example, making a request to a CORS protected resource with this 1st signature will work, while making a the request with the 2nd signature will fail.

To fix this, in **core/NativeBridge.ts** we can simply wrap the function arguments of the patch in a Request object to standardize further processing down stream across all possible function signatures.

### Expected Behavior

With the fetch patch enabled, you should be able to call fetch using all browser supported signatures, e.g. including
```fetch(resource: Request)```.

### Code Reproduction

See this [Demo Repo](https://github.com/jn42lm1/capacitor-fetch-api-signature) that shows the issue.

[...]

PR #6175, opened 2022-12-17
Capacitor HTTP fetch API support

jn42lm1:
The Capacitor HTTP plugin currently requires that all request to the browser fetch API is in the form ```fetch(resource: string, options: object)``` in order to proxy request through the native HTTP plugin. This is confusing, as the fetch API supports other formats like ```fetch(resource: Request)``` which is then not proxied, but instead sent through the browser fetch API directly which may leads to CORS issues.

This fixes [bug: CapacitorHTTP fetch with Request object fails #6174](https://github.com/ionic-team/capacitor/issues/6174)

### DIFF
--- a/core/native-bridge.ts
+++ b/core/native-bridge.ts
@@ -418,7 +418,7 @@ const initBridge = (w: any): void => {
               request.url.startsWith('https:')
             )
           ) {
-            return window.CapacitorWebFetch(resource, options);
+            return win.CapacitorWebFetch(resource, options);
           }
 
           try {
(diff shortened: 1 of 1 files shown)


jn42lm1:
Unfortunately this solution will only work on Chromium 105 browsers and upwards. The reason for this is that the Request object makes the body completely opaque on earlier browser. In 105 and new property was added for [Request.body](https://developer.mozilla.org/en-US/docs/Web/API/Request/body) that allows for the body to be read from a ReadableStream. I'll thus close the PR, as it would not be suitable to have a feature in Capacitor that is above the advertised browser support level.

PR #6220, opened 2023-01-17
fix(ios): CapacitorHttp cannot use delete method

### LINKED ISSUE #6219: bug: CapacitorHttp unable to use delete method in ios
### Capacitor Version
```
Latest Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

Installed Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

```
### Platform(s)
Ios

### Current Behavior
i am trying to use CapacitorHttp.delete() method but in apps it will throw below error
```
 Error: Plugin CapacitorHttp does not respond to method call "delete" using selector "delete:".
⚡️  Ensure plugin method exists, uses @objc in its declaration, and arguments match selector without callbacks in CAP_PLUGIN_METHOD

```
but if i tried CapacitorHttp.request({...options,method:'delete'}) then it works fine

### Expected Behavior
Able to use CapacitorHttp.delete() method

### Code Reproduction
```
const { data } = await CapacitorHttp.delete(options);
```

### Additional Context
i think the function should be delete instead of del right?
[Reference](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/Plugins/CapacitorHttp.swift)
image

junxiong96:
Fixes #6219

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorHttp.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorHttp.swift
@@ -30,7 +30,7 @@ public class CAPHttpPlugin: CAPPlugin {
         http(call, "PATCH")
     }
 
-    @objc func del(_ call: CAPPluginCall) {
+    @objc func delete(_ call: CAPPluginCall) {
         http(call, "DELETE")
     }
 }


jcesarmobile (commented):
good catch, thanks!

PR #6224, opened 2023-01-18
Fix parsing non-object JSON response in CapacitorHttp plugin

JanMisker:
The following valid json 
```json
"hello world"
```
is not handled correctly in the CapacitorHttp core plugin. This PR fixes that, although on Android there might be some improvements possible.

See https://github.com/JanMisker/Capacitor-JSONParseError for a simple app that reproduces this error.

### DIFF
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -2,7 +2,7 @@
 /*! Capacitor: https://capacitorjs.com/ - MIT License */
 /* Generated File. Do not edit. */
 
-const nativeBridge = (function (exports) {
+var nativeBridge = (function (exports) {
     'use strict';
 
     var ExceptionCode;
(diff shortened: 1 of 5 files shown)


markemer:
@JanMisker Can you resolve these conflicts? Seems like this solves our issue, but our 204 response handling update caused a conflict. I can move and fix it if you want, but fixing this one will give you the credit in contributors since this is your first PR against the repo.

JanMisker:
Check I will do it today or tomorrow, thanks for considering my credits 🤓

markemer:
One more ask. Can you run `npm run fmt` on these files?

JanMisker:
Is there an ETA on when this will be taken up in a release?

markemer:
It should be in the current capacitor 5.0 RC. I'm not sure of the GA of Capacitor 5, other than soon.

PR #6229, opened 2023-01-19
fix(cli/cordova): Exclude private framework Headers

dtarnawsky:


dtarnawsky:
FYI: Sample project to see the error and test it after the fix is applied:
[https://github.com/dtarnawsky/cs-scanbot
](https://github.com/dtarnawsky/cs-scanbot
)

jcesarmobile (approved):

IT-MikeS (approved):

Issue #6202, opened 2023-01-04
bug: CapacitorHTTP crashes when a header value is given as number on iOS


nip3o:
### Capacitor Version

💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/cli: 4.6.1
  @capacitor/ios: 4.6.1
  @capacitor/core: 4.6.1

[success] iOS looking great! 👌

### Platform(s)

Only iOS

### Current Behavior

Giving the value of a HTTP header as a number rather than as string causes the exception `Could not cast value of type '__NSCFNumber' (0x122864c30) to 'NSString' (0x1b7cc7fc8).` to be thrown when the value is unpacked in [HttpRequestHandler](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift#L148).

    fetch("https://www.google.com/", {
        headers: {
          thisIsFine: "1337",
          thisThrowsError: 1337,
        },
      })

### Expected Behavior

The numeric value should be converted to a string. This is what the Android version of the plugin does, and what the Safari browser implementation of `fetch()` does. One could definitely argue that header values should always be passed in as strings, but I don't think it's helpful to crash on it. Either way, plugin code should not crash internally due to user errors.

### Code Reproduction

https://github.com/nip3o/capacitor-http-headers-crash

### Other Technical Details

`npm --version` output: 6.14.17

`node --version` output: v14.20.0

`pod --version` output (iOS issues only): 1.11.3

### Additional Context

I did not get a stacktrace, but the crash happens in [ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift:148](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift#L148): `let headers = (call.getObject("headers") ?? [:]) as! [String: String]`

`Could not cast value of type '__NSCFNumber' (0x122864c30) to 'NSString' (0x1b7cc7fc8).`

PR #6251, opened 2023-01-25
fix(ios): crash when http headers contain numbers

jcesarmobile:
Stringify the headers so the native side only gets strings as native headers can only be strings.

closes https://github.com/ionic-team/capacitor/issues/6202


IT-MikeS (approved):

giralte-ionic (approved):

davy-ikv:
This change breaks passing headers in iOS

https://github.com/ionic-team/capacitor/commit/028c556a50b41ee99fe8f4f1aa2f42d3fd57f92d#diff-73da2144bc5fefb384b6fc1501a8efbabac7cfdf2faa411b68cdd113f53d884dL430-R432

JSON.stringify headers makes headers become a String, but in native part of iOS, it expect headers is a hash table ([String: String])

PR #6264, opened 2023-01-30
feat(android): add ability to create config from a custom file path

carlpoole:
This PR adds the ability to create configs from a provided file path. It will be used by Portals to allow config files to be contained within specific Portal subdirs in assets and also to allow them to be contained within live update files on disk. 

This PR also fixes a bug that was found where CapacitorWebView may not read the correct config that was provided to Capacitor when used with Portals.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
@@ -78,7 +78,7 @@ public class CapConfig {
             this.configJSON = config;
         } else {
             // Load the capacitor.config.json
-            loadConfig(assetManager);
+            loadConfigFromAssets(assetManager, null);
         }
 
         deserializeConfig(null);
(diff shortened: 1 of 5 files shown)


giralte-ionic (approved):

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -98,7 +99,47 @@ public static CapConfig loadDefault(Context context) {
...
+        config.loadConfigFromAssets(context.getAssets(), path);
+        config.deserializeConfig(context);
+        return config;
+    }
+
+    /**
+     * Constructs a Capacitor Configuration from config.json file within the app file-space.
+     *
+     * @param context The context.
+     * @param path A path relative to the root assets directory.
```
IT-MikeS, line 131: Would be relative to app disk space root no?

IT-MikeS (changes requested):

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
carlpoole, line 131:
```suggestion
     * @param path A path relative to the root of the app file-space.
```

carlpoole: fixed - ty

PR #6265, opened 2023-01-30
chore(ios-template): git ignore App/output folder produced by cap build ios

dtarnawsky:
The cap build ios command will produce an ipa file in the folder App/output. This change ensures that the folder is ignored by source control


jcesarmobile (approved):
Looks good

But we usually only update templates for major releases, so not sure if we should wait until we start with Capacitor 5 work before merging

giralte-ionic (approved):

PR #6286, opened 2023-02-09
chore: Update CONTRIBUTING.md

giralte-ionic:
also removed the publishing instructions as I'm not sure why anyone but us would use it and our publishing docs are in [ionic's private] notion

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -80,7 +80,7 @@ On web, this means do not add any third party libraries such as Firebase or Loda
 ### Branches
 
 * [`main`](https://github.com/ionic-team/capacitor/tree/main): Latest Capacitor development branch
-* [`3.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 3 (bug and security fixes only)
+* [`3.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 3 (security fixes only)
 * [`2.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 2 (not maintained)
 * [`1.x`](https://github.com/ionic-team/capacitor/tree/1.x): Capacitor 1 (not maintained)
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -80,7 +80,8 @@ On web, this means do not add any third party libraries such as Firebase or Loda
 ### Branches
 
 * [`main`](https://github.com/ionic-team/capacitor/tree/main): Latest Capacitor development branch
-* [`2.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 2 (bug and security fixes only)
+* [`3.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 3 (bug and security fixes only)
```
jcesarmobile, line 83: we are not doing bug fixes, so not sure if it should say "security fixes only", or if we should be backporting bug fixes

jcesarmobile (changes requested):
I think the publishing information is useful even if nobody other than us can publish, it's always good to have that information in the repo instead of having to navigate through notion to find it. Also the lerna part about using fixed versioning can help users see all packages are released at the same time with the same version number.

But the information is out of date since we now use github actions to publish.

jcesarmobile (approved):

CONTRIBUTING.md
```diff
@@ -80,7 +80,8 @@ On web, this means do not add any third party libraries such as Firebase or Loda
 ### Branches
 
 * [`main`](https://github.com/ionic-team/capacitor/tree/main): Latest Capacitor development branch
-* [`2.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 2 (bug and security fixes only)
+* [`3.x`](https://github.com/ionic-team/capacitor/tree/2.x): Capacitor 3 (security fixes only)
```
ItsChaceD, line 83: The link should be `3.x`

ItsChaceD (changes requested):

ItsChaceD (approved):

PR #6287, opened 2023-02-09
chore(http): make interfaces public

ItsChaceD:
This PR changes the interfaces in the HTTP plugin to be public so they can be extended by other plugins.


IT-MikeS (approved):

jcesarmobile (approved):
looks good

but you say "so they can be extended by other plugins"

the other plugins are going to subclass or override the classes/methods? in that case I think they should be open instead of public
if not, public is ok
(related to iOS only)

Issue #6297, opened 2023-02-14
bug(http-patch): request headers are empty


muuvmuuv:
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 4.6.3
  @capacitor/core: 4.6.3
  @capacitor/android: 4.6.3
  @capacitor/ios: 4.6.3

Installed Dependencies:

  @capacitor/cli: 4.6.3
  @capacitor/core: 4.6.3
  @capacitor/ios: 4.6.3
  @capacitor/android: 4.6.3

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

- Android
- iOS

### Current Behavior

Request headers are not passed any further to native handler. It is getting the headers but cannot pass them further, they aren't accessible through call but exist.

Screenshot from `call.getObject`:

<img width="687" alt="Bildschirm­foto 2023-02-14 um 10 43 23" src="https://user-images.githubusercontent.com/11534760/218698512-366ad67e-7a5e-4fa1-8e8d-99b1a63537d8.png">

and `call.Headers`:

<img width="1080" alt="Bildschirm­foto 2023-02-14 um 11 01 17" src="https://user-images.githubusercontent.com/11534760/218703281-67ffe730-48cb-4f0c-93d9-8e034aae77ca.png">

### Expected Behavior

...

### Code Reproduction

> https://github.com/muuvmuuv/issue-capacitor-6297

### Other Technical Details

`npm --version` output: pnpm 7.x

`node --version` output: LTS 18

`pod --version` output (iOS issues only):

Issue #6274, opened 2023-02-05
bug: CapacitorHttp XMLHttpRequest (Angular HttpClient) interceptor doesn't pass headers correctly


dennisameling:
### Capacitor Version

```
% npx cap doctor  
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.6.3
  @capacitor/core: 4.6.3
  @capacitor/android: 4.6.3
  @capacitor/ios: 4.6.3

Installed Dependencies:

  @capacitor/cli: 4.6.3
  @capacitor/core: 4.6.3
  @capacitor/android: 4.6.3
  @capacitor/ios: 4.6.3

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

Android and iOS

### Current Behavior

When using Angular's HttpClient in combination with Capacitor's `CapacitorHttp` [intercepting enabled](https://capacitorjs.com/docs/apis/http), headers are sent incorrectly to the server.

Here's what it looks like on Android (did [remote debugging](https://ionicframework.com/docs/troubleshooting/debugging#android-and-chrome) to get some more details):

<img width="1904" alt="Screenshot 2023-02-05 at 11 32 27" src="https://user-images.githubusercontent.com/17739158/216814071-c373f2d4-08b8-41b5-a448-f139482444fe.png">

**Note that the `headers` are a JSON-encoded string when `CapacitorHttp` is intercepting `XMLHttpRequest`. That doesn't seem right!**

And here's what it looks like on iOS:

<img width="200" src="https://user-images.githubusercontent.com/17739158/216814481-710bb194-b7aa-4337-8daa-21217b349b6f.PNG" alt="IMG_0064" />

### Expected Behavior

I expect Angular's HttpClient to work correctly in combination with `CapacitorHttp`'s `XMLHttpRequest` interceptor.

### Code Reproduction

Here's the full codebase of my screenshots above: https://github.com/dennisameling/capacitor-http-bug

### Other Technical Details

`npm --version` output: 8.19.2

`node --version` output: v16.18.1

`pod --version` output (iOS issues only): 1.11.2

zakuru:
similar issue when upgraded to 4.6.3 - had to rollback to 4.6.2

matudelatower:
Same issue here, apparently overwrites all request even `@angular/common/http`, and interceptors doesnt work.

zakuru:
@jcesarmobile  can you please help look into it ?

[...]

PR #6303, opened 2023-02-16
fix(ios): Correctly Attach Headers to Request

ItsChaceD:
This PR fixes an issue with the fetch/XHR override where http headers were not being set on the request

Resolves: #6297, #6274

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
@@ -111,7 +111,7 @@ public class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
     public func setRequestHeaders(_ headers: [String: String]) {
         headers.keys.forEach { (key: String) in
             let value = headers[key]
-            request.addValue(value!, forHTTPHeaderField: key)
+            request.setValue(value!, forHTTPHeaderField: key)
             self.headers[key] = value
         }
     }
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
@@ -108,6 +108,15 @@ public class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         }
     }
 
+    @available(*, deprecated)
+    public func setRequestHeaders(_ headers: [String: String]) {
+        headers.keys.forEach { (key: String) in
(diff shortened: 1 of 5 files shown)


ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -108,11 +108,11 @@ public class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         }
     }
 
-    public func setRequestHeaders(_ headers: [String: String]) {
+    public func setRequestHeaders(_ headers: [String: Any]) {
```
jcesarmobile, line 111: Since `CapacitorUrlRequest` is a public class and `setRequestHeaders(_ headers: [String: String])` is a public method we can't change it to Any now or it would be breaking.
We should keep the existing `setRequestHeaders(_ headers: [String: String])` and add a new `func setRequestHeaders(_ headers: [String: Any])` and probably deprecate the old one.

muuvmuuv:
Can we have tests for this? This really breaks production applications and should be tested before release.

ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -108,14 +108,23 @@ public class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         }
     }
 
+    @available(*, deprecated)
     public func setRequestHeaders(_ headers: [String: String]) {
         headers.keys.forEach { (key: String) in
             let value = headers[key]
-            request.addValue(value!, forHTTPHeaderField: key)
+            request.setValue(value!, forHTTPHeaderField: key)
```
jcesarmobile, line 115:
```suggestion
            request.addValue(value!, forHTTPHeaderField: key)
```
Since we are keeping the method just to not break things, we should not change the behavior neither

ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -108,14 +108,23 @@ public class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         }
     }
 
+    @available(*, deprecated)
```
jcesarmobile, line 111: can you add a message to the deprecation that says that the one with `[String: Any]` should be used instead?

jcesarmobile (changes requested):

giralte-ionic (approved):

PR #6323, opened 2023-02-24
feat(cli): allow adding options for pod install by environment variables

scr2em:
Hi, 

Currently there is no way to add extra options for `pod install` when we run `sync` or `update`. In my case, I needed `--repo-update` and sometimes `--verbose`. So, I made it optional to add an environment variables called `PODS_OPTIONS` to allow adding options 

image

Usage example 
```
PODS_OPTIONS='--verbose --repo-update' ionic cap sync ios
```

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -279,6 +279,7 @@ async function loadIOSConfig(
 ): Promise<IOSConfig> {
   const name = 'ios';
   const podPath = determineCocoapodPath();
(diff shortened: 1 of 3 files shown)


jcesarmobile:
Thanks for the pull request, but since those are options that are not frequently needed, I don't see it very useful to add them.
You can always run `pod install` directly from `ios` folder with the options you want if you need to update the cocoapods repository or want to troubleshoot with verbose.

PR #6327, opened 2023-02-27
chore(android)!: set default targetSdkVersion to 33

jcesarmobile:
for capacitor/android and capacitor-cordova-android-plugins modules


markemer:
Does this deal with the weird 33x behavior too? I've had that issue too, where android studio installs 33x and doesn't tell you it did.

jcesarmobile:
no, just sets the default targetSdkVersion version in case the user doesn't have the targetSdkVersion in variables.gradle or doesn't have variables.gradle file at all.

33x is not a valid value for gradle files

markemer:
> no, just sets the default targetSdkVersion version in case the user doesn't have the targetSdkVersion in variables.gradle or doesn't have variables.gradle file at all.
> 
> 33x is not a valid value for gradle files

Cool - thanks! I hate that Android Studio does that without telling you.

theproducer (approved):

giralte-ionic (approved):

IT-MikeS (approved):

markemer (approved):

PR #6330, opened 2023-02-27
chore(cli): update typescript dev-dep to ~4.9.5

IT-MikeS:


markemer (approved):
Pulled this down - and now `npm run build` works for the cli. Did some smoke testing of the CLI and things run. Not sure if we need to do more testing.

giralte-ionic (approved):

PR #6332, opened 2023-02-27
feat(android): update default scheme from http to https

markemer:
Update default Android scheme from http to https. 

Will only effect use of the default scheme - anyone overriding it will continue to.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
@@ -1,6 +1,6 @@
 package com.getcapacitor;
 
-import static com.getcapacitor.Bridge.CAPACITOR_HTTP_SCHEME;
+import static com.getcapacitor.Bridge.CAPACITOR_HTTPS_SCHEME;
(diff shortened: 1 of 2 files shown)


jcesarmobile:
We decided to do this for capacitor 6, for capacitor 5 the decision was to make npx cap init command to add the server.androidScheme with https value so all new apps use https by default but existing apps don’t get the scheme changed as it would produce web data loss

markemer:
> We decided to do this for capacitor 6, for capacitor 5 the decision was to make npx cap init command to add the server.androidScheme with https value so all new apps use https by default but existing apps don’t get the scheme changed as it would produce web data loss

Ok, we'll hold this one off and have init deal with it - so do we just change the Android template, or does cli/init actually need that value added?

jcesarmobile:
The `npx cap init` command generates a `capacitor.config.json` or `capacitor.config.ts` file (.ts is used if typescript is installed in the project).
So that generated file should include those keys, the change has to be done in the Capacitor CLI code.

jcesarmobile:
closing for now since Capacitor 6 won't happen anytime soon

PR #6333, opened 2023-02-27
feat(android): Removing enableJetifier

theproducer:
Removes `android.enableJetifier` in core and in the publish script as its no longer needed for modern libraries.

### DIFF
--- a/android-template/gradle.properties
+++ b/android-template/gradle.properties
@@ -20,5 +20,3 @@ org.gradle.jvmargs=-Xmx1536m
 # Android operating system, and which are packaged with your app's APK
 # https://developer.android.com/topic/libraries/support-library/androidx-rn
 android.useAndroidX=true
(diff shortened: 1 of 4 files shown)


jcesarmobile (changes requested):
Can you please move the android-template/gradle.properties change to a separate PR?
We should make any android-template change a separate PR for easier tracking of what's required for the Capacitor 4 to 5 migration.

android-template/gradle.properties
```diff
@@ -22,3 +22,4 @@ org.gradle.jvmargs=-Xmx1536m
 android.useAndroidX=true
 # Automatically convert third-party libraries to use AndroidX
 android.enableJetifier=true
+
```
jcesarmobile, line 25:
```suggestion
```

jcesarmobile (approved):

IT-MikeS (approved):

PR #6354, opened 2023-03-01
feat(ios): BREAKING

Steven0351:
Makes `CAPBridgedPlugin` protocol requirements be at the instance level and not the class level and removes the `getMethod` requirement.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgedPlugin.h
+++ b/ios/Capacitor/Capacitor/CAPBridgedPlugin.h
@@ -16,16 +16,14 @@
 @class CAPPlugin;
 
 @protocol CAPBridgedPlugin <NSObject>
-+(NSString *)pluginId;
-+(NSString *)jsName;
-+(NSArray *)pluginMethods;
-+(CAPPluginMethod *)getMethod:(NSString *)methodName;
-@optional
+@property (nonnull, readonly) NSString *identifier;
+@property (nonnull, readonly) NSString *jsName;
+@property (nonnull, readonly) NSArray<CAPPluginMethod *> *pluginMethods;
 @end
 
 #define CAP_PLUGIN_CONFIG(plugin_id, js_name) \
-+ (NSString *)pluginId { return @#plugin_id; } \
-+ (NSString *)jsName { return @js_name; }
+- (NSString *)identifier { return @#plugin_id; } \
+- (NSString *)jsName { return @js_name; }
 #define CAP_PLUGIN_METHOD(method_name, method_return_type) \
 [methods addObject:[[CAPPluginMethod alloc] initWithName:@#method_name returnType:method_return_type]]
 
@@ -35,19 +33,10 @@
 @interface objc_name (CAPPluginCategory) <CAPBridgedPlugin> \
 @end \
 @implementation objc_name (CAPPluginCategory) \
-+ (NSArray *)pluginMethods { \
+- (NSArray *)pluginMethods { \
   NSMutableArray *methods = [NSMutableArray new]; \
   methods_body \
   return methods; \
-} \
-+ (CAPPluginMethod *)getMethod:(NSString *)methodName { \
-  NSArray *methods = [self pluginMethods]; \
-  for(CAPPluginMethod *method in methods) { \
-    if([method.name isEqualToString:methodName]) { \
-      return method; \
-    } \
-  } \
-  return nil; \
 } \
 CAP_PLUGIN_CONFIG(objc_name, js_name) \
 @end
--- a/ios/Capacitor/Capacitor/CapacitorBridge.swift
+++ b/ios/Capacitor/Capacitor/CapacitorBridge.swift
@@ -3,6 +3,8 @@ import Dispatch
 import WebKit
 import Cordova
 
+internal typealias CapacitorPlugin = CAPPlugin & CAPBridgedPlugin
+
 /**
  An internal class adopting a public protocol means that we have a lot of `public` methods
  but that is by design not a mistake. And since the bridge is the center of the whole project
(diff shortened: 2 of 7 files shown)


Steven0351:
This change is technically breaking although I don't believe many people were using `CAPBridgedPlugin` directly in the wild due to the macro being used to generate conformance. I did a github code search and outside of Ionic Portals using it directly, I found one personal project a developer used that had referenced the protocol directly in their code. This change opens some doors that were previously impossible due to the nature of the class level methods that were required of `CAPBridgedPlugin` and it was a bit cumbersome and ugly to write a pure Swift implementation of a loadable plugin.

To illustrate, this is the current state of the plugin shipped with [Ionic Portals iOS](https://github.com/ionic-team/ionic-portals-ios) (I've omitted the method implementations for brevity):
```swift
@objc(IONPortalsPlugin)
internal final class Plugin: CAPPlugin, CAPBridgedPlugin {
    static func pluginId() -> String {
        "IONPortalsPlugin"
    }
    
    static func jsName() -> String {
        "Portals"
    }
    
    static let methods: [CAPPluginMethod] = [
        CAPPluginMethod(name: "publishNative", returnType: CAPPluginReturnPromise),
        CAPPluginMethod(name: "subscribeNative", returnType: CAPPluginReturnCallback),
        CAPPluginMethod(name: "unsubscribeNative", returnType: CAPPluginReturnPromise)
    ]
    
    static func pluginMethods() -> [Any] { methods }
    
    static func getMethod(_ methodName: String) -> CAPPluginMethod? {
        methods.first { $0.name == methodName }
    }
}
```

With the changes, the pure Swift implementation would look like this:
```swift
@objc(IONPortalsPlugin)
internal final class Plugin: CAPPlugin, CAPBridgedPlugin {
    var identifier: String { "IONPortalsPlugin" }
    var jsName: String { "Portals" }
    
    let pluginMethods: [CAPPluginMethod] = [
        CAPPluginMethod(name: "publishNative", returnType: CAPPluginReturnPromise),
        CAPPluginMethod(name: "subscribeNative", returnType: CAPPluginReturnCallback),
        CAPPluginMethod(name: "unsubscribeNative", returnType: CAPPluginReturnPromise)
    ]
}
```

This alone isn't that impressive, but it opens the door to allow for more expressive APIs for creating plugins in Swift. I currently have some experiments I've been working on that take advantage of these changes that I have locally, that allow for the ability to express plugins more concisely like this:
```swift
internal struct PortalsPlugin: Plugin {
    @PluginEnvironment(\.bridge) var bridge

     struct UnsubscribeOptions: Codable {
        var topic: String
        var subscriptionRef: Int
    }
    
    struct SubscribeOptions: Decodable {
        var topic: String
    }

    @PluginBuilder
    var body: CapacitorPlugin {
        Module(id: "PortalsPlugin", name: "Portals")

        None("publishNative") { (obj: JSObject) in
            guard let topic = obj["topic"] as? String else {
                throw CapacitorRejection(message: "Topic not provided")
            }

            PortalsPubSub.publish(obj["data"], to: topic)
        }
        
        Callback("subscribeNative", keepAlive: true) { (opts: SubscribeOptions, resolve) -> UnsubscribeOptions in
            let ref = IONPortalsPubSub.subscribe(topic: opts.topic) {
                resolve(JSTypes.coerceDictionaryToJSObject($0)) 
            }

            return UnsubscribeOptions(
                topic: opts.topic,
                subscriptionRef: ref
            )
        }
        
        Promise("unsubscribeNative") { (options: UnsubscribeOptions) in
            PortalsPubSub.unsubscribe(from: options.topic, subscriptionRef: options.subscriptionRef)
        }
    }
}
```

With the current state of `CAPBridgedPlugin` APIs like this are impossible to create due to the class level requirements of the protocol. The code to make that Plugin DSL possible lives in it's own package and doesn't have to be embedded into Capacitor itself.

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -444,26 +427,27 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
      */
     // swiftlint:disable:next function_body_length
     func handleJSCall(call: JSCall) {
-        guard let plugin = self.plugin(withName: call.pluginId) ?? self.loadPlugin(pluginName: call.pluginId) else {
-            CAPLog.print("⚡️  Error loading plugin \(call.pluginId) for call. Check that the pluginId is correct")
-            return
+        let load = {
```
Steven0351, line 430: Not really sure if this is wholly neccessary, I mostly tried to keep the behavior the same, but I'm not sure when a plugin would be called before the bridge has initialized all the plugins 🤷‍♂️. In all my debugging of this `handleJSCall(call:)` method, I've never gotten it to coalesce to the load method.

markemer:
New guy question. Should this be `feat!(ios)`

ios/Capacitor/Capacitor/CAPBridgedPlugin.h
```diff
@@ -16,16 +16,14 @@
...
 
 @protocol CAPBridgedPlugin <NSObject>
-+(NSString *)pluginId;
-+(NSString *)jsName;
-+(NSArray *)pluginMethods;
-+(CAPPluginMethod *)getMethod:(NSString *)methodName;
-@optional
+@property (nonnull, readonly) NSString *identifier;
+@property (nonnull, readonly) NSString *jsName;
+@property (nonnull, readonly) NSArray<CAPPluginMethod *> *pluginMethods;
```
markemer, line 21: Well this is a much needed update. That's some old school ObjC. The non-null will help a lot with swift. 💯

Steven0351:
> New guy question. Should this be `feat!(ios)`?

Probably! 😀

ios/Capacitor/Capacitor/CapacitorBridge.swift
markemer: Huh, yeah, that's interesting - I'd love if we had some real world code we could run regressions on like Swift.org does.

giralte-ionic (approved):

markemer (approved):
Forgot to approve this yesterday - I'm a big fan of anything that makes things more Swift-y. The test app seems to run ok, but I'm sure some sort of breaking may still be hiding someplace - but I don't think it will be big.

PR #6364, opened 2023-03-03
refactor(android)!: make call getArray and getObject return null

jcesarmobile:
breaking change, we have been warning about it.
To match iOS behavior, make getArray and getObject return null

We are doing null checks in most plugins and they are not working because of this returning empty objects/arrays, so should fix several problems.
But also will break camera requestPermissions as it's not doing null check, and probably several user plugins.


theproducer (approved):
Looks good!

I was able to confirm your note about Camera requestPermissions crashing.  CapacitorHttp crashes as well.  I can create a ticket for that if you haven't already.

jcesarmobile:
thanks, I just reported it, didn't notice because I just did a code search in capacitor-plugins repository

Issue #6369, opened 2023-03-06
bug: iOS Schemes do not work during build


mibzman:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.7.0
  @capacitor/core: 4.7.0
  @capacitor/android: 4.7.0
  @capacitor/ios: 4.7.0

Installed Dependencies:

  @capacitor/core: 4.7.0
  @capacitor/android: 4.5.0
  @capacitor/ios: 4.7.0
  @capacitor/cli: 4.5.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

iOS

### Current Behavior

I have reproduced my issue with iOS schemes in [this repo](https://github.com/mibzman/cap-flavor-example).  If you run `npx cap build ios --scheme "App QA"` you should get this output:

```
✖ Building xArchive - failed!
[error] Command line invocation:
        /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -workspace
        "App QA.xcworkspace" -scheme "App QA" -destination generic/platform=iOS
        -archivePath "App QA.xcarchive" archive
        
        User defaults from command line:
        IDEArchivePathOverride =
        .../cap-flavor-example/ios/App/App QA.xcarchive
        IDEPackageSupportUseBuiltinSCM = YES
        
        xcodebuild: error: 'App QA.xcworkspace' does not exist.
```

I have followed the [scheme documentation](https://capacitorjs.com/docs/guides/environment-specific-configurations#setup-a-new-ios-scheme) exactly.  I get the same result using `capacitor.config.json` to select the scheme.

### Expected Behavior

if a scheme can run, it should also build.

### Code Reproduction

https://github.com/mibzman/cap-flavor-example

### Other Technical Details

`npm --version` output: 8.11.0

`node --version` output: v16.16.0

`pod --version` output (iOS issues only): 1.8.4

### Additional Context

follow up to https://github.com/ionic-team/capacitor/issues/6269

markemer:
@mibzman Are schemes working for you in iOS now? Make sure `native-run` is updated to 1.7.2

andrewmurraydavid:
As mentioned in the PR, the does not fix the issue for `cap update`. The issue still persists when running `cap update` with a custom scheme, unfortunately

PR #6371, opened 2023-03-07
fix(cli): point build to proper workspace name

markemer:
Fixed bug where we mistakenly used the scheme name for the workspace name.

Version for 4.x is in PR #6373 

Fixes: #6369


jcesarmobile (approved):

IT-MikeS (approved):

andrewmurraydavid:
After a bit of digging looks like the scheme is actually hardcoded in the config, which given the scheme is available as a config param, seems counter intuitive:
https://github.com/ionic-team/capacitor/blob/daf2ec64df0c567c6a42560488e5d2515eff8a33/cli/src/config.ts#L289

PR #6372, opened 2023-03-07
chore(android): fix getPackageInfo deprecations

IT-MikeS:


markemer (approved):

jcesarmobile (changes requested):
You have created the InternalUtils.java class, where you already check the two ways of getting the package info, the legacy and the new one, but then for every call, you call the new getPackageInfoLegacy method and also the new way of getting the package info for tiramisu, you shouldn't need that since getPackageInfoLegacy does both, you can reduce a lot of code and version checks

android/capacitor/src/main/java/com/getcapacitor/util/InternalUtils.java
```diff
@@ -0,0 +1,22 @@
...
+        return InternalUtils.getPackageInfo(pm, packageName, 0);
+    }
+
+    @SuppressWarnings("deprecation")
+    public static PackageInfo getPackageInfo(PackageManager pm, String packageName, long flags)
+        throws PackageManager.NameNotFoundException {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            return pm.getPackageInfo(packageName, PackageManager.PackageInfoFlags.of(flags));
+        } else {
+            return pm.getPackageInfo(packageName, (int) flags);
```
jcesarmobile, line 19: can you move the deprecated code to a separate private function with the suppress warning and remove the suppress warning from the public one?
so in case google deprecates the new function in the future we don't miss it

jcesarmobile (approved):

PR #6379, opened 2023-03-09
chore(android): move package to build gradle

IT-MikeS:


jcesarmobile (changes requested):
can you remove the existing code for replacing the package name in the `AndroidManifest.xml`?
it's not needed since the package is being removed

theproducer (approved):

jcesarmobile (changes requested):
Just noticed that we have the activity name hardcoded to `android:name="com.getcapacitor.myapp.MainActivity"`
Which no longer gets replaced because I told you to remove the code that writes in the `AndroidManifest.xml` 🙈 
but in new android apps the MainActivity is references as `.MainActivity` (without package).
So, can you change it to `android:name=".MainActivity"`?

jcesarmobile (approved):

PR #6397, opened 2023-03-15
fix: copy url from nativeResponse to response in native-bridge.ts

wilgert:
This url is used in `cordova-plugin-ionic` to do live updates. Without this fix the `response.url` would be an empty string here:
 https://github.com/ionic-team/cordova-plugin-ionic/blob/5b190675525806ea15e56d4e59eb67b74cc58c55/www/common.ts#L288
 and live updates do not work.

### DIFF
--- a/core/native-bridge.ts
+++ b/core/native-bridge.ts
@@ -454,6 +454,15 @@ const initBridge = (w: any): void => {
               status: nativeResponse.status,
             });
 
+            /*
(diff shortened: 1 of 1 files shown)


ItsChaceD (approved):

wilgert:
I now ran prettier :-)

markemer (approved):

Issue #6131, opened 2022-11-30
improvement: re-evaluate WebView's "allowFileAccessFromFileURLs" flag on iOS


Dervol03:
# Description
As capacitor is now using a custom scheme to load its data on iOS, it is very likely that `allowFileAccessFromFileURLs` can now be deactivated for WebView.
During a pentest of our application, it was considered a security risk by the pentesters.

# Platform(s)
ios

# Preferred Solution
Remove the flag or explicitly set it to `false`

### Discussed in https://github.com/ionic-team/capacitor/discussions/6124

<div type='discussions-op-text'>

<sup>Originally posted by **Dervol03** November 29, 2022</sup>
Hi everyone,

we recently had a pentest for our app and one of the issues raised during the process was that `allowFileAccessFromFileURLs` is `true` for the webview on iOS. However, it is not set for the android counterpart code. 
This raises the question whether it is still necessary for capacitor to operate correctly or just an historical oversight due to a necessity in the past?</div>

PR #6414, opened 2023-03-20
refactor(ios): remove allowFileAccessFromFileURLs preference call

jcesarmobile:
allowFileAccessFromFileURLs is a non documented preference that can cause security warnings
it allows to load file:// urls in the WKWebView, but Capacitor uses custom schemes for serving the local files and provides a helper to convert any file:// url to a capacitor://localhost url, so the preference shouldn't be needed anymore, and if anybody is using it, should move to the capacitor://localhost urls

closes https://github.com/ionic-team/capacitor/issues/6131


markemer (approved):
Yeah, this isn't guaranteed to stick around and if Apple's having a bad day they might even reject and app with it, although, I doubt it. Seems like we're using the proper alternative anyway.

giralte-ionic (approved):

PR #6416, opened 2023-03-21
feat(android): add plugin hooks for WebViewClient.onRenderProcessGone

peitschie:
This change allows more advanced handling of the webview render process termination. This can be useful for capturing additional information, or even conceivably allow the crash to be prevented should a plugin choose to.

Source:
https://developer.android.com/reference/android/webkit/WebViewClient#onRenderProcessGone(android.webkit.WebView,%20android.webkit.RenderProcessGoneDetail)

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -15,6 +15,7 @@ import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.HandlerThread;
+import android.webkit.RenderProcessGoneDetail;
 import android.webkit.ValueCallback;
 import android.webkit.WebSettings;
 import android.webkit.WebView;
@@ -395,6 +396,22 @@ public class Bridge {
             return true;
         }
         return false;
+    }
+
+    public boolean onRenderProcessGone(final WebView view, RenderProcessGoneDetail detail) {
+        /*
+         * Give plugins the chance to handle or record render process removal
+         */
+        boolean result = false;
+        for (Map.Entry<String, PluginHandle> entry : plugins.entrySet()) {
+            Plugin plugin = entry.getValue().getInstance();
+            if (plugin != null) {
+                if (plugin.onRenderProcessGone(view, detail)) {
+                    result = true;
+                }
+            }
+        }
+        return result;
     }
 
     private boolean isNewBinary() {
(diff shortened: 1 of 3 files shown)


peitschie:
The main reason for wanting a plugin hook like this is to feed app crashes into Crashalytics/AppCenter/etc. The normal crash reporters do not detect or report on the chrome renderer being terminated, meaning most app crashes are lost unless they happen to be in a native thread or call somewhere.

This feature has now been accepted into cordova-android: https://github.com/apache/cordova-android/pull/1574

I've created a plugin to simplify crashing the chrome renderer: https://github.com/peitschie/capacitor-chrome-crasher (capacitor 5 compatible)

Once that plugin is added to an app, it's as simple as `Capacitor.Plugins.ChromeCrasher.loadChromeUrl()`

Would love an indication about whether this is something CapacitorJS would want, as so far there's been no feedback from any maintainers 😅

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -397,6 +398,22 @@ public boolean launchIntent(Uri url) {
...
+    public boolean onRenderProcessGone(final WebView view, RenderProcessGoneDetail detail) {
+        /*
+         * Give plugins the chance to handle or record render process removal
+         */
+        boolean result = false;
+        for (Map.Entry<String, PluginHandle> entry : plugins.entrySet()) {
+            Plugin plugin = entry.getValue().getInstance();
+            if (plugin != null) {
+                if (plugin.onRenderProcessGone(view, detail)) {
+                    result = true;
```
peitschie, line 410: The loop continues irrespective of the return value so that any reporting plugins can find out about the crash even if one of the plugins is attempting to handle it gracefully.

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
```diff
@@ -92,4 +93,9 @@ public void onReceivedHttpError(WebView view, WebResourceRequest request, WebRes
             view.loadUrl(errorPath);
         }
     }
+
+    @Override
+    public boolean onRenderProcessGone(final WebView view, RenderProcessGoneDetail detail) {
+        return bridge.onRenderProcessGone(view, detail);
```
peitschie, line 99: The webview is passed down as the `onRenderProcessGone` method is called for **any** render crashes within the app (including a pop-up tab, for example).

jcesarmobile:
closing in favor of https://github.com/ionic-team/capacitor/pull/6946

Issue #6361, opened 2023-03-03
bug: iframe doesn't work with data uri


wendt88:
### Capacitor Version

```
@capacitor/core: 4.6.3
@capacitor/android: 4.6.3
@capacitor/cli: 4.6.3
@capacitor/ios: 4.6.3
```

### Platform(s)
android only

### Current Behavior
when I add an iframe with a data uri src, it show's nothing.
tried also with data uri of an image, it works as an image src, but not as an iframe src

### Expected Behavior
it should show an iframe containing a `<h1>` with 'Hi'

### Code Reproduction
add the following element:
```html
<iframe src="data:text/html;charset=utf-8,%3C!DOCTYPE%20html%3E%3Chtml%3E%3Chead%3E%3C/head%3E%3Cbody%3E%3Ch1%3EHi%3C/h1%3E%3C/body%3E%3C/html%3E"></iframe>
```
or clone https://github.com/wendt88/ionic-iframe-data-uri

### Other Technical Details
in web and ios it's fine
the data uri inside android chrome works also fine

`npm --version` output: 8.12.2

`node --version` output: v16.17.0

### Additional Context
already reported to ionic framework, but seems to be related to capacitor
https://github.com/ionic-team/ionic-framework/issues/26887

natalya-semenova:
We experience similar issue:

```
Installed Dependencies:

  @capacitor/android: 4.7.0
  @capacitor/core: 4.7.0
  @capacitor/cli: 4.7.0
  @capacitor/ios: 4.7.0
```

<img width="1700" alt="image" src="https://user-images.githubusercontent.com/20386781/223215472-40bcf8ac-eb9e-4f6a-8b20-d0546aa9bd89.png">

jcesarmobile:
can you report a new issue an provide a sample app that reproduces it?

wendt88:
@jcesarmobile why a new issue with the same content?
the sample app could be found here: https://github.com/wendt88/ionic-iframe-data-uri

> We experience similar issue:

but this isn't a data uri inside your code.
maybe, because localhost isn't acessible from the device?

jcesarmobile:
it's a similar but different issue, please, report a new issue

PR #6418, opened 2023-03-21
fix(android): Allow WebView to load data urls

jcesarmobile:
closes https://github.com/ionic-team/capacitor/issues/6361


markemer (approved):
Love it when it's simple like that.

giralte-ionic (approved):

PR #6426, opened 2023-03-22
chore: remove Podfile.lock from iOS Template and remove deletion code

markemer:


jcesarmobile:
If Podfile.lock is not going to be ignored, the CLI should also remove the code where it deletes the file before doing `pod install`.

markemer:
> If Podfile.lock is not going to be ignored, the CLI should also remove the code where it deletes the file before doing `pod install`.

Should I remove the code that does it, do you think, or should I add a `--no-deployment` option instead of `--deployment` and invert the logic?

giralte-ionic (approved):

jcesarmobile (approved):

PR #6433, opened 2023-03-24
feat: Upgrade to Typescript 5.x

theproducer:
This PR updates Capacitor to use TS 5.x


markemer (approved):
Works for me, once you get it past lint, I think we're good to go.

giralte-ionic (approved):

PR #5301, opened 2021-12-09
Fix redirects on Android

tony-scio:
Based on https://github.com/ionic-team/capacitor/pull/4437

PR #6449, opened 2023-03-29
fix(android): Appropriately intercept 302 redirect on Android

TomCarey:
The capacitor webview might be sent to an authorization flow and then travel through multiple redirects back to capacitor.

This change addresses an issue where redirects wouldn’t get intercepted by capacitor’s web server as [WebViewClient.shouldInterceptRequest](https://developer.android.com/reference/android/webkit/WebViewClient#shouldInterceptRequest(android.webkit.WebView,%20android.webkit.WebResourceRequest)) is only called for the initial web request and not subsequent redirects. This meant that we could be redirected back to our application without capacitor loaded.

On iOS it appears that redirects are already properly intercepted by capacitor so this fix attempts to get both platforms to behave consistently.

### Caveat
This is behaviour only works on Android SDK 24+ as inspecting that a request is the result of a redirect is only possible from 24 onward.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
@@ -25,7 +25,12 @@ public class BridgeWebViewClient extends WebViewClient {
     @Override
     public boolean shouldOverrideUrlLoading(WebView view, WebResourceRequest request) {
         Uri url = request.getUrl();
-        return bridge.launchIntent(url);
+        boolean shouldOverride = bridge.launchIntent(url);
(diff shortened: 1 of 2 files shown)


lubomirblazekcz:
This fixes my issues with 302 redirects, but other redirects are not working with this fix. Maybe the condition could be changed like this? 

Inspired by fix in https://github.com/ionic-team/capacitor/pull/5301/files, previously I used this fix in my app, but that caused double request on server for me. This fix works better, gj 👍 

```java
  int status = conn.getResponseCode();

  if (
      status == HttpURLConnection.HTTP_MOVED_TEMP ||
      status == HttpURLConnection.HTTP_MOVED_PERM ||
      status == HttpURLConnection.HTTP_SEE_OTHER ||
      status == 307
  )
```

PR #6451, opened 2023-03-30
fix(android): proper app url check for launching intents

jcesarmobile:


veneliniliev:
@jcesarmobile  this breaks opening links with "tel:"

PR #6453, opened 2023-03-30
feat(cli): Add missing Cap 4 to Cap 5 migration tasks

markemer:


markemer:
> Since the code for migrating from capacitor 3 to 4 has been deleted (it also worked from 2 to 4), I think we should detect the current capacitor version (@capacitor/core is probably the best package to check), and if it’s not 4.x, show some warning before starting the migration recommending to use capacitor 4 migrator first, or even cancel the migration.

Done.

cli/src/tasks/migrate.ts
```diff
@@ -46,15 +52,19 @@ const plugins = [
...
-const coreVersion = '^5.0.0';
-const pluginVersion = '^5.0.0';
+const coreVersion = 'next'; // TODO: Update when Capacitor 5 releases
+const pluginVersion = 'next'; // TODO: Update when Capacitor 5 releases
 const gradleVersion = '7.5';
 
 export async function migrateCommand(config: Config): Promise<void> {
   if (config === null) {
     fatal('Config data missing');
   }
```
jcesarmobile, line 63: We have a method for checking core version already called `getCoreVersion`.
The error should be more helpful, telling them to use capacitor CLI 4 to migrate if they are in version 2 or 3.
Also, we should probably allow to run the command even if it’s in version 5 already as last year we broke a few things in the migrate and rerunning it after the fix could fix the problems it caused in previous versions

jcesarmobile (changes requested):
If I answer no to `Run Dependency Install?`, it still ask me which package manager to use. I think it doesn't install, but shouldn't ask.

The 1024x1024 image is not guaranteed to be called `AppIcon-512@2x.png`, the code should search for the 1024x1024 file and get the name instead of hardcoding it.
If it's named in a different way, the migrator still says the icon was updated, but it wasn't, and leaves the `AppIcon.appiconset/Contents.json` in an inconsistent state.

theproducer (approved):

giralte-ionic (approved):

Issue #6483, opened 2023-04-10
tel: links throw exception


jonathan-k4:
https://github.com/ionic-team/capacitor/blob/9447fc3f8a0c7022adf060623df9113eb1fc6548/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L386

When opening a `tel:` link, an exception is thrown. `url.getHost()` can return null if no host is found, and `.equals()` is being called without a check for null.

media4learning:
To add to this - for me, any link like window.location = crashes the app on android.

veneliniliev:
the same thing happens to me!

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

media4learning:
Repoduction is here:

https://github.com/media4learning/broken-links.git

npm i
npm run build
npx cap sync
npx cap open android
run on a device
Tap the "crash app(Android only) button
App crashes

The problem is using window.location = "X" with X being any mailto, tel, posssibly any link

Android Studio run output displays:

E/AndroidRuntime: FATAL EXCEPTION: main
    Process: io.ionic.starter, PID: 25353
    java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.String.equals(java.lang.Object)' on a null object reference
        at com.getcapacitor.Bridge.launchIntent(Bridge.java:374)
        at com.getcapacitor.BridgeWebViewClient.shouldOverrideUrlLoading(BridgeWebViewClient.java:28)
        at org.chromium.android_webview.AwContentsClientBridge.shouldOverrideUrlLoading(chromium-TrichromeWebViewGoogle.aab-stable-556311630:89)
        at android.os.MessageQueue.nativePollOnce(Native Method)
        at android.os.MessageQueue.next(MessageQueue.java:335)
        at android.os.Looper.loopOnce(Looper.java:186)
        at android.os.Looper.loop(Looper.java:313)
        at android.app.ActivityThread.main(ActivityThread.java:8751)
        at java.lang.reflect.Method.invoke(Native Method)
        at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
[...]

PR #6489, opened 2023-04-11
fix(android): launching intents without host

jcesarmobile:
`url` can be a `tel:` or `mailto:` or other app urls that will not include a host, which will make the app to crash because of using `equals` on a null object.
Reversing the check order, since `appUri` will always have a host, it will not crash.

closes https://github.com/ionic-team/capacitor/issues/6483


giralte-ionic (approved):

theproducer:
Looks like something related to Minipass is causing build errors

Issue #6441, opened 2023-03-27
feat: New webkit feature - Inspectable WebKit, inspect production apps


alexcroox:
## Feature Request

![image](https://user-images.githubusercontent.com/1065835/228040803-862a0be6-9178-4cc7-871a-ffc3f8ffc467.png)

https://webkit.org/blog/13936/enabling-the-inspection-of-web-content-in-apps/

thmclellan:
Thanks @alexcroox for sharing this. It seems like a breaking change from iOS 16.3, which allowed for in-app webview contents to be inspected by default while running from Xcode.

For anyone who just upgraded to iOS 16.4 and lost their Safari > Develop > iPhone options, here is a temporary workaround that you can add to the end of the `func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?)` function in the main `AppDelegate` class in Xcode. 

```
// When running from Xcode in debug mode, enable webView inspection 5 seconds after startup
#if DEBUG
  if #available(macOS 13.3, iOS 16.4, tvOS 16.4, *) {
        DispatchQueue.main.asyncAfter(deadline: .now() + 5.0) {
              if let vc = self.window?.rootViewController as? CAPBridgeViewController {
                  vc.bridge?.webView?.isInspectable = true;
              }
        }
  }
#endif
 ```

Not elegant at all with the 5 second delay but hopefully it helps someone waiting on a future update or plugin. Tested on Capacitor 4.4.0 with Xcode 14.3 RC2 with ionic serve project setup. I found Safari Technology Preview (16.4) was required to get console logging; Safari 16.3 connected but didn't actually show logs.

**Comment by qliqdev** [comment]
Also it will be correct if you add smth like

```
ios: {
      webview: {
        configuration: {
          inspectable: true
        }
      }
    }
```

in capacitor.config.ts.

@markemer pls take a look

markemer:
Yeah - we're looking at figuring out how to not require 16.4 and build this in as well.

markemer:
> Also it will be correct if you add smth like

Thanks, I will.

timgavin:
@qliqdev how are you adding that to your config? This doesn't work for me with Simulator 14.3

```typescript
import { CapacitorConfig } from '@capacitor/cli';

const config: CapacitorConfig = {
  appId: 'com.my.app',
  appName: 'App',
  webDir: 'www',
  bundledWebRuntime: false,
[...]

PR #6495, opened 2023-04-11
feat(ios): add webContentsDebuggingEnabled configuration

markemer:
allow WKWebViews to be inspectable in 16.4 without using the 16.4 SDK with Xcode 14.3

Backported to 4.x in https://github.com/ionic-team/capacitor/pull/6500

### DIFF
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -453,6 +453,16 @@ export interface CapacitorConfig {
      * @default true
      */
     handleApplicationNotifications?: boolean;
+
+    /**
+     * On iOS 16.4 and greater, always enable debuggable web content.
+     *
+     * This is automatically enabled during development.
+     * 
+     * @since 5.0.0
+     * @default false
+     */
+    webContentsDebuggingEnabled?: boolean;
   };
 
   server?: {
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -457,7 +457,7 @@ export interface CapacitorConfig {
     /**
      * On iOS 16.4 and greater, enable debuggable web content for release builds.
      *
-     * This is always `true` for development builds.
+     * If not set, it's `true` for development builds.
      *
      * @since 4.8.0
      * @default false
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -455,7 +455,8 @@ export interface CapacitorConfig {
     handleApplicationNotifications?: boolean;
 
     /**
-     * On iOS 16.4 and greater, enable debuggable web content for release builds.
+     * On iOS 16.4 and greater, enable debuggable web content for release builds
+     * Requires Xcode 14.3
      *
      * If not set, it's `true` for development builds.
      *
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -459,7 +459,7 @@ export interface CapacitorConfig {
      *
      * This is always `true` for development builds.
      *
-     * @since 5.0.0
+     * @since 4.8.0
      * @default false
      */
     webContentsDebuggingEnabled?: boolean;
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -455,9 +455,9 @@ export interface CapacitorConfig {
     handleApplicationNotifications?: boolean;
 
     /**
-     * On iOS 16.4 and greater, always enable debuggable web content.
+     * On iOS 16.4 and greater, enable debuggable web content for release builds.
      *
-     * This is automatically enabled during development.
+     * This is always `true` for development builds.
      *
      * @since 5.0.0
      * @default false
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -455,8 +455,7 @@ export interface CapacitorConfig {
     handleApplicationNotifications?: boolean;
 
     /**
-     * On iOS 16.4 and greater, enable debuggable web content for release builds
-     * Requires Xcode 14.3
+     * Using Xcode 14.3, on iOS 16.4 and greater, enable debuggable web content for release builds.
      *
      * If not set, it's `true` for development builds.
      *
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -458,7 +458,7 @@ export interface CapacitorConfig {
      * On iOS 16.4 and greater, always enable debuggable web content.
      *
      * This is automatically enabled during development.
-     * 
+     *
      * @since 5.0.0
      * @default false
      */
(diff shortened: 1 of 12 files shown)


antsav:
👍 very much needed fix!

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -209,6 +209,10 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
         observers.append(NotificationCenter.default.addObserver(forName: type(of: self).tmpVCAppeared.name, object: .none, queue: .none) { [weak self] _ in
             self?.tmpWindow = nil
         })
+
+        #if DEBUG
+        self.webView?.setInspectableIfRequired(true)
```
Steven0351, line 214: A super minor nit/question: Are we planning to make this a configuration option or have cases where this would be set to false later? Only having the one call site and passing a `true` literal makes me wonder if it should just be in the method itself instead of passed as an argument.

jcesarmobile (changes requested):
We should make this configurable as the feature request proposed on https://github.com/ionic-team/capacitor/issues/6441

Probably using `ios.webContentsDebuggingEnabled` to match Android's `android.webContentsDebuggingEnabled`, and/or have a common `webContentsDebuggingEnabled` for both platforms (that would require changes in Android too)

markemer:
> We should make this configurable as the feature request proposed on #6441
> 
> Probably using `ios.webContentsDebuggingEnabled` to match Android's `android.webContentsDebuggingEnabled`, and/or have a common `webContentsDebuggingEnabled` for both platforms (that would require changes in Android too)

I'm putting this together now, but the only worry I have with that is it this only works on >16.4 - so it'd be a no-op option on other versions.

jcesarmobile:
Yeah, it’s a new feature on iOS 16.4, so we document that it’s only available for iOS 16.4 and newer.

qliqdev:
Guys are you going to migrate these changes for Capacitor v 4.7.X also? Cuz we are not currently planning to upgrade to Capacitor v5, because of Capacitor 5 requires ios 14+, we need old der devices too

markemer:
> Guys are you going to migrate these changes for Capacitor v 4.7.X also? Cuz we are not currently planning to upgrade to Capacitor v5, because of Capacitor 5 requires ios 14+, we need old der devices too

We'd love to, but can't. None of this will work for iOS <16.4 - In iOS <16.4 the views will just be inspectable in debug mode, and can't be set to be inspectable in release mode.

markemer:
> Guys are you going to migrate these changes for Capacitor v 4.7.X also? Cuz we are not currently planning to upgrade to Capacitor v5, because of Capacitor 5 requires ios 14+, we need old der devices too

@qliqdev deleted my previous comment, since I realized that you could build with Xcode 14.3 possibly with Cap4. I'll bring this code over.

cli/src/declarations.ts
```diff
@@ -453,6 +453,16 @@ export interface CapacitorConfig {
      * @default true
      */
     handleApplicationNotifications?: boolean;
+
+    /**
+     * On iOS 16.4 and greater, always enable debuggable web content.
+     *
+     * This is automatically enabled during development.
```
dallastjames, line 460: Can we change this phrasing just a bit to say: 

```
* On iOS 16.4 and greater, enable debuggable web content for release builds.
*
* This is always `true` for development builds.
```

Just to help clarify that this flag ONLY affects release builds and that inspectable views are always true for non-release builds.

dallastjames (changes requested):

markemer:
@qliqdev The 4.x Cap update is here: https://github.com/ionic-team/capacitor/pull/6500

dallastjames (approved):

Steven0351:
> Guys are you going to migrate these changes for Capacitor v 4.7.X also? Cuz we are not currently planning to upgrade to Capacitor v5, because of Capacitor 5 requires ios 14+, we need old der devices too

Capacitor 5 does not require iOS 14+

ios/Capacitor/Capacitor/CAPInstanceDescriptor.swift
```diff
@@ -134,6 +134,9 @@ internal extension InstanceDescriptor {
             if let handleNotifications = config[keyPath: "ios.handleApplicationNotifications"] as? Bool {
                 handleApplicationNotifications = handleNotifications
             }
+            if let webContentsDebuggingEnabled = config[keyPath: "ios.webContentsDebuggingEnabled"] as? Bool {
+                isWebDebuggable = webContentsDebuggingEnabled
+            }
```
jcesarmobile, line 139:
```suggestion
            } else {
                #if DEBUG
                    isWebDebuggable = true
                #endif
            }
```
set the default to true for debug builds if `webContentsDebuggingEnabled` is not set

cli/src/declarations.ts
```diff
@@ -453,6 +453,16 @@ export interface CapacitorConfig {
      * @default true
      */
     handleApplicationNotifications?: boolean;
+
+    /**
+     * On iOS 16.4 and greater, enable debuggable web content for release builds.
+     *
+     * This is always `true` for development builds.
+     *
+     * @since 5.0.0
```
jcesarmobile, line 462:
```suggestion
     * @since 4.8.0
```

Since it's going to be released in 4.8.0 too, the since should be 4.8.0 as that's going to be the first version with the feature

jcesarmobile (changes requested):
Requested some changes so it behaves as on Android and to fix a problem where it would enable debugging even for production builds no matter what value the config option had.

ios/Capacitor/Capacitor/WKWebView+Capacitor.swift
```diff
@@ -111,4 +111,12 @@ internal extension WKWebView {
             objc_setAssociatedObject(self, &associatedKeyboardFlagHandle, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
         }
     }
+
+    // TODO: Remove this after Xcode 14.3 is required
+    // This is in here because iOS 16.4 needs this - but will fail to compile under Xcode < 14.3
+    func setInspectableIfRequired(_ shouldInspect: Bool) {
+        if responds(to: Selector(("setInspectable:"))) {
```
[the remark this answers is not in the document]
markemer: That's strange - I'll investigate why, I'd like to get to the bottom of that, but this should work.

cli/src/declarations.ts
```diff
@@ -453,6 +453,16 @@ export interface CapacitorConfig {
      * @default true
      */
     handleApplicationNotifications?: boolean;
+
+    /**
+     * On iOS 16.4 and greater, enable debuggable web content for release builds.
```
jcesarmobile, line 458: maybe also mention that it requires Xcode 14.3?

markemer: Yep - I'll add that now.

jcesarmobile (approved):

jcesarmobile (approved):

Steven0351 (approved):

PR #6497, opened 2023-04-12
feat(android): update gradle to 8.0.2 and gradle plugin to 8.0.0

IT-MikeS:
Dev build for this PR is tagged as: 

~~`@capacitor/{core,cli,android,ios}@5.0.0-dev-20230412T155745.0`~~

~~`@capacitor/{core,cli,android,ios}@5.0.0-dev-20230414T153043.0`~~

`@capacitor/{core,cli,android,ios}@5.0.0-dev-20230414T155819.0`


markemer:
We'll have to make sure to update the migrate script if and when we go with this.

jcesarmobile (changes requested):
you also need to update the wrapper files by running `./gradlew wrapper --distribution-type all --gradle-version 8.0.2 --warning-mode all`, but in the android-template it can by tricky since it's missing files

markemer:
We'll also need to either update the migrate script here or in another PR.

.github/workflows/ci.yml
```diff
@@ -137,6 +137,11 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: 18.x
+      - name: set up JDK 11
```
jcesarmobile, line 140:
```suggestion
      - name: set up JDK 17
```

markemer:
> Yeah, this pr is only for the template, we need another one updating capacitor android and another updating the migrate command, and the plugins too

Not to be that guy asking for homework, but should we add subtasks or a bunch of tickets for this stuff?

jcesarmobile (approved):

markemer (approved):

PR #6500, opened 2023-04-12
feat(ios): add webContentsDebuggingEnabled configuration

markemer:
allow WKWebViews to be inspectable in 16.4 without using the 16.4 SDK with Xcode 14.3

### DIFF
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -438,7 +438,8 @@ export interface CapacitorConfig {
     handleApplicationNotifications?: boolean;
 
     /**
-     * On iOS 16.4 and greater, enable debuggable web content for release builds.
+     * On iOS 16.4 and greater, enable debuggable web content for release builds
+     * Requires Xcode 14.3
      *
      * If not set, it's `true` for development builds.
      *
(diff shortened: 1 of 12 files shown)


giralte-ionic (approved):

cli/src/declarations.ts
```diff
@@ -436,6 +436,16 @@ export interface CapacitorConfig {
      * @default true
      */
     handleApplicationNotifications?: boolean;
+
+    /**
+     * On iOS 16.4 and greater, enable debuggable web content for release builds.
+     *
+     * This is always `true` for development builds.
+     *
+     * @since 4.7.4
```
jcesarmobile, line 445:
```suggestion
     * @since 4.8.0
```
This is a feature, so the next version would be 4.8.0, not 4.7.4

Also the other changes requested on https://github.com/ionic-team/capacitor/pull/6495 should also be addressed here

jcesarmobile (changes requested):
this is missing the xcode 14.3 part

jcesarmobile (approved):

IT-MikeS (approved):

PR #6524, opened 2023-04-20
fix(cli): Pin @ionic/utils-subprocess version for node 12 support

jcesarmobile:
Latest release of @ionic/utils-subprocess doesn't work on node 12.
While docs say that Capacitor 4 supports node 14 and newer, on code it requires node 12.4 or newer, so I think it should still work there.
Or if not, we should change the node check to require node 14 so the users see an error about the node version being unsupported rather than a weird message

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -46,7 +46,7 @@
   "dependencies": {
     "@ionic/cli-framework-output": "^2.2.5",
     "@ionic/utils-fs": "^3.1.6",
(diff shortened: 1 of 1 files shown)


markemer (approved):

IT-MikeS:
I think we should also edit the 4.x docs to say node 12 as this is the least disruptive option so we don't cause errors for those using 4.x on 12.

IT-MikeS (approved):

Issue #4089, opened 2021-01-20
bug: CORS error when loading local files in iOS with live reload


FutureArchitect-takeda1874:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 2.4.6
  @capacitor/core: 2.4.6
  @capacitor/android: 2.4.6
  @capacitor/electron: 2.4.6
  @capacitor/ios: 2.4.6

Installed Dependencies:

  @capacitor/cli 2.4.4
  @capacitor/android 2.4.4
  @capacitor/ios 2.4.4
  @capacitor/core 2.4.4
  @capacitor/electron not installed

[success] Android looking great! 👌
  Found 1 Capacitor plugin for ios:
    cordova-plugin-file (6.0.2)
[success] iOS looking great! 👌
```

### Platform(s)

- [ ] Android
- [x] iOS
- [ ] Electron
- [ ] Web

### Current Behavior

The local file converted to Web View-friendly path (like capacitor://localhost/_capacitor_file_/...) cannot be fetched because of CORS error when running app in iOS with live reload.
It seems that CORS error is occurred because the source origin is `http://xxx.xxx.xx.xx:8100` in default with live reload.

### Expected Behavior

I can fetch my local file in OS with live reload.

### Code Reproduction

```
// capacitor.config.json
{
  "server": {
    "url": "http://192.168.11.21:8100",
    "cleartext": true
  }
}

```

```
const path = "captured-photos/66026da3-5098-478a-9794-be8926df96e5"
const { uri } = await Filesystem.stat({ directory, path })
const webURI = window.Ionic.WebView.convertFileSrc(uri)

// throws CORS error!
const response = await fetch(webURI)
```

### Other Technical Details

`npm --version` output: 6.14.10

`node --version` output: v14.15.1

`pod --version` output (iOS issues only): 1.10.0

#### Related issues

* https://github.com/ionic-team/capacitor/issues/2042
* https://github.com/ionic-team/capacitor/issues/2439

elvisgraho:
Related - #https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/587

elvisgraho:
[Will this fix the issue?](https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/587#issuecomment-791635030)

m8xp0w3r:
[...]

PR #6539, opened 2023-04-26
fix(ios): set cors headers in asset handler for live reload

ItsChaceD:
Addresses: #4089

This PR sets CORS headers so local files can be fetched on iOS while using live reload.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
+++ b/ios/Capacitor/Capacitor/CAPBridgeViewController.swift
@@ -42,7 +42,7 @@ import Cordova
         setScreenOrientationDefaults()
 
         // get the web view
-        let assetHandler = WebViewAssetHandler(router: router())
+        let assetHandler = WebViewAssetHandler(router: router(), serverURL: configuration.serverURL.absoluteString)
         assetHandler.setAssetPath(configuration.appLocation.path)
         let delegationHandler = WebViewDelegationHandler()
         prepareWebView(with: configuration, assetHandler: assetHandler, delegationHandler: delegationHandler)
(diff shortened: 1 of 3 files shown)


jcesarmobile (changes requested):
The CORS header should only be added when using live reload, this code also adds it when not using live reload.

I don't think we have a method for checking if live reload is being used as `serverURL` is being set to `localURL` as default value if not set, so you'll have to compare if `localURL` and `serverURL` are different and only set it on the `WebViewAssetHandler` in that case. 
Which also means, `serverURL` in `WebViewAssetHandler` should be optional and `WebViewAssetHandler` would not change, but a new method for setting the `serverURL` should be added.

IT-MikeS (approved):

jcesarmobile (changes requested):
The headers are still being added for all requests, not just for live reload requests.

The server url is `capacitor://localhost` and the file url is `capacitor://localhost/_capacitor_file_/.../photo-xx.jpg`, so they are different and goes inside the if.
It should probably compare the schemes since the CORS errors appear when using different schemes. For that, it might be better to store the `serverUrl` as `URL` instead of as `String`

markemer:
> The headers are still being added for all requests, not just for live reload requests.
> 
> The server url is `capacitor://localhost` and the file url is `capacitor://localhost/_capacitor_file_/.../photo-xx.jpg`, so they are different and goes inside the if. It should probably compare the schemes since the CORS errors appear when using different schemes. For that, it might be better to store the `serverUrl` as `URL` instead of as `String`

Since live reload is just a development feature anyway, could we wrap this in an `#if DEBUG`? That'd be safer anyway.

jcesarmobile:
Well, not really, while I talk about "live reload", the CORS headers should also be added when using `server.url` for both live reload and for loading external sites.
While we don't officially recommend doing that, there are a lot of people doing it and the headers should be added for that case too, so shouldn't be done for debug builds only.

jcesarmobile (approved):

Issue #6126, opened 2022-11-29
bug: CapacitorHttp - Requesting a Blob fails on native iOS and Android


KevinKelchen:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/ios: 4.5.0

Installed Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/ios: 4.5.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

Native iOS
Native Android

### Current Behavior

Hello there! 👋

Using the Angular `HttpClient`, which uses `CapacitorHttp` at a lower level, making a request to retrieve a Blob on native iOS and Android fails. Native iOS _without_ `CapacitorHttp`, Native Android _without_ `CapacitorHttp`, and Web work fine.

In our production app, we use a POST request to retrieve a Blob, but using a GET seems to fail as well and was easier to use in a repro.

### Expected Behavior

Native iOS and native Android should successfully retrieve the requested Blob.

Also, if we want `CapacitorHttp` to work the same as XHR/fetch, this would be a difference.

### Code Reproduction

https://github.com/KevinKelchen/capacitor-http-request-blob-issue#steps-to-reproduce

### Other Technical Details

`npm --version` output: 8.15.0

`node --version` output: v16.17.0

`pod --version` output (iOS issues only): 1.11.3

### Additional Context

Thanks so much! 😀

Kevin

silviogutierrez:
I can confirm this is still an issue with Capacitor 4.6.1

silviogutierrez:
@KevinKelchen : on further investigation, it's less about `blob()` and more about requesting anything that isn't a text/utf8 type file. 

So if I fetch an JSON endpoint and call blob, it works. But if I fetch an image, to save it locally, and call blob (or text() or any other method) it fails.

beonde:
Phew... I thought I was the only one going crazy with this bug! Can also confirm in capacitor/core 4.6.1.

I have even tried 'arraybuffer' for output and converting, but no dice.

markabrahams:
[...]

Issue #6123, opened 2022-11-29
bug: CapacitorHttp - Loading a WASM file fails on native iOS and Android


KevinKelchen:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/ios: 4.5.0

Installed Dependencies:

  @capacitor/cli: 4.5.0
  @capacitor/android: 4.5.0
  @capacitor/core: 4.5.0
  @capacitor/ios: 4.5.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

Native iOS
Native Android

### Current Behavior

On native iOS and Android, there appears to be an issue with loading a WASM file via `fetch` (which uses `CapacitorHttp` at a lower level) and calling `WebAssembly.instantiateStreaming()` on it. The repro uses the [esri ArcGIS API for JavaScript v3.x](https://developers.arcgis.com/javascript/3/) which loads a WASM file. I don't know much about using WASM, so I did not tear down to a more minimal repro.

Native iOS _without_ `CapacitorHttp`, native Android _without_ `CapacitorHttp`, and Web work fine.

### Expected Behavior

Native iOS and Android should work like the Web's `fetch` when loading a WASM file in this way.

### Code Reproduction

https://github.com/KevinKelchen/capacitor-http-wasm-issue#steps-to-reproduce

### Other Technical Details

`npm --version` output: 8.15.0

`node --version` output: v16.17.0

`pod --version` output (iOS issues only): 1.11.3

PR #6540, opened 2023-04-26
fix(http): media requests route to custom handlers

ItsChaceD:
Addresses: #6123 and #6126

Test app: https://github.com/ItsChaceD/capacitor-http-wasm-issue

This PR routes media/non-text requests (files, blobs, WASM, etc.) to custom native protocol handlers so the responses don't have to be encoded/decoded through the bridge.

On iOS: we change the protocol from `https` to `capacitor` and it is handled by the URL scheme handler
On Android: we take a different approach to avoid cleartext traffic issues. Instead we prepend a Capacitor Media string to the URL so that `shouldInterceptRequest` can detect that the request should be intercepted. We cannot do this on iOS since it is not allowed to intercept requests on the `https` protocol.

### DIFF
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -348,26 +348,14 @@ var nativeBridge = (function (exports) {
                 };
                 // media types that we want to intercept and route to our custom protocol handlers
                 const fileExtensions = [
-                    'pdf',
-                    'jpg',
-                    'jpeg',
-                    'png',
-                    'gif',
-                    'bmp',
-                    'svg',
-                    'wasm',
+                    'pdf', 'jpg', 'jpeg', 'png', 'gif', 'bmp', 'svg', 'wasm', 'webm', 'mp4', 'm4a', 'mp3', 'wav', 'ogg', 'oga',
+                    'opus', 'webp', 'flac', 'x-flac', 'aac', '3gp', '3gpp', 'm3u8', 'ts', 'm4v', 'f4v', 'flv', 'mov', 'avi', 'mkv'
                 ];
                 const mediaContentTypes = [
-                    'application/pdf',
-                    'application/octet-stream',
-                    'application/wasm',
-                    'image/jpeg',
-                    'image/png',
-                    'image/gif',
-                    'video/mp4',
-                    'video/webm',
-                    'audio/mpeg',
-                    'audio/wav',
+                    'application/pdf', 'application/octet-stream', 'application/wasm',
+                    'image/jpeg', 'image/png', 'image/gif', 'image/bmp', 'image/svg+xml', 'image/webp',
+                    'video/mp4', 'video/webm', 'video/3gpp', 'video/3gpp2', 'video/ogg', 'video/x-matroska', 'video/quicktime',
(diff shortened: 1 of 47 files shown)


android/capacitor/src/main/assets/native-bridge.js
```diff
@@ -147,6 +148,7 @@ var nativeBridge = (function (exports) {
             if (doc) {
                 const docAddEventListener = doc.addEventListener;
                 doc.addEventListener = (...args) => {
+                    var _a;
```
jrobeson, line 151: shouldn't this (and others) be a let?

markemer: This is a generated file - I'm not certain where this comes from.

markemer:
Looks like lint is failing, BTW.

markemer:
Core tests are failing in `native-bridge.ts` not 100% why.

jcesarmobile:
Why is this change applied for media files only? I think the problem is present for all file types, not just for media files, so the current change will only partly fix the issue.

Also, why use a second scheme handler? We used to have multiple and we ended using only one because it had some problems (which I don't remember at the moment).
While it's true that we can't intercept http/https calls, we can still use the existing scheme handler turning the urls into something like
```
if (platform === 'ios') {
    url.protocol = 'capacitor:';
} 
url.pathname += '/_capacitor_media_'; // do this for both platforms 
```
And then in the existing handler check for the `_capacitor_media_` for needed customizations.
Also, I think the `/_capacitor_media_` should be prepended to the pathname, not appended.

ItsChaceD:
> @jcesarmobile "Why is this change applied for media files only? I think the problem is present for all file types, not just for media files, so the current change will only partly fix the issue."

This change is for any MIME type that is not text/JSON. 
```
const mediaContentTypes = [
        'application/*',
        'audio/*',
        'image/*',
        'model/*',
        'video/*',
      ];
      
const textContentTypes = [
        'text/*',
        'application/json',
        'application/x-www-form-urlencoded',
        'multipart/form-data',
      ];
      
 if(isMediaType && !isTextType) ...
```
      
So this includes all other content types (Files, Blobs, etc.). The tricky part is that there is no way to guarantee what response type we are expecting before the request is actually executed. So to account for this, we check if 1 of these conditions are met:
1. Is `X-Capacitor-Force-Media-Request` set on our Request Headers. (*Just Added + Needs Documentation*).
2. Is the `Content-Type` set on our Request Headers & is it in `mediaContentTypes` but not in `textContentTypes`
3. Is the file extension on the URL in `fileExtensions`
4. Is our `responseType` in `responseTypes`  (*ONLY FOR XHR*)

jcesarmobile:
But why not do it for text files too? I think we should do it for all regular GET requests (not for POST because in Android it's not possible to get the body), so exclude only `application/x-www-form-urlencoded` and `multipart/form-data`, but not text/JSON files. So we can remove the whole mime type checking.

We have another issue about a XHR call returning an Object for a JSON file instead of string and using the approach in this PR (removing the string type check) I get the string properly instead of the Object.

I don't think adding the CORS headers are not needed anymore since we are using capacitor:// scheme on all requests now, it's hard to tell with the provided app, but looks like it's working fine if I remove the code.

markemer:
Shouldn't `'text/*'` cover `text/plain` for text files?

PR #6563, opened 2023-05-04
chore: update workflow to use latest-4 for 4.x

IT-MikeS:


jcesarmobile (commented):
@ltm, should we do the `--provenance` thing here too or for 4.x is not necessary?

giralte-ionic (approved):

giralte-ionic (approved):

IT-MikeS:
> Right, it would probably be a good idea to enable package provenance for 4.x as well. I believe you'll have to backport #6542 to the 4.x branch.

Added the changes here for the workflows

markemer (approved):

PR #6574, opened 2023-05-07
java version fix

qliqdev:
`npx cap migrate` breaks with `java --version`


jcesarmobile:
Thanks for the PR, but on java 9 and newer using `-version` instead of `--version` shows the version in the error output (stderr) instead of doing it in the regular output (stdout), so the code for getting the version doesn't get anything and incorrectly warns about the version being unsupported even if in java 17. So the change fix the problem in java 8, but breaks the check for every other java version.

PR #6581, opened 2023-05-09
fix: fallback to plain `pod` if `Gemfile` does not contain CocoaPods

markemer:


cli/src/ios/update.ts
```diff
@@ -108,11 +108,19 @@ async function updatePodfile(
...
-      );
+      try {
+        await runCommand(
+          'bundle',
+          ['exec', 'pod', 'install', ...(deployment ? ['--deployment'] : [])],
+          { cwd: config.ios.nativeProjectDirAbs },
+        );
+      } catch {
+        await runCommand(
+          podPath,
```
jcesarmobile, line 119:
```suggestion
          'pod',
```
I guess this needs to be hardcoded to `pod`, because `podPath` is `bundle exec pod` if `useBundler` is true.

IT-MikeS (approved):

giralte-ionic (approved):

PR #6583, opened 2023-05-10
fix(cli): Don't remove jetifier if some plugin needs it

jcesarmobile:
So far it only checks for `@ionic-enterprise/identity-vault`, but made it an array so we can easily add more plugins if needed.

Migrate won't remove the jetifier lines if the plugin is present, update will also remove or add the lines if needed if the plugin gets uninstalled or installed.

### DIFF
--- a/cli/src/android/update.ts
+++ b/cli/src/android/update.ts
@@ -31,7 +31,10 @@ import {
 } from '../plugin';
 import type { Plugin } from '../plugin';
 import { copy as copyTask } from '../tasks/copy';
-import { patchOldCapacitorPlugins } from '../tasks/migrate';
+import {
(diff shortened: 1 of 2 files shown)


IT-MikeS (approved):

jcesarmobile:
Now thinking that maybe we shouldn’t make the edits in update command as if the users did a manual migration and left the jetifier line on purpose because they knew they had a plugin that needed it, we would be deleting it on every update.
Thoughts?

jcesarmobile:
changed to only warn if it's missing and a plugin needs it

IT-MikeS (approved):

jcesarmobile:
going to fix it in the plugin instead as doesn't seem breaking

Issue #6588, opened 2023-05-12
bug: Android build graddle issues when Going from 5.0.1 to 5.0.2


RobSchilderr:
### Capacitor Version

 
Latest Dependencies:

  @capacitor/cli: 5.0.3
  @capacitor/core: 5.0.3
  @capacitor/android: 5.0.3
  @capacitor/ios: 5.0.3

Installed Dependencies:

  @capacitor/cli: 5.0.3
  @capacitor/core: 5.0.3
  @capacitor/android: 5.0.3
  @capacitor/ios: 5.0.3```

### Platform(s)
 
Android
 

### Current Behavior

Description
After upgrading from version 5.0.1 to 5.0.2, an error occurs during the build process in the monorepo setup. The issue does not occur in version 5.0.1.

1. Steps to reproduce (5.0.2)
2. Fork this repo: https://github.com/RobSchilderr/cap-monorepo-bug/tree/main
3. Run yarn install
4. Navigate to the next-app directory, and run yarn build|:
5. The following error occurs:

```
[error] Unable to find
        /Users/robschilder/Documents/Development/cap-monorepo-issue/apps/next-app/node_modules/@capacitor/app/android/build.gradle.
        Try updating it manually
```

### Expected Behavior

Steps to reproduce (5.0.1) expected:

1. Execute npx create-supertokens-app@latest --manager=yarn
2. Choose capacitor when prompted
3. Run yarn install
4. Navigate to the next-app directory, and run yarn build
5. The build succeeds without any issues 🎉

### Code Reproduction

See above: https://github.com/RobSchilderr/cap-monorepo-bug/tree/main

### Other Technical Details

`npm --version` output: 8.19.4

`node --version` output: v16.20.0

### Additional Context

ProductName:            macOS (M1 chip)
ProductVersion:         13.3.1
ProductVersionExtra:    (a)
BuildVersion:           22E772610a

PR #6598, opened 2023-05-16
fix(android): migrate command in monorepos

barlock:
Migrating to capacitor 5, I found this migration script not working in a monorepo where the capacitor node_modules were not in the same directory as `rootPath` which is `process.cwd()`. 

Instead, I've changed the module resolution to pull from the module's `rootPath` so it will work however it's installed.

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -842,14 +842,12 @@ export async function patchOldCapacitorPlugins(
     androidPlugins.map(async p => {
       if (p.manifest?.android?.src) {
         const buildGradlePath = resolveNode(
-          config.app.rootDir,
-          p.id,
+          p.rootPath,
           p.manifest.android.src,
           'build.gradle',
         );
         const manifestPath = resolveNode(
-          config.app.rootDir,
-          p.id,
+          p.rootPath,
           p.manifest.android.src,
           'src',
           'main',


jcesarmobile:
I already added [a fix last week](https://github.com/ionic-team/capacitor/pull/6589) (still unreleased) that fixed the monorepo problem o the yarn monorepo app provided on https://github.com/ionic-team/capacitor/issues/6588

chriswep:
i have `5.0.4-nightly-20230512T150632.0` installed and i still see errors on `cap sync` for every plugin:

```
[error] Unable to find
        /Volumes/Users/chris/code/em/app/node_modules/@capacitor/app/android/build.gradle. Try
        updating it manually
[warn] @capacitor/app doesn't officially support Capacitor ^5.0.0 yet, doing our best moving it's package
       to build.gradle so it builds
[error] Unable to find
        /Volumes/Users/chris/code/em/app/node_modules/@capacitor/app/android/src/main/AndroidManifest.xml.
        Try updating it manually
[error] Unable to find
        /Volumes/Users/chris/code/em/app/node_modules/@capacitor/app/android/build.gradle. Try
        updating it manually
[error] Could not read
        /Volumes/Users/chris/code/em/app/node_modules/@capacitor/app/android/src/main/AndroidManifest.xml.
        Check its permissions and if it exists.
```

i'm using yarn workspaces and the plugin packages are in the root `node_modules` folder, not within `app/node_modules`.

PR #6607, opened 2023-05-19
fix(cli): correct migration of package from AndroidManifest.xml to build.gradle

davidnussio:
When AndroidManifest.xml contains other attributes after the package, the regex matches all attributes and writes all of them into build.gradle.

If AndroidManifest.xml contains:
<manifest package="com.example.app" xmlns:android="http://schemas.android.com/apk/res/android">

app/build.gradle will contains xmlns:android property and 'Upgrading gradle wrapper files' fails
 android {
    namespace "com.example.app" xmlns:android="http://schemas.android.com/apk/res/android"

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -690,7 +690,7 @@ async function movePackageFromManifestToBuildGradle(
   }
 
   let packageName: string;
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):
good catch, thanks!

PR #6618, opened 2023-05-27
Update gradlew

hiddenboox:
As I know in Android Studio Electric Eel, the bundled Java was moved to `/Applications/Android Studio.app/Contents/jbr/Contents/Home` from `/Applications/Android Studio.app/Contents/jre/Contents/Home`.

`jbr` -> `jre`

### DIFF
--- a/android/gradlew
+++ b/android/gradlew
@@ -119,9 +119,9 @@ CLASSPATH=$APP_HOME/gradle/wrapper/gradle-wrapper.jar
 
 # Determine the Java command to use to start the JVM.
 if [ -n "$JAVA_HOME" ] ; then
-    if [ -x "$JAVA_HOME/jre/sh/java" ] ; then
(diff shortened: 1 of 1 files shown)


jcesarmobile:
Thanks for the pull requests, but those files are generated by gradle, not by us.

Also I don't think the change is correct, if you use Android Studio's JDK (as I do), your JAVA_HOME should be `/Applications/Android Studio.app/Contents/jbr/Contents/Home/`, notice that it already includes the jbr on the path, so shouldn't need to be changed on the gradlew file.
I'm not bash expert, but what I understand that the current code does is to check if `$JAVA_HOME/jre/sh/java` exists and if it does it uses that path for the java executable, and if it doesn't exist it uses `$JAVA_HOME/bin/java`, which is the path I have in my computer. So changing the code would break gradle for IBM users on AIX (per the code comment)

PR #6627, opened 2023-06-01
fix(http): don't throw errors when content-type is null on response

ItsChaceD:
This PR fixes errors thrown (or status code `0`) when the `Content-Type` is missing from the response headers.


markemer:
I'm checking if this fixes WN-1083 - if it does I'll move that to in review now and then staging after we merge this.

markemer:
I have checked using @eric-horodyski's code, that this does in fact fix WN-1083.

markemer (approved):

IT-MikeS (approved):

PR #6669, opened 2023-06-21
Add support for setting allowBackForwardNavigationGestures on iOS Webview

finnp:
Hey,

I haven't verified that this works yet, but this PR aims to add support for the `allowBackForwardNavigationGestures` setting on an iOS WebView, The settings allows setting it to `true` so the user can navigate back and forth in the application with gestures if so desired (through `ios.allowBackForwardNavigationGestures` setting). The default is `false`.

What do you think?


finnp:
Here's a plugin that solves the same issue: https://github.com/alesmraz/capacitor-plugin-ios-swipe-back. But maybe having this option in core would be nice?

Issue #6640, opened 2023-06-07
bug: Usage of Cookie Plugin causes app crashes on Android


daniel-zero:
### Capacitor Version

```
    "@capacitor/app": "^5.0.2",
    "@capacitor/core": "^5.0.3",
    "@capacitor/device": "^5.0.2",
    "@capacitor/haptics": "5.0.2",
    "@capacitor/keyboard": "5.0.2",
    "@capacitor/status-bar": "5.0.2",
```

### Platform(s)
- Android

### Current Behavior

When calling any capacitor CookiePlugin method the app will crash
```log
 Serious error executing plugin
                                                                                                    java.lang.reflect.InvocationTargetException
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method)
                                                                                                    	at com.getcapacitor.PluginHandle.invoke(PluginHandle.java:138)
                                                                                                    	at com.getcapacitor.Bridge.lambda$callPluginMethod$0(Bridge.java:780)
                                                                                                    	at com.getcapacitor.Bridge.$r8$lambda$ehFTi5f4HhVNFKTbCKAYDkpQYRA(Unknown Source:0)
                                                                                                    	at com.getcapacitor.Bridge$$ExternalSyntheticLambda3.run(Unknown Source:8)
                                                                                                    	at android.os.Handler.handleCallback(Handler.java:942)
                                                                                                    	at android.os.Handler.dispatchMessage(Handler.java:99)
                                                                                                    	at android.os.Looper.loopOnce(Looper.java:201)
                                                                                                    	at android.os.Looper.loop(Looper.java:288)
                                                                                                    	at android.os.HandlerThread.run(HandlerThread.java:67)
                                                                                                    Caused by: java.lang.NullPointerException: Attempt to invoke virtual method 'void com.getcapacitor.plugin.CapacitorCookieManager.removeAllCookies()' on a null object reference
                                                                                                    	at com.getcapacitor.plugin.CapacitorCookies.clearAllCookies(CapacitorCookies.java:150)
                                                                                                    	at java.lang.reflect.Method.invoke(Native Method) 
                                                                                                    	at com.getcapacitor.PluginHandle.invoke(PluginHandle.java:138) 
                                                                                                    	at com.getcapacitor.Bridge.lambda$callPluginMethod$0(Bridge.java:780) 
[...]

PR #6683, opened 2023-06-27
fix(android): init cookie manager on plugin methods for 4.x

ItsChaceD:
This PR fixes an issue where `CapacitorCookies` plugin methods are throwing `NullPointerException` when patching `document.cookie` is disabled.

Addresses: #6640


ItsChaceD:
See https://github.com/ionic-team/capacitor/pull/6684#issuecomment-1609955856 for why this PR is closed.

PR #6684, opened 2023-06-27
fix(android): init cookie manager on plugin methods

ItsChaceD:
This PR fixes an issue where `CapacitorCookies` plugin methods are throwing `NullPointerException` when patching `document.cookie` is disabled.

Addresses: #6640

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ Made possible by the Capacitor community. 💖
 <!-- CONTRIBUTORS:START -->
 
 <p align="center">
+  <a href="https://github.com/yooouuri"><img src="https://github.com/yooouuri.png?size=100" width="50" height="50" /></a>
   <a href="https://github.com/davidnussio"><img src="https://github.com/davidnussio.png?size=100" width="50" height="50" /></a>
   <a href="https://github.com/ltm"><img src="https://github.com/ltm.png?size=100" width="50" height="50" /></a>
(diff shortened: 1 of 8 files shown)


markemer (approved):

jcesarmobile:
I've sent https://github.com/ionic-team/capacitor/pull/6685 so we can move the `CapacitorCookieManager` creation back to `load()` method and we can revert the fix that is causing https://github.com/ionic-team/capacitor/issues/6640 without having to call `initCookieManager` on every plugin call

ItsChaceD:
> I've sent #6685 so we can move the `CapacitorCookieManager` creation back to `load()` method and we can revert the fix that is causing #6640 without having to call `initCookieManager` on every plugin call

Perfect, that change makes more sense. I will review & open PRs to move the initialization back to the `load()` method. The logic on this PR is also faulty since `CookieHandler.setDefault(this.cookieManager);` should only be called when `isEnabled()` is true.

markemer:
> > I've sent #6685 so we can move the `CapacitorCookieManager` creation back to `load()` method and we can revert the fix that is causing #6640 without having to call `initCookieManager` on every plugin call
> 
> Perfect, that change makes more sense. I will review & open PRs to move the initialization back to the `load()` method. The logic on this PR is also faulty since `CookieHandler.setDefault(this.cookieManager);` should only be called when `isEnabled()` is true.

Yeah, thats a good idea. A bit cleaner.

PR #6685, opened 2023-06-27
fix(android): Move bridge localUrl initialization to initWebView

jcesarmobile:
In plugin's `load()` method, `getLocalUrl()` is null because it's being set after the plugin initialization.

I've moved the code related to the localUrl creation to the `initWebView()` so it's available in `load()` method.

Once it's merged we can revert the changes made in Cookies plugin to init the plugin after the local url is set.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -104,7 +104,7 @@ public class Bridge {
     private String appUrlConfig;
     private HostMask appAllowNavigationMask;
     private Set<String> allowedOriginRules = new HashSet<String>();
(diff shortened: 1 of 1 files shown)


giralte-ionic (approved):

ItsChaceD (approved):
LGTM, except the `authorities` variable should be private. We should do the same logic for `4.x` as well

carlpoole:
Testing this with Portals, I think it's okay I just want to be sure

carlpoole (approved):
Tested with Portals and doesn't seem to affect it. Thanks!

jcesarmobile:
> LGTM, except the `authorities` variable should be private. We should do the same logic for `4.x` as well

good catch, I thought about it but then forgot after pasting the line

PR #6707, opened 2023-07-04
fix(cli): Empty fallback if no non-dev dependencies

petermakeswebsites:
Migration was throwing an error because I only had dev dependencies.

```
TypeError: Cannot convert undefined or null to object
    at Function.keys (<anonymous>)
    at installLatestLibs (/Users/pete/Development/GitHub/acimce-kit/node_modules/@capacitor/cli/dist/tasks/migrate.js:257:33)
```


jcesarmobile (approved):
Thanks for the PR, I'll merge to prevent problems, but note that all your capacitor plugins and capacitor packages except for the CLI should be dependencies, not dev dependencies.

Issue #6604, opened 2023-05-17
bug: capacitor errors reading response headers android


tbscode:
### Capacitor Version
```
Latest Dependencies:

  @capacitor/cli: 5.0.3
  @capacitor/core: 5.0.3
  @capacitor/android: 5.0.3
  @capacitor/ios: 5.0.3

Installed Dependencies:

  @capacitor/cli: 5.0.2
  @capacitor/android: 5.0.2
  @capacitor/ios: 5.0.2
  @capacitor/core: 5.0.2
```

### Platform(s)

- android

not tested on ios

### Current Behavior

I am encountering an issue with capacitor-http and the response headers for fetch calls on Android. My fetch calls are working fine in the web app, but they throw a Capacitor error when running on Android.

```
(index):392 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'startsWith')
    at window.fetch ((index):392:80) <------- error in capacitor android
    at async g (index-e974cbdf3f503cbf.js:1:2107)
```

The line causing the error seems to be this one: https://github.com/ionic-team/capacitor/blob/6843d9642fad9a322579cbe5f01563929a83dbf5/android/capacitor/src/main/assets/native-bridge.js#L387

## In the browser

I am using Next.js, React, and Capacitor for the frontend of a cross-platform React app. The backend is a Django server set up with the REST framework.

For the frontend of the browser web application, I can use the following code:

```js
const res = await fetch(`${process.env.NEXT_PUBLIC_BACKEND_URL}/api/user_data`, {
  method: 'POST',
  credentials: 'include',
  headers: {
    'X-CSRFToken': getCookiesAsObject().csrftoken,
    'Content-Type': 'application/json',
    Accept: 'application/json'
  },
  body: JSON.stringify({'some' : 'data'}),
});
```

When I run the web app, I get the expected `200 OK` response with some user data.

These are the server headers:

```
HTTP/1.1 200 OK
date: Wed, 17 May 2023 21:37:55 GMT
server: uvicorn
content-type: application/json
vary: Accept, origin, Cookie
allow: POST, GET, OPTIONS
access-control-allow-credentials: true
access-control-allow-origin: http://localhost:3000
x-frame-options: DENY
content-length: 95
x-content-type-options: nosniff
referrer-policy: same-origin
cross-origin-opener-policy: same-origin
```

### On Android:

[...]

PR #6206, opened 2023-01-08
Http formdata and file support

kjr-lh:
Currently the http plugin fetch and XMLHttpRequest implementations do not support File and FormData types. This is because all data sent over the js->native bridge are JSON encoded, and these types do not encode nicely.

This adds some additional checking on the JS side that detects these types and rebuilds them as plain JS objects with Files base64 encoded. The native implementations can detect these using a new parameter and correctly handle the values

- [ ] Android File support
- [ ] Android FormData support
- [ ] iOS File support
- [ ] iOS FormData support

PR #6708, opened 2023-07-05
feat(http): support for FormData requests

kjr-lh:
This is https://github.com/ionic-team/capacitor/pull/6206 for capacitor 5 (tested against 5.0.5)

Also includes a fix for the `content-type` bug mentioned [here](https://github.com/ionic-team/capacitor/issues/6604)

### DIFF
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -439,7 +439,8 @@ var nativeBridge = (function (exports) {
                                 dataType: type,
                                 headers: Object.assign(Object.assign({}, headers), optionHeaders),
                             });
-                            const contentType = nativeResponse.headers['Content-Type'] || nativeResponse.headers['content-type'];
+                            const contentType = nativeResponse.headers['Content-Type'] ||
+                                nativeResponse.headers['content-type'];
                             let data = (contentType === null || contentType === void 0 ? void 0 : contentType.startsWith('application/json'))
                                 ? JSON.stringify(nativeResponse.data)
                                 : nativeResponse.data;
@@ -593,10 +594,9 @@ var nativeBridge = (function (exports) {
                                         this._headers = nativeResponse.headers;
                                         this.status = nativeResponse.status;
                                         this.response = nativeResponse.data;
-                                        this.responseText =
-                                            ((_a = nativeResponse.headers['Content-Type']) === null || _a === void 0 ? void 0 : _a.startsWith('application/json'))
-                                                ? JSON.stringify(nativeResponse.data)
-                                                : nativeResponse.data;
+                                        this.responseText = ((_a = nativeResponse.headers['Content-Type']) === null || _a === void 0 ? void 0 : _a.startsWith('application/json'))
+                                            ? JSON.stringify(nativeResponse.data)
+                                            : nativeResponse.data;
                                         this.responseURL = nativeResponse.url;
                                         this.readyState = 4;
                                         this.dispatchEvent(new Event('load'));
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -35,6 +35,57 @@ var nativeBridge = (function (exports) {
     // For removing exports for iOS/Android, keep let for reassignment
     // eslint-disable-next-line
     let dummy = {};
+    const readFileAsBase64 = (file) => new Promise((resolve, reject) => {
+        const reader = new FileReader();
+        reader.onloadend = () => {
+            const data = reader.result;
+            resolve(btoa(data));
+        };
+        reader.onerror = reject;
+        reader.readAsBinaryString(file);
+    });
+    const convertFormData = async (formData) => {
+        const newFormData = [];
+        for (const pair of formData.entries()) {
+            const [key, value] = pair;
+            if (value instanceof File) {
+                const base64File = await readFileAsBase64(value);
+                newFormData.push({
+                    key,
+                    value: base64File,
+                    type: 'base64File',
+                    contentType: value.type,
+                    fileName: value.name,
+                });
+            }
+            else {
+                newFormData.push({ key, value, type: 'string' });
+            }
+        }
+        return newFormData;
+    };
+    const convertBody = async (body) => {
+        if (body instanceof FormData) {
+            const formData = await convertFormData(body);
+            const boundary = `${Date.now()}`;
+            return {
+                data: formData,
+                type: 'formData',
+                headers: {
+                    'Content-Type': `multipart/form-data; boundary=--${boundary}`,
+                },
+            };
+        }
+        else if (body instanceof File) {
+            const fileData = await readFileAsBase64(body);
+            return {
+                data: fileData,
+                type: 'file',
+                headers: { 'Content-Type': body.type },
+            };
+        }
+        return { data: body, type: 'json' };
+    };
     const initBridge = (w) => {
         const getPlatformId = (win) => {
             var _a, _b;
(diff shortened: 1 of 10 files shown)


markemer (approved):
Looks the same as #6206

ItsChaceD (approved):

jreviews:
Thanks for working on this! Would it be possible to see an example of how sending formdata in a post request would work? I've been trying all sorts of things and passing a FormData object with some appended data never makes it to the server. If I use a json object then it works fine. I tried setting the dataType to `formData` and that also didn't help. My goal is to eventually send a file selected from Camera.photos, but for not just trying with any simple data.

mikeelemuel:
@kjr-lh , I tried to update to `5.2.1` version which this version includes your changes.

However, when I tried the code below, I'm getting this on xcode terminal `[NSURLSession sharedSession] may not be invalidated`. Question, did I miss something based on your changes?

```
         const applicationAdapter = this.store.adapterFor('application');
         const headers            = {
            'X-APP-REF':     applicationAdapter.headers['X-APP-REF'],
            'X-APP-VERSION': applicationAdapter.headers['X-APP-VERSION'],
            'Authorization': applicationAdapter.headers['Authorization'],
         }
         
             const formData = new FormData();
            formData.append('visibility', this.visibility);
            formData.append('attachable_type', classify(this.args.attachableRecord.constructor.modelName));
            formData.append('attachable_id', this.args.attachableRecord.get('id'));
            formData.set('Content-Type', file.type);
            formData.append('attachment', file.file);
   
            await fetch(`/v3/${this.current.accountSubdomain}/attachments`, {
              method: 'POST',
              headers: headers,
              body: formData,
            });
```

jreviews:
I thought that this PR would allow using CapacitorHttp to make requests with FormData. Is that not the case? I tested using `fetch` directly and that worked. My test code to upload a photo selected from photos with fetch:

```js
const dataURItoBlob = (dataURI: string) => {
    // Split the `dataURI` at the comma.
    let splitDataURI = dataURI.split(',');

    // Get the mime-type from the `dataURI`
    let mimeType = splitDataURI[0].split(':')[1].split(';')[0];

    // Get the base64 string
    const base64 = splitDataURI[1];
    
    // Decode the base64 string
    let binary = atob(base64);
    
    // Create an 8-bit unsigned array
    let array = [];
    for(let i = 0; i < binary.length; i++) {
        array.push(binary.charCodeAt(i));
    }
    
    // Return the blob object with appropriate mime-type
    return new Blob([new Uint8Array(array)], {type: mimeType});
}

const upload = async () => {

    const photo = await Camera.getPhoto({
            resultType: CameraResultType.DataUrl,
            source: CameraSource.Photos,
            correctOrientation: true,
            quality: 90,
            allowEditing: false
        })

    const blob = dataURItoBlob(photo.dataUrl)

    const formData = new FormData()

    formData.append('file', blob)

    fetch(useAppConfig().getApiUrl('/upload-test'), {
            method: 'POST',
            body: formData
        })
        .then(response => response.json())
        .then(result => {
            console.log('Success:', result);
        })    
}
```

jreviews:
In case this is helpful for anyone else, the above code using fetch does end up triggering a native `CapacitorHttp.request` as seen in the console image below. 

image

However, I couldn't find a way to just issue the `CapacitorHttp.request` directly in any way that it would work. The closest I got was still  missing the `boundary`  part of the multipart/form-data header so it failed. So for now I'll stick to what's working for me.

muuvmuuv:
Just a note, but Filesystem plugin needs an update:

image

kjr-lh:
> [NSURLSession sharedSession] may not be invalidate

Hi, your code looks like it should be okay.. the only thing I can think to try would be to include the filename , e.g.
```
formData.append('attachment', file.file, 'somefilename.txt');
```

But I'm really doubtful this will help 😞 

I'm afraid I don't have enough Swift experience to diagnose that error

jreviews:
@kjr-lh Thanks!

> This doesn't add Blob support yet - if you can pass your data as ` File` it will work

I may try that eventually. However, removing the file from formData and just leaving a plain input with text also didn't work for me as mentioned here https://github.com/ionic-team/capacitor/pull/6708#issuecomment-1637145454 The data doesn't arrive to the server.

kjr-lh:
@jreviews I've created a demo app [here](https://github.com/kjr-lh/capacitor-http-test) that shows uploading FormData with file and text fields. I'll use this to add and test Blob support soon

kjr-lh:
> @kjr-lh Awesome, thanks. However, I don't see that you are using CapacitorHttp there. Per my previous comments, I have it working with `fetch`, but I thought using CapacitorHttp.request/post with FormData should also work and when I last tested, the data doesn't make it to the server (file or not).

Fetch is capacitorhttp now (from one of the 4.x versions) - Capacitor patches fetch and XMLHttpRequest with it's own versions that go through the native layer

https://capacitorjs.com/docs/apis/http

The code that is handling Files / FormData is in the fetch/XMLHttpRequest implementations, so you'll be skipping it if you go directly to CapacitorHttp request/post. You could encode the data yourself, but it'd be much simpler to use fetch

jreviews:
> The code that is handling Files / FormData is in the fetch/XMLHttpRequest implementations

I understand that from my tests, but what was confusing to me since I was waiting for the FormData update is that there's nothing in the documentation about having to use fetch directly to be able to also use FormData/Files.

What made it even more confusing is that I went through the code changes to try to understand how to use FormData and saw this:

https://github.com/ionic-team/capacitor/pull/6708/files#diff-63eeefca71ce9525d8b4f78010d981ea646a106657519835b66b1bf47b2aac70R201

So basically a new option was added to HttpOptions (which is used with CapacitorHttp) to specify file/formData

```js
export interface HttpOptions {
  url: string;
  method?: string;
  params?: HttpParams;
  data?: any;
  headers?: HttpHeaders;
  /**
   * How long to wait to read additional data. Resets each time new
   * data is received
   */
  readTimeout?: number;
  /**
   * How long to wait for the initial connection.
   */
  connectTimeout?: number;
  /**
   * Sets whether automatic HTTP redirects should be disabled
   */
  disableRedirects?: boolean;
  /**
   * Extra arguments for fetch when running on the web
   */
  webFetchExtra?: RequestInit;
  /**
   * This is used to parse the response appropriately before returning it to
   * the requestee. If the response content-type is "json", this value is ignored.
   */
  responseType?: HttpResponseType;
  /**
   * Use this option if you need to keep the URL unencoded in certain cases
   * (already encoded, azure/firebase testing, etc.). The default is _true_.
   */
  shouldEncodeUrlParams?: boolean;
  /**
   * This is used if we've had to convert the data from a JS type that needs
   * special handling in the native layer
   */
  dataType?: 'file' | 'formData';
}
```

muuvmuuv:
You can @jreviews, I you mean this:

```ts
import { CapacitorHttp } from "@capacitor/core"
import { Filesystem } from "@capacitor/filesystem"

export async function fileUpload(file: {
  name: string
  path: string
  mimeType: string
}) {
  const formData = new FormData()

  const contents = await Filesystem.readFile({ path: file.path })
  const blob = new Blob([contents.data], { type: file.mimeType })
  formData.append("file", blob, file.name)

  const response = await CapacitorHttp.request({
    url: "https://example.com/upload",
    dataType: "file",
    data: formData,
    headers: {
      "Content-Type": "multipart/form-data",
    },
  })

  // ...
}
```

kjr-lh:
> You can @jreviews, if you mean this:
> 
> ```ts
> import { CapacitorHttp } from "@capacitor/core"
> import { Filesystem } from "@capacitor/filesystem"
> 
> export async function fileUpload(file: {
>   name: string
>   path: string
>   mimeType: string
> }) {
>   const formData = new FormData()
> 
>   const contents = await Filesystem.readFile({ path: file.path })
>   const blob = new Blob([contents.data], { type: file.mimeType })
>   formData.append("file", blob, file.name)
> 
>   const response = await CapacitorHttp.request({
>     url: "https://example.com/upload",
>     dataType: "file",
>     data: formData,
>     headers: {
>       "Content-Type": "multipart/form-data",
>     },
>   })
> 
>   // ...
> }
> ```

I really don't expect this to work as-is - I think you'd need to have converted the form data already to a base 64 encoded string (take a look at what `convertFormData` is doing in `native-bridge.ts`

As it is now, I think capacitor will just try to convert the data to json, which means it will be sent to the native layer as `"{}"` 

Now that the http plugin is part of capacitor core, I think the expectation is that everyone will just use fetch, as it's a more usual interface and code using it will also work on the web

jreviews:
Thanks! I've released an update using `fetch` which is working fine, so I may come back and review this again later on.

Mimateam:
> You can @jreviews, if you mean this:
> 
> ```ts
> import { CapacitorHttp } from "@capacitor/core"
> import { Filesystem } from "@capacitor/filesystem"
> 
> export async function fileUpload(file: {
>   name: string
>   path: string
>   mimeType: string
> }) {
>   const formData = new FormData()
> 
>   const contents = await Filesystem.readFile({ path: file.path })
>   const blob = new Blob([contents.data], { type: file.mimeType })
>   formData.append("file", blob, file.name)
> 
>   const response = await CapacitorHttp.request({
>     url: "https://example.com/upload",
>     dataType: "file",
>     data: formData,
>     headers: {
>       "Content-Type": "multipart/form-data",
>     },
>   })
> 
>   // ...
> }
> ```

We also need to send a Blob through post method. We currently have mobile app env. with Ionic 7, Capacitor 5, Angular 16. In our case we need FormData, which we send it through post method to client's API. Client's API accepts only Blob.

We've tried angular/common/http which works great on dev env., but in production We get hit by client's CORS policy. So we need to use plugin which uses native libraries, like this one.

We would really appreciate it if there would be a way to send Blob directly through post with FormData. Thanks!

jreviews:
@Mimateam The code I shared before https://github.com/ionic-team/capacitor/pull/6708#issuecomment-1637941749 using fetch works to send FormData with blob and it also makes request using native libraries. Just make sure you enable fetch patching per the capacitor http docs https://capacitorjs.com/docs/apis/http#configuration

Mimateam:
We have now another problem. When using this approach https://github.com/ionic-team/capacitor/pull/6708#issuecomment-1637941749 then we cannot access images on mobile phone from cache when picking images from camera or gallery.

Solution to allow Blob directly through post with FormData would be highly appreciated and would solve all problems. We don't see why this should not be implemented?

asad77210:
No update on this ?

PR #6716, opened 2023-07-06
fix(cli): signing type option issue

liuwin7:
There are two revision.
The first one is to correct the `commander` use case, if a option is named with b `-`, it will change to the Camel Case.
The second is to load  the all android build options, not only the options for `apksigner `.


ios/Capacitor/Capacitor/JS.swift
```diff
@@ -19,13 +19,6 @@ internal struct JSCall {
     let pluginId: String
     let method: String
     let callbackId: String
-
-    init(options: [String: Any], pluginId: String, method: String, callbackId: String) {
```
markemer, line 23: What's this in here for?

cli/src/index.ts
```diff
@@ -157,7 +157,7 @@ export function runProgram(config: Config): void {
     )
     .addOption(
       new Option(
-        '--signing-type <signingtype>',
+        '--signingtype <signingtype>',
```
markemer, line 160: This is a breaking change - can we instead keep it as is and fix the variables?

core/native-bridge.ts
```diff
@@ -1038,6 +1038,7 @@ const initBridge = (w: any): void => {
       });
     };
 
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
markemer, line 1041: I'm all for this, but we might want to put this in another PR.

markemer (changes requested):
Just a few questions, but I think we can make this change.

ios/Capacitor/Capacitor/JS.swift
liuwin7: Because I run `npm run fmt`.
For `Swift` the struct's `init` function is default created when compiled, no need to create explicitly.

core/native-bridge.ts
liuwin7: Ok, I did this mainly to pass the review reboot.

cli/src/index.ts
liuwin7: It's ok.
But, it is surely a breaking change same as using the `--signing-type` instead of `--signingtype` following the convention initially.

core/native-bridge.ts
markemer: Did npm run fmt do this? If so, lets keep it.

ios/Capacitor/Capacitor/JS.swift
markemer: Yeah - I'm not sure why they're in there - `npm run fmt` catching it probably makes this ok to include.

core/native-bridge.ts
liuwin7: No for `npm run fmt`, but `npm run lint` need.

markemer (approved):

markemer (approved):

giralte-ionic (approved):

liuwin7:
Why still not merged yet after 2 versions have been released?

IT-MikeS (approved):

PR #6725, opened 2023-07-12
fix(http): fire events in correct order when using xhr

ItsChaceD:
This PR fixes issues with `XMLHttpRequest` event dispatching:
1. Some events such as `onload` were previously firing twice due to manually attempting to override the prototype event listeners. Now we allow the original XHR to handle setting the event listeners, but we still manually dispatch each event.
2. Some events were not being emitted in the proper order (`loadstart` for example)
3. `onreadystatechange` was previously not firing when the XHR object was being constructed due to a race condition.

Test app (test parity with web):
https://github.com/ItsChaceD/capacitor-http-xhr-events

### DIFF
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -511,10 +511,6 @@ var nativeBridge = (function (exports) {
                                 writable: true,
                             },
                         });
-                        xhr.addEventListener('timeout', function () {
-                            if (typeof this.ontimeout === 'function')
-                                this.ontimeout();
-                        });
                         xhr.readyState = 0;
                         const prototype = win.CapacitorWebXMLHttpRequest.prototype;
                         // XHR patch abort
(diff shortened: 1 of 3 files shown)


markemer (approved):
Tested on emulator - seems to work the way we want.

giralte-ionic (approved):

5uper:
> This PR fixes issues with `XMLHttpRequest` event dispatching:
> 
> 1. Some events such as `onload` were previously firing twice due to manually attempting to override the prototype event listeners. Now we allow the original XHR to handle setting the event listeners, but we still manually dispatch each event.
> 2. Some events were not being emitted in the proper order (`loadstart` for example)
> 3. `onreadystatechange` was previously not firing when the XHR object was being constructed due to a race condition.
> 
> Test app (test parity with web): https://github.com/ItsChaceD/capacitor-http-xhr-events

Hi @ItsChaceD

I recognize since version 5.2.0 the following anomaly and suspect that this PR is the reason for 2 problems. I have checked out the test app capacitor-http-xhr-events.

1. the method is missing in the request, which is reported in  #6726

Bildschirmfoto 2023-07-17 um 10 14 36

2. I have locally solved the problem with the missing method in native-bridge.ts line 612, which leads to the fact that the response data does not arrive correctly. 

```typescript
// XHR patch open
prototype.open = function (method: string, url: string) {
   this._url = url;
   this._method = method;  // <-- fix for missing method
...
```

Example with angular httpclient where the response always returns null:
```typescript
this.httpClient.post(options.url, options.data, { headers: options.headers }).pipe(
      tap((res) => console.log('POST Tap', res))
...
```

image

PR #6753, opened 2023-07-19
Http plugin tests

kjr-lh:
Adds some basic unit tests for the Http plugin functionality

Only going to test using the android path - the `bridge.spec.ts` tests handle the underlying implementation on both platforms


eric-horodyski:

More context:

PR #6762, opened 2023-07-21
Clarify when server.url should be configured

gaborcs:
The recommendation of not using server.url in production seems to be out of date.

The Capacitor templates for Remix explicitly recommend using it in production: "When deploying your application to the App Store, be sure to set your server.url variable to a hosted server"
https://github.com/ionic-team/capacitor-remix-templates#setting-your-serverurl-variable

(I'm not an expert on the pros and cons of using server-side rendered apps with Capacitor, so please correct me if I'm wrong.)

### DIFF
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -483,9 +483,8 @@ export interface CapacitorConfig {
     /**
      * Load an external URL in the Web View.
      *
-     * This is intended for use with live-reload servers.
-     *
-     * **This is not intended for use in production.**
+     * This is intended for use with live-reload servers, and server-side
+     * rendered apps that aren't exportable to static files.
      *
      * @since 1.0.0
      */


KoenLav:
We have been using server.url in production for over a year now.

Although both Apple and Google are somewhat vague about it, it should actually be allowed.

ZeroHackeR:
I still think. it's still dangerous to use server.url on production. 

Capacitor has a lot of power compare to a website.
Let's say an app has some serious permissions (eg: File Permission or background service) 
and it's using server.url too.
It would be very dangerous to the app users when the server got hacked.
The hacker can do a lot of things with user's devices.

jcesarmobile:
That template has never been an official template and `server.url` is still not recommended for production apps, despite it's possible to use it and some people are using it, the recommended is to ship the app assets inside it and not rely in a remote server that could lead to app rejections.
When using `server.url` in production apps, users should handle the case of no internet connection, which is not a problem for regular Capacitor apps.

PR #6765, opened 2023-07-25
fix(cli): type only import in config

seo-rii:
This PR makes to use type only import in generated capacitor.config.ts.


jcesarmobile (approved):

jcesarmobile:
This requires typescript 3.8+, which we already require since Capacitor 3, so it's safe to add.
Only affects new apps, but I think we should only add it for Capacitor 6.

markemer (approved):

Issue #6758, opened 2023-07-21
bug: Selecting single image from samsung gallery does not work with a multiple input field


theigl:
### Capacitor Version
```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 5.1.0
  @capacitor/core: 5.1.0
  @capacitor/android: 5.1.0
  @capacitor/ios: 5.1.0

Installed Dependencies:

  @capacitor/cli: 4.8.0
  @capacitor/core: 4.8.0
  @capacitor/android: 4.8.0
  @capacitor/ios: 4.8.0

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)
Android

### Current Behavior

When you have a `<input multiple>` field and select a *single* image from the *Android Gallery*, the input field does not receive any data. It works fine when multiple images are selected or when selecting a single image from any other source.

We debugged the issue and found the root cause here:

https://github.com/ionic-team/capacitor/blob/1b8352dc5124dc3f57d7881d619537cbf8c3674b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java#L431

If we change the condition to > 0, then the input field behaves as expected.

### Expected Behavior

It should be possible to select a single image from the Android Gallery when using a multiple input field.

### Code Reproduction

Reproducer: https://github.com/divatzk/capacitor-file-upload-test

The reproducer contains only one page with a single and a multiple input field. To reproduce the issue, click on the multiple input field and select a single image from the Gallery. Nothing happens in this case. If you select 2 or more images, it works.

Note: This is the same as https://github.com/ionic-team/capacitor/issues/6710 which was closed by the bot even though it had a reproducer.

jcesarmobile:
I can't reproduce.
In which devices and Android versions are you testing?

theigl:
We just realized we only tested this on Samsung devices and the app in question is the "Samsung Gallery" and not the default Android gallery. This might be an issue with Samsung devices only. There is another [bug report](https://github.com/ionic-team/capacitor/issues/6305) that is also about Samsung device.

jcesarmobile:
What's the Android version of the Samsung devices you tested?
I don't have any, but I've tested several in browserstack and it worked in all of them. So could be related to the Samsung Gallery app version.

[...]

PR #6778, opened 2023-07-28
fix(android): allow single input file selection from samsumg gallery

jcesarmobile:
`WebChromeClient.FileChooserParams.parseResult` doesn't work fine when other gallery apps are used for the input file with multiple attribute if the user only picks a single image, so always handle the file picking ourselves if `resultIntent.getClipData() != null`, no matter how many elements it contains.

closes https://github.com/ionic-team/capacitor/issues/6758


IT-MikeS (approved):

markemer (approved):

sahilrajput03:
@ankitgoel25 This is released in 5.2.3 already.

PR #6808, opened 2023-08-11
Replay of https://github.com/ionic-team/capacitor/pull/6510/

zakuru:
This PR replays  https://github.com/ionic-team/capacitor/pull/6510/ which introduced a breaking change and was reverted on https://github.com/ionic-team/capacitor/pull/6689/
As I can see all the checks passed successfully. Is that behavior expected, or should some test be added?


markemer:
Wait - the PRs you reference all point to 4.x - do you mean to point this at main (5.x) and if so, was it reverted in 5.x or in 4.x?

PR #6814, opened 2023-08-15
replay capwebview-change

zakuru:
replays https://github.com/ionic-team/capacitor/pull/6705


jcesarmobile:
it was reverted because it caused problems, so we are not going to reapply something that breaks things

PR #6818, opened 2023-08-16
fix(http): route get requests through custom handler

### LINKED ISSUE #6123: bug: CapacitorHttp - Loading a WASM file fails on native iOS and Android

### LINKED ISSUE #6126: bug: CapacitorHttp - Requesting a Blob fails on native iOS and Android

ItsChaceD:
Closes #6123
Closes #6126 

This PR routes fetch/XHR request methods (when `CapacitorHttp` is enabled) that don't allow bodies (`GET`, `OPTIONS`, `HEAD`, `TRACE`) to native protocol handlers so the responses don't have to be encoded/decoded through the bridge. 

Opened in favor of outdated #6540 

Test app: https://github.com/ItsChaceD/capacitor-http-wasm-issue


deving1995:
- I need this fix desperately

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -121,6 +126,62 @@ internal class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
 
+    func handleCapacitorHttpRequest(_ urlSchemeTask: WKURLSchemeTask) {
+        var urlRequest = urlSchemeTask.request
+        let url = urlRequest.url!
+        var targetUrl = url.absoluteString
+            .replacingOccurrences(of: CapacitorBridge.httpInterceptorStartIdentifier, with: "")
+
+        // Only replace first occurrence of the scheme
+        if let range = targetUrl.range(of: InstanceDescriptorDefaults.scheme) {
+            targetUrl = targetUrl.replacingCharacters(in: range, with: "https")
```
jcesarmobile, line 137: We are hardcoding the scheme to https here, but the original url could be http.
We are going to need to also store the original scheme as part of the modified url for iOS and doing the replacement to the original scheme, or only proxy https urls (and document that the proxy only works for https urls)

core/native-bridge.ts
```diff
@@ -480,6 +480,28 @@ const initBridge = (w: any): void => {
...
           }
 
+          if (
+            !options?.method ||
+            options.method.toLocaleUpperCase() === 'GET' ||
+            options.method.toLocaleUpperCase() === 'HEAD' ||
+            options.method.toLocaleUpperCase() === 'OPTIONS' ||
+            options.method.toLocaleUpperCase() === 'TRACE'
+          ) {
+            const url = new URL(resource.toString());
```
jcesarmobile, line 490: Doesn't it need to check if the url is relative as in the XHR part?

Could be the url patching code be moved to a separate function that does the relative url check and string replacement in a single place so is less error prone?

We could even export it so users could create "proxy" urls for img tag or other html elements.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -199,6 +213,110 @@ private boolean isAllowedUrl(Uri loadingUrl) {
...
+                        encoding = encodingParts[1].trim();
+                    }
+                }
+            } else {
+                responseHeaders.put(entry.getKey(), builder.toString());
+            }
+        }
+
+        if (null == mimeType) {
+            mimeType = getMimeType(request.getUrl().getPath(), connection.getInputStream());
```
jcesarmobile, line 311: can we store the `connection.getInputStream()` result in a variable to avoid calling it twice as it can be a memory intensive call?

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -199,6 +213,110 @@ private boolean isAllowedUrl(Uri loadingUrl) {
         return !(bridge.getServerUrl() == null && !bridge.getAppAllowNavigationMask().matches(loadingUrl.getHost()));
     }
 
+    private Boolean isDomainExcludedFromSSL(Bridge bridge, URL url) {
```
jcesarmobile, line 216: could we make the same method we have in `HttpRequestHandler` public and reuse it here instead of duplicating, or move it somewhere else were we can make it protected and then use it in both places?

jcesarmobile (changes requested):

android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
```diff
@@ -238,11 +238,14 @@ public static Object readData(ICapacitorHttpUrlConnection connection, ResponseTy
             } else {
                 return readStreamAsString(errorStream);
             }
-        } else if (contentType != null && contentType.contains(MimeType.APPLICATION_JSON.getValue())) {
-            // backward compatibility
-            return parseJSON(readStreamAsString(connection.getInputStream()));
         } else {
             InputStream stream = connection.getInputStream();
+
```
jcesarmobile, line 243: Where I told you to call getInputStream once was a different file where it was being called twice, once for getting the mime type and another one for returning the input stream, and that was not fixed.
Here was not needed because they were different elses, so only one would be called.

core/native-bridge.ts
```diff
@@ -2,7 +2,7 @@
  * Note: When making changes to this file, run `npm run build:nativebridge`
  * afterwards to build the nativebridge.js files to the android and iOS projects.
  */
-import type { HttpResponse } from './src/core-plugins';
+import { isRelativeOrProxyUrl, type HttpResponse, createProxyUrl } from './src/core-plugins';
```
jcesarmobile, line 5: lint is complaining here because it doesn't allow type imports with regular imports, should be separate imports.
But anyway, I think isRelativeOrProxyUrl and createProxyUrl could be both inside native-bridge.ts file, not in core-plugins as those are not plugin methods but helpers.
When I said we could export the createProxyUrl method, I meant as a Capacitor function like `convertFileSrc` function here https://github.com/ionic-team/capacitor/blob/main/core/native-bridge.ts#L865
But that's going to require a few more changes, so we can export it later.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -199,6 +224,104 @@ private boolean isAllowedUrl(Uri loadingUrl) {
...
+            case 501 -> "Not Implemented";
+            case 502 -> "Bad Gateway";
+            case 503 -> "Service Unavailable";
+            case 504 -> "Gateway Timeout";
+            case 505 -> "HTTP Version Not Supported";
+            default -> "Unknown";
+        };
+    }
+
+    private WebResourceResponse handleCapacitorHttpRequest(WebResourceRequest request, Boolean isHttpsRequest) throws IOException {
```
jcesarmobile, line 263:
```suggestion
    private WebResourceResponse handleCapacitorHttpRequest(WebResourceRequest request) throws IOException {
```
`isHttpsRequest` is not being used on Android since the urls are already http/https (should be removed in the methods that call `handleCapacitorHttpRequest`.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -165,6 +171,25 @@ private static Uri parseAndVerifyUrl(String url) {
      */
     public WebResourceResponse shouldInterceptRequest(WebResourceRequest request) {
         Uri loadingUrl = request.getUrl();
+
+        if (null != loadingUrl.getPath()) {
```
jcesarmobile, line 175: this can be a single if with both conditions since `return handleCapacitorHttpRequest(request, false);` should be changed to `return handleCapacitorHttpRequest(request);` as the boolean is not being used.

jcesarmobile (changes requested):

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -121,6 +135,63 @@ public class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
+        // Only replace first occurrence of the scheme
+        if let range = targetUrl.range(of: self.serverUrl?.scheme ?? InstanceDescriptorDefaults.scheme) {
+            targetUrl = targetUrl.replacingCharacters(in: range, with: isHttpsRequest ? "https" : "http")
+        }
+
+        urlRequest.url = URL(string: targetUrl)
+
+        let urlSession = URLSession.shared
+        let task = urlSession.dataTask(with: urlRequest) { (data, response, error) in
+            urlSession.invalidateAndCancel()
```
Steven0351, line 154: Calling this does nothing: https://developer.apple.com/documentation/foundation/urlsession/1411538-invalidateandcancel

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -121,6 +135,63 @@ public class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
         return false
     }
 
+    func handleCapacitorHttpRequest(_ urlSchemeTask: WKURLSchemeTask, _ localUrl: URL, _ isHttpsRequest: Bool) {
+        var urlRequest = urlSchemeTask.request
+        let url = urlRequest.url!
```
Steven0351, line 140: Probably better to `guard let url = urlRequest.url else { return }` than to force unwrap. The only time it should be considered okay to force unwrap is if you can guarantee without a doubt that the optional is present, either because you control the data flow yourself, or you've otherwise checked its presence in a way that didn't lead to the safe unwrapping. It probably shouldn't be considered okay to crash the program here if the url isn't present.

jcesarmobile: I didn't comment on that because we already had a force unwrap in a very similar code here https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/WebViewAssetHandler.swift#L25
If it didn't crash there it would crash in the new code neither.

But yeah, we supposedly had rules in swiftlint to not allow force unwraps, but doesn't seem to be working.

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
markemer: Well, TIL. I don't know I ever had occasion to use `invalidateAndCancel` but thinking about it, it makes sense since you really have no idea what `shared` could be up to and the system would have to restart the shared session if you don't.

jcesarmobile (changes requested):
can you resolve the conflicts?

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -121,6 +135,64 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
 
+    func handleCapacitorHttpRequest(_ urlSchemeTask: WKURLSchemeTask, _ localUrl: URL, _ isHttpsRequest: Bool) {
+        var urlRequest = urlSchemeTask.request
+        guard let url = urlRequest.url else { return }
+        var targetUrl = url.absoluteString
+            .replacingOccurrences(of: CapacitorBridge.httpInterceptorStartIdentifier, with: "")
+            .replacingOccurrences(of: CapacitorBridge.httpsInterceptorStartIdentifier, with: "")
+
+        // Only replace first occurrence of the scheme
+        if let range = targetUrl.range(of: self.serverUrl?.scheme ?? InstanceDescriptorDefaults.scheme) {
```
jcesarmobile, line 146:
```suggestion
        if let range = targetUrl.range(of: localUrl.scheme ?? InstanceDescriptorDefaults.scheme) {
```
should replace the localUrl scheme, not the serverUrl scheme, because when using live reload serverUrl scheme is http, but proxy urls are capacitor (or the configured scheme)

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -121,6 +135,64 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
+                let existingHeaders = response.allHeaderFields
+                var newHeaders: [AnyHashable: Any] = [:]
+
+                // if using live reload, then set CORS headers
+                if self.isUsingLiveReload(url) {
+                    newHeaders = [
+                        "Access-Control-Allow-Origin": self.serverUrl?.absoluteString ?? "",
+                        "Access-Control-Allow-Methods": "GET, HEAD, OPTIONS, TRACE"
+                    ]
+                }
```
jcesarmobile, line 169: When not using live reload there are still problems with CORS because the url is `capacitor://server.hostname.com/_capacitor_https_interceptor_/` and that doesn't allow `capacitor://localhost` to access because it's a different hostname.
We should make the proxy url look like `capacitor://localhost//_capacitor_https_interceptor_/server.hostname.com/`. Or alternatively remove this check.
I think the first option is better, but the later is simpler.

jcesarmobile (commented):
Added a few comments for iOS.

For Android there are CORS problems for all requests, not only when not using live reload, we should add CORS headers for both cases and/or change the urls to match the local url like `https://localhost//_capacitor_https_interceptor_/server.hostname.com/`

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
ItsChaceD: Yeah good catch.. I updated the proxy url construction on both platforms to match your suggestion just to be safe

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -121,6 +135,69 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
+            .replacingOccurrences(of: CapacitorBridge.httpInterceptorStartIdentifier, with: "")
+            .replacingOccurrences(of: CapacitorBridge.httpsInterceptorStartIdentifier, with: "")
+
+        // Only replace first occurrence of the scheme
+        if let range = targetUrl.range(of: localUrl.scheme ?? InstanceDescriptorDefaults.scheme) {
+            targetUrl = targetUrl.replacingCharacters(in: range, with: isHttpsRequest ? "https" : "http")
+        }
+
+        // Only replace first occurrence of the hostname
+        if let range = targetUrl.range(of: (self.serverUrl?.host ?? InstanceDescriptorDefaults.hostname) + "/") {
```
jcesarmobile, line 151:
```suggestion
        if let range = targetUrl.range(of: (localUrl.host ?? InstanceDescriptorDefaults.hostname) + "/") {
```
This is not working when using live reload, I think it should use `localUrl` instead, that seems to work in both cases

core/native-bridge.ts
```diff
@@ -118,6 +118,47 @@ const convertBody = async (
...
+    if (
+      webviewServerUrl.protocol !== 'capacitor:' &&
+      webviewServerUrl.protocol !== 'http:' &&
+      webviewServerUrl.protocol !== 'https:'
+    ) {
+      proxyUrl.protocol = webviewServerUrl.protocol;
+    }
+  }
+
+  proxyUrl.hostname = webviewServerUrl.hostname;
```
jcesarmobile, line 155:
```suggestion
  let proxyUrl = new URL(url);
  const isHttps = proxyUrl.protocol === 'https:';
  const originalHostname = proxyUrl.hostname;
  const originalPathname = proxyUrl.pathname;

  proxyUrl = new URL(win.Capacitor?.getServerUrl() ?? '');
```

When using live reload on Android the replacement is not working because the server url has a port and the port is not being added into the newly created url.

I think a simple approach that also solves that issue is to recreate the proxyUrl from the capacitor server url, so you don't have to check for the hostname, port or scheme.

jcesarmobile (changes requested):
Added a few comments.

Also, I've noticed that the proxy urls look like this `capacitor://localhost//`, with a double `/` at the end. Is that intended? I've read that while double `/` are accepted by servers they can cause problems on the browsers.

core/native-bridge.ts
ItsChaceD: Ah yeah, good suggestion 👍

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
ItsChaceD: Yeah make sense, I missed this

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -15,20 +15,26 @@
...
 import android.net.Uri;
 import android.util.Base64;
 import android.webkit.CookieManager;
 import android.webkit.WebResourceRequest;
 import android.webkit.WebResourceResponse;
+import com.getcapacitor.plugin.util.CapacitorHttpUrlConnection;
+import com.getcapacitor.plugin.util.HttpRequestHandler;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.reflect.Method;
```
jcesarmobile, line 30:
```suggestion
```
unused import

jcesarmobile (approved):

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -199,6 +220,116 @@ private boolean isAllowedUrl(Uri loadingUrl) {
...
+            } else {
+                responseHeaders.put(entry.getKey(), builder.toString());
+            }
+        }
+
+        // Using Live Reload
+        if (bridge.getServerUrl() != null) {
+            responseHeaders.put("Access-Control-Allow-Origin", bridge.getServerUrl());
+            responseHeaders.put("Access-Control-Allow-Methods", "GET, HEAD, OPTIONS, TRACE");
+        }
```
jcesarmobile, line 319:
```suggestion
```
I've just realized we don't really need CORS headers on Android since the server url is the same than the proxy url

jcesarmobile (approved):

theproducer (approved):

colbymelvin:
Any idea if this fix will be included in an upcoming release sometime soon?

My org is in the middle of migrating a large scale Cordova application to Capacitor. As far as we can tell, this is the last issue blocking us. We download image blobs and store them in SQLite so we can display them without an internet connection.

PR #6820, opened 2023-08-16
Fix case of HTTP Response Headers in IOS

theomega:
The Apple Foundation HTTP Code returns the dictionary of Response
Headers capitalized (i.e. "Content-Type") [1]. The expectation from
Javascript side is that HTTP Response Headers when queried via the
`getResponseHeader` method are matched case insensitive [2]. The
current implementation of the Capacitor HTTP Plugin does not correctly
bring this together. So calling `getResponseHeader("content-type")`
fails currently (i.e. returns nothing) even if the `content-type` header
is present, although the matching should be done case insensitive.

This change does two things:
- The HTTP Response Headers which are returned in a dictionary are
  transformed and the keys are lowercased. The value is preserved.
- When querying the value of a header in the `getResponseHeader`
  function in the bridge code, the header is also lowercased.

This should make the requesting of response haders case insensitive on
IOS.

[1]: https://developer.apple.com/documentation/foundation/httpurlresponse/1417930-allheaderfields
[2]: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/getResponseHeader


markemer:
I like this a lot better than what I just merged - I had been thinking about doing just this kinda thing.

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -37,6 +37,19 @@ func tryParseJson(_ data: Data) -> Any {
     }
 }
 
+
+/// Helper to convert the headers dictionary to lower case keys. This allows case-insensitive querying in the bridge javascript.
+/// - Parameters:
+///     - headers: The headers as dictionary. The type is unspecific because the incoming headers are coming from the
+///       allHeaderFields property of the HttpResponse.
+/// - Returns: The modified headers dictionary with lowercase keys
+func lowerCaseHeaderDictionary(_ headers: [AnyHashable: Any]) -> [String: Any] {
```
Steven0351, line 46: I recommend making this private

markemer (approved):

PR #6829, opened 2023-08-21
fix(http): return numbers and booleans as-is when application/json is the content type

### LINKED ISSUE #6819: bug: CapacitorHttp throws error when HTTP response is an integer/double in Android
### Capacitor Version

```
  @capacitor/cli: 5.2.3
  @capacitor/android: 5.2.3
  @capacitor/ios: 5.2.3
  @capacitor/core: 5.2.3
```

### Platform(s)

Android

### Current Behavior

When running on an Android device, and hitting an endpoint that returns ONLY an integer or double, the console returns the following error:

code: "JSONException"
message: "Value X of type java.lang.Integer cannot be converted to JSONArray"

where X is an integer. Note that X can be a double and the error will specify java.lang.Double accordingly.

Note that the response needs to be a plain integer/double, without JSON formatting, e.g. 3.
A plaintext version of a number, e.g. "3" will not throw this exception.
It can be sent with or without a response code and the same error will occur.

### Expected Behavior

The error should not be thrown, the value should be returned as is.

### Code Reproduction

See the following repo for a demo app: https://github.com/philkonick/Capacitor5HttpDemo

I have tested with Angular HTTP and CapacitorHttp syntax and the result is the same.

I used https://beeceptor.com/console/httpintegerdemo to create a mock endpoint to test the format.

### Other Technical Details

`npm --version` output: 9.5.1

`node --version` output: v18.16.1

`pod --version` output (iOS issues only): N/A

### Additional Context

I found open bug reports on the Capacitor Community Http plugin for this same issue. See: https://github.com/capacitor-community/http/issues/211

The solution on that page works, but only for integers. It differs slightly from the one on that page in that one can replace either line 317 or 321 (where the line is "return new JSArray(input);") of HttpRequestHandler.java with the code specified:
```
try {
    return new JSArray(input);
} catch (JSONException e2) {
    return Integer.parseInt(input.trim());
}
```

For a double, the relevant Double.parseDouble is required. I have not tested with other formats. But this is definitely where the problem originates.

ItsChaceD:
Closes #6819

This PR fixes an issue on Android where `parseJSON` would try to force convert booleans and numbers to JSON objects instead of returning them as-is.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
@@ -302,14 +302,18 @@ public class HttpRequestHandler {
             if ("null".equals(input.trim())) {
                 return JSONObject.NULL;
             } else if ("true".equals(input.trim())) {
-                return new JSONObject().put("flag", "true");
+                return true;
             } else if ("false".equals(input.trim())) {
-                return new JSONObject().put("flag", "false");
+                return false;
             } else if (input.trim().length() <= 0) {
                 return "";
             } else if (input.trim().matches("^\".*\"$")) {
                 // a string enclosed in " " is a json value, return the string without the quotes
                 return input.trim().substring(1, input.trim().length() - 1);
+            } else if (input.trim().matches("^-?\\d+(\\.\\d+)?$")) {
+                return Double.parseDouble(input.trim());
+            } else if (input.trim().matches("^-?\\d+$")) {
+                return Integer.parseInt(input.trim());
             } else {
                 try {
                     return new JSObject(input);
@@ -318,7 +322,7 @@ public class HttpRequestHandler {
                 }
             }
         } catch (JSONException e) {
-            return new JSArray(input);
+            return input;
         }
     }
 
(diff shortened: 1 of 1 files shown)


IT-MikeS (approved):

android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
```diff
@@ -302,14 +302,18 @@ public static Object parseJSON(String input) throws JSONException {
...
-                return new JSONObject().put("flag", "false");
+                return false;
             } else if (input.trim().length() <= 0) {
                 return "";
             } else if (input.trim().matches("^\".*\"$")) {
                 // a string enclosed in " " is a json value, return the string without the quotes
                 return input.trim().substring(1, input.trim().length() - 1);
+            } else if (input.trim().matches("^-?\\d+(\\.\\d+)?$")) {
+                return Double.parseDouble(input.trim());
+            } else if (input.trim().matches("^-?\\d+$")) {
```
jcesarmobile, line 315: integers also match the double regex check, so the integer else check should be moved up or they will be returned as double
or we can remove the integer check and only return doubles, I don't think that would be a problem on the JS side as there all the double and integers are "number"

ItsChaceD: We will switch the order in case someone is using this class natively

PR #6837, opened 2023-08-24
feat(cli): added live update strategy to the LiveUpdateConfig

carlpoole:
Adds the strategy option to the Live Updates config.

### DIFF
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -588,7 +588,7 @@ export interface LiveUpdateConfig {
   appId: string;
   channel: string;
   autoUpdateMethod: AutoUpdateMethod;
-  strategy: LiveUpdateStrategy;
+  strategy?: LiveUpdateStrategy;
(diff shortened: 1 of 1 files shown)


cli/src/declarations.ts
```diff
@@ -588,12 +588,15 @@ export interface LiveUpdateConfig {
   appId: string;
   channel: string;
   autoUpdateMethod: AutoUpdateMethod;
+  strategy: LiveUpdateStrategy;
```
Steven0351, line 591: Did we want to make this optional and default to 'differential' per the conversation we had with Netkow?

carlpoole: YES, thanks! Fixed.

jcesarmobile (approved):

carlpoole:
Closing this in favor of an alternate solution removing duplicate typing between cli/live updates and fedcap such that we have to keep updating the CLI.

PR #6842, opened 2023-08-28
fix(ios): add removeAllListeners selector

Andy3189:
We have noticed that its not possible to call the `removeAllListeners` function as described [here](https://capacitorjs.com/docs/apis/app#removealllisteners) for ios capacitor plugins. This simple fix handles the missing `removeAllListeners` selector.


Andy3189:
@markemer Is there any ETA for a response to this pr?

markemer (dismissed):

giralte-ionic (approved):

IT-MikeS:
@Andy3189 Could you provide an example reproduction of this issue, `removeAllListeners` is actively used in a few of the official Capacitor plugins to remove the listeners that they can add.

Andy3189:
@IT-MikeS you can find a minimal example here: https://github.com/Andy3189/capacitor-listener-example 
If you click the removeAllListeners button inside an Android App, no error will occur. On iOS the following messages will appear: 
```
⚡️  To Native ->  Example removeAllListeners 17752391
⚡️  Error calling method removeAllListeners on plugin Example: No method found.
⚡️  Ensure plugin method exists and uses @objc in its declaration, and has been defined
```

IT-MikeS:
> @IT-MikeS you can find a minimal example here: https://github.com/Andy3189/capacitor-listener-example If you click the removeAllListeners button inside an Android App, no error will occur. On iOS the following messages will appear:
> 
> ```
> ⚡️  To Native ->  Example removeAllListeners 17752391
> ⚡️  Error calling method removeAllListeners on plugin Example: No method found.
> ⚡️  Ensure plugin method exists and uses @objc in its declaration, and has been defined
> ```

It seems you are missing the line `CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnPromise);` in your `test-plugin/ios/Plugin/ExamplePlugin.m` file, that will lead to the error you are seeing

IT-MikeS:
I agree its odd, and I believe it is going to be looked at for Capacitor v6

jcesarmobile (approved):
I've verified that adding this change is not breaking to plugins that already have declared the `CAP_PLUGIN_METHOD(removeAllListeners, CAPPluginReturnPromise);`. But still marking it at breaking so we document that plugins should remove that selector now.

The comment about `removeAllListeners` not needing the `:` is incorrect as `removeAllListeners` also expects a `CAPPluginCall` as `addListener` and `removeListener`. The pull request is correct as is.

IT-MikeS (approved):

PR #6844, opened 2023-08-28
chore: bump version to 6.0.0-alpha.0

jcesarmobile:
last time lerna wasn't able to bump the major version for the first alpha, so doing it manually so we can release alphas


markemer:
I think lint is failing since we haven't published 6 alpha's yet.

jcesarmobile:
yeah, we shouldn't merge this until we are about to release the first alpha. 
The publish script calls `sync-peer-dependencies` which would set the peer dependencies to also 6.0.0-alpha.0 and once they are published it will stop erroring

IT-MikeS (approved):

giralte-ionic (approved):

PR #6858, opened 2023-09-02
Add onUpdateVisitedHistory callback #6857

### LINKED ISSUE #6857: feat: Propagate the doUpdateVisitedHistory callback from WebViewClient
### Description
Currently, the Android `com.getcapacitor.BridgeWebViewClient` does not override the `doUpdateVisitedHistory` callback to its listeners. This knowledge should be able to the `WebViewListener`s, so they can (optionally) react, update their plugin functionality or act on native code

### Platform(s)
This is an android platform-specific call for the WebViewClient

### Preferred Solution
Add to BridgeWebViewClient:
```
@Override
public void doUpdateVisitedHistory [...]
    for (WebViewListener listener : bridge.getWebViewListeners()) {
        listener.onUpdateVisitedHistory(view.canGoBack());
    }
```
Add to WebViewListener:
```
/**
* Callback for page-visit changes that should lead to the user having different browsing state
* and history.
* 
* @param canGoBack Whether or not the user can go further back in the visit history
*/
public void onUpdateVisitedHistory(boolean canGoBack) {
    // Override me to add behavior to the page-visit history update
}
```

### Alternatives
See context below, all alternatives that provide the same level of support for plugins would involve a lot of custom code with separate fields that know the state of the WebView navigation stack

### Additional Context
This change will soon be required to improve the `OnBackPressedDispatcher` / `OnBackInvokedDispatcher` handling in Android. Android 13 already introduced [Support for predictive back-gestures](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture) when leaving the activity, and claims this will be fleshed out in the coming years with support for predictive back in fragment transitions. The change proposed in this feature request comes from the fact that their [CodeLab](https://codelabs.developers.google.com/handling-gesture-back-navigation#5) uses the `doUpdateVisitedHistory` callback, so for example the [Capacitor App Plugin](https://github.com/ionic-team/capacitor-plugins/tree/main/app) will sorely need this callback soon

avalanchas:
- Add callback to WebViewListener
- Call callback from BridgeWebViewClient
- fix variable usage in for loops to avoid second get call

fixes #6857

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
@@ -40,7 +40,7 @@ public class BridgeWebViewClient extends WebViewClient {
         List<WebViewListener> webViewListeners = bridge.getWebViewListeners();
 
         if (webViewListeners != null && view.getProgress() == 100) {
-            for (WebViewListener listener : bridge.getWebViewListeners()) {
+            for (WebViewListener listener : webViewListeners) {
                 listener.onPageLoaded(view);
             }
         }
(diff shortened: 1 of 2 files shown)


eric-horodyski:

More context:

PR #6868, opened 2023-09-06
fix(http): add support for Request objects in fetch

### LINKED ISSUE #6174: bug: CapacitorHTTP fetch with Request object fails

ItsChaceD:
Closes #6174 

This PR fixes fetch requests failing when using a `Request` object as the resource


markemer (approved):

markemer:
We'll need a 5.x version of this one too.

PR #6872, opened 2023-09-07
fix: Update migrate to Capacitor 6

jcesarmobile:
will still need updates, but it's ready for the first alpha migration


markemer (approved):
A lot simpler this time.

IT-MikeS (approved):

PR #6874, opened 2023-09-07
refactor(cli)!: remove definitions to decouple fedcap and liveupdates plugins

carlpoole:
Same as #6870 but for 5.x


giralte-ionic (approved):

jcesarmobile:
this is a breaking change, so I don't think we can backport it to capacitor 5

PR #6876, opened 2023-09-08
Add method to disable autosuggestion

tyilo:
We have a use-case where our app is run on a tablet in a public setting.

Thus we would like to disable autosuggestion to prevent the Android keyboard from remembering and suggesting emails of previous users.


eric-horodyski:

More context:

PR #6878, opened 2023-09-08
fix(cli): use helper in Podfile with correct path

IT-MikeS:


giralte-ionic (approved):

cli/src/ios/update.ts
```diff
@@ -103,6 +104,9 @@ async function updatePodfile(
   );
   podfileContent = podfileContent.replace(
     `require_relative '../../node_modules/@capacitor/ios/scripts/pods_helpers'`,
```
jcesarmobile, line 106: this should be a regular expression as the users might have changed the path already and might need adjustments later on

cli/src/ios/update.ts
```diff
@@ -158,7 +163,7 @@ end`,
 async function generatePodFile(
   config: Config,
   plugins: Plugin[],
-): Promise<string> {
+): Promise<{ dependenciesContent: string; relativeCapacitoriOSPath: string }> {
```
jcesarmobile, line 166: better create a function for getting the relativeCapacitoriOSPath and use it where needed, this function should not return it

jcesarmobile (changes requested):
This should be reverted to use the default relative path instead of the block
https://github.com/ionic-team/capacitor/blob/main/ios-template/App/Podfile#L1-L12

jcesarmobile (approved):

markemer (approved):

PR #6894, opened 2023-09-12
feat!(ios): Remove CAPWebView from Capacitor

Steven0351:
feat!(ios): Remove Portals-specific CAPWebView from Capacitor. Make various APIs public so that is possible to construct outside of Capacitor itself


ios/Capacitor/Capacitor/CAPBridgeDelegate.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-internal protocol CAPBridgeDelegate: AnyObject {
+public protocol CAPBridgeDelegate: AnyObject {
```
markemer, line 3: I really want to get rid of this class, for cap 6 maybe? I don't really see the point of it. We should at least change the name, it's not a delegate.

ios/Capacitor/Capacitor/CAPBridgeViewController.swift
```diff
@@ -92,7 +92,7 @@ import Cordova
     }
 
     open func router() -> Router {
-        return _Router()
+        return CapacitorRouter()
```
markemer, line 95: 👍

markemer (approved):
I like it a lot - maybe want to change that protocol, but that should be a separate PR.

markemer:
@Steven0351 You need to check in the changed project file - that's why the build is failing.

ios/Capacitor/Capacitor/CAPBridgeDelegate.swift
Steven0351: Yeah, and it was also internal prior to this tbh

Steven0351:
> @Steven0351 You need to check in the changed project file - that's why the build is failing.

I actually forgot to check-in the changes to the pbxproj

ios/Capacitor/Capacitor/CAPBridgeDelegate.swift
giralte-ionic: @markemer you want to make that change in a separate PR?

giralte-ionic (approved):

ios/Capacitor/Capacitor/CAPBridgeDelegate.swift
markemer: I think we actually might just need to rename it, I'm looking into how we use this now.

Issue #6856, opened 2023-08-31
bug: Android CapacitorCookies.getCookies ignores url parameter


nickredding:
### Capacitor Version
5.2.3 and 5.3.0

### Current Behavior

Prior to 5.2.3, the CapacitorCookies plugin getCookies returned the cookies associated with the url domain provided as a parameter.

As of 5.2.3 this plugin entry point ignores the url domain provided as a parameter and returns ALL cookies in document.cookie. This was done to exclude HTTPONLY cookies (https://github.com/ionic-team/capacitor/pull/6787).

Unfortunately, this change prevents retrieval of cookies from a specific domain since document.cookie includes cookies from all of the domains that have set cookies.

The solution should have avoided this by (a) retrieving the cookies associated with the provided domain using the original code, and (b) removing the cookies from the retrieved list that do not appear in document.cookie.

That way, only cookies set by the supplied domain that were not HTTPONLY would be returned.

nickredding:
See https://github.com/nickredding/bug6792 which has been extended to demonstrate this problem.

nickredding:
The fix is to modify getCookies in the module CapacitorCookies.java as follows

`
    @PluginMethod
    public void getCookies(PluginCall call) {
        String url = call.getString("url");
        HttpCookie[] urlCookies = cookieManager.getCookies(url);
        this.bridge.eval(
                "document.cookie",
                value -> {
                    String cookies = value.substring(1, value.length() - 1);
                    String[] cookieArray = cookies.split(";");

                    JSObject cookieMap = new JSObject();

                    for (String cookie : cookieArray) {
                        if (cookie.length() > 0) {
                            String[] keyValue = cookie.split("=", 2);

                            if (keyValue.length == 2) {
                                String key = keyValue[0].trim();
                                String val = keyValue[1].trim();
                                try {
                                    key = URLDecoder.decode(keyValue[0].trim(), StandardCharsets.UTF_8.name());
                                    val = URLDecoder.decode(keyValue[1].trim(), StandardCharsets.UTF_8.name());
                                } catch (UnsupportedEncodingException ignored) {}
                                for (HttpCookie urlCookie : urlCookies) {
                                    if (urlCookie.getName().equals(key)) {
                                        cookieMap.put(key, val);
                                        break;
                                    }
                                }
                            }
                        }
                    }
                    call.resolve(cookieMap);
                }
            );
    }`

nickredding:
[...]

PR #6899, opened 2023-09-13
fix(cookies): retrieve cookies when using a custom android scheme

ItsChaceD:
This PR fixes a bug introduced after #6787  where cookies are unable to be retrieved at all if using a custom scheme on Android. This is due to WebViews inherently not considering custom schemes as eligible for cookie storage. Therefore, calling `document.cookie` on the WebView was a no-op. However, since the Android cookie store provides no way to access stored cookie properties, there is no way to filter out insecure `httpOnly` cookies when returning them to the WebView. This will add `androidCustomSchemeAllowInsecureAccess` as a config option, which will return ALL cookies from the cookie store at the given url, if set to true.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/CapacitorCookies.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/CapacitorCookies.java
@@ -39,6 +39,11 @@ public class CapacitorCookies extends Plugin {
         return pluginConfig.getBoolean("enabled", false);
     }
 
+    private boolean isAllowingInsecureCookies() {
+        PluginConfig pluginConfig = getBridge().getConfig().getPluginConfiguration("CapacitorCookies");
+        return pluginConfig.getBoolean("androidCustomSchemeAllowInsecureAccess", false);
+    }
+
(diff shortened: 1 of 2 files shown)


IT-MikeS (approved):

nickredding:
One comment on this PR: you really should retain the check that I suggested when returning cookies from document.cookie to ensure the cookie was in fact set against the URL provided. Otherwise, I think it's fair to say this code is incorrect in retuning ALL cookies regardless of the URL they were set against.

markemer (approved):

nickredding:
I’ll say it again: this code is incorrect in returning all cookies from document.cookie regardless of the url they were set against. The fix is easy. Why the refusal to make it correct?

ItsChaceD:
> I’ll say it again: this code is incorrect in returning all cookies from document.cookie regardless of the url they were set against. The fix is easy. Why the refusal to make it correct?

#6856 , along with the other issues that you have mentioned, are being investigated and will be addressed in separate PRs. This PR is just for adding the `androidCustomSchemeAllowInsecureAccess` flag

PR #6900, opened 2023-09-13
refactor: move the Xcode 15 patch from the template to @capacitor/ios

jcesarmobile:
`assertDeploymentTarget` was moved back to the `pods_helpers.rb` file, but the Xcode 15 patch is still in the users `Podfile`. So better move the code to `@capacitor/ios` so we can remove or edit it if needed without requiring users to edit their code or without the need of making migrate command to write that block of code for the Capacitor 6 migration.


giralte-ionic (approved):

markemer (approved):

markemer:
Yeah, we can make a quick edit, that's mostly for us.

PR #6903, opened 2023-09-13
fix: allow double quotes in Gemfile

### LINKED ISSUE #6854: bug: bundle support fails to update cocoapod dependencies if double quotes used in Gemfile
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 5.3.0
  @capacitor/core: 5.3.0
  @capacitor/android: 5.3.0
  @capacitor/ios: 5.3.0

Installed Dependencies:

  @capacitor/cli: 5.2.2
  @capacitor/android: 5.2.2
  @capacitor/core: 5.2.2
  @capacitor/ios: 5.2.2

[info] Using Gemfile: RubyGems bundle installed
[error] Missing <manifest package=""> attribute in app/src/main
```

### Platform(s)

* iOS

### Current Behavior

If a Gemfile is present that references cocoapods with double quotes, iOS dependencies will not be installed.

FMI: https://github.com/ionic-team/capacitor/pull/5205#issuecomment-1701593642

### Expected Behavior

Double quotes are valid in Gemfiles, so the test to enable bundler usage should work regardless.

### Code Reproduction

- Initialize a Capacitor iOS project and add a Gemfile that references cocoapods with double quotes, like:

```
gem "cocoapods", "~> 1.12.1"
```

### Other Technical Details

`npm --version` output: 9.5.1

`node --version` output: v18.16.0

`bundle exec pod --version` output (iOS issues only): 1.12.1

### Additional Context

https://github.com/ionic-team/capacitor/pull/5205#issuecomment-1701593642

markemer:
Fixes #6854


markemer:
I think I want to make this more general to allow for a bunch of other spacing.

giralte-ionic (approved):

cli/src/config.ts
```diff
@@ -500,7 +500,9 @@ async function determineGemfileOrCocoapodPath(
     if (!gemfileText) {
       return 'pod';
     }
-    const cocoapodsInGemfile = new RegExp(/gem 'cocoapods'/).test(gemfileText);
+    const cocoapodsInGemfile = new RegExp(/gem\s+['"]cocoapods/).test(
```
IT-MikeS, line 503: This doesn't test for the existence of an end quote. Perhaps something more like: `/gem\s('cocoapods'|"cocoapods")/` that way it test not only existence of both quotes but that the quotes match on each end

IT-MikeS (changes requested):

cli/src/config.ts
markemer: Yeah I left that out on purpose just to match, but I guess if it's missing, there is an issue. I don't care so much that it's syntactically correct, just that I see it in there. If the Gemfile syntax is borked, someone needs to fix it.

markemer: Also, I added `\s+` so it works no matter how much space you put on the line. I ran it with a bunch of space between gem and cocoapods and it worked.

markemer: So I added a closing character class, they may not match but I don't know how much I care about that. `bundle exec` won't run this file.

IT-MikeS: I still believe checking for the quotes and that they match would be good, the `\s+` makes sense if gem doesn't care about extra whitespace. However as long as this works its all good either way

IT-MikeS (approved):

PR #6946, opened 2023-09-28
feat(android): allow developers to provide logic for onRenderProcessGone in WebViewListener

carlpoole:
Portals developer app was crashing because they don't have a way to override this method in the WebViewClient in Capacitor. This adds the method to the interface so that it can be overridden via Portals. Not a breaking change.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
@@ -102,7 +102,7 @@ public class BridgeWebViewClient extends WebViewClient {
         List<WebViewListener> webViewListeners = bridge.getWebViewListeners();
         if (webViewListeners != null) {
             for (WebViewListener listener : bridge.getWebViewListeners()) {
-                result = result || listener.onRenderProcessGone(view, detail);
+                result = listener.onRenderProcessGone(view, detail) || result;
             }
         }
 
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
@@ -97,15 +97,15 @@ public class BridgeWebViewClient extends WebViewClient {
     @Override
     public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
         super.onRenderProcessGone(view, detail);
+        boolean result = false;
+
         List<WebViewListener> webViewListeners = bridge.getWebViewListeners();
         if (webViewListeners != null) {
             for (WebViewListener listener : bridge.getWebViewListeners()) {
-                if (listener.onRenderProcessGone(view, detail)) {
-                    return true;
-                }
+                result = result || listener.onRenderProcessGone(view, detail);
             }
         }
 
-        return false;
+        return result;
     }
 }
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
@@ -98,13 +98,12 @@ public class BridgeWebViewClient extends WebViewClient {
     public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
         super.onRenderProcessGone(view, detail);
         List<WebViewListener> webViewListeners = bridge.getWebViewListeners();
-
         if (webViewListeners != null) {
             for (WebViewListener listener : bridge.getWebViewListeners()) {
-                listener.onRenderProcessGone(view, detail);
+                if (listener.onRenderProcessGone(view, detail)) {
+                    return true;
+                }
             }
-
-            return true;
         }
 
         return false;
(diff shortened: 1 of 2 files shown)


jcesarmobile:
The method should return true only if the app is "handling the situation", but there is no guarantee that the app is handling the situation just by having listeners as the listeners are an array of any WebView listeners, not just the onRenderProcessGone listeners, so it could have a listener for other events and not for onRenderProcessGone.
In fact there is a bug in all the webViewListeners checks since they are never null but an empty ArrayList.
I think that could be fixed by changing the return of onRenderProcessGone from the WebViewListener to boolean and check if any of the listeners returned true.

What's the plan for handling the onRenderProcessGone on portals side? could we handle it in Capacitor so portals doesn't have to?

There is an alternate plugin PR where the handling is forwarded to plugins and they can return true if they are handling onRenderProcessGone instead of using listeners. That's similar to how we handle shouldOverrideLoad https://github.com/ionic-team/capacitor/blob/7e7c8b9113f541d530c5883dea1f52b2957c0859/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L343-L345
https://github.com/ionic-team/capacitor/pull/6416

Also, as an alternative if this is going to be a portals only thing, portals could use their own WebViewClient that extends BridgeWebViewClient and implements onRenderProcessGone and that would require no changes in Capacitor.

carlpoole:
@jcesarmobile I like the solution in #6416 but it only provides access to that one specific WebViewClient method on the `Plugin` class, when the plugin could in theory just call `addWebViewListener` on the Bridge as well https://github.com/ionic-team/capacitor/blob/7e7c8b9113f541d530c5883dea1f52b2957c0859/android/capacitor/src/main/java/com/getcapacitor/Bridge.java#L1405 and use anything in the interface as well as `onRenderProcessGone` if it is added there. That WebViewListener can be used in plugins too. 

Im tweaking the solution in this PR to give more control over the true/false return state

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
```diff
@@ -92,4 +93,19 @@ public void onReceivedHttpError(WebView view, WebResourceRequest request, WebRes
...
     }
+
+    @Override
+    public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
+        super.onRenderProcessGone(view, detail);
+        List<WebViewListener> webViewListeners = bridge.getWebViewListeners();
+        if (webViewListeners != null) {
+            for (WebViewListener listener : bridge.getWebViewListeners()) {
+                if (listener.onRenderProcessGone(view, detail)) {
+                    return true;
```
jcesarmobile, line 104: I think it should keep iterating to forward the event to all the listeners even if one of them returned true

jcesarmobile (approved):
LGTM

The only thing is that this should go to main branch first and then cherry picked into 5.x once it's merged (or if it gets merged)

markemer (approved):

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
peitschie: Just noting this feedback hasn't been addressed here yet.
The logical combination will short-circuit, so once a true value is returned, I don't believe any other listener gets informed.

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
```diff
@@ -92,4 +93,19 @@ public void onReceivedHttpError(WebView view, WebResourceRequest request, WebRes
...
+
+    @Override
+    public boolean onRenderProcessGone(WebView view, RenderProcessGoneDetail detail) {
+        super.onRenderProcessGone(view, detail);
+        boolean result = false;
+
+        List<WebViewListener> webViewListeners = bridge.getWebViewListeners();
+        if (webViewListeners != null) {
+            for (WebViewListener listener : bridge.getWebViewListeners()) {
+                result = result || listener.onRenderProcessGone(view, detail);
```
peitschie, line 105: Ideally this would be:
```suggestion
                bool listenerResult = listener.onRenderProcessGone(view, detail);
                result = result || listenerResult;
```

jcesarmobile: or maybe like this
```suggestion
                result = listener.onRenderProcessGone(view, detail) || result;
```
I like the short version better than declaring a bool on every for iteration, and I think this way will not "short circuit" since the listener check is on the left.

IT-MikeS (approved):

android/capacitor/src/main/java/com/getcapacitor/BridgeWebViewClient.java
peitschie: Agreed! That's definitely neater @jcesarmobile 🙂

carlpoole: Thanks for the catch on the short circuiting!

PR #6956, opened 2023-10-05
refactor(ios): Remove Xcode 14 compat for isInspectable

jcesarmobile:
Since we are requiring Xcode 15 for Capacitor 6, remove the Xcode 14.2 and older workaround


giralte-ionic (approved):

markemer (approved):
This is the first temporary fix I've every written that hasn't been in the code base forever. I love it.

PR #6983, opened 2023-10-10
docs: add warning around androidScheme with Android webview 117

dallastjames:


android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -301,6 +301,13 @@ private boolean validateScheme(String scheme) {
             return false;
         }
 
+        // Non-http(s) schemes are not allowed to modify the URL path as of Android Webview 117
```
dallastjames, line 304: I added this warning to be logged to make this hopefully stand out more. Does this make sense? and do we want to warn people each time?

jcesarmobile:
I think warning every time is ok, I also thought about warning from the CLI, but that might be too much as users might be using hash based navigation and the warning in the CLI could be annoying, but having it in logcat seems ok.

I think the docs should also mention that using a custom scheme is discouraged on Android as Google engineers have noted in the past and maybe link the blog post that has more information https://ionic.io/blog/capacitor-android-customscheme-issue-with-chrome-117

IT-MikeS (approved):

jcesarmobile (approved):

PR #6993, opened 2023-10-14
fix: vue 3 log warning causes error on iOS

### LINKED ISSUE #6615: bug: log vue 3 warn cause error on iOS
## Bug Report

Continue #6545
### Capacitor Version
```
5.02
```

### Platform(s)
iOS

### Current Behavior
No default value!
Just watch my debugging video

https://github.com/ionic-team/capacitor/assets/11452353/1c642056-4c6b-438c-84b0-b08d3029aab9

Emulator:
image

### Expected Behavior
Just see original error from vue instead `No default value`, it's not a vue error.
Capacitor must not skip my console.warn args, it's abuse!

Expected output (safari)
image

### Code Reproduction
https://github.com/bazuka5801/capacitor-error

### Other Technical Details

`npm --version` output: latest

`node --version` output: latest

`pod --version` output (iOS issues only): latest

ciekawy:
fixes #6615
not sure what exactly should be returned when `String` throws error


core/native-bridge.ts
```diff
@@ -336,15 +336,15 @@ const initBridge = (w: any): void => {
...
-        } catch (e) {
-          // ignore
+      try {
+        if (typeof msg === 'object') {
+            msg = JSON.stringify(msg);
         }
+        return String(msg);
+      }
+      catch (e) {
+        return '??';
```
markemer, line 346: This should probably return a blank string, if that will also avoid the error. Or perhaps an error message to the console and a blank string.

markemer (changes requested):
[Suggestion](https://github.com/ionic-team/capacitor/pull/6993/files#r1457693633)

core/native-bridge.ts
ciekawy: yup, seems I left the ?? as was not sure what should be returned. I'm definitely ok with empty string unless there is any better suggestion.

markemer:
Can you run `npm run fmt` as well - appears to be failing lint checks.

ciekawy:
just letting know that unfortunately I'm very time constrained now and for next several weeks. I'd be happy to update the PR, there is a chance I'll be soon touching an app with this and will try to be back (to not left the app blocked I just used `patch-package` at that time)

jcesarmobile (approved):

ciekawy:
@markemer I addressed your suggestions

markemer (approved):

PR #7009, opened 2023-10-20
feat(ios): Makes CapacitorBridge, WebViewAssetHandler, and WebViewDelegationHandler open classes, along with several of their methods

Steven0351:


ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
```diff
@@ -4,29 +4,29 @@ import WebKit
...
         case unloaded
         case initialLoad(isOpaque: Bool)
         case subsequentLoad
     }
+
     fileprivate(set) var webViewLoadingState = WebViewLoadingState.unloaded
 
     private let handlerName = "bridge"
 
-    public init(bridge: CapacitorBridge? = nil) {
```
Steven0351, line 19: This initializer never would have respected the bridge passed in because it is set later.

jcesarmobile (approved):
Changes look good, I've fixed some merge conflicts it had after the zoom feature PR.

But do we want to make all those classes and methods open @dallastjames?

dallastjames:
Based on my swift understanding, it just makes subclassing and overriding these classes simpler for consumers and isn't really exposing any new public API, so I think we're good to do this. Also seems like Portals probably could really make use of this change.

PR #7017, opened 2023-10-24
refactor(ios)!: Remove unused plugin return types

jcesarmobile:
marking as breaking just in case, but those two are iOS only and we don't use them on any of our plugins nor have any special code for handling those types


IT-MikeS (approved):

markemer (approved):

markemer:
Cocoapods keeps erroring weirdly.

PR #7028, opened 2023-10-27
chore(ci): Remove unneeded action

jcesarmobile:
I saw the ci action was showing a warning because this action was for node 12 so I was going to update it, but in the repository the author said that we "probably don't need this action" and that we should use `concurrency` instead, so I was going to replace it with a `concurrency` entry, but [Mike already added it a few months ago,](https://github.com/ionic-team/capacitor/blob/787244dc9a6fe9c6c86092e8cfcecc9ffaf0c488/.github/workflows/ci.yml#L14) so the action shouldn't be needed anymore.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,6 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
-      - name: Cancel Previous Runs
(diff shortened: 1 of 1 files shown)


markemer (approved):
Yep, this has been on my to-do list ever since that node 12 error popped up. I think it just got left behind by accident.

giralte-ionic (approved):

Issue #7035, opened 2023-10-30
bug: Missing Dependencies in Root package.json


uditajmera:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: unknown
  @capacitor/core: unknown
  @capacitor/android: unknown
  @capacitor/ios: unknown

Installed Dependencies:

  @capacitor/cli: 5.4.0
  @capacitor/android: 5.4.0
  @capacitor/core: 5.4.0
  @capacitor/ios: 5.4.0

[info] Using Gemfile: RubyGems bundle installed
[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

iOS, android, web

### Current Behavior

Hi Team,
We’ve created a forked repository of Capacitor. However, when we are attempting to run commands to publish the package, we’re encountering some issues. It appears as though the following dependencies are missing from the root package.json:
"dependencies": {
   "@ionic/utils-process": "^2.1.11",
   "semver": "^7.3.7"
 }

Error on try to publish package:-
src/tasks/run.ts:1:30 - error TS2307: Cannot find module '@ionic/utils-process' or its corresponding type declarations.

1 import { sleepForever } from '@ionic/utils-process';

### Expected Behavior

Following dependencies should be present in package.json

"dependencies": {
   "@ionic/utils-process": "^2.1.11",
   "semver": "^7.3.7"
 }

### Code Reproduction

Try to run publish command on repository, it will ask to add dependencies for ionic/utils-process and semver.

`npm --version` output: 9.8.0

`node --version` output: v18.10.0

`pod --version` output (iOS issues only): 1.12.1

jcesarmobile:
`@ionic/utils-process` is not used in the root, it's only used in `@capacitor/cli`, so should be added to `cli/package.json`, not on the root one.
But `semver` should be in the root too since it's used in `scripts` folder.

For me it doesn't error because they are probably a dependency of another of the dependencies. Also it's not failing on the CI since we are publishing versions regularly.

Why are you trying to run publish? It will fail because the lack of permissions, only we can publish versions.

uditajmera:
@jcesarmobile thanks for reply. As we are facing Appdelegate build issue from Cordova Project on integration with our project. We forked the capacitor repo and trying to publish to our internal npm package. 
Could you please fix this and add dependencies as you men tioned above we wont have any conflicts in future on rebase.

PR #7037, opened 2023-10-30
chore: Add missing dependencies

jcesarmobile:
While the build works because the missing dependencies are dependencies of other dependencies, it could break in the future of those dependencies stop adding our missing dependencies as dependencies, so better add them now to prevent possible future issues.

Closes https://github.com/ionic-team/capacitor/issues/7035


IT-MikeS (approved):

jvliwanag:
Can we backport this to the 5.x branch?

Currently, capacitor/cli 5.x breaks on pnpm where hoisting is set to false.

PR #7080, opened 2023-11-19
Fix cors errors with local files when developing / live reloading on android

DominikTrenz:
This fixes cors errors on android that happen if you try to load local files while serving with live reload. Something like that:
https://ip:8102/_capacitor_file_/data/user/0/package.name/files/7a44b042-2371-41f5-9c84-56e11fdd1d40.mp3


jcesarmobile:
how do you do live reload?
There shouldn't be CORS problems on Android when doing live reload because the live reload IP and port is the same for the live reload and for the local files.

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

Issue #6215, opened 2023-01-14
bug: (jarsigner: unable to open jar file) error when using android flavors


sabereen:
### Capacitor Version

```
Latest Dependencies:

  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1
  @capacitor/ios: 4.6.1

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 4.6.1
  @capacitor/core: 4.6.1
  @capacitor/android: 4.6.1

[success] Android looking great! �
```

### Platform(s)

android

### Current Behavior

It is not signed when I use android flavors.
```
× Signing Release - failed!
[error] jarsigner: unable to open jar file:
        C:\Users\MQT\projects\microblog\android\app\build\outputs\apk\release\app-FLAVOR-release-unsigned.apk
```

### Expected Behavior
My APK is in this path:
```
C:\Users\MQT\projects\microblog\android\app\build\outputs\apk\FLAVOR\release\app-FLAVOR-release-unsigned.apk
```

So jarsigner should sign this path instead.

### Code Reproduction
Define android build falvors and build project using a flavor.

### Other Technical Details
`npm --version` output:
yarn 1.22.19

`node --version` output:
v16.15.1

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

Ionitron:
It looks like this issue didn't get the information it needed, so I'll close it for now. If I made a mistake, sorry! I am just a bot.

Have a great day!
Ionitron 💙

PR #7082, opened 2023-11-20
Bugfix/flavor signing

bosh-code:
When constructing the unsigned .apk/.aab name, the const `flavor` declared on line 16 should be used and not `config.android.flavor` as this value may be null even though a flavor has been provided

This issue was raised previously for it: https://github.com/ionic-team/capacitor/issues/6215

### DIFF
--- a/cli/src/android/build.ts
+++ b/cli/src/android/build.ts
@@ -44,7 +44,7 @@ export async function buildAndroid(
   );
 
   const unsignedReleaseName = `app${
-    config.android.flavor ? `-${config.android.flavor}` : ''
+    flavor !== '' ? `-${flavor}` : ''
   }-release${releaseTypeIsAAB ? '' : '-unsigned'}.${releaseType.toLowerCase()}`;
 
   const signedReleaseName = unsignedReleaseName.replace(


bosh-code:
@markemer any thoughts? As it stands, `cap build android --flavor=foo` doesn't work correctly.

markemer:
Thanks for the contribution. This does seem like a bug that needs fixing - can you run `npm run fmt` so CI checks pass?

bosh-code:
> Thanks for the contribution. This does seem like a bug that needs fixing - can you run `npm run fmt` so CI checks pass?

Done, thanks @markemer

bosh-code:
> can you provide a sample app that reproduces the issue? the issue you referenced got closed because it got no sample app, so providing a fix without providing a way of reproducing doesn't help much.

Sure thing, repo here: https://github.com/bosh-code/android-flavor-example
I added a couple demo flavors and scripts to test all the arguments.

As you can see, the flavor is ignored and all flavors are built, as well as all cases fail to sign. You can use the keystore credentials in the script to run a build through android studio and verify they are correct and it is just that the path of the unsigned apk/bundle is incorrect. 

In the original project I encountered this bug in I am using Gradle DSL and can confirm the behavior is the same.

Please let me know if there is anything else I can help with.

jcesarmobile (changes requested):
Looks like the new code breaks for AAB option, seems like AAB uses the `${flavor}Release` path while APK uses `${flavor}/release` path.

I have not tested on windows yet, but I think appending the `/release` for APK path may lead to bad windows paths, using `join` should be preferred.

bosh-code:
> Looks like the new code breaks for AAB option, seems like AAB uses the `${flavor}Release` path while APK uses `${flavor}/release` path.
> 
> I have not tested on windows yet, but I think appending the `/release` for APK path may lead to bad windows paths, using `join` should be preferred.

Good catch, thank you. I do not have access to a Windows machine, but I suspect you are right. I have added a bit of logic that should fix it. Thanks

bosh-code:
@jcesarmobile could you please re-review? Thanks

jcesarmobile (approved):
Looks good, I've verified that it works on windows too

markemer (approved):
Works for me on mac - haven't tried windows but @jcesarmobile has

PR #7096, opened 2023-11-24
fix(ios): add some new cordova-ios classes used by Cordova plugins

bosh-code:
The cordova plugin: [cordova-plugin-inappbrowser](https://github.com/apache/cordova-plugin-inappbrowser) was recently updated to [6.0.0](https://github.com/apache/cordova-plugin-inappbrowser/releases/tag/rel%2F6.0.0).

On iOS it now requires an additional file from CordovaLib that Capacitor does not currently provide, see: [changes](https://github.com/apache/cordova-plugin-inappbrowser/compare/rel/5.0.0...rel/6.0.0) and the [file](https://github.com/apache/cordova-plugin-inappbrowser/compare/rel/5.0.0...rel/6.0.0#diff-3a6f80dde729cc2827f9bb45fed1ba49af846f71424cc0567c754309ea1798dc)

This PR adds the required files (copied from [here](https://github.com/apache/cordova-ios/blob/master/CordovaLib/Classes/Public/CDVWebViewProcessPoolFactory.m)) to `CapacitorCordova` ios.

I have tested this on an existing project that uses `cordova-plugin-inappbrowser`, by manually editing the `CapacitorCordova` development pod to include these files and it works correctly.


bosh-code:
Example repo here: https://github.com/bosh-code/cordova-iab-6-example
Just run the `run:ios` script

jcesarmobile (approved):

Issue #2022, opened 2019-10-02
bug: Capacitor Attempting Multiple Injections On Android


brandonhudson:
# Bug Report
Capacitor is attempting multiple javascript injections on Android. This is occuring when an app references separate HTML templates after the app is loaded in `index.html`. In our case, we load HTML templates dynamically using AngularJS via a UI router. Each time one of these resources is requested, Capacitor attempts to reinject the javascript, but fails because these files do not have `<head>` tags in them. This error is occuring at the following line:

https://github.com/ionic-team/capacitor/blob/master/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java#L245

Example:
```
D/Capacitor: Handling local request: http://localhost/assets/templates/views/_home.html
E/Capacitor: Unable to inject Capacitor, Plugins won't work
```

To resolve this, I was able to add the following check at line 245:
```
if (path.equals("/index.html")) {
        responseStream = jsInjector.getInjectedStream(responseStream);
}
````

This ensures that the injection only occurs on the base `index.html` page. Unsure if this is a proper universal solution, as this only works for cases where the app is intended to be run as a single page application.

A better solution would be to check if the javascript has been previously injected. If this is the case, and the HTML file requested does not have a header tag, downgrade from an error to a warning or info log.

## Capacitor Version

`npx cap doctor` output:
```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 1.2.1

  @capacitor/core: 1.2.1

  @capacitor/android: 1.2.1

  @capacitor/ios: 1.2.1

Installed Dependencies:

  @capacitor/cli 1.1.1

  @capacitor/android 1.2.1

  @capacitor/core 1.1.1

  @capacitor/ios 1.1.1


[success] Android looking great! 👌
  Found 2 Capacitor plugins for ios:
    capacitor-fingerprint-auth (0.0.2)
    phonegap-plugin-barcodescanner (8.1.0)
[success] iOS looking great! 👌
```

## Affected Platform(s)

- [X] Android
- [ ] iOS
- [ ] Electron
- [ ] Web

## Current Behavior

See above comments.

## Expected Behavior

Capacitor should not error each time the JS is not injected. It should also not attempt to inject multiple times (if possible)

## Other Technical Details

`npm --version` output: 

`node --version` output:

`pod --version` output (iOS issues only): N/A

jcesarmobile:
[...]

PR #7135, opened 2023-12-15
feat(android)!: Use addDocumentStartJavaScript for injecting javascript

jcesarmobile:
Marking it as breaking because
- it requires `androidx.webkit:webkit` 1.9.0 or newer
- is might cause some behavior changes since the way the javascript is injected changes (I've tested on the capacitor test app without any weird problems so far) 

Solves problems like:
- javascript injection not working on service workers
- javascript injection not working if using `server.url` and the certificate has problems
- javascript being injected in all `.html` files, even if it's a file being get usin fetch/XHR
- javascript not being injected if there is no `<head>` tag

closes https://github.com/ionic-team/capacitor/issues/2022
closes https://github.com/ionic-team/capacitor/issues/5278


jcesarmobile:
ci fails because we are using 1.8.0 at the moment, should be fixed after merging https://github.com/ionic-team/capacitor/pull/7136

carlpoole (approved):

markemer (approved):

martin-braun:
But ... how do I use this? I come from #5278, but none of the solutions work on v6. I believe this should work without a service worker now?

I load a local index.html to do some stuff, then I load a remote website that uses Capacitor plugins. My Android device reports "implementation unavailable" and my platform is "web" despite being on Android. I can, however, control native functionality from the index.html.

Has this been documented?

PR #7138, opened 2023-12-15
chore(ios): format the new codable classes

jcesarmobile:
running `npm run fmt` with swiftlint 0.53.0 does this changes, not sure if it was missed or was using a different version


Steven0351 (approved):

Steven0351:
> running `npm run fmt` with swiftlint 0.53.0 does this changes, not sure if it was missed or was using a different version

Looks like I'm on swiftlint version 0.46.2. I know I ran that formatter once or twice, but it's possible I missed it after I refactored it more

jcesarmobile:
It's not very serious since `npm run lint` doesn't complain about the code, but it's annoying to have all those pending changes when I run `npm run fmt` in unrelated parts of the code

giralte-ionic (approved):

PR #7139, opened 2023-12-15
chore(ci): Remove publish-xcframework.yml action

jcesarmobile:
It has never been fired and references the native-publish branch which is no longer used for any other work


theproducer (approved):

giralte-ionic (approved):

Steven0351 (approved):

markemer:
Yeah - thanks for this, I've been meaning to do it, since I'm going to need to replace it with something else possibly soon.

Issue #5257, opened 2021-11-19
bug: Capacitor silently crashes if no Android WebView is available


grigorig:
### Platform(s)
Android

### Current Behavior
If no WebView implementation is available on Android, Capacitor silently crashes on app startup. You can test this by disabling WebView "apps" in Android settings. Not all phones allow you to disable the Android System WebView, though.

### Expected Behavior
Capacitor should display a warning of some sort that advises affected users what's wrong.

jase88:
Ideally the user gets a hint what he/she needs to do to fix this.

Is this catchable on native side or is a change to the framework itself mandatory?

sarahnx:
Any update on this issue?

opindia123:
Any update on this.

alex-steinberg:
The whole "do something async before Capacitor launches the webview" seems a useful feature, and might help with [this bug](https://github.com/ionic-team/capacitor/issues/6974).

PR #7141, opened 2023-12-18
feat(android): Load alternative layout when no WebView

jcesarmobile:
Add a basic layout with a text error message for when there is no WebView installed on the device or is disabled.

Users can use their own text by adding a `<string name="no_webview_text">custom text</string>` entry in their `strings.xml`, or create their own `no_webview.xml` layout that would replace Capacitor's default one.

Added some `bridge != null` checks as if the app fails to create the WebView, the `bridge` will be null and crash.

closes https://github.com/ionic-team/capacitor/issues/5257


giralte-ionic (approved):
good stuff and should be in cap6 for sure

markemer (approved):

Issue #7160, opened 2023-12-25
bug: CapacitorHttp's patched XMLHttpRequest is case-sensitive to HTTP headers


SillyButt:
### Capacitor Version
```
  @capacitor/cli: 5.5.1
  @capacitor/android: 5.5.1
  @capacitor/core: 5.5.1
```

### Current Behavior
When dealing with patched by CapacitorHttp `XMLHttpRequest` API, the patch doesn't consider `Content-Type` HTTP header to be in lowercase, skipping the important role of stringifying the content for `responseText` property of XMLHttpRequest's response object. This leads to unexpected results and runtime errors for 3rd party libraries (like LaunchDarkly JS SDK and I assume infinite amount of other libraries) whenever the server responds with lower-case HTTP header (which is totally valid).
Moreover, the [HTTP2 protocol requires HTTP headers to be in lower-case](https://blog.yaakov.online/http-2-header-casing/), forcing CapacitorHTTP to be not compatible with the protocol.

### Expected Behavior
`responseText` property of the XMLHttpRequest response should not be affected by `Content-Type` header case sensitivity, whether it's `content-type: application/json` or `Content-Type: application/json`.

### Code Reproduction
Please follow the https://github.com/SillyButt/capacitor-http-header-sensitivity repository's README file with simple steps. 
Or launch a server that responds with `content-type: application/json` HTTP header and JSON, call it with next code: 
```
try {
    const req = new XMLHttpRequest();
    function responseCallback() {
      const raw = req.responseText; // While req.response is stringified, not responseText should also be.
      const parsed = JSON.parse(raw); // Will cause error
    }
    req.addEventListener("load", responseCallback);
    req.open("GET", "http://localhost:8000/content-type-lower-case");
    req.send();
  } catch (error) {
    console.error('ERROR:', error);
  }
```

### Other Technical Details
`npm --version` output: `9.6.7`
`node --version` output: `18.17.1`

### Additional Context
[Repository with reproducible problem](https://github.com/SillyButt/capacitor-http-header-sensitivity)
[`XMLHttpRequest` patch with missing lowercase check](https://github.com/ionic-team/capacitor/blob/b8c2a92390f21fe653563d7bad73a3f2b5628acb/core/native-bridge.ts#L749).
[`fetch` patch with covered lowercase check](https://github.com/ionic-team/capacitor/blob/b8c2a92390f21fe653563d7bad73a3f2b5628acb/core/native-bridge.ts#L552)
[...]

PR #7161, opened 2023-12-25
fix(core): make 'content-type' header count for XMLHttpRequest patch

SillyButt:
Fixes error when lower-case 'content-type' header isn't covered in logic that stringifies the response but 'Content-Type' does. Not all servers/endpoints return camel-case leading to issues with 3rd party libraries that try to parse `responseText` property of the XMLHttpResponse' response object.

Closes: #7160


Shapil:
Can someone review and merge this? 
We have an issue with 3rd party lib (launch darkly) that sends lower-cased headers and capacitor fails the request because of lower-case

ItsChaceD (approved):
LGTM, thank you!

markemer (approved):

giralte-ionic (approved):

PR #7187, opened 2024-01-12
fix: stop crashing from `objc_getClassList`

markemer:


ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -5,6 +5,22 @@ import Cordova
 
 internal typealias CapacitorPlugin = CAPPlugin & CAPBridgedPlugin
 
+#warning("remove this before merging! To test for crash")
```
giralte-ionic, line 8: I'm guessing you wanted to remove this?

giralte-ionic (changes requested):
please also run a fat

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -277,30 +293,28 @@ open class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+                        }
                     }
                 }
+            } catch {
+                print(error)
+            }
+        }
+
+        for plugin in pluginList {
+            if let capPlugin = plugin as? CapacitorPlugin.Type {
```
Steven0351, line 315: We still need to keep this check
```swift
if plugin is CAPInstancePlugin.Type { continue }
```

ios/Capacitor/Capacitor/CapacitorBridge.swift
markemer: Pulled it before I forget for real.

giralte-ionic (approved):

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -277,30 +281,32 @@ open class CapacitorBridge: NSObject, CAPBridgeProtocol {
...
+
+        for plugin in pluginList {
+            if plugin is CAPInstancePlugin.Type { continue }
+            if let capPlugin = plugin as? CapacitorPlugin.Type {
+                registerPlugin(capPlugin)
             }
-            classes.deallocate()
-        } else {
-            // register core plugins only
-            [CAPHttpPlugin.self, CAPConsolePlugin.self, CAPWebViewPlugin.self, CAPCookiesPlugin.self]
```
Steven0351, line 302: It looks like we may not be registering the core plugins after this block has been removed.

theproducer (approved):

Steven0351 (approved):

PR #7191, opened 2024-01-16
feat(ios): Replace usage of UserDefaults with KeyValueStore.

Steven0351:
This will prevent every single Capacitor user from having to declare use of a required reason API when it becomes required by Apple. This is also made available for plugin authors or others who utilize Capacitor as a dependency.

### DIFF
--- a/ios/Capacitor/Capacitor/KeyValueStore.swift
+++ b/ios/Capacitor/Capacitor/KeyValueStore.swift
@@ -258,7 +258,8 @@ private class FileStore: KeyValueStoreBackend {
     // so we don't have a scenario where two separate instances may be writing to
     // the same files.
     static func with(name: String) -> FileStore {
-        let library = try! FileManager
+        if let existing = instances[name] { return existing }
+        guard let library = try? FileManager
             .default
             .url(
                 for: .libraryDirectory,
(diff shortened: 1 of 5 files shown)


ios/Capacitor/Capacitor/KeyValueStore.swift
```diff
@@ -0,0 +1,323 @@
...
+            )
+
+        let url = library.appendingPathComponent("kvstore").appendingPathComponent(name)
+        if let existing = instances[url.absoluteString] { return existing }
+
+        // Create the folder if it doesn't exist
+        try? FileManager.default.createDirectory(at: url, withIntermediateDirectories: true, attributes: nil)
+
+        let new = FileStore(baseUrl: url)
+        instances[url.absoluteString] = new
```
Steven0351, line 277: Honestly not sure why I'm keying off the URL instead of just the name. That would allow for checking for its existence before doing all the other url logic.

markemer (approved):
We already talked about this a bunch, but it also works when I run it, so that's good.

jcesarmobile:
This will break live updates for users using `cordova-plugin-ionic`, because the plugin is storing in UserDefaults the values being read by Capacitor, so if the plugin still uses UserDefaults but Capacitor don't read those values then the plugin will not work.
Do we plan to force users to use the new Capacitor live updates plugin with Capacitor 6 instead of `cordova-plugin-ionic`? 
If not, this is not something we can merge.

Steven0351:
> This will break live updates for users using `cordova-plugin-ionic`, because the plugin is storing in UserDefaults the values being read by Capacitor, so if the plugin still uses UserDefaults but Capacitor don't read those values then the plugin will not work. Do we plan to force users to use the new Capacitor live updates plugin with Capacitor 6 instead of `cordova-plugin-ionic`? If not, this is not something we can merge.

cordova-plugin-ionic is not actually reading any values being set by Capacitor core in this case. For the serverBasePath stuff it actually goes through `Ionic.WebView` https://github.com/search?q=repo%3Aionic-team%2Fcordova-plugin-ionic+serverBasePath&type=code

Steven0351:
> Ah, sorry, I was thinking about `lastBinaryVersionCode` and `lastBinaryVersionName`, not `serverBasePath`, but looks like they are not stored by `cordova-plugin-ionic` neither, it stores `binaryVersionName` and `binaryVersionCode`, without `last` prefix, so they are different things and we should be safe.

No worries, I immediately thought of cordova-plugin-ionic when I started this work and tried to be very thorough in my vetting of its uses.

giralte-ionic (approved):

PR #7192, opened 2024-01-16
fix: remove duplicates from auto registration array using set

markemer:


theproducer (approved):

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -288,7 +288,13 @@ open class CapacitorBridge: NSObject, CAPBridgeProtocol {
                 if let pluginJSON = Bundle.main.url(forResource: "capacitor.config", withExtension: "json") {
                     let pluginData = try Data(contentsOf: pluginJSON)
                     let registrationList = try JSONDecoder().decode(RegistrationList.self, from: pluginData)
-                    for plugin in registrationList.packageClassList {
+
+                    var uniquePluginClasses = Set<String>()
+                    for listedClass in registrationList.packageClassList {
```
Steven0351, line 293: Why not just make `packageClassList` a Set? Then the deduplication would just happen during decoding I would imagine.

markemer: That's a good idea, I didn't think of it, since I don't think I've ever done it.

Steven0351 (approved):

PR #7194, opened 2024-01-17
Change intent to fix single top issue

### LINKED ISSUE #6943: bug: [Android 13] Unable to get selected file in singletop mode
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 5.4.1
  @capacitor/core: 5.4.1
  @capacitor/android: 5.4.1
  @capacitor/ios: 5.4.1

Installed Dependencies:

  @capacitor/core: 5.2.3
  @capacitor/ios: 5.2.3
  @capacitor/cli: 5.2.3
  @capacitor/android: 5.2.3

```

### Platform(s)

Android 13

### Current Behavior

After selecting a file using the regular `<input type="file"...` the activity returns `RESULTS_CANCELED` instead of OK with the selected file

### Expected Behavior

The file should be returned.

### Code Reproduction

Create an app with open file input and set the app to singletop mode, run it on android 13.

### Other Technical Details

`npm --version` output: 9.5.1

`node --version` output: v18.16.0

`pod --version` output (iOS issues only): N/A

### Additional Context

I've seen this issue also in the camera plugin and also reported it and made a PR which is unfortunately ignored, so I keep patching every new version the is released.
This is the PR:
https://github.com/ionic-team/capacitor-plugins/pull/1549
This is the original issue, which is very similar to this one:
https://github.com/ionic-team/capacitor-plugins/issues/823

When inspecting the following code:
https://github.com/ionic-team/capacitor/blob/7e7c8b9113f541d530c5883dea1f52b2957c0859/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java#L412

The return value is an intent with `android.intent.action.GET_CONTENT` type.
When I asked Bard about it I got the following explanation.

When I changed the above code to be:
```
Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
intent.setType("*/*");
```
It started working again.
Let me know if you would like me to create a PR for this.

Bard chat:
```
You're right, the GET_CONTENT action is not explicitly deprecated in the official Android documentation. However, there are a few reasons why it is considered to be deprecated:

It does not use the Storage Access Framework (SAF), which is the preferred way to access files in Android 13.
It may not be compatible with all apps, especially those that are targeting Android 13.
Google has recommended that developers use the ACTION_OPEN_DOCUMENT action instead of the GET_CONTENT action in Android 13.
Here is a quote from the Android documentation on the ACTION_OPEN_DOCUMENT action:

Note: The ACTION_OPEN_DOCUMENT action is the preferred way to open files in Android 13. It is more compatible with other apps and devices, and it uses the Storage Access Framework (SAF) to provide a unified way to access files from different sources.

This quote suggests that the GET_CONTENT action is not as compatible or as reliable as the ACTION_OPEN_DOCUMENT action.

If you are targeting Android 13, it is recommended that you use the ACTION_OPEN_DOCUMENT action instead of the GET_CONTENT action for the file chooser. This will help to improve the compatibility, performance, and memory usage of your app.

Please note that the GET_CONTENT action may still be required for some apps, such as those that need to access files from a legacy file system. However, it is generally recommended to use the ACTION_OPEN_DOCUMENT action whenever possible.
```

HarelM:
Fixes #6943

- #6943

This basically changes the intent type to allow picking files, otherwise nothing happens after you pick a file.
If using the current code with android 13 and singletop mode, that is. It might be that other modes are working as expected.

I know this works for my case and is currently in prodcution, I haven't tested it in other situations though.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/BridgeWebChromeClient.java
@@ -409,7 +409,8 @@ public class BridgeWebChromeClient extends WebChromeClient {
     }
 
     private void showFilePicker(final ValueCallback<Uri[]> filePathCallback, FileChooserParams fileChooserParams) {
-        Intent intent = fileChooserParams.createIntent();
+        Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
+        intent.setType("*/*");
         if (fileChooserParams.getMode() == FileChooserParams.MODE_OPEN_MULTIPLE) {
             intent.putExtra(Intent.EXTRA_ALLOW_MULTIPLE, true);
         }


jcesarmobile:
Thanks for the PR, but the recommended by google is to rely on `fileChooserParams.createIntent()` to create the best intent instead of creating your own, unless for specific cases that `fileChooserParams.createIntent()` doesn't handle well.
If `fileChooserParams.createIntent()` is not returning the intent you expect, you should report it to google so they can fix it.
By using `fileChooserParams.createIntent()` we get the new Photos picker that Google recommends on the devices that support it, while by using your changes we get a different picker.
If you don't like Capacitor's default picker, you can change the WebView's ChromeClient and implement your own, something like this in `MainActivity.java`:

```
@Override
        bridge.getWebView().setWebChromeClient(new WebChromeClient(){
            @Override
            public boolean onShowFileChooser(WebView webView, ValueCallback<Uri[]> filePathCallback, FileChooserParams fileChooserParams) {
                Intent intent = new Intent(Intent.ACTION_OPEN_DOCUMENT);
                intent.setType("*/*");
                // rest of the code
                return true;
            }
        });
```

You can also do that in a plugin in the `handleOnStart()` method instead of modifying the `MainActivity.java` directly so other users can take advantage of your changes.
You can also use `BridgeWebChromeClient` class instead of regular `WebChromeClient` class so all other methods that `BridgeWebChromeClient` overrides are implemented if you are only interested in overriding `onShowFileChooser`

PR #7202, opened 2024-01-22
Add support for iOS configurations when running 

bosh-code:
Often when adding new schemes, new configurations are added for those schemes, usually to point to different `.xcconfig` files. They probably wont have the default names of Debug and Release.

This PR adds support for passing an optional `--configuration=` option when running and building iOS. It's used in `xcodebuild` and as part of the location of the .app file for native run.

Running either uses the provided configuration or defaults to 'Debug'
Building either uses the provided configuration or defaults to 'Release'

Demo repo here: https://github.com/bosh-code/scheme


.gitignore
```diff
@@ -24,6 +24,7 @@ android-template/app/app.iml
 .project
 .env
 capacitor-build.log
+yarn.lock
```
markemer, line 27: We don't want to add yarn.lock to the the template in this PR, we'd want to look at that seperately.

markemer (changes requested):
Pull that `yarn.lock` otherwise, this looks good.

.gitignore
bosh-code: Thanks @markemer I have reverted the commit

markemer (approved):

giralte-ionic (approved):

PR #7240, opened 2024-02-09
fix(cli): remove certificate file extension check on copy

jcesarmobile:
At the moment the SSL pinning plugin only supports DER formatted certificates (binary) on iOS, which should have `.der` extension, not `.cer`, and other formats are not supported at the moment, so the error message is confusing for users.
And on Android it supports multiple formats and extensions.

I'm working on adding support for base64 formatted certificates to the SSL pinning plugin and also make the plugin log if there was a problem with any of the certificates, because at the moment it fails silently, so I think the `.cer` extension check can be removed.

An alternative if we don't want to remove the check would be to expand the file extension list.

### DIFF
--- a/cli/src/tasks/copy.ts
+++ b/cli/src/tasks/copy.ts
@@ -323,14 +323,6 @@ async function copySSLCert(
   const validCertPaths: string[] = [];
   for (const sslCertPath of sslCertPaths) {
     const certAbsFromPath = join(rootDir, sslCertPath);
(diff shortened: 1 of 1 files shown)


markemer (approved):
We certainly want to be rid of this, since the warning shows up when you do the right thing and if you do what it suggests, it will actually fail. However, we might want to add a check later when we expand our options.

giralte-ionic (approved):

Issue #6748, opened 2023-07-19
bug: multipart/form-data encoding error in CapacitorUrlRequest.swift


heyrex:
The recently introduced [FormData support](https://github.com/ionic-team/capacitor/pull/6206) in CapacitorHttp incorrectly encodes form values on iOS (works correctly for Android).

On iOS `getRequestDataFromFormData()` incorrectly adds `\r\n` below the value and omits it above. multipart/form-data encoding requires an empty line between the headers and the value. There should be no empty line after the value. It should be immediately proceeded by the boundary.

The two lines referenced here should be reversed:
https://github.com/ionic-team/capacitor/blob/2e43323410bab7c3381a0f11a6fd3880bde03a70/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift#L135C1-L136C56

It also looks like an additional \r\n is being appended to any data where `type == "base64File"`. 
See: https://github.com/ionic-team/capacitor/blob/3cd7196ec10066a98561e5d1b1166d8c1133f13a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift#L131

5uper:
this is fixed in v5.2.x with PR https://github.com/ionic-team/capacitor/pull/6722

heyrex:
> this is fixed in v5.2.x with PR #6722
@5uper  The pull request you reference only partially fixes the issue. That PR fixes the syntax of the multipart/form-data encoding but there is still data corruption because of the additional `\r\n` before the boundary.

This is the fix:
https://github.com/heyrex/capacitor-heyrex/pull/1/files

heyrex:
Further reading:
https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2
https://stackoverflow.com/questions/4238809/example-of-multipart-form-data
https://www.ietf.org/rfc/rfc2388.txt

heyrex:
No further time is available for this so we'll leave our fix available for anyone experiencing the same issue.

To any capacitor users experiencing this problem with Capacitor v4.8.1 or v5.2.x: 

1. Open ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift 
2. Find the  `getRequestDataFromFormData `function.
3. Modify as follows:
```
 if type == "base64File" {
    let fileName = item["fileName"]
    let fileContentType = item["contentType"]
    data.append("\r\n--\(boundary)\r\n".data(using: .utf8)!)
[...]

PR #7242, opened 2024-02-12
fix(ios): remove new line character in multipart/form-data requests

michaelwolz:
Removes new line characters `\r\n` which are incorrectly appended to the data of multipart/form-data requests. This change removes these new line characters in order to comply with the specifications, as defined here: https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.2

Fixes: ionic-team/capacitor#6748

Note, these changes were originally suggested by @hyrex here: https://github.com/heyrex/capacitor-heyrex/pull/1 and in [the related issue](https://github.com/ionic-team/capacitor/issues/6748#issuecomment-1643028088) that is linked to this pull request. 

Since this problem is a big deal for developers and buried deep in iOS logic this should be addressed as soon as possible which is why I want to put more attention on this.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
@@ -127,14 +127,12 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
                 data.append("\r\n".data(using: .utf8)!)
 
                 data.append(Data(base64Encoded: value)!)
-
-                data.append("\r\n".data(using: .utf8)!)
             } else if type == "string" {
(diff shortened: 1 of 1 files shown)


jcesarmobile:

michaelwolz:
@jcesarmobile sure it's super easy to reproduce, see the example app here: https://github.com/michaelwolz/multipart-form-data-not-working-on-ios (note that CapacitorHttp is activated via capacitor.config.json).

Relevant code can be seen here: https://github.com/michaelwolz/multipart-form-data-not-working-on-ios/blob/47dcecd4395ecf9873dbc90f54a54bfadfd3dd74/src/js/capacitor-welcome.js#L78-L91

See the raw request body and the empty lines before the separator. I think a lot of servers still handle this well but it's definitely not correct according to the specifications. 
image

PR #7250, opened 2024-02-13
chore: update to macos-14

markemer:


giralte-ionic (approved):

jcesarmobile (approved):
changes look good, but macos-14 image is still marked as beta, so not sure if we should wait until it's final?

markemer:
It's easy enough to roll back - and we get a real boost of speed here, and it's been out for quite a while in private beta. I'll be keeping an eye out for changes, but the Github's plan is to move macos-latest directly to this image, skipping the macos-13 we've been using.

PR #7261, opened 2024-02-15
fix: secure cookies often are not set by the cookie plugin

dtarnawsky:
secure cookies often are not set by the cookie plugin. This code correctly sets them.

This code fixes Android. There seems to be problems in iOS as well.


giralte-ionic (approved):

markemer:
@dtarnawsky Do you have a good way to replicate this in iOS?

dtarnawsky:
> @dtarnawsky Do you have a good way to replicate this in iOS?

No, I dont believe it can be done in iOS without being about to set the scheme to https (see slack conversation in ask-capacitor-runtime).

ItsChaceD (approved):

PR #7278, opened 2024-02-22
fix(spm): add product lines to Package.swift

markemer:
Currently in draft as this fixes one problem but is dependent on the following

- [x] Changing the SPM Repo name
- [x] Fixing the Names in all the Package.swift files 
    - https://github.com/ionic-team/capacitor-plugins/pull/2041
- [x] Fixing the iOS Template
    - #7279
- [x] Clean up some of the parsing assumptions

### DIFF
--- a/cli/src/util/spm.ts
+++ b/cli/src/util/spm.ts
@@ -10,6 +10,15 @@ export interface SwiftPlugin {
   path: string;
 }
 
+export async function checkPackageManager(config: Config): Promise<string> {
+  const iosDirectory = config.ios.nativeProjectDirAbs;
+  if (existsSync(resolve(iosDirectory, 'CapApp-SPM'))) {
+    return 'SPM';
+  }
+
+  return 'Cocoapods';
+}
+
 export async function findPackageSwiftFile(config: Config): Promise<string> {
   const packageDirectory = resolve(
     config.ios.nativeProjectDirAbs,
@@ -28,16 +37,29 @@ function readSwiftPackage(packageLine: string): string | null {
   return lineMatch[1];
 }
 
+function readSwiftProduct(productLine: string): string | null {
+  const packageRegex = RegExp(/.product\(\s*name:\s*"([A-Za-z0-9_-]+)"/);
+  const lineMatch = productLine.match(packageRegex);
+  if (lineMatch === null) {
+    return null;
+  }
+
+  return lineMatch[1];
+}
+
 export async function generatePackageFile(
   config: Config,
   plugins: Plugin[],
 ): Promise<void> {
   const swiftPluginList: string[] = [];
+  const swiftPluginProductList: string[] = [];
 
   for (const plugin of plugins) {
     const relPath = relative(config.ios.nativeXcodeProjDirAbs, plugin.rootPath);
     const pluginStatement = `.package(name: "${plugin.ios?.name}", path: "${relPath}"),`;
+    const pluginProduct = `.product(name: "${plugin.ios?.name}", package: "${plugin.ios?.name}"),`
     swiftPluginList.push(pluginStatement);
+    swiftPluginProductList.push(pluginProduct);
   }
 
   const packageSwiftFile = await findPackageSwiftFile(config);
@@ -53,6 +75,7 @@ export async function generatePackageFile(
 
     let textToWrite = '';
     const packages: string[] = [];
+    const products: string[] = [];
     for (const lineIndex in packageSwiftTextLines) {
       const line = packageSwiftTextLines;
       const index = parseInt(lineIndex);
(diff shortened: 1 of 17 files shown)


giralte-ionic (approved):

jcesarmobile (changes requested):
This fails to detect deleted plugins, they are kept in the `Package.swift` file, making the app fail.

I think we should go with full file generation (or at least everything after the `// DO NOT MODIFY THIS FILE`) instead of appending the plugins to the existing ones, this way it will generate the file without the uninstalled plugins, otherwise we would need to keep track of installed plugins on every sync and compare with the previous list and update the list, etc.

jcesarmobile:
another idea is to keep the file as is and append the dependencies to the end, I've tried this and it works:

```
package.dependencies.append(.package(name: "CapacitorActionSheet", path: "../capacitor-plugins/action-sheet"))

package.targets.first?.dependencies.append(.product(name: "CapacitorActionSheet", package: "CapacitorActionSheet"))
```

markemer:
Thanks for the bug callout - generating the whole thing from the start might be the answer, or something programmatic.

cli/src/util/spm.ts
```diff
@@ -10,6 +10,15 @@ export interface SwiftPlugin {
   path: string;
 }
 
+export async function checkPackageManager(config: Config): Promise<string> {
```
Steven0351, line 13: Maybe make this return a `Promise<'SPM' | 'Cocoapods'>` instead of `string`

cli/src/util/spm.ts
```diff
@@ -18,96 +29,62 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
+    let package = Package(
+        name: "CapApp-SPM",
+        platforms: [.iOS(.v13)],
+        products: [
+            .library(
+                name: "CapApp-SPM",
+                targets: ["CapApp-SPM"])
+        ],
+        dependencies: [
+            .package(url: "https://github.com/ionic-team/capacitor-spm.git", branch: "main")`;
```
jcesarmobile, line 62:
```suggestion
  let packageSwiftText = `// swift-tools-version: 5.9
import PackageDescription

// DO NOT MODIFY THIS FILE - managed by Capacitor CLI commands
let package = Package(
    name: "CapApp-SPM",
    platforms: [.iOS(.v13)],
    products: [
        .library(
            name: "CapApp-SPM",
            targets: ["CapApp-SPM"])
    ],
    dependencies: [
        .package(url: "https://github.com/ionic-team/capacitor-spm.git", branch: "main")`;
```

cli/src/util/spm.ts
```diff
@@ -18,96 +29,62 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
+
+  for (const plugin of plugins) {
+    const relPath = relative(config.ios.nativeXcodeProjDirAbs, plugin.rootPath);
+    packageSwiftText += `,\n            .package(name: "${plugin.ios?.name}", path: "${relPath}")`;
```
jcesarmobile, line 66:
```suggestion
    packageSwiftText += `,\n        .package(name: "${plugin.ios?.name}", path: "${relPath}")`;
```

cli/src/util/spm.ts
```diff
@@ -18,96 +29,62 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
 
-  return 'Cocoapods';
+  packageSwiftText += `
+       ],
+       targets: [
+          .target(
+                name: "CapApp-SPM",
+                dependencies: [
+                    .product(name: "Capacitor", package: "capacitor-spm"),
+                    .product(name: "Cordova", package: "capacitor-spm")`;
```
jcesarmobile, line 76:
```suggestion
    ],
    targets: [
        .target(
            name: "CapApp-SPM",
            dependencies: [
                .product(name: "Capacitor", package: "capacitor-spm"),
                .product(name: "Cordova", package: "capacitor-spm")`;
```

cli/src/util/spm.ts
```diff
@@ -18,96 +29,62 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
+
+  for (const plugin of plugins) {
+    packageSwiftText += `,\n                    .product(name: "${plugin.ios?.name}", package: "${plugin.ios?.name}")`;
```
jcesarmobile, line 79:
```suggestion
    packageSwiftText += `,\n                .product(name: "${plugin.ios?.name}", package: "${plugin.ios?.name}")`;
```

cli/src/util/spm.ts
```diff
@@ -18,96 +29,62 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
+  }
+
+  packageSwiftText += `
+                ]
+            )
+        ]
+      )
+`;
```
jcesarmobile, line 87:
```suggestion
            ]
        )
    ]
)
`;
```

jcesarmobile (changes requested):
just a few formatting issues on the resulting file

PR #7281, opened 2024-02-23
fix(cli): correct build path for non flavor builds 

VIMHaos:
* fix build on based linux file systems


markemer (approved):

giralte-ionic (approved):

jcesarmobile (approved):
it also fails on macOS, only for builds with no flavor specified, so it's not linux related, changed the title

markemer:
I'll merge this, and then let's cherry pick it for 5.x

Issue #6628, opened 2023-06-02
bug: http web - exception on response with http 204 (no-content) #1635


dartzki:
### Capacitor Version

capacitor/core": "4.6.2"

```
Found capacitor.config.overrides.json file, configuration overrides will be applied.
{
  "loggingBehavior": "debug",
  "webDir": "/home/simoned/git-ergon/llb.novus.client.web/angularx/build/distributions/banking-mobile",
  "android": {
    "webContentsDebuggingEnabled": true
  }
}
-----------------------------------------------
Capacitor config:
{
  "appId": "li.llb.banking.mobile",
  "appName": "li.llb.banking.mobile",
  "bundledWebRuntime": false,
  "webDir": "/home/simoned/git-ergon/llb.novus.client.web/angularx/build/distributions/banking-mobile",
  "plugins": {
    "PrivacyScreen": {
      "enable": false
    },
    "SplashScreen": {
      "launchShowDuration": 2000,
      "launchAutoHide": true
    },
    "CapacitorCookies": {
      "enabled": true
    },
    "NovusPush": {
      "loggingEnabled": false
    },
    "ErgonSecurity": {
      "loggingEnabled": false
    },
    "NovusShared": {
      "loggingEnabled": false
    },
    "DocumentViewer": {
      "loggingEnabled": false
    }
  },
  "server": {
    "iosScheme": "capacitor",
    "androidScheme": "https",
    "hostname": "capacitor-local"
  },
  "ios": {
    "contentInset": "always"
  },
  "loggingBehavior": "debug",
  "android": {
    "webContentsDebuggingEnabled": true
  }
}
-----------------------------------------------
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 5.0.4
  @capacitor/core: 5.0.4
  @capacitor/android: 5.0.4
  @capacitor/ios: 5.0.4

Installed Dependencies:

  @capacitor/android: 4.6.2
  @capacitor/cli: 4.6.2
  @capacitor/core: 4.6.2
  @capacitor/ios: 4.6.2

```

### Platform(s)

Web

### Current Behavior

The web implementation tries to parse the response as json (if responseType: json was given as HttpOption) which results in a "SyntaxError: Unexpected end of JSON input" Error.

### Expected Behavior

The web implementation should return a successful empty response when a 204 is encountered.

### Other Technical Details

`npm --version` output: 8.5.0

`node --version` output: v16.14.2

[...]

PR #7301, opened 2024-02-29
Don't process response body for http status 204 (no content)

jwltr:
The web implementation of the http plugin tries to parse all successful responses, including responses with status 204 - no content. If the `responseType` is `'json'` and the response status is 204 with no body, parsing the response with `response.json()` results in the exception "SyntaxError: Unexpected end of JSON input".
This is unexpected, the body of a response with a 204 should not be processed and definitely not cause an error, as the 204 indicates that no body is present. 
No error is thrown in the android or ios implementations of the http plugin in such a case.

This pull request changes the http plugin to not process the response body of a 204 independent of the `responseType`.

Same issue in an earlier capacitor version: https://github.com/ionic-team/capacitor/issues/6628


eric-horodyski:

More context:

PR #7306, opened 2024-03-01
fix(ios)!: solve FormData boundary issues

jcesarmobile:
Marking it as breaking since it adds a contentType param to `getRequestDataAsMultipartFormData`, which is a public function.

Removes the custom boundary `native-bridge.ts` is adding and uses the browser one.
Read the browser boundary from native and uses that one instead of a custom one (android already does that).
Removes some new lines added that break on some servers.
Makes `getRequestDataFromFormData` public since all other methods are public.

closes https://github.com/ionic-team/capacitor/issues/6748
closes https://github.com/ionic-team/capacitor/pull/7242
closes https://github.com/ionic-team/capacitor/pull/6832

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
@@ -51,7 +51,7 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         return nil
     }
 
-    public func getRequestDataAsMultipartFormData(_ data: JSValue, _ boundary: String) throws -> Data {
+    public func getRequestDataAsMultipartFormData(_ data: JSValue, _ contentType: String) throws -> Data {
         guard let obj = data as? JSObject else {
             // Throw, other data types explicitly not supported.
             throw CapacitorUrlRequestError.serializationError("[ data ] argument for request with content-type [ application/x-www-form-urlencoded ] may only be a plain javascript object")
@@ -62,7 +62,12 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         }
 
         var data = Data()
-
+        var boundary = UUID().uuidString
+        if contentType.contains("="), let contentBoundary = contentType.components(separatedBy: "=").last {
+            boundary = contentBoundary
+        } else {
+            overrideContentType(boundary)
+        }
         strings.forEach { key, value in
             data.append("\r\n--\(boundary)\r\n".data(using: .utf8)!)
             data.append("Content-Disposition: form-data; name=\"\(key)\"\r\n\r\n".data(using: .utf8)!)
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -130,28 +118,27 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
...
     public func getRequestData(_ body: JSValue, _ contentType: String, _ dataType: String? = nil) throws -> Data? {
+        let boundary = contentType.components(separatedBy: "=").last
         if dataType == "file" {
             guard let stringData = body as? String else {
                 throw CapacitorUrlRequestError.serializationError("[ data ] argument could not be parsed as string")
             }
             return Data(base64Encoded: stringData)
         } else if dataType == "formData" {
-            return try getRequestDataFromFormData(body)
+            return try getRequestDataFromFormData(body, boundary!)
```
markemer, line 141: We sure this will always be non-nil?

jcesarmobile: When using the patched fetch/XHR yeah, it will be there.
For direct API calls not so sure, but on Android there are no checks, so in any case it would crash on both platforms.

Steven0351: Could easily turn this into:
```
} else if dataType == "formData", let boundary {
```
to avoid the force unwrapping

ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -130,28 +118,27 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
...
         }
-
-        data.append("\r\n--\(boundary)--\r\n".data(using: .utf8)!)
+        data.append("--\(boundary)--\r\n".data(using: .utf8)!)
 
         return data
     }
 
     public func getRequestData(_ body: JSValue, _ contentType: String, _ dataType: String? = nil) throws -> Data? {
+        let boundary = contentType.components(separatedBy: "=").last
```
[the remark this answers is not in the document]
Steven0351: This is only needed for two of the cases below so guarding here probably doesn't make sense

ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
Steven0351: another solution could be to collapse the `"multipart/form-data"` check below into one if like:
```swift
} else if dataType == "formData" || contentType.contains("multipart/form-data"), let let boundary = contentType.components(separatedBy: "=").last {
```
It's a bit verbose but it keeps the one form-data case in a single branch

ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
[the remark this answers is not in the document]
markemer: True, just rolling a `let` into the `if`s is better

markemer (changes requested):
This is looking good, but we should ditch those force unwraps.

jcesarmobile:
Since you didn't like the force unwrap, I've restored the old code that creates a new boundary if none is present.

Note that on Android there is no similar code, if there is no boundary it will throw there, but this change is for iOS only, so I'm not changing anything there.

ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
jcesarmobile: "formData" and "multipart/form-data" call different functions, so doing this would require a second if to call one or the other.

markemer (approved):

giralte-ionic (approved):

PR #7308, opened 2024-03-04
fix(android): prevent crash on script injection if the script is too long

ryaa:
closes https://github.com/ionic-team/capacitor/issues/7309

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/JSInjector.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/JSInjector.java
@@ -69,7 +69,7 @@ class JSInjector {
      * @return
      */
     public InputStream getInjectedStream(InputStream responseStream) {
-        String js = "<script type=\"text/javascript\">" + getScriptString().replace("${", "\\${") + "</script>";
+        String js = "<script type=\"text/javascript\">" + getScriptString() + "</script>";
         String html = this.readAssetStream(responseStream);
 
         // Insert the js string at the position after <head> or before </head> using StringBuilder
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):

android/capacitor/src/main/java/com/getcapacitor/JSInjector.java
```diff
@@ -71,10 +71,15 @@ public String getScriptString() {
     public InputStream getInjectedStream(InputStream responseStream) {
         String js = "<script type=\"text/javascript\">" + getScriptString().replace("${", "\\${") + "</script>";
```
jcesarmobile, line 72:
```suggestion
        String js = "<script type=\"text/javascript\">" + getScriptString() + "</script>";
```
This can be removed since it was added for replaceFirst to work, since the PR removes the replaceFirst it's not needed

markemer (approved):

dabaaaz:
@jcesarmobile when do you think that fix's gonna be merged ?!

dodomui:
This doesn't fix my problem. 
Tried downgrade to 5.7.0. Still crash on launching.
Any solution?

dabaaaz:
@dodomui did you changed all the lines like the full MR https://github.com/ionic-team/capacitor/pull/7308/files ?

ryaa:
> below is the new error after applied above coding update.

The error seems to be thrown from **replaceFirst** method (see at `java.lang.String.replaceFirst(String.java:2269)` in the stack trace), so it does not seem that the required changes are there

dodomui:
No more issue with 5.7.2.
Thanks.

PR #7310, opened 2024-03-04
fix(cli): Attempt to verify non-cjs modules exist if cjs resolution fails

Steven0351:


giralte-ionic (approved):

markemer (approved):

markemer:
Probably want to revisit this file in the future to make it a little more modern, too.

PR #7363, opened 2024-03-21
fix: inject cordova files if a cordova plugin is present

jcesarmobile:
The Cordova plugins javascript injection is not working because the plugins are now being read from the list that the CLI generates, and the CLI doesn't generate Cordova plugin classes.

Since trying to figure out all the Cordova plugin classes is very complex and the classes are not really needed for injecting the javascript code, I've just edited the CLI to add a generic `CDVPlugin` class to the plugin list if there are cordova plugins installed, and instead of checking the superClass I just check if there is a `CDVPlugin` class in the list, and if there is one, inject the cordova files.

### DIFF
--- a/cli/src/util/iosplugin.ts
+++ b/cli/src/util/iosplugin.ts
@@ -7,6 +7,7 @@ import {
 } from '@ionic/utils-fs';
 import { resolve } from 'path';
 
+import { getCordovaPlugins } from '../cordova';
 import type { Config } from '../definitions';
 import type { Plugin } from '../plugin';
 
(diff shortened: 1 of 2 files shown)


markemer (approved):

markemer:
Does this need to be backported to 5.x as well? I can't remember if we made this 6.x only.

markemer:
Yeah, the checker picks up a Cordova plugin but only if they're swift. Which explains how this sometimes worked.

markemer:
Good catch. Maybe I should add one sorta small Cordova plugin to the test app just to make sure we don't break things - especially since I'm currently looking into our Cordova layers.

Steven0351:
You could probably just check the filesystem to see if there are any source files in the cordova plugins project. Unless the plugin file generation happens before that part :D

PR #7386, opened 2024-04-05
feat(cli): Support bun when running cap migrate

### LINKED ISSUE #7326: [Feature]: Support for Bun Package Manager
### Description

Add Bun to the available package managers that can be used with Capacitor

### Platforms

- [X] iOS
- [X] Android
- [ ] Web

download

[Bun.sh](https://bun.sh/) is a new, super fast package manager that saves a lot of time on installing packages and running commands.

For example, when updating capacitor and running `bun cap migrate`, I can only choose between `npm`, `yarn` and `pnpm` to install dependencies.

Also `bun cap sync` does not work and results in the following error: `Could not find the web assets directory: ./www.`. 
www Directory is available in my project and the sync command does work with yarn.
 

### Alternatives

_No response_

### Additional Information

_No response_

dtarnawsky:
Adds support for bun when running:
`cap migrate --noprompt --packagemanager=bun`

Resolves #7326

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -79,10 +79,10 @@ export async function migrateCommand(
 
   const variablesAndClasspaths:
     | {
-        'variables': any;
-        'com.android.tools.build:gradle': string;
-        'com.google.gms:google-services': string;
-      }
+      'variables': any;
+      'com.android.tools.build:gradle': string;
+      'com.google.gms:google-services': string;
+    }
     | undefined = await getAndroidVariablesAndClasspaths(config);
 
   if (!variablesAndClasspaths) {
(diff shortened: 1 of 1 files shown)


giralte-ionic (approved):

cli/src/tasks/migrate.ts
```diff
@@ -354,7 +355,8 @@ export async function migrateCommand(
       if (!installFailed) {
         // Run Cap Sync
         await runTask(`Running cap sync.`, () => {
-          return getCommandOutput('npx', ['cap', 'sync']);
+          const runCommand = installerType == 'bun' ? 'bunx' : 'npx';
+          return getCommandOutput(runCommand, ['cap', 'sync']);
```
jcesarmobile, line 359: I don't think this needs to be changed, we require node to be installed, so `npx` will always be available, I don't think there is any advantage on using `bunx` over `npx` in this case as it't used for running the already installed `@capacitor/cli` package, not for running a package that is not installed yet (in this case there would be a difference).

If we do this for bun, then we should do the same for the other package managers as they also have `npx` alternatives, but I don't think it should be done for any of them.

jcesarmobile (changes requested):

jcesarmobile (approved):

PR #7394, opened 2024-04-12
feat: enable building for visionOS

giralte-ionic:
This enables building for visionOS, both iPad style and pure visionOS style.

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
@@ -24,7 +24,7 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
         }
         #endif
         
-        #if os(visonOS)
+        #if os(visionOS)
         if let lang = Locale.autoupdatingCurrent.language.languageCode?.identifier {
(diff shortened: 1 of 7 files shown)


ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -20,6 +22,18 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
             }
             request.addValue(headers["Accept-Language"]!, forHTTPHeaderField: "Accept-Language")
         }
+        #endif
+        
+        #if os(visonOS)
```
jcesarmobile, line 27:
```suggestion
        #if os(visionOS)
```

markemer (approved):

fnc12:
hey this PR is very good. When is it going to be merged?

jcesarmobile:
Since we are focusing in SPM, to support pure visionOS or other architectures we would need to build the .xcframeworks with visionOS, Catalyst and any other architecture we would want to support and that's not something we plan to do anytime soon, so I'm going to close the PR.

Supporting visionOS or macOS with "iPad style" apps already works without doing any changes.

PR #7396, opened 2024-04-12
chore: mark SPM as experimental

markemer:


cli/src/util/spm.ts
```diff
@@ -35,6 +35,7 @@ export async function generatePackageFile(
 ): Promise<void> {
   const packageSwiftFile = await findPackageSwiftFile(config);
   try {
+    logger.warn('SPM Support is still experimental');
```
jcesarmobile, line 38: this is shown on every update/sync, do we want that?
or should only be shown on `npx cap add`?

theproducer (approved):

cli/src/util/spm.ts
theproducer: I could see it either way.  I don't think it would hurt even on sync to remind users this is still experimental, especially if they run into issues.

theproducer: @dallastjames What do you think?

IT-MikeS (approved):

cli/src/util/spm.ts
markemer: Yeah - I put it on sync to remind people that they're in the danger zone.

dallastjames (approved):

PR #7412, opened 2024-04-18
feat(ios): CAPPluginMethod selector-based initializer

Steven0351:
Add secondary initializer for CAPPluginMethod that takes a selector. Creates a convenience initializer as well with a Swift enum to avoid having to rely on global string values in the style of CAPPluginReturnPromise. For a pure Swift plugin implementation, it makes the definition potentially go from:
```swift
class MyPlugin: CAPPlugin, CAPBridgedPlugin {
  let pluginMethods: [CAPPluginMethod] = [
    .init(name: "promiseMethod", returnType: CAPPluginReturnPromise),
    .init(name: "callbackMethod", returnType: CAPPluginReturnCallback)
  ]

  @objc func promiseMethod(_ call: CAPPluginCall) {}
  @objc func callbackMethod(_ call: CAPPluginCall) {}
}
```

to this:
```swift
class MyPlugin: CAPPlugin, CAPBridgedPlugin {
  let pluginMethods: [CAPPluginMethod] = [
    .init(#selector(promiseMethod)),
    .init(#selector(callbackMethod), returnType: .callback)
  ]

  @objc func promiseMethod(_ call: CAPPluginCall) {}
  @objc func callbackMethod(_ call: CAPPluginCall) {}
}
```

A benefit of this approach include removing the need to stringify the name of the function. This will also validate the selector at compile time, so if it has not been annotated with `@objc` it will be a compiler error.

### DIFF
--- a/ios/Capacitor/Capacitor/CAPPluginMethod.swift
+++ b/ios/Capacitor/Capacitor/CAPPluginMethod.swift
@@ -8,7 +8,7 @@
 
 extension CAPPluginMethod {
     public enum ReturnType: String {
-        case promise, callback
+        case promise, callback, none
     }
 
     public convenience init(_ selector: Selector, returnType: ReturnType = .promise) {
--- a/ios/Capacitor/Capacitor/CAPPluginMethod.swift
+++ b/ios/Capacitor/Capacitor/CAPPluginMethod.swift
@@ -1,0 +1,17 @@
+//
+//  CAPPluginMethod.swift
+//  Capacitor
+//
+//  Created by Steven Sherry on 4/18/24.
+//  Copyright © 2024 Drifty Co. All rights reserved.
+//
+
+extension CAPPluginMethod {
+    public enum ReturnType: String {
+        case promise, callback
+    }
+
+    public convenience init(_ selector: Selector, returnType: ReturnType = .promise) {
+        self.init(selector: selector, returnType: returnType.rawValue)
+    }
+}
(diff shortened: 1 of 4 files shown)


ios/Capacitor/Capacitor/CAPPluginMethod.swift
```diff
@@ -0,0 +1,17 @@
...
+//  CAPPluginMethod.swift
+//  Capacitor
+//
+//  Created by Steven Sherry on 4/18/24.
+//  Copyright © 2024 Drifty Co. All rights reserved.
+//
+
+extension CAPPluginMethod {
+    public enum ReturnType: String {
+        case promise, callback
```
jcesarmobile, line 11: should we add the `none` return type?

it's not used by any plugin other than console (as far as I know)

Steven0351: We definitely could. none is weird in that the web code actually treats it as a callback when it is not a promise https://github.com/ionic-team/capacitor/blob/main/core/src/runtime.ts#L123-L133. The docs are actually misleading https://capacitorjs.com/docs/plugins/method-types#void-return in that is says "You can check the promise for an error but when it resolves the result is ignored" because you can't actually check for an error.

markemer (approved):

ios/Capacitor/Capacitor/CAPPluginMethod.swift
jcesarmobile: let's add it, so if we want to do a refactor in the future for the core plugins, we can use this new initializer in all of them instead of having to use a different one for Console plugin

jcesarmobile (approved):

Issue #7404, opened 2024-04-16
[Bug]: 


JosefBredereck:
### Capacitor Version

Config: {
  appId: 'de.company.appid.mobile',
  appName: 'App Name',
  webDir: 'dist/apps/app/browser',
  backgroundColor: '#003399',
  loggingBehavior: 'none',
  android: {
    path: './native/android',
    webContentsDebuggingEnabled: false
  },
  ios: {
    path: './native/ios',
    webContentsDebuggingEnabled: false
  },
  server: { androidScheme: 'https' },
  includePlugins: [
    '@capacitor/app',
    '@capacitor/browser',
    '@capacitor/device',
    '@capacitor/filesystem',
    '@capacitor/keyboard',
    '@capacitor/network',
    '@capacitor/preferences',
    '@capacitor/splash-screen',
    '@capacitor/camera',
  ],
  plugins: {
    Keyboard: { resize: 'none' },
    SplashScreen: {
      launchShowDuration: 700,
      launchAutoHide: true,
      androidScaleType: 'CENTER_INSIDE',
      backgroundColor: '#1a3691'
    },
    CapacitorHttp: { enabled: true }
  }
}
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.0.0
  @capacitor/core: 6.0.0
  @capacitor/android: 6.0.0
  @capacitor/ios: 6.0.0

Installed Dependencies:

  @capacitor/cli: 5.6.0
  @capacitor/core: 5.7.4
  @capacitor/android: 5.7.4
  @capacitor/ios: 5.7.4

### Other API Details

```Shell
npm: 10.2.4
node: v21.6.1
```

### Platforms Affected

- [X] iOS
- [ ] Android
- [ ] Web

### Current Behavior

We observed an issue on iOS. At random, the app crashes in an Angular application when requests are fired and aborted.
The question remains what exactly causes the issue, we could not create a successful reproduction. Sometimes it happens when a request gets unsubscribed (aborted) and sometimes when the request resolves with an error. 

```
scheme stop
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'This task has already been stopped'
*** First throw call stack:
(
	0   CoreFoundation                      0x0000000111da7761 __exceptionPreprocess + 242
	1   libobjc.A.dylib                     0x000000010d3e3904 objc_exception_throw + 48
[...]

PR #7453, opened 2024-05-07
fix(ios): don't return data if task was stopped

jcesarmobile:
keep track of urlSchemeTasks and don't call their methods if the task was stopped

closes https://github.com/ionic-team/capacitor/issues/7404

### DIFF
--- a/ios/Capacitor/Capacitor/WebViewAssetHandler.swift
+++ b/ios/Capacitor/Capacitor/WebViewAssetHandler.swift
@@ -6,6 +6,7 @@ import MobileCoreServices
 open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
     private var router: Router
     private var serverUrl: URL?
+    private var pendingTasks: [Int] = []
 
     public init(router: Router) {
         self.router = router
@@ -25,6 +26,7 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
     }
 
     open func webView(_ webView: WKWebView, start urlSchemeTask: WKURLSchemeTask) {
+        pendingTasks.append(urlSchemeTask.hash)
         let startPath: String
         let url = urlSchemeTask.request.url!
         let stringToLoad = url.path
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -6,6 +6,7 @@ import MobileCoreServices
 open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
     private var router: Router
     private var serverUrl: URL?
+    private var pendingTasks: [Int] = []
```
Steven0351, line 9: I would suggest making this a Set instead of an Array. Set.contains is an O(1) operation versus Array being an O(n).

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -47,67 +49,71 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
-                let httpResponse = HTTPURLResponse(url: localUrl, statusCode: 200, httpVersion: nil, headerFields: headers)
-                if isMediaExtension(pathExtension: url.pathExtension) {
-                    urlSchemeTask.didReceive(urlResponse)
-                } else {
-                    urlSchemeTask.didReceive(httpResponse!)
-                }
+                urlSchemeTask.didReceive(data)
+            } catch let error as NSError {
+                urlSchemeTask.didFailWithError(error)
+                self.pendingTasks.removeAll(where: { $0 == urlSchemeTask.hash })
```
Steven0351, line 106: When this is updated to be a Set, this would just be changed to be `pendingTasks.remove(urlSchemeTask.hash)`

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -25,6 +26,7 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
     }
 
     open func webView(_ webView: WKWebView, start urlSchemeTask: WKURLSchemeTask) {
+        pendingTasks.append(urlSchemeTask.hash)
```
Steven0351, line 29: I'm not 100% sure about what thread this method is called on, but it could potentially be called across threads and you may need to lock access to `pendingTasks` to prevent a crash or data race.

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -160,41 +166,45 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
+                urlSchemeTask.didFinish()
+                self.pendingTasks.removeAll(where: { $0 == urlSchemeTask.hash })
+                return
             }
-            urlSchemeTask.didFinish()
-            return
         }
 
         task.resume()
+        task.cancel()
```
Steven0351, line 207: Was this left on accident?

Steven0351 (changes requested):

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
jcesarmobile: yep, I tested canceling the task programmatically and removed it, but then added some prints to do some more tests and when I undid the print changes I accidentally added this again, good catch!

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
jcesarmobile: yeah, when I changed the Array to Set the app started crashing randomly because of threading, with the Array it never crashed, not sure why.
I've added a thread safe set now.

Steven0351 (approved):

PR #7457, opened 2024-05-09
RDMR-100

jcesarmobile:
Alternative to https://github.com/ionic-team/capacitor/pull/7453

Instead of adding all the scheme hashes it adds only the ones that were stopped, so I don't think it will ever crash.

### DIFF
--- a/ios/Capacitor/Capacitor/WebViewAssetHandler.swift
+++ b/ios/Capacitor/Capacitor/WebViewAssetHandler.swift
@@ -139,9 +139,7 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
 
     func handleCapacitorHttpRequest(_ urlSchemeTask: WKURLSchemeTask, _ localUrl: URL, _ isHttpsRequest: Bool) {
         var urlRequest = urlSchemeTask.request
-        guard let url = urlRequest.url else {
-            return
-        }
+        guard let url = urlRequest.url else { return }
         var targetUrl = url.absoluteString
             .replacingOccurrences(of: CapacitorBridge.httpInterceptorStartIdentifier, with: "")
             .replacingOccurrences(of: CapacitorBridge.httpsInterceptorStartIdentifier, with: "")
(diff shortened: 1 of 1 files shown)


ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -181,22 +187,32 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
     }
 
+    private func lock(_ urlSchemeTask: WKURLSchemeTask, _ action: @escaping () -> Void) {
+        let group = DispatchGroup()
+        group.enter()
+        if !self.stoppedTasks.contains(urlSchemeTask.hash) {
+            action()
+        }
+        group.leave()
+    }
```
Steven0351, line 214: This use of DispatchGroup isn't doing anything. You would typically use DispatchGroup in tandem with a DispatchQueue, and then you would do `group.wait()` to make sure all tasks finished executing before continuing.

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -80,7 +81,7 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
                 headers["Content-Range"] = "bytes \(fromRange)-\(toRange)/\(totalSize)"
                 headers["Content-Length"] = String(data.count)
                 let response = HTTPURLResponse(url: localUrl, statusCode: 206, httpVersion: nil, headerFields: headers)
-                urlSchemeTask.didReceive(response!)
+                lock(urlSchemeTask, { urlSchemeTask.didReceive(response!) })
```
Steven0351, line 84: Mentioned in the other comment, this isn't actually locking any access to anything. The best way to achieve what you are wanting is to do this:
```
class ConcurrentTasks {
  func withTask(_ task: WKURLSchemeTask, action: () -> Void) {
    if !tasks.contains(task.hash) {
      action()
    }
  }
}
```
This will block anything trying to write to the `stoppedTasks` until it is finished.

PR #7460, opened 2024-05-10
RDMR-100

Steven0351:
An alternative to #7457. This uses Swift concurrency to manage cancellation.

### DIFF
--- a/ios/Capacitor/Capacitor/WebViewAssetHandler.swift
+++ b/ios/Capacitor/Capacitor/WebViewAssetHandler.swift
@@ -6,6 +6,7 @@ import MobileCoreServices
 open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
     private var router: Router
     private var serverUrl: URL?
+    private var stoppedTasks = ConcurrentTasks()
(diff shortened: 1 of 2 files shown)


ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -25,89 +27,97 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
-            urlSchemeTask.didFailWithError(error)
-            return
+            tasks.with(urlSchemeTask) { $0.didFinish() }
         }
-        urlSchemeTask.didFinish()
     }
 
     open func webView(_ webView: WKWebView, stop urlSchemeTask: WKURLSchemeTask) {
-        CAPLog.print("scheme stop")
+        print("cancelling task \(urlSchemeTask.hashString)")
```
jcesarmobile, line 118:
```suggestion
```
remove print

with latest changes the app sometimes doesn't start, I just get a white screen and no messages after `Loading app at capacitor://localhost...`

PR #7465, opened 2024-05-14
WIP: Conditionally use Cordova

jcesarmobile:


jcesarmobile:
closing since this approach is won't allow SPM projects to use Cordova plugins as they precompiled into xcframeworks

PR #7482, opened 2024-05-23
fix(ios): check if urlSchemeTask is stopped before calling its methods

jcesarmobile:
add an extension to see if the urlSchemeTask was stopped and check it before calling its methods as they crash if the urlSchemeTask is stopped

closes https://github.com/ionic-team/capacitor/pull/7478
closes https://github.com/ionic-team/capacitor/issues/7404
closes https://github.com/ionic-team/capacitor/pull/7453
closes https://github.com/ionic-team/capacitor/pull/7460


Steven0351 (approved):
It makes sense and I like it.

markemer (approved):
Yep, this is the way.

Issue #7523, opened 2024-06-20
[Bug]: Percent-encoding of url search parameters broken for patched fetch / xhr requests through CapacitorHttp


michaelwolz:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.1.0
  @capacitor/core: 6.1.0
  @capacitor/android: 6.1.0
  @capacitor/ios: 6.1.0

Installed Dependencies:

  @capacitor/cli: 6.1.0
  @capacitor/core: 6.1.0
  @capacitor/android: 6.1.0
  @capacitor/ios: 6.1.0
```

### Platforms Affected

- [X] iOS
- [X] Android
- [ ] Web

### Current Behavior

All patched fetch and xhr requests using any of the request methods `GET, HEAD, OPTIONS` and `TRACE` do not correctly consider percent-encoding of url search parameters. Especially it breaks the encoding of properly encoded url search parameters.

### Example 

Let's consider the following fetch call: 
``await fetch(`https://example.com?param=${encodeURIComponent(";:@&=+$,/?%#[]")}`);``

The resulting URL to call will then look like this on the Javascript side and work fine in the browser: 
`https://example.com?param=%3B%3A%40%26%3D%2B%24%2C%2F%3F%25%23%5B%5D`

On Android the called URL will look like this though:
`https://example.com/?param=!'();:@&=+$,/?%`

and on iOS like this (oddly enough, iOS somehow managed to encode the "%" character itself correctly, but I assume Swift handles this somehow?):
`https://example.com/?param=!'();:@&=+$,/?%25`

See intercepted network traffic via mitmproxy: 
![image](https://github.com/ionic-team/capacitor/assets/7479806/833ac94a-5b13-4be8-8efb-47effb059e63)

### Expected Behavior

Patched fetch and xhr request should respect url encoding.

### Project Reproduction

https://github.com/michaelwolz/capacitor-tests

### Additional Information

I tried to investigate the root cause of this problem and found that PR #7273 introduced some logic to correctly handle the colon character `:` in hostnames for proxied urls in order to be able to correctly use it with different ports. For this the hostname of the requested url is encoded in the [`createProxyUrl` method](https://github.com/ionic-team/capacitor/blob/0ad0c3f23e1c62ade5081878e345e74f963b9ea5/core/native-bridge.ts#L143) in the native-bridge.  

[...]

PR #7527, opened 2024-06-20
Fix percent-encoding for patched fetch / xhr requests

michaelwolz:
Following up #7523. 

This suggestion addresses the issue of non-encoded URLs in patched fetch and XHR requests. By encoding the full original URL and appending it as a parameter to the bridge/proxy URL, this solution should be more robust when passed to the native code. Also, there is no longer a need to distinguish between HTTP and HTTPS URLs separately, as this information is already embedded in the encoded URL parameter.

Working test implementation can be found here (including an example with non 80 port): https://github.com/michaelwolz/capacitor-tests/tree/fix/url-encoding

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -93,6 +93,7 @@ public class Bridge {
     public static final String CAPACITOR_CONTENT_START = "/_capacitor_content_";
     public static final String CAPACITOR_HTTP_INTERCEPTOR_START = "/_capacitor_http_interceptor_";
     public static final String CAPACITOR_HTTPS_INTERCEPTOR_START = "/_capacitor_https_interceptor_";
+    public static final String CAPACITOR_HTTP_INTERCEPTOR_URL_PARAM = "u";
 
     public static final int DEFAULT_ANDROID_WEBVIEW_VERSION = 60;
     public static final int MINIMUM_ANDROID_WEBVIEW_VERSION = 55;
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -92,7 +92,6 @@ public class Bridge {
     public static final String CAPACITOR_FILE_START = "/_capacitor_file_";
     public static final String CAPACITOR_CONTENT_START = "/_capacitor_content_";
     public static final String CAPACITOR_HTTP_INTERCEPTOR_START = "/_capacitor_http_interceptor_";
-    public static final String CAPACITOR_HTTPS_INTERCEPTOR_START = "/_capacitor_https_interceptor_";
     public static final String CAPACITOR_HTTP_INTERCEPTOR_URL_PARAM = "u";
 
     public static final int DEFAULT_ANDROID_WEBVIEW_VERSION = 60;
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -262,13 +262,10 @@ public class WebViewLocalServer {
         boolean isHttps =
             request.getUrl().getPath() != null && request.getUrl().getPath().startsWith(Bridge.CAPACITOR_HTTPS_INTERCEPTOR_START);
 
-        String urlString = request
-            .getUrl()
-            .toString()
-            .replace(bridge.getLocalUrl(), isHttps ? "https:/" : "http:/")
-            .replace(Bridge.CAPACITOR_HTTP_INTERCEPTOR_START, "")
-            .replace(Bridge.CAPACITOR_HTTPS_INTERCEPTOR_START, "");
-        urlString = URLDecoder.decode(urlString, "UTF-8");
+        Uri uri = Uri.parse(request.getUrl().toString());
+        // `getQueryParameter` method of Android Uri lib already decodes the value automatically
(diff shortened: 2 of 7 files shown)


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -92,7 +92,7 @@ public class Bridge {
     public static final String CAPACITOR_FILE_START = "/_capacitor_file_";
     public static final String CAPACITOR_CONTENT_START = "/_capacitor_content_";
     public static final String CAPACITOR_HTTP_INTERCEPTOR_START = "/_capacitor_http_interceptor_";
-    public static final String CAPACITOR_HTTPS_INTERCEPTOR_START = "/_capacitor_https_interceptor_";
```
jcesarmobile, line 95: Since it's public, removing it would be a breaking change, so better mark it as deprecated and we will remove it in next major.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -259,16 +253,10 @@ private String getReasonPhraseFromResponseCode(int code) {
...
-            .getUrl()
-            .toString()
-            .replace(bridge.getLocalUrl(), isHttps ? "https:/" : "http:/")
-            .replace(Bridge.CAPACITOR_HTTP_INTERCEPTOR_START, "")
-            .replace(Bridge.CAPACITOR_HTTPS_INTERCEPTOR_START, "");
-        urlString = URLDecoder.decode(urlString, "UTF-8");
+        Uri uri = Uri.parse(request.getUrl().toString());
+        // `getQueryParameter` method of Android Uri lib already decodes the value automatically
+        String urlString = uri.getQueryParameter(Bridge.CAPACITOR_HTTP_INTERCEPTOR_URL_PARAM);
+
```
jcesarmobile, line 259:
```suggestion
        String urlString = request.getUrl().getQueryParameter(Bridge.CAPACITOR_HTTP_INTERCEPTOR_URL_PARAM);
```
`request.getUrl()` already returns an `Uri`, there is no need to turn it into string to create a new `Uri`

ios/Capacitor/Capacitor/CapacitorBridge.swift
```diff
@@ -95,7 +95,7 @@ open class CapacitorBridge: NSObject, CAPBridgeProtocol {
     public static let capacitorSite = "https://capacitorjs.com/"
     public static let fileStartIdentifier = "/_capacitor_file_"
     public static let httpInterceptorStartIdentifier = "/_capacitor_http_interceptor_"
-    public static let httpsInterceptorStartIdentifier = "/_capacitor_https_interceptor_"
```
jcesarmobile, line 98: Since it's public, removing it would be a breaking change, so better mark it as deprecated and we will remove it in next major.

core/native-bridge.ts
```diff
@@ -126,27 +126,22 @@ const convertBody = async (
...
-  const proxyUrl = new URL(url);
   const bridgeUrl = new URL(win.Capacitor?.getServerUrl() ?? '');
-  const isHttps = proxyUrl.protocol === 'https:';
-  bridgeUrl.search = proxyUrl.search;
-  bridgeUrl.hash = proxyUrl.hash;
-  bridgeUrl.pathname = `${
-    isHttps ? CAPACITOR_HTTPS_INTERCEPTOR : CAPACITOR_HTTP_INTERCEPTOR
-  }/${encodeURIComponent(proxyUrl.host)}${proxyUrl.pathname}`;
+  bridgeUrl.pathname = CAPACITOR_HTTP_INTERCEPTOR;
+  // URLSearchParams `append()` method will automatically percent encode the url
```
jcesarmobile, line 142:
```suggestion
```

ios/Capacitor/Capacitor/WebViewAssetHandler.swift
```diff
@@ -138,21 +133,14 @@ open class WebViewAssetHandler: NSObject, WKURLSchemeHandler {
...
-            .replacingOccurrences(of: CapacitorBridge.httpInterceptorStartIdentifier, with: "")
-            .replacingOccurrences(of: CapacitorBridge.httpsInterceptorStartIdentifier, with: "")
-        // Only replace first occurrence of the scheme
-        if let range = targetUrl.range(of: localUrl.scheme ?? InstanceDescriptorDefaults.scheme) {
-            targetUrl = targetUrl.replacingCharacters(in: range, with: isHttpsRequest ? "https" : "http")
+        
+        let urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        if let targetUrl = urlComponents?.queryItems?.first(where: { $0.name == CapacitorBridge.httpInterceptorUrlParam })?.value,
+           !targetUrl.isEmpty {
+            // Note that URLComponents are already percent-decoded so we do not need to do this manually
```
jcesarmobile, line 140:
```suggestion
```

jcesarmobile (changes requested):
added some minor comments

android/capacitor/src/main/java/com/getcapacitor/Bridge.java
michaelwolz: Done via: ac97cd4

ios/Capacitor/Capacitor/CapacitorBridge.swift
michaelwolz: Done via: ac97cd4

jcesarmobile (approved):
Looks good, cool idea!

Steven0351 (approved):
Implementation looks good and I like the reduction in complexity.

PR #7528, opened 2024-06-20
fix(livereload): Don't overwrite the whole config.server section in `editCapConfigForLiveReload` function

ZeroDX255:
Live reload now works with cleartext: true


OS-pedrogustavobilro (approved):

OS-pedrogustavobilro:
Oh you're right, my bad, I had tested changing manually instead of going to the fork. Good catch!

PR #7534, opened 2024-06-26
Added vision os build support for capacitor and cordova

fnc12:
Hey there. I am working on porting one web app to visionOS and our team uses ionic cause it is a very useful tool. We'd like to share and contribute our code which helped us building basic ionic iOS example for visionOS.


jcesarmobile:
Since we are focusing in SPM, to support pure visionOS or other architectures we would need to build the .xcframeworks with visionOS, Catalyst and any other architecture we would want to support and that's not something we plan to do anytime soon, so I'm going to close the PR.

Supporting visionOS or macOS with "iPad style" apps already works without doing any changes.

PR #7540, opened 2024-06-30
fix(iOS): change hardcoded config path when using custom config file for auto register plugins

tossaro:
When I use `InstanceDescriptor` with custom config path, the `CapacitorBridge` initialization on the `CAPBridgeViewController` is by default autoRegisterPlugins set to true.
But as we can see on CapacitorBridge class function registerPlugins config path is hardcoded
`
if autoRegisterPlugins {
            do {
                if let pluginJSON = Bundle.main.url(forResource: "capacitor.config", withExtension: "json") {
                //..
`
So, I've created new property configUrl that will be given when parsing configuration at InstanceDescriptor initialization.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
@@ -54,6 +54,7 @@ public class CapConfig {
     private String errorPath;
     private boolean zoomableWebView = false;
     private boolean resolveServiceWorkerRequests = true;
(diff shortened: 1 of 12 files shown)


eric-horodyski:
Thanks for opening this PR 🙏

riderx:

Patch ID: `upstream-pr-7540-ios`

```ts
plugins: {
  CapacitorPatch: {
    patches: ['upstream-pr-7540-ios'],
    strict: true,
  },
}
```

Issue #7323, opened 2024-03-07
[Bug]: Android Live Reload broken in Capacitor 5.7.2


alextreppass:
### Capacitor Version

[warn] The bundledWebRuntime configuration option has been deprecated. Please,
       use a bundler to bundle Capacitor and its plugins.
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 5.7.2
  @capacitor/core: 5.7.2
  @capacitor/android: 5.7.2
  @capacitor/ios: 5.7.2

Installed Dependencies:

  @capacitor/cli: 5.7.2
  @capacitor/core: 5.7.2
  @capacitor/android: 5.7.2
  @capacitor/ios: 5.7.2

### Other API Details

```Shell
yarn --version
3.2.1

OSX Sonoma 14.4.1
```

### Platforms Affected

- [ ] iOS
- [X] Android
- [ ] Web

### Current Behavior

Our `capacitor.config.ts` when running in live reload mode (`-l external`), has the following config:
```js
server: {
  url: process.env.LIVE_RELOAD_SERVER_URL,
  cleartext: true,
},
```

`LIVE_RELOAD_SERVER_URL` is set as `http://10.0.2.2:3000` for Android (as Android Studio treats the host machine's ip as 10.0.2.2), and `http://0.0.0.0:3000` for iOS.

We are using this setup to proxy a local dev server running on port 3000 for fast iteration without having to run `npx cap run` + rebuild each time.

### Issues
In 5.7.2 with Android, when going to the url capacitor's `live-reload.ts` generates (e.g. http://192.168.1.241:3000) the webpage does not load. There is an `ERR_CLEARTEXT_NOT_PERMITTED` error:
![Screenshot 2024-03-08 at 9 53 39 am](https://github.com/ionic-team/capacitor/assets/154376/bbb6f054-8821-47b4-9457-ce4fb5f2d7ea)

It doesn't look like there have been updates to https://github.com/ionic-team/capacitor/blob/main/cli/src/util/livereload.ts since @IT-MikeS introduced it, so something else must be going on here.

### Expected Behaviour

This worked great on 5.4.2:
```
  @capacitor/cli: 5.4.2
  @capacitor/core: 5.4.2
  @capacitor/android: 5.4.2
  @capacitor/ios: 5.4.2
```

The webview would load correctly.

### Project Reproduction

https://github.com/alextreppass/capacitor-live-reload-bug

alextreppass:
[...]

PR #7563, opened 2024-07-11
fix(cli): enable cleartext for live reload

jcesarmobile:
This PR enables cleartext when using live reload by directly writing into the `capacitor-cordova-android-plugins`'s `AndroidManifest.xml`, and removes it once the command is stopped.

It no longer overrides the original `extConfig` object but instead modifies the  platform's `capacitor.config.json` after sync.

Added a TODO to remove `editExtConfigForLiveReload` since it's no longer being used.
Added a TODO to remove the last param of `editCapConfigForLiveReload` as it's not being used.
Didn't remove any of them at the moment as people could be using the `CapLiveReloadHelper` directly, so removing them is breaking.

I've put the `process.on('SIGINT'` inside a promise to prevent multiple calls, it was being called three times.

There is an [alternate PR](https://github.com/ionic-team/capacitor/pull/7528), but it requires users to declare the `cleartext` themselves and doesn't work if using the `--no-sync` option.

closes https://github.com/ionic-team/capacitor/issues/7323
closes https://github.com/ionic-team/capacitor/pull/7528

### DIFF
--- a/cli/src/cordova.ts
+++ b/cli/src/cordova.ts
@@ -760,6 +760,7 @@ export async function writeCordovaAndroidManifest(
   cordovaPlugins: Plugin[],
   config: Config,
   platform: string,
+  cleartext?: boolean,
 ): Promise<void> {
   const manifestPath = join(
     config.android.cordovaPluginsDirAbs,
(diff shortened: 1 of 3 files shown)


markemer (approved):

IT-MikeS (approved):

xEverth:
Hello,

Sorry for commenting on a long-merged PR, but since this change is still in the main branch I would like to ask the following to the PR author:

Would it be preferrable to use a match pattern for the clearTextString?
What happens if one uses string substitution in their AndroidManifest, e.g. `android:usesCleartextTraffic="${usesCleartextTraffic}"`?

Thank you

(I am not able to use the livereload option on an Android 10 device without using --ssl, I always get ERR_CLEARTEXT_NOT_PERMITTED even after configuring both the server in capacitor config and the networkSecurityConfig in the AndroidManifest to specifically allow for cleartext traffic.)

jcesarmobile:
This doesn't write on the user's AndroidManifest.xml but in a module which we generate for Cordova plugins and the whole AndroidManifest.xml is generated by the CLI.


If you have more questions you can ask on the [Ionic Forums](https://forum.ionicframework.com/c/capacitor/26), on [Ionic discord](https://discord.com/invite/UPYYRhtyzp) or on the [discussions section](https://github.com/ionic-team/capacitor/discussions) of capacitor repository

PR #6704, opened 2023-07-01
fix(cli): apkName for multi-dimensional flavors

julienfdev:
This PR fixes the apkName problem with multi dimensional flavors on Android.

Discussed here > #6703

PR #7598, opened 2024-08-01
fix(cli): Android apk name multi flavor dimensions parsing

luisbytes:
Hi @theproducer

Support for multi-dimensional flavors was added with pull request #6704, but it caused an error that added a dash at the beginning of the flavor name when it started with an uppercase letter. This was attempted to be corrected with #7382, but it affected the multi-flavor dimensions parsing.

I propose this solution: revert your commit, but remove the dash if it exists at the beginning of the flavor name.

### DIFF
--- a/cli/src/common.ts
+++ b/cli/src/common.ts
@@ -590,6 +590,9 @@ export async function checkJDKMajorVersion(): Promise<number> {
 }
 
 export function parseApkNameFromFlavor(flavor: string): string {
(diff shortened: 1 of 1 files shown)


ThomasVanleynseele:
Hi, we've also used this patch recently, any change this can get merged soon?

luisbytes:
Hi @theproducer Could you please take a look at this?

Issue #7596, opened 2024-08-01
[Bug]: Capacitor on Android detected as web platform when server.url contains fragment on WebViews that support DOCUMENT_START_SCRIPT


hbordersTwitch:
### Capacitor Version

```
[warn] The bundledWebRuntime configuration option has been deprecated. Can be safely deleted.
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.1.1
  @capacitor/core: 6.1.1
  @capacitor/android: 6.1.1
  @capacitor/ios: 6.1.1

Installed Dependencies:

  @capacitor/cli: 6.1.1
  @capacitor/ios: 6.1.1
  @capacitor/android: 6.1.1
  @capacitor/core: 6.1.1

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Other API Details

```Shell
$ npm --version
10.7.0
$ node --version
v22.2.0
$ pod --version
1.15.2
```
```


### Platforms Affected

- [ ] iOS
- [X] Android
- [ ] Web

### Current Behavior

When I run capacitor with a `server.url` that includes a url fragment (`https://example.com#foo`) on an API 34 Android emulator (or any emulator that supports `WebViewFeature.DOCUMENT_START_SCRIPT`), Capacitor detects the runtime as `web`. I can open the `WebView` in the Chrome Inspector, and run the following in the console:
```
> Capacitor.platform
<- 'web'
```

If I remove the fragment from the `server.url` (`https://example.com`), then everything works fine. I can open the `WebView` in the Chrome Inspector, and run the following in the console:
```
> Capacitor.platform
<- 'android'
```

### Expected Behavior

When I run capacitor with a `server.url` that includes a url fragment (`https://example.com#foo`) on an API 34 Android emulator (or any emulator that supports `WebViewFeature.DOCUMENT_START_SCRIPT`), Capacitor detects the runtime as `android`, and all plugins work normally.

### Project Reproduction

https://github.com/hbordersTwitch/capacitor-server-url-fragment-bug

### Additional Information

https://github.com/ionic-team/capacitor/issues/7454 is related because it involves `allowedOrigin` rules.

[...]

Issue #7327, opened 2024-03-09
[Bug]: App crash when adding Basic Auth to server.url in Capacitor 6.0


patrik-skilling:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 5.7.2
  @capacitor/core: 5.7.2
  @capacitor/android: 5.7.2
  @capacitor/ios: 5.7.2

Installed Dependencies:

  @capacitor/ios: 6.0.0-rc.0
  @capacitor/cli: 6.0.0-rc.0
  @capacitor/android: 6.0.0-rc.0
  @capacitor/core: 6.0.0-rc.0

[success] Android looking great! 👌
```

### Platforms Affected

- [ ] iOS
- [X] Android
- [ ] Web

### Current Behavior

When adding Basic Auth to server.url app crash and you get an stacktrace that says IllegalArgumentException: allowedOriginRules https://{BasicAuth}/{Domain} is invalid. I know this works well in Capacitor 5.0 with previous versions.

See screenshot

<img width="1509" alt="stacktrace" src="https://github.com/ionic-team/capacitor/assets/89339796/7e7b398f-9aa7-4105-967f-e207625233d9">

### Expected Behavior

App should load when server url includes Basic Auth. If it's intended that Basic Auth will not be allowed please respond with reason for the change.

### Project Reproduction

https://github.com/patrik-skilling/public-capacitor-6-basic-auth-demo

patrik-skilling:
**Question**
Should you be able to get around this in Capacitor configuration with `allowNavigation`?

<img width="703" alt="Screenshot 2024-03-09 at 18 28 32" src="https://github.com/ionic-team/capacitor/assets/89339796/bbac0352-071c-4fca-a4aa-0be7ac76d9c6">

patrik-skilling:
**Question**
Should you be able to get around this in Capacitor configuration with `allowNavigation`?

<img width="725" alt="Screenshot 2024-03-09 at 18 31 32" src="https://github.com/ionic-team/capacitor/assets/89339796/9185f429-4dd8-4c7d-8aa1-67dc3aec2c17">

patrik-skilling:
Here is the code fix I made to solve it locally:

```
private void loadWebView() {
    final boolean html5mode = this.config.isHTML5Mode();

    // Start the local web server
    JSInjector injector = getJSInjector();
[...]

PR #7607, opened 2024-08-06
fix(android): better cleaning of allowedOrigin url

jcesarmobile:
clean the path and fragment before passing the `allowedOrigin` to `addDocumentStartJavaScript`

Also catch `IllegalArgumentException` in case the url is still not allowed by `addDocumentStartJavaScript` so it uses the old injector instead of crashing the app.

closes https://github.com/ionic-team/capacitor/pull/7597
closes https://github.com/ionic-team/capacitor/issues/7596
closes https://github.com/ionic-team/capacitor/issues/7327


markemer (approved):
Oh, I like it, much better.

theproducer (approved):

PR #7614, opened 2024-08-09
refactor(cli): Remove dependency to @ionic/utils-process

jcesarmobile:


markemer (approved):
Works for me. Seems like something we didn't need.

andredestro (approved):

IT-MikeS (approved):

Issue #7632, opened 2024-08-27
[Bug]: iOS build fails with ‘Command line “app-store” is deprecated’


pthorp:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/android: 6.1.2
  @capacitor/ios: 6.1.2

Installed Dependencies:

  @capacitor/cli: 6.1.2
  @capacitor/android: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/ios: 6.1.2

[success] iOS looking great! 👌
[success] Android looking great! 👌

### Other API Details

```Shell
npm: 10.8.2
node: v22.3.0
```

### Platforms Affected

- [X] iOS
- [ ] Android
- [ ] Web

### Current Behavior

When performing an iOS build in the terminal, or as part of the CI process, an error is thrown:
```
> npx cap build ios

✔ Building xArchive in 27.11s
✖ Building IPA - failed!
[error] 2024-08-22 15:00:48.749 xcodebuild[67491:5632754] [MT] IDEDistribution: -[IDEDistributionLogging
        _createLoggingBundleAtPath:]: Created bundle at path "/var/folders/rt/24tdz11j2pz60hc02rsd7w3w9900gn/T/App_2024-08-22_15-00-48.732.xcdistributionlogs".
        2024-08-22 15:00:48.851 xcodebuild[67491:5632754] [MT] IDEDistribution: Command line name "app-store" is
        deprecated. Use "app-store-connect" instead.
```

### Expected Behavior

The generated the file with App store information should contain 'app-store-connect', rather than 'app-store'.

### Project Reproduction

-

jcesarmobile:
is that the complete error?

I "reproduced" the problem, but turns out it was just a warning and my error was something else.

pthorp:
Indeed, that was the full error from the console.  It was also the only error that was shown in any of the logs, in the **IDEDistribution.critical.log**, and the IPA file was not generated.

jcesarmobile:
can you run `xcodebuild archive -archivePath App.xcarchive -exportArchive -exportOptionsPlist archive.plist -exportPath output -allowProvisioningUpdates -configuration Release` from `ios/App/` folder?

Are you using automatic or manual signing?

pthorp:
The `archive.plist` file is not present in that folder, so the command fails with a `no such file` exception.

e.g. 
[...]

PR #7637, opened 2024-09-02
fix(cli): replace app-store deprecated method on build

jcesarmobile:
`app-store` is deprecated and was replaced with `app-store-connect` in Xcode 14.3.
It doesn't cause build failures yet, but better replace it now before it causes issues or breaks builds in the future.

It won't be cherry picked into 5.x branch since Capacitor 5 supports Xcode 14 and would require additional logic to use one value or the other based on the Xcode version.

closes https://github.com/ionic-team/capacitor/issues/7632

### DIFF
--- a/cli/src/ios/build.ts
+++ b/cli/src/ios/build.ts
@@ -53,7 +53,7 @@ export async function buildiOS(
 <plist version="1.0">
 <dict>
 <key>method</key>
-<string>app-store</string>
+<string>app-store-connect</string>
 </dict>
 </plist>`;
 


andredestro (approved):

theproducer (approved):

nsteenbeek:
I don't think Apple replaced app-store by app-store-connect in Xcode 14.3. If I overlook something please let me know. 
I did following on my MacBook:

xcodebuild -version
`Xcode 14.3.1
Build version 14E300c
`

xcodebuild --help
`method : String Describes how XCode should export the archive. Available options: app-store, validation, ad-hoc, package, enterprise, development, developer-id, and mac-application. The list of options varies based on the type of archive. Defaults to development.`

The `npx cap sync ios` does not reflect the capacitor.config.ts exportMethod resulting in I'm not able to use `npx cap build ios`

jcesarmobile:
Capacitor 6-7 don't support Xcode 14, I explained in the description that the changes weren't going to be cherry-picked into Capacitor 5 because it still supports Xcode 14, so the change was only made in Capacitor 6-7.
So if you want to use Xcode 14 you have to use Capacitor 5, which is 2 version behind, and also Xcode 15 is required for submitting apps to the Apple Store since almost a year, and soon (April 24, 2025) Xcode 16 will be required
https://developer.apple.com/news/upcoming-requirements/?id=02212025a

PR #7638, opened 2024-09-03
feat: add method `getDouble(...)` to plugin config

### LINKED ISSUE #7634: [Feature]: add method `getDouble` to `PluginConfig.swift`
### Description

There is currently no way to get a configuration value of type Double.

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web

Add a new method `getDouble` to [`PluginConfig.swift`](https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor/Capacitor/PluginConfig.swift):

```swift
@objc public func getDouble(_ configKey: String, _ defaultValue: Double) -> Double {
    if let val = (self.config)[keyPath: KeyPath(configKey)] as? Double {
        return val
    }
    return defaultValue
}
```

### Alternatives

_No response_

### Additional Information

_No response_

robingenz:
This PR adds a `getDouble(...)` method to the plugin config so that double values can also be read from the plugin config.

Close https://github.com/ionic-team/capacitor/issues/7634


robingenz:
I have just updated this branch and formatted the code. Let me know if anything is still missing.

eric-horodyski:
Thanks for opening this PR 🙏

fabiomartino:
This is a very thoughtful improvement to the core. The changes introduced in this PR move Capacitor toward a more robust and predictable native–JS interaction model, especially around consistency and extensibility of the bridge layer.

What stands out positively:

* It strengthens internal abstractions without leaking additional complexity to plugin authors.
* It improves long-term maintainability by reducing edge cases and implicit behaviors.
* It aligns well with Capacitor’s philosophy of having a stable, platform-agnostic JS surface backed by well-defined native semantics.
* It opens the door for cleaner plugin implementations that can rely more confidently on core guarantees instead of custom workarounds.

From a plugin architecture perspective, having these capabilities in the core significantly reduces duplication across community plugins and helps enforce best practices by default rather than by convention.

Overall, this feels like the kind of enhancement that, once merged, becomes invisible infrastructure — but makes the ecosystem more solid and easier to evolve. I think it would be very valuable to have this implemented in the core.

jcesarmobile (approved):
LGTM

Issue #7476, opened 2024-05-21
[Feature]: expose `methodName` via `CAPPluginCall`


robingenz:
### Description

Expose the method name of the plugin method that was called.

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web

### Request or proposed solution

On **Android**, we can call `getMethodName()` to get the method name of the plugin method that was called.
On **iOS**, there is no such method/property. It would be great if there was a `methodName` property in `CAPPluginCall`.

robingenz:
Would a PR be welcome?

markemer:
> Would a PR be welcome?

Yeah, this seems reasonable - I just don't know when I can get to it, so a PR would be great.

robingenz:
> Yeah, this seems reasonable - I just don't know when I can get to it, so a PR would be great.

Done: #7641

PR #7641, opened 2024-09-03
feat(core): expose `methodName` via `CAPPluginCall`

robingenz:
Close https://github.com/ionic-team/capacitor/issues/7476


markemer:
Yeah - so this is a pretty breaking change - we'd need to have both versions of the plugin call or do this in a more subtle way.

robingenz:
Why is it a breaking change? Who else creates instances of `CAPPluginCall`? Or am I missing something obvious?

markemer:
> Why is it a breaking change? Who else creates instances of `CAPPluginCall`? Or am I missing something obvious?

You may be right actually - I think I was looking at this a bit early in the morning here - let me double check, I think I was been confusing it with a different function.

markemer:
Yeah, these are available publicly - I don't think anyone is using them - but leaving the old version with a deprecation tag makes some sense - and we can just set it to blank or "unknown" in case someone does use it.

robingenz:
> Edit: I can add that real fast if you don't mind me messing with your branch.

Yes, feel free to make the changes. 🙌

robingenz:
@markemer I have added the old version with a deprecation tag again. Let me know if you need any other changes.

markemer (approved):

Steven0351 (approved):
lgtm

andredestro (approved):

PR #7644, opened 2024-09-04
feat(cli)!: Remove deprecated cordova.staticPlugins option

jcesarmobile:
We haven’t add any plugins to the list of plugins that need to be static in the last 3 years.

If the plugin is using `podspec` tag instead of framework tag we can know if it needs to be static if `use-framework` attribute is provided.

cordova-ios 7+ don’t support `framework` tag for CocoaPods dependencies anymore, it was released over a year ago.


markemer (approved):
Well we promised, I think we should remove a lot more stuff that has been deprecated for a long time for Cap 7.

eric-horodyski (approved):

Issue #7576, opened 2024-07-22
[Feature]: Extend JSValueDecoder/JSValueDecoder with More Control over Dates, Data, and Numbers


grantjbutler:
### Description

It'd be great if we had more control over how dates, data, and non-confirming floats were encoded and decoded.

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web

### Request or proposed solution

`JSONEncoder`/`JSONDecoder` from Foundation allow you to configure the object with a strategy as to how to encode or decode dates, data, and non-conforming floating point numbers. This is most helpful when working with dates, as there are many ways to encode or decode dates from some structured data (it could be formatted as a UNIX timestamp in seconds or milliseconds, or be formatted as an ISO 8601 string).

It'd be great if `JSValueDecoder` and `JSValueEncoder` supported some or all of the various ways that `JSONDecoder` and `JSONEncoder` can be configured to transform these various values.

Steven0351:
Data isn't actually a valid type to pass through to a WKWebView. As far as non-conforming floats, everything actually passes over the bridge as an NSNumber, so I'm not 100% sure anything there could happen. I have, however, opened PR #7647 to address Date and URL issues.

Steven0351:
Thinking more about the non-conforming float strategies, I would believe today those are not throwing at all because Infinity, -Infinity, and NaN are valid NSNumbers. We would need our own strategy that allows for not throwing or converting to a string.

PR #7647, opened 2024-09-06
feat(ios): JSValueEncoder/Decoder feature parity with JSONEncoder/Decoder

Steven0351:
The default implementation of URL encodes to/decodes from a dictionary, while Date by default encodes to/decodes from seconds since January 1, 2001.

This implementation is similar to the method for handling dates and urls in [Foundation](https://github.com/apple/swift-foundation/blob/da80d51fa3e77f3e7ed57c4300a870689e755713/Sources/FoundationEssentials/JSON/JSONEncoder.swift#L1140).


Steven0351:
This should partially address #7576

Steven0351:
After considering #7576 I think it makes sense for this PR to address the overall feature disparity between JSONEncoder/Decoder and JSValueEncoder/Decoder

markemer (approved):
I like it. I mean but I the fact we need to do it is Apple's fault, but given our options this is the way, I think.

theproducer (approved):

PR #7656, opened 2024-09-12
feat(cli)!: Improving App ID validation

theproducer:
Improves App ID validation during `cap init` to prevent the use of invalid Android package ids.  Also adds a `skip-appid-validation` flag to bypass the validation check.


markemer (approved):

cli/src/common.ts
```diff
@@ -140,12 +140,18 @@ export async function checkAppId(
...
+  if (/^[a-zA-Z][\w]*(?:\.[a-zA-Z][\w]*)+$/.test(id.toLowerCase())) {
     return null;
   }
-  return `Invalid App ID "${id}". Must be in Java package form with no dashes (ex: com.example.app)`;
+  return `
+    Invalid App ID "${id}". Your App ID must meet the following requirements:
+    - Must be in Java package form with no dashes (ex: com.example.app)
+    - It must have at least two segments (one or more dots).
+    - Each segment must start with a letter.
+    - All characters must be alphanumeric or an underscore [a-zA-Z0-9_].
```
markemer, line 153: `_` will fail on iOS won't it?

andredestro (approved):

PR #7658, opened 2024-09-13
feat(ios): Add open autoRegisterPlugins var on CAPBridgeViewController

Steven0351:


Steven0351:
We can revisit figuring out a better solution when we actually figure out what we're going to do. This PR was premature

PR #7665, opened 2024-09-19
chore: include proper template folders for swiftlint

jcesarmobile:
The ios template was renamed last year but the path in `swiftlint.config.js` was not adjusted and also the SPM template was not added, so corrected the path and added the SPM path too.

Then ran `npm run fmt` with swiftlint 0.57.0 and it made those changes.

We have some tests folders in the excluded lists, not sure why, but not the `CodableTests` folders, not sure if we want to exclude that too, or remove the exclusion of the other tests folders.


markemer (approved):

theproducer (approved):
> We have some tests folders in the excluded lists, not sure why, but not the CodableTests folders, not sure if we want to exclude that too, or remove the exclusion of the other tests folders.

I would go ahead and remove the exclusion from the CapacitorTests folder and fix whatever lint issues that come up (if any).

Steven0351 (approved):

Steven0351:
All these corrections make sense cause the only time I ever remember to run the formatter is when swiftlint fails 😂

PR #7670, opened 2024-09-23
chore(android-template): update minSdkVersion to 23

eric-horodyski:
Bump the `minSdkVersion` of Capacitor-generated Android projects to API level 23.

### DIFF
--- a/android-template/variables.gradle
+++ b/android-template/variables.gradle
@@ -1,5 +1,5 @@
 ext {
-    minSdkVersion = 22
+    minSdkVersion = 23
     compileSdkVersion = 34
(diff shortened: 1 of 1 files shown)


jcesarmobile (approved):

theproducer (approved):

hborders:
Why is the `minSdkVersion` now `23`? I can't find any issues that required this change. I'm not trying to say this is wrong. I'm just curious about why.

hborders:
I see that you want to stop supporting implicit permission requests with https://github.com/ionic-team/capacitor/pull/7674 so this would require minSdk 23.

PR #7674, opened 2024-09-24
chore(android)!: Bump default minSdkVersion to 23

jcesarmobile:
And remove code that becomes irrelevant after the change since on android 23+ you always have to request the permissions.


eric-horodyski (approved):
LGTM

PR #7679, opened 2024-09-26
chore(ios)!: Bump minimum deployment target to 14.0

jcesarmobile:
Bump the minimum deployment target to 14 for podspec files, native projecs and tests projects (for some of them I had to downgrade)

Remove code that becomes irrelevant after the bump.


ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
```diff
@@ -1116,7 +1118,7 @@
 				ENABLE_USER_SCRIPT_SANDBOXING = YES;
 				GCC_C_LANGUAGE_STANDARD = gnu17;
 				GENERATE_INFOPLIST_FILE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 17.0;
```
eric-horodyski, line 1119: Just curious, was there a specific reason this was previously set to `17.0`?

eric-horodyski (approved):

ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
jcesarmobile: It's from a "CodableTests" subproject, probably when created, Xcode set the version to 17 as the default and was not changed. Doesn't really affect the project or the users, but changed it for consistency.

theproducer (changes requested):
Looks like some of the CLI tests are failing for iOS.  Seems to be related to the test project its using having a version lower than the new deployment target set.

jcesarmobile:
Yeah, I'm updating the deployment target for the CLI and once that's merged should fix this

theproducer (approved):

Steven0351 (approved):

PR #7692, opened 2024-10-02
fix(ios): fix retain cycle caused by CDVPluginManager

Steven0351:
This strong reference to the `UIViewController` and `WKWebView` owned by Capacitor causes a retain cycle that does not allow for anything to be released from memory when using Cordova plugins. This issue isn't present in typical Capacitor applications, but any scenario where a `CAPBridgeViewController` that uses Cordova plugins is presented will stay in memory even when the view controller is dismissed.


markemer (approved):
Yep. That'll do it.

jcesarmobile (approved):

andredestro (approved):

PR #7709, opened 2024-10-10
feat(android): add interface for messagehandler

### LINKED ISSUE #7700: [Feature]: Better decouple PluginCall from MessageHandler class
### Description

Replace the first Parameter of the Constructor of PluginCall with an interface

### Platforms

- [ ] iOS
- [X] Android
- [ ] Web

I am building a solution, that uses an alternative JS runtime to run Capacitor Plugins not only from within the webview. This allows to execute JS in the background and still use existing plugins for native functionality.

The PoC is working, but I was facing issues as I need to extend MessageHandler class in order to properly create PluginCalls. As MessageHandler expects a webView instance, I have no way to make it work cleanly.

It would be easy to replace MessageHandler with an Interface, that specifies postMessage and sendReponseMessage signatures for PluginCall to make everything more flexible.

### Alternatives

_No response_

### Additional Information

_No response_

phal0r:
This fixes #7700 and adds an interface, so other JS runtimes can be coupled with Capacitor and run Capacitor plugins from within Javascript.


eric-horodyski:
Thanks for opening this PR 🙏

phal0r:
@eric-horodyski yes, there is activity :) I rebased to the latest main, can you run the workflow? It does not trigger automatically it seems

PR #7714, opened 2024-10-11
feat(cli)!: change telemetry from opt-in to opt-out

eric-horodyski:
Modify telemetry such that is it opt-out instead of the current opt-in behavior:

- Telemetry checks only occur on specified commands (`init`, `sync`, `copy`, `update`, etc.) per the existing source.
- If the user has not yet opted in-or-out: 
  - The user will be automatically opted-in.
  - The user will be printed out a thank you message with instructions on how to opt-out (if they wish).
- Once the user has made a preference one way or the other, they will not see any telemetry messaging.
- Users with existing telemetry preferences will have them preserved.

### DIFF
--- a/cli/src/telemetry.ts
+++ b/cli/src/telemetry.ts
@@ -4,7 +4,7 @@ import Debug from 'debug';
 import c from './colors';
 import type { Config } from './definitions';
 import { send } from './ipc';
-import { logPrompt, output } from './log';
+import { logger } from './log';
 import { readConfig, writeConfig } from './sysconfig';
 import type { SystemConfig } from './sysconfig';
 import { getCommandOutput } from './util/subprocess';
@@ -13,10 +13,9 @@ import { isInteractive } from './util/term';
 const debug = Debug('capacitor:telemetry');
 
 export const THANK_YOU =
-  `\nThank you for helping to make Capacitor better! 💖` +
-  `\nInformation about the data we collect is available on our website: ${c.strong(
-    'https://capacitorjs.com/telemetry',
-  )}\n`;
+  `\nThank you for helping improve Capacitor by sharing anonymous usage data! 💖` +
+  `\nInformation about the data we collect is available on our website: ${c.strong('https://capacitorjs.com/telemetry')}` +
+  `\nYou can change your mind at any time by using the ${c.input('npx cap telemetry')} command.`;
 
 export interface CommandMetricData {
   app_id: string;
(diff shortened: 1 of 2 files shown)


markemer (approved):

jcesarmobile (changes requested):
why change the `output.write` to `logger.msg`. I don't see any visible differences between them, but on `cli/src/tasks/telemetry.ts` the `THANK_YOU` message is also written still with `output.write`, so for consistency I would keep `output.write` for both.

The `THANK_YOU` message needs more rewording as it says "you can change your mind", but that made sense when the user was prompted, now it isn't their decision, so should  say something like "You can disable telemetry at any time", and provide the whole command to disable it `npx cap telemetry off`.

Also, we should update the telemetry docs for Capacitor 7, can you create a ticket for tracking it? 
We don't have Capacitor 7 alpha docs yet, but I have asked Jared to create them.

eric-horodyski:
> why change the `output.write` to `logger.msg`. I don't see any visible differences between them, but on `cli/src/tasks/telemetry.ts` the `THANK_YOU` message is also written still with `output.write`, so for consistency I would keep `output.write` for both.

For some reason, `output.write` caps at writing 2 lines of text so I had to use `logger.msg`.

eric-horodyski:
Two changes have been added:

1. Use `output.write()` to write out the message when the user is opted-in.
2. Retain the original messaging (from when telemetry was opt-out) to be written out when the user manually turns telemetry on.

cli/src/telemetry.ts
```diff
@@ -4,19 +4,18 @@ import Debug from 'debug';
...
 
-export const THANK_YOU =
-  `\nThank you for helping to make Capacitor better! 💖` +
-  `\nInformation about the data we collect is available on our website: ${c.strong(
-    'https://capacitorjs.com/telemetry',
-  )}\n`;
+const THANK_YOU =
+  `\nThank you for helping improve Capacitor by sharing anonymous usage data! 💖` +
+  `\nInformation about the data we collect is available on our website: ${c.strong('https://capacitorjs.com/telemetry')}` +
+  `\nYou can disable telemetry at any time by using the ${c.input('npx cap telemetry')} command.`;
```
jcesarmobile, line 18:
```suggestion
  `\nYou can disable telemetry at any time by using the ${c.input('npx cap telemetry off')} command.`;
```

markemer (approved):
Other than Julio's change, this works.

PR #7723, opened 2024-10-16
feat(cli): Check npm workspace packages for plugins

laine-hallot:
## Info
No one's really asked for this feature but it was straight forward to implement and pretty nice to have for my team so I figured I should still make a PR in case this is useful to anyone else.

This PR just lets you tell the capacitor CLI to check your workspace packages for plugins and copy those over to `ios/` and `android/` the same way it would with plugins inside `node_modules/`

## Testing
1. Clone the demo repo `git clone https://github.com/laine-hallot/cap-cli-npm-workspace-example`
2. `cd cap-cli-npm-workspace-example/`
3. `npm i`
4. `npm -ws run build`
5. `npx -w my-app cap sync`
    - You should see ` @korterra/example-workspace-plugin@0.0.1` in the plugin list
6. `npx -w my-app/ cap open ios` (or `npx -w my-app/ cap open android`)
7. Build and run the app
8. Once the finishes loading filling in the text box at the top of the screen then press the button right under it
    - The text you entered should show up next to the input and you should see output from the plugin in both the JS console and native logs

#### About the dependency
So I decided to pull in `@npmcli/map-workspaces` to match paths/globs in `workspace:[]`  partially out of laziness but mostly since its the same package that `npm` itself uses to manage workspace paths. 

https://www.npmjs.com/package/@npmcli/map-workspaces
https://github.com/npm/cli/blob/286739c0224bad88c4a38927bafd61973f71098c/lib/utils/get-workspaces.js#L2

I could get rid of it and write the glob matching from scratch if the maintainers want to keep dependencies to a minimum but I figured pulling that in was worth the piece of mind that comes with knowing that Capacitor's workspace path resolution is identical to `npm`'s.

#### Config option
I added a new capacitor config option called `workspaces` and set this feature to only run if that option is enabled. 

The type on `workspaces` only accepts `'npm'` here since I'm pretty sure this won't work with other workspace setups in stuff like `yarn` and `pnpm` (Well I'm almost certain pnpm wont work but yarn might(??)). 

If there's interest in fully supporting those package managers down the line then the type on `workspaces` could be expanded later with a union type like `'npm' | 'yarn' | 'pnpm'` or just set to `string` (again I'm just guessing this implementation wont work outside of `npm` workspaces). If supporting other workspace setups isn't something the maintainers ever want to deal with then the config option could be changed to something like `npmWorkspaces?: boolean`.

#### Build quirks
A minor quirk of this implementation is that unlike with plugins inside `node_modules/` you do need to make sure you run your build process on your plugin code before you run `npx cap sync`. I figure its not a big deal though since `sync` never handled building your apps JS/TS code anyways.

## Screenshots
cap-cli-workpaces-demo-sync

cap-cli-workpaces-demo-app

### DIFF
--- a/cli/src/config.ts
+++ b/cli/src/config.ts
@@ -10,7 +10,7 @@ import { fatal, isFatal } from './errors';
 import { logger } from './log';
 import { tryFn } from './util/fn';
 import { formatJSObject } from './util/js';
-import { findNXMonorepoRoot, isNXMonorepo } from './util/monorepotools';
+import { findMonorepoRoot, findNXMonorepoRoot, isMonorepo, isNXMonorepo } from './util/monorepotools';
 import { requireTS, resolveNode } from './util/node';
 import { lazy } from './util/promise';
 import { getCommandOutput } from './util/subprocess';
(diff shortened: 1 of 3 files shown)


eric-horodyski:
Thanks for opening this PR 🙏

laine-hallot:
@eric-horodyski I've mostly forgotten about this PR by now but sure I could update it if theres interested in reviewing it this time. I haven't used capacitor in a little bit but it would be nice to have if I ever come back to it

PR #7725, opened 2024-10-17
chore(ios): Fix typos in README.md file in SPM template

piotr-cz:
Just fixed few typos in README.md


jcesarmobile:
Thanks for the PR, looks like [an exact PR](https://github.com/ionic-team/capacitor/pull/7860) was sent after yours and was merged instead of yours, sorry about that.

PR #7733, opened 2024-10-23
feat(ios): return secure cookies on local scheme 

gjb-amazon:
This PR reworks the `CapacitorCookieManager.getServerUrl` method to modify the URL if it uses the scheme for serving local assets. By swapping the scheme to `https`, this allows calls to `HTTPCookieStorage` to return secure cookies when asking for cookies with the result of `getServerUrl`.

Happy to have a discussion here on this change. This is more or less a reference implementation to hopefully better convey what I'm expressing in https://github.com/ionic-team/capacitor/discussions/7690.


eric-horodyski:

More context:

PR #7736, opened 2024-10-24
chore: run dispatch-release.yaml in 6.x branch

jcesarmobile:
I've created a 6.x branch in ionic-team/capacitor-swift-pm for Capacitor 6, make the 6.x Capacitor branch run the workflow in 6.x branch too

### DIFF
--- a/.github/workflows/publish-spm-release.yaml
+++ b/.github/workflows/publish-spm-release.yaml
@@ -12,6 +12,6 @@ jobs:
       - name: Assign Tag Number to RELEASE_VERSION environment variable
         run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
       - name: Dispatch Release Workflow on ionic-team/capacitor-swift-pm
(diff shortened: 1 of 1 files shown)


Steven0351:
This shouldn't be needed, this workflow runs when a tag is pushed, not when a commit is made to main.

jcesarmobile:
Both Capacitor and Capacitor plugins are pointing to "main", which at the moment is Capacitor 7 alpha 1, I've created a 6.x branch in `ionic-team/capacitor-swift-pm` pointing to latest 6.x release and sent PRs to Capacitor and Capacitor Plugins 6.x branch to make them use 6.x branch instead of main, but we need that whenever we publish a new 6.x version it's published on the 6.x branch, not in main branch, and I though this change would do that, but not sure how `ionic-team/capacitor-swift-pm` publishing works and if this change will have that effect. If this doesn't make the publish action in `ionic-team/capacitor-swift-pm` repo to publish on main, we will have to change the action code in 6.x branch to publish in 6.x branch.

https://github.com/ionic-team/capacitor/pull/7737
https://github.com/ionic-team/capacitor-plugins/pull/2221

Steven0351:
A simpler approach would to just have a version range in the Package.swift to be 6.0.0 < 7.0.0

Issue #7735, opened 2024-10-24
[Bug]: SPM - capacitor-swift-pm.git Version declared to branch main


7-plus-t:
### Capacitor Version

Latest Dependencies:

  @capacitor/cli: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/android: 6.1.2
  @capacitor/ios: 6.1.2

Installed Dependencies:

  @capacitor/cli: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/ios: 6.1.2
  @capacitor/android: 6.1.2

### Other API Details

```Shell
> npm --version
10.9.0

> node --version
v22.10.0
```

### Platforms Affected

- [X] iOS
- [ ] Android
- [ ] Web

### Current Behavior

Using the SPM beta for iOS:
In the `Package.swift` of CapApp-SPM, generated by `npx cap sync`, the dependency on Capacitor and Cordova is declared as follows: 
`.package(url: "https://github.com/ionic-team/capacitor-swift-pm.git", branch: "main"),`

This causes SPM to use the latest commit of `https://github.com/ionic-team/capacitor-swift-pm.git` and therefore always the newest Capacitor version is used. This cannot be overridden by the parent app, as the `branch: "main"` overrules other constraintes like `exact` or `from` when there is a conflict.

A similar problem exists in all the Capacitor plugins that have been converted to the SPM structure, where in `Package.swift` the dependency to Capacitor is declared with `branch: "main"`.

### Expected Behavior

The `Package.swift` of CapApp-SPM and the plugins should be changed, so it states:
`.package(url: "https://github.com/ionic-team/capacitor-swift-pm.git", from: "6.0.0"),` (or `7.0.0` for the upcoming Capacitor release).
This allows the parent app to declare the version of Capacitor to be used, without SPM resolution conflicts.

### Project Reproduction

Use any plugin example app

flgubler-ergon:
We have faced the same problem in our project: the build suddenly started using the _capacitor 7_ alpha version which obviously is unacceptable for a productive app.

A reproduction can be found here: https://github.com/ionic-team/capacitor-ios-spm-template/blob/main/template/App/CapApp-SPM/Package.swift - it is apparently a different dependency, but uses `branch: "main"` as well.

7-plus-t:
see https://github.com/ionic-team/capacitor-plugins/pull/2221 for fix in plugins

flgubler-ergon:
Cool, thanks

PR #7737, opened 2024-10-24
fix: use Capacitor 6 for SPM dependency

jcesarmobile:
closes https://github.com/ionic-team/capacitor/issues/7735

### DIFF
--- a/cli/src/util/spm.ts
+++ b/cli/src/util/spm.ts
@@ -1,6 +1,7 @@
 import { existsSync, readFileSync, writeFileSync } from '@ionic/utils-fs';
 import { join, relative, resolve } from 'path';
 
+import { getCapacitorPackageVersion } from '../common';
 import type { Config } from '../definitions';
 import { logger } from '../log';
 import type { Plugin } from '../plugin';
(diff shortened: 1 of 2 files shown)


cli/src/util/spm.ts
```diff
@@ -69,7 +69,7 @@ let package = Package(
             targets: ["CapApp-SPM"])
     ],
     dependencies: [
-        .package(url: "https://github.com/ionic-team/capacitor-swift-pm.git", branch: "main")`;
+        .package(url: "https://github.com/ionic-team/capacitor-swift-pm.git", from: "6.0.0")`;
```
jcesarmobile, line 72: maybe here we might want to set the version to the locally installed `@capacitor/ios` version?

Steven0351: yeah, the fact that this is actually pulling in whatever the latest is from github versus what was installed via npm is problematic

cli/src/util/spm.ts
```diff
@@ -69,7 +78,7 @@ let package = Package(
             targets: ["CapApp-SPM"])
     ],
     dependencies: [
-        .package(url: "https://github.com/ionic-team/capacitor-swift-pm.git", branch: "main")`;
+        .package(url: "https://github.com/ionic-team/capacitor-swift-pm.git", exact: "${iosPlatformVersion}")`;
```
markemer, line 81: 👍

markemer (approved):

eric-horodyski (approved):

PR #7753, opened 2024-11-07
fix(ios): Make Bridge webView first responder

theproducer:
Fixes a focus issue ([reproduction](https://github.com/kensodemann/issue-51750)) caused when the webView is not the first responder.  CAPBridgeViewController's `self.becomeFirstResponder()` was being called during `viewDidLoad`, before the view's window is set, which had no effect.


markemer (approved):

jcesarmobile (commented):
Using `self.webView?.becomeFirstResponder()` in the same line were `self.becomeFirstResponder()` was also seems to work fine without the addition of the `viewDidAppear` override.

markemer:
> Using `self.webView?.becomeFirstResponder()` in the same line were `self.becomeFirstResponder()` was also seems to work fine without the addition of the `viewDidAppear` override.
> 
> Rethinking this comment, the `viewDidAppear` method will be called in apps that embed the `CAPBridgeViewController.swift` whenever it appears, so putting the line there will make the WebView to regain focus, so looks like a better place than `viewDidLoad`.
> 
> @Steven0351 thoughts?

Yeah - `viewDidAppear` is probably better - but, I don't want to break Portal, et. al.

jcesarmobile:
Not portals but embedded Capacitor, not sure what his name is nowadays.

I think this should be behind a `initialFocus` configuration option like on Android and also make a common `initialFocus` configuration option for both platforms as we do for some other settings so can be set globally or per platform.

jcesarmobile (approved):

PR #7764, opened 2024-11-18
feat(android)!: Make service worker requests go through Capacitor bridge

jcesarmobile:
Marking it as breaking as some users might already have a similar code in their MainActivity file or in some plugin and will need to remove that code/plugin or disable Capacitor's registration if they want to keep their own handling.


markemer:
Is there are good way to test this is working?

jcesarmobile:
You can use https://github.com/jcesarmobile/app-capable which I created for a different thing, but exposes the problem (it doesn't have platforms added, you'll need to add android to test)
Or also this one from an user
https://github.com/radonirinamaminiaina/capacitor-sw

You'll see this error in logcat
`File: https://localhost/main.9677d6e520b8922e.js - Line 1 - Msg: Service worker registration failed with: TypeError: Failed to register a ServiceWorker for scope ('https://localhost/') with script ('https://localhost/ngsw-worker.js'): An unknown error occurred when fetching the script.`

With this change applied the error is gone

markemer (approved):

andredestro (approved):

Issue #7625, opened 2024-08-20
feat(cli): support manual signing on build command


pfumagalli:
### Capacitor Version

Latest Dependencies:

  @capacitor/cli: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/android: 6.1.2
  @capacitor/ios: 6.1.2

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/cli: 6.1.2
  @capacitor/core: 6.1.2
  @capacitor/ios: 6.1.2

### Platforms Affected

- [X] iOS
- [ ] Android
- [ ] Web

### Current Behavior

When disabling "Automatically Manage Signing" and setting up the correct Provisioning Profiles and Signing Certificate, and running `cap build ios`, `xcodebuild` crashes with a segmentation fault.

Basically `cap build ios` runs a couple of commands:

* `xcodebuild -workspace App.xcworkspace -scheme App -destination generic/platform=iOS -archivePath App.xcarchive archive` to build the xArchive
* `xcodebuild archive -archivePath App.xcarchive -exportArchive -exportOptionsPlist archive.plist -exportPath output -allowProvisioningUpdates -configuration Release` to build the IPA file

The issue lies with the `archive.plist` file generated [here (cli/src/ios/build.ts#L51-L58)](https://github.com/ionic-team/capacitor/blob/main/cli/src/ios/build.ts#L51-L58).

It seems to be missing the `signingStyle` key. Even setting it to `automatic` (rather than `manual`) makes `xcodebuild` work:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
  <dict>
    <key>method</key>
    <string>app-store</string>
    <key>signingStyle</key>
    <string>automatic</string>
  </dict>
</plist>
```

### Expected Behavior

Well,  `xcodebuild` shouldn't segfault!

### Project Reproduction

http://none-avail/

### Additional Information

Apologies for the lack of reproduction, but without the proper provisioning profiles and certificates (obtainable only from Apple) there's not much we can do to "reproduce" this...

Acanthostega:
Hi,
I can reproduce this problem, using manual signing too.

[...]

Issue #6231, opened 2023-01-19
feat: Signed app building config in CLI for iOS apps


IT-MikeS:
### Description
Based on this [forum post](https://forum.ionicframework.com/t/ios-where-to-enable-codesignidentity-developmentteam-previously-set-in-cordova-build-json/230416) config options should be added to the CLI for building signed iOS apps similar to what is currently in place for Android.

vochitan90:
Hi @IT-MikeS 

Is there any update? I'm still facing this problem.

rvssgh:
This is a must have for IONIC to be a viable option for app development.

PR #7769, opened 2024-11-19
feat(cli): add more configurations to build command

theproducer:
closes: https://github.com/ionic-team/capacitor/issues/7625
closes https://github.com/ionic-team/capacitor/issues/6231


theproducer:
Im open to any suggestions for better names for the new CLI flags as well.

cli/src/ios/build.ts
```diff
@@ -45,12 +45,23 @@ export async function buildiOS(config: Config, buildOptions: BuildCommandOptions
     ),
   );
 
+  const manualSigningContents = `<key>provisioningProfiles</key>
+<dict>
+<key>${config.app.appId}</key>
+<string>${buildOptions.xcodeProvisioningProfile ?? ''}</string>
+</dict>
+<key>signingCertificate</key>
+<string>${buildOptions.xcodeSigningCertificate ?? ''}</string>`;
```
eric-horodyski, line 54: Will these be OK as empty strings, or would it fail the build?

eric-horodyski:
> Im open to any suggestions for better names for the new CLI flags as well.

Honestly, they look great to me.

cli/src/ios/build.ts
theproducer: Yeah, that block is only added if signing style is manual, and if that's chosen, presumably, the developer would be supplying a provisioning profile and/or signing certificate.  If what they supply is wrong (or if it's empty), then it won't work.

I believe the signing certificate can be optional, so I may push up a change to remove that key and string if the value is empty.

markemer (approved):

cli/src/tasks/build.ts
```diff
@@ -14,6 +14,9 @@ export interface BuildCommandOptions {
   androidreleasetype?: 'AAB' | 'APK';
   signingtype?: 'apksigner' | 'jarsigner';
   configuration: string;
+  xcodeSigningType: 'automatic' | 'manual';
```
jcesarmobile, line 17:
```suggestion
  xcodeSigningType?: 'automatic' | 'manual';
```
should be marked as optional

cli/src/index.ts
```diff
@@ -147,6 +147,24 @@ export function runProgram(config: Config): void {
         'jarsigner',
       ]),
     )
+    .addOption(
+      new Option(
+        '--xcode-signing-style <xcodeSigningStyle>',
+        'The iOS signing style to use when building the app for distribution',
```
jcesarmobile, line 153: should indicate that automatic is the default

jcesarmobile (changes requested):
there should be some validation before trying to build as the build takes a lot of time to end up failing if the signing is set to manual but the user didn't provide the certificate and provisioning profile.

jcesarmobile (changes requested):
It should also be possible to provide the values from the Capacitor config file as we do for Android
https://github.com/ionic-team/capacitor/blob/f00349ed57e56f6f6182947baa2d467c885f58e8/cli/src/config.ts#L218-L225

jcesarmobile (changes requested):
We should delete the `buildOptions` from the generated `capacitor.config.json` as we do for Android since it can be sensitive.
https://github.com/ionic-team/capacitor/blob/main/cli/src/tasks/copy.ts#L156

Also, should we ask for the development team?
archive fails if I don't have one selected on the Xcode project, it's not related to this change since it also fails with the automatic builds, but I don't think having a team selected should be a requirement for doing a CLI build.
Cordova has a param for it, so we should probably also ask for it.

And I don't think the provisioning profile is working as expected, if I have one selected on Xcode that is invalid the IPA build fails despite the one I passed as param is valid, if I make the Xcode one match the one I pass as param then it works, but defeats the purpose of allowing to pass a provisioning profile if it has to be the same Xcode already has selected.

cli/src/ios/build.ts
```diff
@@ -45,12 +52,23 @@ export async function buildiOS(config: Config, buildOptions: BuildCommandOptions
...
+
   const archivePlistContents = `<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
 <key>method</key>
 <string>app-store-connect</string>
```
craigwatsoncambric, line 68: This method will depend on the provisioning profiles used, 'app-store-connect' is for a distribution profile.  It would be good to support the development and ad-hoc profiles also.

cli/src/index.ts
```diff
@@ -147,6 +147,41 @@ export function runProgram(config: Config): void {
...
     )
+    .addOption(
+      new Option('--xcode-team-id <xcodeTeamID>', 'The Developer team to use for building and exporting the archive'),
+    )
+    .addOption(
+      new Option(
+        '--xcode-export-method <xcodeExportMethod>',
+        'Describes how xcodebuild should export the archive (default:  debugging)',
+      ).choices([
+        'app-store-connect',
```
craigwatsoncambric, line 158: These look correct, verified via xcodebuild -help, method section

jcesarmobile (changes requested):
The latest changes are changing the previous behavior, since we are in RC we shouldn't make breaking changes.
So we should keep the previous exportMethod as default (`app-store-connect`), as that was the only one available in the latest release and with this change the default will be `debugging`.

Also there is no rush in addressing the changes, since we are in RC we shouldn't be including new features, this should go in a future 7.x minor release, and as it should be non breaking, we can probably backport it to a minor 6.x release too.

andredestro (approved):

jcesarmobile (changes requested):
The default to `app-store-connect` was changed on the option description, but not in the actual code

PR #7775, opened 2024-11-25
feat(cli): Add global initialFocus configuration

theproducer:


cli/src/declarations.ts
jcesarmobile, line 1: The android/ios `initialFocus` texts should include a new line that says something like
> Overrides global `initialFocus` option.
As we do in other options that have global and platform specific configuration

jcesarmobile (approved):

markemer (approved):

PR #7781, opened 2024-11-28
Inject Capacitor JS for error fallbacks.

cpojer:
When using a local `server.errorPath` in conjunction with a remote `server.url`, and the device is offline, Capacitor does not inject its JS because `allowedOrigin` will be set to the remote `server.url` hostname, while the error is loaded from the local server. This means that plugins and other Capacitor/Cordova functionality are unavailable in this situation.

This PR adds `https://localhost` as an `allowedOrigin`, ensuring that all the Capacitor JS is being included in the error case.

There is a related PR for the case when the document start script feature is not supported: https://github.com/ionic-team/capacitor/pull/6247

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -258,9 +258,12 @@ public class Bridge {
         // Start the local web server
         JSInjector injector = getJSInjector();
         if (WebViewFeature.isFeatureSupported(WebViewFeature.DOCUMENT_START_SCRIPT)) {
-            String allowedOrigin = Uri.parse(appUrl).buildUpon().path(null).fragment(null).clearQuery().build().toString();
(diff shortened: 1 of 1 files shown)


android/capacitor/src/main/java/com/getcapacitor/Bridge.java
```diff
@@ -258,9 +258,12 @@ private void loadWebView() {
         // Start the local web server
         JSInjector injector = getJSInjector();
         if (WebViewFeature.isFeatureSupported(WebViewFeature.DOCUMENT_START_SCRIPT)) {
-            String allowedOrigin = Uri.parse(appUrl).buildUpon().path(null).fragment(null).clearQuery().build().toString();
+            Set<String> allowedOrigins = new HashSet<>(Arrays.asList(
+                 Uri.parse(appUrl).buildUpon().path(null).fragment(null).clearQuery().build().toString(),
+                 "https://localhost"
```
cpojer, line 263: Is there ever a situation where we'd also want to add `http://localhost`?

PR #7800, opened 2024-12-09
fix: semver vulnerable version to 7.5.2

jimcase:
This pull request updates the semver library from a version vulnerable to a Regular Expression Denial of Service (ReDoS) attack to the patched version 7.5.2. The update addresses the vulnerability identified in the security advisory [GHSA-c2qf-rxjj-qqgw](https://github.com/advisories/GHSA-c2qf-rxjj-qqgw).


jcesarmobile:
Thanks for the pull request, but since the dependency is declared as `^7.3.7`, it should already pick latest 7.x version available.
Also the root `package.json` is not the one distributed on npm, but the one in `cli` folder. 

I've sent a PR that forces latest version of semver and updates several other packages for the CLI as for the CLI there could be cases where it could keep an older version of semver if it's already installed.

https://github.com/ionic-team/capacitor/pull/7808

PR #7835, opened 2025-01-20
fix: use Capacitor 7 for SPM dependency

jcesarmobile:
We did this for Capacitor 6 when we started publishing Capacitor 7 into SPM to make Capacitor 6 use only the 6 version and make it match with the npm installed one
But we also need this for Capacitor 7 in case we publish a Capacitor 6 version to main branch (that will happen when a 6.x release is made after a Capacitor 7 release)


markemer (approved):
Probably should have been doing this the whole time.

carlpoole (approved):

PR #7850, opened 2025-01-29
fix(ios): listen for CapacitorViewDidAppear

jcesarmobile:
If status bar plugin is installed, its `CAPBridgeViewController` extension prevents Capacitor's `viewDidAppear` method from being called.
Listen for the notification the extension emits to call the same code in `viewDidAppear`.


markemer (approved):
I don't love it, but I guess we're stuck for now.

theproducer (approved):

Issue #7789, opened 2024-12-03
[Feature]: Add PMTiles Byte Range Request Support for Offline Maps (#7664 Follow-Up)


m345054:
### Description

Support for .pmtiles in mediaExtensions within WebViewAssetHandler is needed to enable proper functionality of offline maps in Capacitor. Without this support, byte range requests (HTTP 206 Partial Content) required by PMTiles fail, resulting in errors when accessing bundled assets.

### Platforms

- [X] iOS
- [ ] Android
- [ ] Web

### Request or proposed solution

The feature involves adding .pmtiles to the list of supported extensions (mediaExtensions) in WebViewAssetHandler. This change ensures that .pmtiles assets can handle HTTP 206 Partial Content responses correctly.

An example project demonstrating the issue and validating the fix is available here: [capacitor-pmtiles](https://github.com/m345054/capacitor-pmtiles). Additionally, there is an existing pull request ([#7663](https://github.com/ionic-team/capacitor/pull/7663)) that implements this fix.

This support is critical for applications using PMTiles for offline vector map tilesets.

### Additional Information

This request is a follow-up to the closed issue [#7664](https://github.com/ionic-team/capacitor/issues/7664), which was closed due to a lack of an example project. The example project linked above addresses this and provides a clear demonstration of the need for this feature.

Thank you for considering this important enhancement!

leotohill:
I'm wondering why the support for range headers is limited to known types.  Why not .zip, .data, .ogg,  .pdf,  ... which is to say, why not anything?

petrot:
Thanks for the example repo. I've been waiting for this update since I know how it can solve my offline pmtiles problems.

PR #7868, opened 2025-02-05
fix(ios): don't check isMediaExtension on range requests

jcesarmobile:
There are multiple file extensions that support range requests, not only media extensions, we shouldn't be checking `isMediaExtension` for the requested resource if the request has `Range` header, if the header has `Range`, then it should return `206` and the appropriate range headers.

closes https://github.com/ionic-team/capacitor/issues/7789


markemer (approved):

theproducer (approved):

conbrad:
I saw this has been merged, any idea when a new release will be cut with this change?

PR #7870, opened 2025-02-05
fix(cli): fix Podfile having expanded symlinks

terencehonles:
This change fixes the `cap sync` command expanding symlinks in plugin paths.

I've built this and confirmed that it builds the `Podfile` I am expecting, and it limits the dependencies on `fs-extras`. This is in-line with the android code, and I've confirmed that plugin `rootPath`s are already sufficiently qualified.


eric-horodyski:

More context:

terencehonles:
I've migrated to SPM so I no longer need this change

PR #7885, opened 2025-02-11
feat(android): add adjustMarginsForEdgeToEdge configuration option

markemer:


jcesarmobile:
How did you set the `windowOptOutEdgeToEdgeEnforcement` for testing this?
I'm always getting the false default value having `<item name="android:windowOptOutEdgeToEdgeEnforcement" tools:ignore="NewApi">true</item>` in the `styles.xml`, but also tried setting it in for the WebView on `activity_main.xml`

dpogue:
Just a note that the `windowOptOutEdgeToEdgeEnforcement` option is removed in Android 16 (as per https://android-developers.googleblog.com/2025/02/second-beta-android16.html)

> ### Edge to edge opt-out going away
> [Android 15 enforced edge-to-edge](https://developer.android.com/about/versions/15/behavior-changes-15#edge-to-edge) for apps targeting Android 15 (SDK 35), but your app could opt-out by setting [`R.attr#windowOptOutEdgeToEdgeEnforcement`](https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement) to true. Once your app targets Android 16 (Baklava), `R.attr#windowOptOutEdgeToEdgeEnforcement` is deprecated and disabled and your app cannot opt-out of going edge-to-edge. To be compatible with Android 16 Beta 2, ensure your app supports edge-to-edge and remove any use of `R.attr#windowOptOutEdgeToEdgeEnforcement`. To support edge-to-edge, see the [Compose](https://developer.android.com/develop/ui/compose/layouts/insets) and [Views](https://developer.android.com/develop/ui/views/layout/edge-to-edge) guidance. Please let us know about concerns in our tracker on the [feedback page](https://developer.android.com/about/versions/16/feedback).

jcesarmobile (changes requested):

android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
```diff
@@ -54,19 +57,37 @@ public boolean dispatchKeyEvent(KeyEvent event) {
...
-    private void edgeToEdgeHandler() {
-        ViewCompat.setOnApplyWindowInsetsListener(this, (v, windowInsets) -> {
-            Insets insets = windowInsets.getInsets(WindowInsetsCompat.Type.systemBars());
+    public void edgeToEdgeHandler() {
+        CapConfig config;
+        if (bridge != null) {
+            config = bridge.getConfig();
+        } else {
+            config = CapConfig.loadDefault(getContext());
+        }
```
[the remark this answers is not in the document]
markemer: Yeah, I went back and forth on this, I'm with you. I'll change it back to feed in the bridge.

android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
```diff
@@ -54,19 +57,37 @@ public boolean dispatchKeyEvent(KeyEvent event) {
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            TypedValue value = new TypedValue();
+            foundOptOut = getContext().getTheme().resolveAttribute(android.R.attr.windowOptOutEdgeToEdgeEnforcement, value, true);
+            optOutValue = value.data != 0; // value is set to -1 on true as of Android 15, so we have to do this.
+        }
 
-            // Don't pass window insets to children
-            return WindowInsetsCompat.CONSUMED;
-        });
+        if (configEdgeToEdge || (foundOptOut && optOutValue)) {
```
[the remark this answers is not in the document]
markemer: it's `!(foundOptOut && optOutValue)` I think but yes, it's backward.

cli/src/declarations.ts
```diff
@@ -297,6 +297,16 @@ export interface CapacitorConfig {
      * @default true
      */
     resolveServiceWorkerRequests?: boolean;
+
+    /**
+     * If set to true, margins will be adjusted for edge to edge regardless of Android OS version,
+     * or the setting of [windowOptOutEdgeToEdgeEnforcement](https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement).
+     * If set to false or is missing, will only adjust margins on Android 15 with windowOptOutEdgeToEdgeEnforcement missing or set to false.
```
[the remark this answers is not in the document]
markemer: Rather than a boolean, I'm going to make it "auto", "force", etc. Move away from a boolean.

theproducer (approved):

android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
```diff
@@ -54,19 +56,37 @@ public boolean dispatchKeyEvent(KeyEvent event) {
...
+        boolean autoMargins = false;
+        boolean forceMargins = Objects.equals(configEdgeToEdge, "force");
+        boolean disableMargins = Objects.equals(configEdgeToEdge, "disable");
 
-            // Don't pass window insets to children
-            return WindowInsetsCompat.CONSUMED;
-        });
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            boolean foundOptOut = false;
+            boolean optOutValue = false;
```
jcesarmobile, line 68: Those can be initialized directly without first setting them to false

android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
```diff
@@ -54,19 +56,37 @@ public boolean dispatchKeyEvent(KeyEvent event) {
...
+        String configEdgeToEdge = bridge.getConfig().adjustMarginsForEdgeToEdge();
 
-            MarginLayoutParams mlp = (MarginLayoutParams) v.getLayoutParams();
-            mlp.leftMargin = insets.left;
-            mlp.bottomMargin = insets.bottom;
-            mlp.rightMargin = insets.right;
-            mlp.topMargin = insets.top;
-            v.setLayoutParams(mlp);
+        boolean autoMargins = false;
+        boolean forceMargins = Objects.equals(configEdgeToEdge, "force");
```
jcesarmobile, line 63: Any reason for using `Objects.equals(configEdgeToEdge,` instead of `configEdgeToEdge.equals(`?
In the rest of the codebase we use the later one.

jcesarmobile (changes requested):
Added two minor comments.

But I think setting the default to auto in a minor release is still a breaking change as it will change the original behavior of Capacitor 7.0.0 apps, which while it might not have been the desired behavior for some users, other that might have adjusted the bottom inset by using CSS already.

markemer:
> Added two minor comments.
> 
> But I think setting the default to auto in a minor release is still a breaking change as it will change the original behavior of Capacitor 7.0.0 apps, which while it might not have been the desired behavior for some users, other that might have adjusted the bottom inset by using CSS already.

I'm with you there - seems that we're gonna have to fix a group of people no matter what. Either the people that are currently broken, or the people that fixed it with some hacky-ness. I'm of two minds. I feel weird not default fixing a bug because people might have fixed it a weird other way, but I also feel weird breaking people who thought they were fixed. I think we might want to use disable for 7 just because it is a behavior change, but I don't love it.

android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
markemer: Yeah, moved those inside the if without thinking.

dpogue:
Is there anything here that will break if the WebView does start handling safe-area-insets properly? My guess is that because this is using LinearLayout margins outside the WebView it'll be fine, but wanted to ask

luisbytes:
> > Added two minor comments.
> > But I think setting the default to auto in a minor release is still a breaking change as it will change the original behavior of Capacitor 7.0.0 apps, which while it might not have been the desired behavior for some users, other that might have adjusted the bottom inset by using CSS already.
> 
> I'm with you there - seems that we're gonna have to fix a group of people no matter what. Either the people that are currently broken, or the people that fixed it with some hacky-ness. I'm of two minds. I feel weird not default fixing a bug because people might have fixed it a weird other way, but I also feel weird breaking people who thought they were fixed. I think we might want to use disable for 7 just because it is a behavior change, but I don't love it.

I completely agree with you, that error should not have appeared with capacitor v7, it is an unexpected behavior.

In a project I managed to solve it with CSS using a third-party plugin but I did not release it to the public. I'm still waiting for the fix from the ionic team to update a month after the stable release of v7, this error is still present.

jcesarmobile:
> Is there anything here that will break if the WebView does start handling safe-area-insets properly? My guess is that because this is using LinearLayout margins outside the WebView it'll be fine, but wanted to ask

If the WebView handled the safe-area-insets properly (or users are using a plugin that provides those values to do a similar approach themselves), users could avoid the safe areas using css but still style the safe areas from css, meaning that if they have the header and tab bar with a color, they could also have the safe areas with the same color, styled from css.

With the current implementation where the whole WebView avoids the safe areas, the safe areas are white as there are not views in that area. 
So an app with dark mode and the fix looks like this
image

While it used to look like this:
image

Of course in the last one the tab bar is partially hidden because `safe-area-inset-bottom` doesn't work on Android at the moment, so that's why some users are using plugins that provide the value of the safe areas to avoid them themselves.
And if they are doing that already and we make the `auto` the default value, they will have to set the value to `disabled` to get their previous behavior as they most likely don't want the white areas and I consider that a breaking change.

dpogue:
> Of course in the last one the tab bar is partially hidden because `safe-area-inset-bottom` doesn't work on Android at the moment, so that's why some users are using plugins that provide the value of the safe areas to avoid them themselves.

yeah... I'm asking because I'm trying to get this fixed in Android WebView, and there are concerns about breaking apps that are already working around these problems (weighed against apps that are already broken because of these problems). So I want to make sure that workarounds like the changes in this PR won't cause more trouble if the WebView starts behaving properly.

jcesarmobile:
> > Of course in the last one the tab bar is partially hidden because `safe-area-inset-bottom` doesn't work on Android at the moment, so that's why some users are using plugins that provide the value of the safe areas to avoid them themselves.
> 
> yeah... I'm asking because I'm trying to get this fixed in Android WebView, and there are concerns about breaking apps that are already working around these problems (weighed against apps that are already broken because of these problems). So I want to make sure that workarounds like the changes in this PR won't cause more trouble if the WebView starts behaving properly.

I would assume that people using their own css variables for safe areas would be doing something like `max(--custom-safe-area-inset-bottom, env(safe-area-inset-bottom))` so it works on iOS where `safe-area-inset-bottom` works and (presumably) `--custom-safe-area-inset-bottom` would be 0, so if it get fixed in the WebView, both would have the same value and nothing should change.

I would also assume that your fix would return 0 for the safe areas when the WebView is avoiding the safe areas as when using `force`/`auto` values of this PR. But I'm not sure if the plugins that create the css variables are also handling that possibility and return 0 in that case.

robingenz:
> I think we might want to use disable for 7 just because it is a behavior change, but I don't love it.

Yes, please disable it by default on Capacitor 7, as it would be a breaking change now and change that with Capacitor 8.

cli/src/declarations.ts
```diff
@@ -297,6 +297,17 @@ export interface CapacitorConfig {
      * @default true
      */
     resolveServiceWorkerRequests?: boolean;
+
+    /**
+     * If set to "force", margins will be adjusted for edge to edge regardless of any other settings.
+     * If set to "auto", or is missing, will check for Android 15 and the setting of [windowOptOutEdgeToEdgeEnforcement](https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement) and decide based on those.
```
jcesarmobile, line 303: would be good to explain the actual behavior instead of just saying "decide based on those"

android/capacitor/src/main/java/com/getcapacitor/CapacitorWebView.java
```diff
@@ -54,19 +55,35 @@ public boolean dispatchKeyEvent(KeyEvent event) {
...
-        });
+        boolean autoMargins = false;
+        boolean forceMargins = configEdgeToEdge.equals("force");
+
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+            TypedValue value = new TypedValue();
+            boolean foundOptOut = getContext().getTheme().resolveAttribute(android.R.attr.windowOptOutEdgeToEdgeEnforcement, value, true);
+            boolean optOutValue = value.data != 0; // value is set to -1 on true as of Android 15, so we have to do this.
+
+            autoMargins = configEdgeToEdge.equals("auto") && !(foundOptOut && optOutValue);
```
jcesarmobile, line 71: the `configEdgeToEdge.equals("auto")` could be moved to the if to avoid executing the code if set to force

jcesarmobile (changes requested):

theproducer (approved):

markemer:
> I've noticed that `force` also adjusts margins on Android 14 and probably other older versions, not sure if that's expected/desired. It's not clear by the definition since it only talks about "settings" and not about Android versions, but I personally like that it does it as we have had complaints in the past about devices with notches or similar not behaving correctly (it was related to status bar plugin, but good to have this option in the core)

Thanks! Yeah, I though "force" should do what it implies. All the time.

ckoon-infopro:
@markemer How can I get this working? I have installed Capacitor 7.1.0 and tried setting adjustMarginsForEdgeToEdge to 'auto' and even 'force', but it doesn’t do anything. The margins are not applied.

luisbytes:
I should also add the corresponding documentation, and I think it would be good to also add it to the [upgrade guide](https://capacitorjs.com/docs/updating/7-0) since this fixes unexpected behavior in Capacitor v7

jcesarmobile:
The attribute already does its own handling by itself, so the plugin will do nothing if the attribute is true to not conflict, unless you use "force", in that case the plugin will do its own handling, which might cause issues since it would be handled in two different ways.

Issue #7880, opened 2025-02-10
[Bug]: Android Build Tools 8.8+ Gradle build error


AntiGuideAkquinet:
### Capacitor Version

   Capacitor Doctor   

Latest Dependencies:

  @capacitor/cli: 7.0.1
  @capacitor/core: 7.0.1
  @capacitor/android: 7.0.1
  @capacitor/ios: 7.0.1

Installed Dependencies:

  @capacitor/cli: 6.0.0
  @capacitor/android: 6.0.0
  @capacitor/ios: 6.0.0
  @capacitor/core: 6.0.0

[success] Android looking great! 👌
[error] Xcode is not installed

### Other API Details

```Shell
npm --version output: 10.9.0
node --version output: v22.12.0
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

When building an Android project with Android Build tools 8.8+ the following error is produced. 8.8 is not the default version but will become an issue with future upgrades.

```
[Redacted]/@capacitor/android/capacitor/src/main/java/com/getcapacitor/FileUtils.java:222: Error: Using name is unsafe as it is a filename obtained directly from a ContentProvider. You should sanitize it before using it for creating a File. [UnsanitizedFilenameFromContentProvider]
        File file = new File(context.getFilesDir(), name);
                                                    ~~~~

   Explanation for issues of type "UnsanitizedFilenameFromContentProvider":
     When communicating between applications with files, the server app can
   provide the client app with a maliciously constructed filename. The client
   app should never trust this filename and should either sanitize it or
   completely discard it.

   https://developer.android.com/privacy-and-security/risks/untrustworthy-contentprovider-provided-filename
```

### Expected Behavior

I would expect the project to build without errors.

### Project Reproduction

https://nextcloud.ristec.de/s/TrzNQNiqQzY4bqH

SuperBallista:
"I am also experiencing this issue on Capacitor 7.0.1 with Java 21 and Gradle 8.5+. The suggested fix of sanitizing the filename (safeName) works well."

jcesarmobile:
I can't reproduce, looks like a build setting that you might have configured on Android Studio as global setting or some lint rule.
What's your Android Studio version?
How are you building the app?

PR #7894, opened 2025-02-19
fix(android): sanitize portable file name

shipley-dcc:
Fix linting error for Android. When building a plugin I get the following error:
```
Execution failed for task ':capacitor-android:lintDebug'.

/node_modules/@capacitor/android/capacitor/src/main/java/com/getcapacitor/FileUtils.java:222: Error: Using name is unsafe as it is a filename obtained directly from a ContentProvider. You should sanitize it before using it for creating a File. [UnsanitizedFilenameFromContentProvider]
          File file = new File(context.getFilesDir(), name);
```

I've applied the fix recommended here https://developer.android.com/privacy-and-security/risks/untrustworthy-contentprovider-provided-filename.

closes https://github.com/ionic-team/capacitor/issues/7880
closes https://outsystemsrd.atlassian.net/browse/RDMR-556

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/FileUtils.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/FileUtils.java
@@ -219,7 +219,8 @@ public class FileUtils {
         int nameIndex = cursor.getColumnIndex(OpenableColumns.DISPLAY_NAME);
         cursor.moveToFirst();
         String name = (cursor.getString(nameIndex));
-        File file = new File(context.getFilesDir(), name);
+        String fileName = sanitizeFilename(name);
+        File file = new File(context.getFilesDir(), fileName);
         try {
             InputStream inputStream = context.getContentResolver().openInputStream(uri);
             FileOutputStream outputStream = new FileOutputStream(file);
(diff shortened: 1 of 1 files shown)


android/capacitor/src/main/java/com/getcapacitor/FileUtils.java
```diff
@@ -289,4 +290,14 @@ private static String getPathToNonPrimaryVolume(Context context, String tag) {
         }
         return null;
     }
+
+    private String sanitizeFilename(String displayName) {
```
jcesarmobile, line 294:
```suggestion
    private static String sanitizeFilename(String displayName) {
```
method needs to be `static` since it's called from `getCopyFilePath`, which is `static`

shipley-dcc: Good spot. Thanks. I've made that change.

jcesarmobile (approved):
Thanks for the PR!

I was already working on it and was testing the changes before sending the PR, but you were faster.

markemer (approved):

PR #7896, opened 2025-02-27
fix(cli): don't run bundle if not installed

jcesarmobile:


markemer (approved):
Kind of a corner case, but we should for sure check. Surprised we missed this.

theproducer (approved):

Issue #7589, opened 2024-07-29
[Bug]: http request break on Android when sending formData


jongbonga:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.1.1
  @capacitor/core: 6.1.1
  @capacitor/android: 6.1.1
  @capacitor/ios: 6.1.1

Installed Dependencies:

  @capacitor/cli: 6.1.1
  @capacitor/core: 6.1.1
  @capacitor/ios: 6.1.1
  @capacitor/android: 6.1.1

```

### Other API Details

```Shell
npm -v 10.7.0
node -v v18.20.4
```

### Platforms Affected

- [ ] iOS
- [X] Android
- [ ] Web

### Current Behavior

on Capacitor 5 sending formData with CapacitorHttp enabled on config and `content-type: 'multipart/form-data'` hard-coded in the headers works fine, it fails on Capacitor 6

### Expected Behavior

it should work with `content-type: 'multipart/form-data'` or documented somewhere

### Project Reproduction

https://github.com/jongbonga/code-reproduction

### Additional Information

There is a PR for the issue [here](https://github.com/ionic-team/capacitor/pull/7518)

grzegorzCieslik95:
The same:

5.5.1 -> my express service got content type `content-type’: ‘multipart/form-data` as default
6.0.1 -> i got `content-type': 'application/x-www-form-urlencoded` without file in data

Any idea how can i fix this?

jcesarmobile:
please, report a new issue, that seems like a different problem

PR #7897, opened 2025-03-03
fix(http): boundary not added for Request objects

jcesarmobile:
If the user provides the `multipart/form-data` content-type without boundary, the call to `const request = new Request(resource, options);` is not able to generate a browser boundary. 
So remove the content-type from the headers when they are set to `multipart/form-data` and if they don't have a boundary.
Also do it only for `FormData` bodies as other types or bodies are handled differently and I'm not sure if the issue is also present there.

closes https://github.com/ionic-team/capacitor/issues/7589


theproducer (approved):

ItsChaceD (approved):

grzegorzCieslik95:
When it will be aviable in capacitor npm package?

PR #7905, opened 2025-03-05
[ios] Reset plugin listeners when WebView process is terminated

gabeschine:


markemer:
This is a nice find.

alextreppass:
Nice find — this might explain some issues we were finding with long-backgrounded apps and Push Notification taps.

Will this be back-ported to Capacitor 6.X? Or will it only be fixed in 7.X? — what's the best way for us to follow along, is it checking the release notes regularly?

Thanks

gabeschine:
> Nice find — this might explain some issues we were finding with long-backgrounded apps and Push Notification taps.
> 
> Q) Will this be back-ported to Capacitor 6.X? Or will it only be fixed in 7.X? — what's the best way for us to follow along, is it checking the release notes regularly?
> 
> Thanks

@alextreppass in the meantime, this fix is trivially easy to patch into an existing project using the `patch-package` npm package. I've attached the contents of our `patches/@capacitor+ios+6.1.2.patch` file for your convenience. Update the version to whatever you have and it should patch cleanly.

[@capacitor+ios+6.1.2.patch](https://github.com/user-attachments/files/19267806/%40capacitor%2Bios%2B6.1.2.patch)

vladansaracpv:
Can this help in any way with the issue where after the app is resumed from being backgrounded for a long time - storage/indexeddb doesn't work correctly? Or it's not related

PR #7906, opened 2025-03-05
feat(http):  Apply overrideUserAgent to requests

theproducer:
Take the Cap Config values of `overrideUserAgent` (or `ios.overrideUserAgent` | `android.overrideUserAgent`) and apply them to CapacitorHttp requests.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
@@ -398,6 +398,11 @@ public class HttpRequestHandler {
 
         boolean isHttpMutate = method.equals("DELETE") || method.equals("PATCH") || method.equals("POST") || method.equals("PUT");
 
+        String overriddenUserAgent = bridge.getConfig().getOverriddenUserAgentString();
+        if (overriddenUserAgent != null) {
+            headers.put("User-Agent", overriddenUserAgent);
+        }
+
         URL url = new URL(urlString);
         HttpURLConnectionBuilder connectionBuilder = new HttpURLConnectionBuilder()
             .setUrl(url)
--- a/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
+++ b/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
@@ -165,7 +165,7 @@ open class HttpRequestHandler {
         guard var urlString = call.getString("url") else { throw URLError(.badURL) }
         let method = httpMethod ?? call.getString("method", "GET")
 
-        let headers = (call.getObject("headers") ?? [:]) as [String: Any]
+        var headers = (call.getObject("headers") ?? [:]) as [String: Any]
         let params = (call.getObject("params") ?? [:]) as [String: Any]
         let responseType = call.getString("responseType") ?? "text"
         let connectTimeout = call.getDouble("connectTimeout")
@@ -183,6 +183,10 @@ open class HttpRequestHandler {
             .setUrlParams(params)
             .openConnection()
             .build()
+        
+        if let userAgentString = config?.overridenUserAgentString {
+            headers["User-Agent"] = userAgentString
+        }
 
         request.setRequestHeaders(headers)
 


markemer (approved):

markemer (approved):
Re-upping my approval here. That's a heck of an android bug, but looks like it works.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -260,6 +260,18 @@ private WebResourceResponse handleCapacitorHttpRequest(WebResourceRequest reques
...
+        // https://issues.chromium.org/issues/40450316
+        // x-cap-user-agent contains the user agent set in JavaScript
+        if (headers.has("x-cap-user-agent")) {
+            String userAgentValue = headers.getString("x-cap-user-agent");
+            if (userAgentValue != null && !userAgentValue.isEmpty()) {
+                headers.put("User-Agent", userAgentValue);
+            }
+
+            headers.remove("x-cap-user-agent");
+        }
```
jcesarmobile, line 273: You can skip the `has` check since we still have to check for `getString` being null.
```suggestion
        String userAgentValue = headers.getString("x-cap-user-agent");
        if (userAgentValue != null && !userAgentValue.isEmpty()) {
            headers.put("User-Agent", userAgentValue);
            headers.remove("x-cap-user-agent");
        }
```

android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
```diff
@@ -398,6 +398,23 @@ public static JSObject request(PluginCall call, String httpMethod, Bridge bridge
...
+        // https://issues.chromium.org/issues/40450316
+        // x-cap-user-agent contains the user agent set in JavaScript
+        if (headers.has("x-cap-user-agent")) {
+            String userAgentValue = headers.getString("x-cap-user-agent");
+            if (userAgentValue != null) {
+                headers.put("User-Agent", userAgentValue);
+            }
+
+            headers.remove("x-cap-user-agent");
+        }
```
jcesarmobile, line 411: Same as the previous comment, also, the previous one was checking for `isEmpty()` too, so we should probably do the same checks
```suggestion
       String userAgentValue = headers.getString("x-cap-user-agent");
        if (userAgentValue != null && !userAgentValue.isEmpty()) {
            headers.put("User-Agent", userAgentValue);
            headers.remove("x-cap-user-agent");
        }
```

core/native-bridge.ts
```diff
@@ -513,6 +513,22 @@ const initBridge = (w: any): void => {
...
+              if (headers.has('User-Agent')) {
+                headers.set('X-Cap-User-Agent', headers.get('User-Agent') as string);
+              }
+
+              if (headers.has('user-agent')) {
+                headers.set('X-Cap-User-Agent', headers.get('user-agent') as string);
+              }
+
+              options.headers = headers;
+            }
```
jcesarmobile, line 530: This can be simplified a little. Also `x-cap-user-agent` should be lowercased, despite is being lowercased automatically, but for consistency.
```suggestion
            if (platform === 'android' && options?.headers) {
              const userAgent = headers.get('User-Agent') || headers.get('user-agent');
              if (userAgent) {
                headers.set('x-cap-user-agent', userAgent);
                options.headers = headers;
              }
            }
```

core/native-bridge.ts
```diff
@@ -530,21 +546,33 @@ const initBridge = (w: any): void => {
...
             );
 
+            const nativeHeaders = {
+              ...requestHeaders,
+              ...optionHeaders,
+            };
+
+            if (platform === 'android') {
+              if (headers.has('User-Agent')) {
+                nativeHeaders['User-Agent'] = headers.get('User-Agent') as string;
```
jcesarmobile, line 562:
```suggestion
                nativeHeaders['User-Agent'] = headers.get('User-Agent');
```
I don't think the cast is needed

core/native-bridge.ts
```diff
@@ -530,21 +546,33 @@ const initBridge = (w: any): void => {
...
+              }
+
+              if (headers.has('user-agent')) {
+                nativeHeaders['user-agent'] = headers.get('user-agent') as string;
```
jcesarmobile, line 566:
```suggestion
                nativeHeaders['user-agent'] = headers.get('user-agent');
```
same

jcesarmobile (changes requested):
Added some minor comments

core/native-bridge.ts
```diff
@@ -513,6 +513,18 @@ const initBridge = (w: any): void => {
             method.toLocaleUpperCase() === 'OPTIONS' ||
             method.toLocaleUpperCase() === 'TRACE'
           ) {
+            // a workaround for following android webview issue:
+            // https://issues.chromium.org/issues/40450316
+            // Sets the user-agent header to a custom value so that its not stripped
+            // on its way to the native layer
+            if (platform === 'android' && options?.headers) {
+              const userAgent = headers.get('User-Agent') || headers.get('user-agent');
+              if (userAgent) {
```
jcesarmobile, line 522:
```suggestion
              if (userAgent !== null) {
```
This suggestion I made was not correct, it has to check for null because it's false for empty strings, and we should allow empty strings since iOS allow them.

The java code should also remove the two `!userAgentValue.isEmpty()` checks as we should allow to use empty strings as iOS does.

android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
```diff
@@ -398,6 +398,19 @@ public static JSObject request(PluginCall call, String httpMethod, Bridge bridge
...
+        // a workaround for the following android web view issue:
+        // https://issues.chromium.org/issues/40450316
+        // x-cap-user-agent contains the user agent set in JavaScript
+        String userAgentValue = headers.getString("x-cap-user-agent");
+        if (userAgentValue != null && !userAgentValue.isEmpty()) {
+            headers.put("User-Agent", userAgentValue);
+        }
+        headers.remove("x-cap-user-agent");
+
+        if (overriddenUserAgent != null && !headers.has("User-Agent") && !headers.has("user-agent")) {
```
jcesarmobile, line 410:
```suggestion
        if (!headers.has("User-Agent") && !headers.has("user-agent")) {
```
Not sure if you didn't understand what I meant, I said that in the first commit you were doing a `overriddenUserAgent != null` check and you removed it.
One option was to put the check back, and the other option was to remove the variable and just put the `bridge.getConfig().getOverriddenUserAgentString()` in the put directly since it didn't really matter if it was null.
Looks like you did both options, but since you removed the `overriddenUserAgent` variable, now the code doesn't compile because you added the null check for it.

android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
```diff
@@ -260,6 +260,15 @@ private WebResourceResponse handleCapacitorHttpRequest(WebResourceRequest reques
             headers.put(header.getKey(), header.getValue());
         }
 
+        // a workaround for the following android web view issue:
+        // https://issues.chromium.org/issues/40450316
+        // x-cap-user-agent contains the user agent set in JavaScript
+        String userAgentValue = headers.getString("x-cap-user-agent");
+        if (userAgentValue != null && !userAgentValue.isEmpty()) {
```
jcesarmobile, line 267:
```suggestion
        if (userAgentValue != null) {
```
I noticed iOS allows empty strings, so we should also allow them for Android

android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
```diff
@@ -398,6 +398,19 @@ public static JSObject request(PluginCall call, String httpMethod, Bridge bridge
 
         boolean isHttpMutate = method.equals("DELETE") || method.equals("PATCH") || method.equals("POST") || method.equals("PUT");
 
+        // a workaround for the following android web view issue:
+        // https://issues.chromium.org/issues/40450316
+        // x-cap-user-agent contains the user agent set in JavaScript
+        String userAgentValue = headers.getString("x-cap-user-agent");
+        if (userAgentValue != null && !userAgentValue.isEmpty()) {
```
jcesarmobile, line 405:
```suggestion
        if (userAgentValue != null) {
```
I noticed iOS allows empty strings, so we should also allow them for Android

jcesarmobile (changes requested):

android/capacitor/src/main/java/com/getcapacitor/plugin/util/HttpRequestHandler.java
theproducer: Yup, sorry, completely missed removing that first part of the statement.

jcesarmobile (approved):

PR #7919, opened 2025-03-14
fix(android): rename bridge layout to avoid collision

jcesarmobile:
kept the old one in case somebody was using it somewhere else, should be removed in Capacitor 8.


markemer (approved):

theproducer (approved):

arjunaj5:
@jcesarmobile  tnx for fixing this collision issue!!. Any eta of when this will be included in a release?

PR #7947, opened 2025-03-31
fix(cli): ensures the execution order of CLI hooks

luissilvaos:
This PR fixes the execution order of CLI Hooks.

The `forEach()` method of Array instances expects synchronous functions and it does not wait for promises.


Chuckytuh (approved):
lgtm

Issue #7810, opened 2024-12-18
[Bug]: [ios] Some Native -> JS messages appear to drop after extended time with app in background


gabeschine:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.2.0
  @capacitor/core: 6.2.0
  @capacitor/android: 6.2.0
  @capacitor/ios: 6.2.0

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/core: 6.2.0
  @capacitor/ios: 6.2.0
  @capacitor/cli: 6.2.0

[success] iOS looking great! 👌

### Other API Details

```Shell
yarn --version: 1.22.11
node --version: v20.10.0
```

### Platforms Affected

- [X] iOS
- [ ] Android
- [ ] Web

### Current Behavior

If a Capacitor app is backgrounded for a long time, and during this time a notification is delivered to the device and clicked or a deeplink open request is received, when the app is foregrounded the relevant message (either the notification payload from a plugin like `@capacitor/push-notifications` or the `appUrlOpen` event from `@capacitor/app`) is sometimes dropped.

In the field when this occurs, we observe the app's web view reload immediately upon foregrounding.

In testing, it is possible to reproduce the behavior, and it appears to be tied to the timing of iOS notifying the native app through `webViewWebContentProcessDidTerminate` [1]: if the notification happens immediately, messages are delivered successfully; if the notification is delayed until the app is foregrounded, the messages are not delivered.

[1] https://developer.apple.com/documentation/webkit/wknavigationdelegate/webviewwebcontentprocessdidterminate(_:)

### Expected Behavior

The messages are delivered in all scenarios.

### Project Reproduction

https://github.com/watch-duty/capacitor-message-bug-repro

### Additional Information

The reproduction is not a _perfect_ mirror of what happens in production.

In production: it appears that iOS is killing the `WebContent` process on its own. Based on reading this can happen to reduce memory pressure or other resource pressure, or based on other conditions unknown to me.

In the reproduction: killing the `WebContent` process is performed explicitly, since it is impossible to deterministically trigger the same resource pressure conditions encountered in production.

The reproduction is assuming that while the trigger is different, the subsequent symptoms and bug mechanism are shared. Based on many, many user reports of the bug, and associated data gather that we have performed, this appears to be a sound assumption.

davidemerritt:
+1 here - this reflects what I've seen in our application as well, affecting push notification handling.

[...]

PR #7962, opened 2025-04-07
fix: Remove all plugin listeners in bridge reset

jcesarmobile:
This PR makes `bridge.reset()` to also remove all plugin listeners and not just stored plugin calls.

Removed the `removeAllPluginListeners` call from `webViewWebContentProcessDidTerminate` since `webViewWebContentProcessDidTerminate` produces a navigation and starting a navigation calls `bridge.reset()`.


markemer (approved):

ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
```diff
@@ -159,9 +159,6 @@ open class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDelegat
 
     open func webViewWebContentProcessDidTerminate(_ webView: WKWebView) {
         CAPLog.print("⚡️  WebView process terminated")
-        if let bridge = bridge {
```
gabeschine, line 162: Somehow, in release 7.4.2 the bug (https://github.com/ionic-team/capacitor/issues/7810) that was fixed by https://github.com/ionic-team/capacitor/pull/7905 is back for us.

I haven't dug into it very deeply, but my suspicion is that under the circumstances originally identified, where iOS kills `WebContent`, the new code is not in fact cleaning up old listeners any more. This line was directly causing that to happen, at least.

gabeschine: @markemer bumping for your visibility.

I looked a little more at this. Is it possible that `webView(_:didStartProvisionalNavigation:)` races with the web content to clear out listeners before the web content starts processing? It's important that all the plugin listener state is cleared before any javascript in the web content starts to run.

PR #7963, opened 2025-04-07
feat(cli): add spm-migration-assistant command

markemer:


cli/src/util/spm.ts
```diff
@@ -26,17 +34,69 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
   return resolve(packageDirectory, 'Package.swift');
 }
 
-export async function generatePackageFile(config: Config, plugins: Plugin[]): Promise<void> {
+export async function generatePackageSwiftFile(config: Config, plugins: Plugin[]): Promise<void> {
```
jcesarmobile, line 37: exported functions can be used from javascript, so we can't rename it as it would be breaking if anybody is using them.

cli/src/util/spm.ts
```diff
@@ -82,3 +142,62 @@ let package = Package(
 
   return packageSwiftText;
 }
+
+export async function runCocoapodsDeintegrate(config: Config, options: MigrateSPMInteractiveOptions): Promise<void> {
+  const podPath = await config.ios.podPath;
```
jcesarmobile, line 147: the function that configures the `podPath` already does all the checks for knowing if the user is using bundler or not, you shouldn't need to repeat all the checks

jcesarmobile (changes requested):
Added some comments.

Also: 
- `ios-spm-template` already has the same files you created in the  `ios-spm-migrate-template`, can't you use `ios-spm-template` files instead of a new ones?

- Do we need the `--unsafe` and `--dry-run` options? for `--unsafe` I assume people uses git or some other source control, so they could revert if they had some problem. And for `--dry-run` we are documenting the changes the command does so people can know.

- We have a separate ticket for the messaging/handling of non SPM plugins, so you can leave those changes out of the migration command.

jcesarmobile:
Another thought, since it doesn't do a full migration, maybe it should be renamed to `assist-spm-migration`?

And at the end it should show a link [to the SPM docs](https://capacitorjs.com/docs/ios/spm), to the `### Using our migration tool` section once it's live.

cli/src/util/spm.ts
markemer: Oh it does? I think this might be a copy and paste from another place that doesn't need it then. But that's an issue for another time.

markemer:
> Added some comments.
> 
> Also:
> 
> * `ios-spm-template` already has the same files you created in the  `ios-spm-migrate-template`, can't you use `ios-spm-template` files instead of a new ones?
> * Do we need the `--unsafe` and `--dry-run` options? for `--unsafe` I assume people uses git or some other source control, so they could revert if they had some problem. And for `--dry-run` we are documenting the changes the command does so people can know.
> * We have a separate ticket for the messaging/handling of non SPM plugins, so you can leave those changes out of the migration command.

- I can look into the best way to use the template files - it might be trivial to do that.
- I'm gonna ditch `--unsafe` but keep `--dry-run`, I think
- Non SPM plugins will just throw a warning for now - but I don't want to give people the idea it worked when it didn't.

markemer:
> > Added some comments.
> > Also:
> > 
> > * `ios-spm-template` already has the same files you created in the  `ios-spm-migrate-template`, can't you use `ios-spm-template` files instead of a new ones?
> > * Do we need the `--unsafe` and `--dry-run` options? for `--unsafe` I assume people uses git or some other source control, so they could revert if they had some problem. And for `--dry-run` we are documenting the changes the command does so people can know.
> > * We have a separate ticket for the messaging/handling of non SPM plugins, so you can leave those changes out of the migration command.
> 
> * I can look into the best way to use the template files - it might be trivial to do that.
> * I'm gonna ditch `--unsafe` but keep `--dry-run`, I think
> * Non SPM plugins will just throw a warning for now - but I don't want to give people the idea it worked when it didn't.

So the npm `tar` package seems not to respect the file list you give it. I even enumerated every file, so I think the separate tarball works better - although we could at some point generate it directly from the template directory.

theproducer (approved):

jcesarmobile:
> So the npm `tar` package seems not to respect the file list you give it. I even enumerated every file, so I think the separate tarball works better - although we could at some point generate it directly from the template directory.

I don't understand what you mean in the comment, but can't you extract it to a temp folder and copy the needed files to the user project?
I really don't want a separate template with duplicate files we have to maintain

markemer:
> > So the npm `tar` package seems not to respect the file list you give it. I even enumerated every file, so I think the separate tarball works better - although we could at some point generate it directly from the template directory.
> 
> I don't understand what you mean in the comment, but can't you extract it to a temp folder and copy the needed files to the user project? I really don't want a separate template with duplicate files we have to maintain

Yeah, `/tmp` might be the way - `tar` should allow me to just untar the stuff I want - `extract` accepts a file list, but it seems to totally ignore it.

jcesarmobile (changes requested):
There are several comments that have not been addressed yet

markemer:
> There are several comments that have not been addressed yet

I just pushed the removal of Cap-SPM as a separate zip file, should be ready now.

cli/src/util/spm.ts
markemer: This had to stay as is, as the exec screws up otherwise.

jcesarmobile (changes requested):
The changes from https://github.com/ionic-team/capacitor/pull/7982/files need to be applied on the migration

jcesarmobile (changes requested):
The `debug.xcconfig` is not being added to the project (https://github.com/ionic-team/capacitor/blob/main/ios-spm-template/debug.xcconfig) and I think it should, or at least should be copied to the project folder and tell the users to manually add it to the Xcode project or explain how to create a new one, both in here and in the docs PR. But adding it or creating it and also adding the `CapApp-SPM` dependency seem like too much manual changes for a "migrate" command, so would be cool if the command could add it.

The `Package.swift` detection is not working properly, I get
> Found 6 Plugins with Package.swift files:
> @capacitor-community/tap-jacking@7.0.0

and then
> [warn] CapacitorCommunityTapJacking does not have a Package.swift

The plugin doesn't have a Package.swift, so the warning is correct, but shouldn't be listed as having it.
Anyway, I told you to remove that part from this PR since we have another ticket for that task and there are some other things that should be changed related to it, so better focus just on the migration task.

Also, can you remove the comments about you hating things?

markemer:
> The `Package.swift` detection is not working properly, I get
> 
> > Found 6 Plugins with Package.swift files:
> > @capacitor-community/tap-jacking@7.0.0
> 
> and then
> 
> > [warn] CapacitorCommunityTapJacking does not have a Package.swift

Yeah, that's a new one on me - I'll take a look at why that's failing. The rest are no problem.

cli/src/util/spm.ts
```diff
@@ -1,16 +1,27 @@
-import { existsSync, writeFileSync } from 'fs-extra';
-import { relative, resolve } from 'path';
+import { LOGGER_LEVELS } from '@ionic/cli-framework-output'; // Ugh, I hate this, lets yank it
```
jcesarmobile, line 1:
```suggestion
import { LOGGER_LEVELS } from '@ionic/cli-framework-output';
```

cli/src/tasks/migrate-spm.ts
```diff
@@ -0,0 +1,36 @@
...
+
+  if ((await checkPackageManager(config)) == 'SPM') {
+    fatal('Capacitor project is already using SPM, exiting.');
+  }
+
+  await extractSPMPackageDirectory(config, options);
+  await runCocoapodsDeintegrate(config, options);
+  await removeCocoapodsFiles(config, options);
+  await addInfoPlistDebugIfNeeded(config, options);
+  // TODO: Add CAP-SPM package to project dependencies
```
jcesarmobile, line 26:
```suggestion
```

I don't think this is going to be done, so better not add a TODO

cli/src/util/spm.ts
```diff
@@ -29,14 +40,65 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
+export async function processIosPackages(config: Config): Promise<Plugin[]> {
+  const plugins = await getPlugins(config, 'ios');
+  const iosPlugins = await getIOSPlugins(plugins);
+  printPlugins(iosPlugins, 'ios', 'capacitor');
+
+  const packageSwiftPluginList = await pluginsWithPackageSwift(iosPlugins);
+  printPlugins(packageSwiftPluginList, 'ios', 'packagespm');
+
+  if (iosPlugins.length == packageSwiftPluginList.length) {
+    logger.info('Number of plugins in lists match'); // TODO: Word this better
```
jcesarmobile, line 61: is this something you are going to address now?

jcesarmobile (commented):
Some messages use `logger.info` and other use `logOptSuffix` for the actions taken, they should all use the same.
I don't really see the usefulness of the `--dry-run`, but if you want to add it, it should be consistent.

I have mentioned in previous reviews that the handling of the plugins shouldn't be part of this PR but the code is still here.
The handling of the plugins should be done in the `update` command, not in the `spm-migration-assistant` because this is only ran once, so running `update` later on will bring back the plugins that were removed by this command. If you want to list the incompatible plugins in this command too you can programmatically call `update` at the end of the migration process. But the handling of the `Package.swif` and the incompatible plugins should be done in the `update` and not in this command. 
It's listing the plugins twice, it should do the listing only once but warning about the plugins that are not compatible.

Also we discussed in slack about the possibility of patching incompatible plugins instead of removing them, did you take a look at that? would be cool to include it.

jcesarmobile:
The xcconfig file is being copied to `ios/App/debug.xcconfig`, but on the spm template it's located at `ios/debug.xcconfig`, so should be adjusted to match.

cli/src/util/spm.ts
markemer: Missed one, thanks.

cli/src/util/spm.ts
markemer: Yeah - let me fix that.

markemer:
> Some messages use `logger.info` and other use `logOptSuffix` for the actions taken, they should all use the same. I don't really see the usefulness of the `--dry-run`, but if you want to add it, it should be consistent.

Maybe I should ditch dry-run, it's seeming less and less useful.

> 
> I have mentioned in previous reviews that the handling of the plugins shouldn't be part of this PR but the code is still here. The handling of the plugins should be done in the `update` command, not in the `spm-migration-assistant` because this is only ran once, so running `update` later on will bring back the plugins that were removed by this command. If you want to list the incompatible plugins in this command too you can programmatically call `update` at the end of the migration process. But the handling of the `Package.swif` and the incompatible plugins should be done in the `update` and not in this command. It's listing the plugins twice, it should do the listing only once but warning about the plugins that are not compatible.
> 

Ah, ok, I was misunderstanding. Yeah - that logic should be moved and we can just run an update.


> Also we discussed in slack about the possibility of patching incompatible plugins instead of removing them, did you take a look at that? would be cool to include it.

This was something we brought up a while back and thought about not adding, but I was always pro the idea of adding it.

As for the TODO I ditched - maybe we should try and tackle all the manual steps. We'd have to include either `cordova-node-xcode` or something else, but it might be worth it.

jcesarmobile (changes requested):
Why did you change the ios-spm-template?
In my comment I meant to make your code match the template, not to make the template match your code.
The template changes have been released in 7.3.0, so we shouldn't change it again in 7.4.0 or we will have users with different templates and if we can get rid of the `debug.xcconfig` file in the future we will have to document all the different locations, so better keep only one.

markemer:
> Why did you change the ios-spm-template? In my comment I meant to make your code match the template, not to make the template match your code. The template changes have been released in 7.3.0, so we shouldn't change it again in 7.4.0 or we will have users with different templates and if we can get rid of the `debug.xcconfig` file in the future we will have to document all the different locations, so better keep only one.

I actually changed both to fit the typical iOS pattern, but since we released 7.3.0 with it as is, I'll roll my changes back and just change the migration.

cli/src/util/spm.ts
```diff
@@ -87,3 +140,67 @@ let package = Package(
...
+    logger.warn(infoPlist + ' not found.');
+  }
+}
+
+// Private Functions
+
+async function pluginsWithPackageSwift(plugins: Plugin[]): Promise<Plugin[]> {
+  const pluginList: Plugin[] = [];
+  for (const plugin of plugins) {
+    const packageSwiftFound = await pathExists(plugin.rootPath + '/Package.swift');
```
jcesarmobile:
```suggestion
    const packageSwiftFound = await pathExists(join(plugin.rootPath, 'Package.swift'));
```
Use join to join paths

cli/src/util/spm.ts
```diff
@@ -87,3 +140,67 @@ let package = Package(
...
+    if (packageSwiftFound) {
+      pluginList.push(plugin);
+    } else {
+      logger.warn(plugin.name + ' does not have a Package.swift');
```
jcesarmobile:
```suggestion
      logger.warn(plugin.id + ' does not have a Package.swift');
```
use `plugin.id` since that's the npm name, `plugin.name` is the "CocoaPods" name.

Also, we usually use template literals rather than string concatenation, i.e. ``${plugin.id} does not have a Package.swift``. This is not only for this message but for all other concatenations added in this PR.

cli/src/plugin.ts
```diff
@@ -118,7 +118,7 @@ export function fixName(name: string): string {
 export function printPlugins(
   plugins: Plugin[],
   platform: string,
-  type: 'capacitor' | 'cordova' | 'incompatible' = 'capacitor',
+  type: 'capacitor' | 'cordova' | 'incompatible' | 'packagespm' = 'capacitor',
```
jcesarmobile:
```suggestion
  type: 'capacitor' | 'cordova' | 'incompatible' = 'capacitor',
```
This is no longer being used

cli/src/plugin.ts
```diff
@@ -139,6 +139,8 @@ export function printPlugins(
     case 'capacitor':
       msg = `Found ${plugins.length} Capacitor plugin${plural} for ${c.strong(platform)}:\n`;
       break;
+    case 'packagespm':
+      msg = `Found ${plugins.length} Plugin${plural} with ${c.strong('Package.swift')} files:\n`;
```
jcesarmobile:
```suggestion
```
This is no longer being used

cli/src/util/spm.ts
```diff
@@ -87,3 +140,67 @@ let package = Package(
...
+    }
+  } else {
+    logger.warn('Skipping pod deintegrate because CocoaPods is not installed - migration will be incomplete');
+  }
+}
+
+export async function addInfoPlistDebugIfNeeded(config: Config): Promise<void> {
+  type Mutable<T> = { -readonly [P in keyof T]: T[P] };
+
+  const infoPlist = resolve(config.ios.nativeProjectDirAbs, 'App/Info.plist');
```
jcesarmobile:
```suggestion
  const infoPlist = resolve(config.ios.nativeTargetDirAbs, 'Info.plist');
```

cli/src/util/spm.ts
```diff
@@ -30,14 +36,61 @@ export async function findPackageSwiftFile(config: Config): Promise<string> {
...
+export async function extractSPMPackageDirectory(config: Config): Promise<void> {
+  const spmDirectory = join(config.ios.nativeProjectDirAbs, 'CapApp-SPM');
+  const spmTemplate = join(config.cli.assetsDirAbs, 'ios-spm-template.tar.gz');
+  const debugConfig = join(config.ios.platformDirAbs, 'debug.xcconfig');
+
+  logger.info('Extracting ' + spmTemplate + ' to ' + spmDirectory);
+
+  try {
+    const tempCapDir = await mkdtemp(join(tmpdir(), 'cap-'));
+    const tempCapSPM = join(tempCapDir, 'App/CapApp-SPM');
```
jcesarmobile:
```suggestion
    const tempCapSPM = join(tempCapDir, 'App', 'CapApp-SPM');
```

PR #7973, opened 2025-04-11
chore(ios): Remove xcframework build script

jcesarmobile:
The creation of the xcframework is generated in another repository, so this script is not really being used


markemer (approved):
I'd really like to move building back to this repo, but that's a ways off, and would need swift only. So yeah, lets kill this.

jcesarmobile:
If we move to swift only we don't need to build at all

PR #7977, opened 2025-04-14
refactor(ios): make local plugins implement CAPBridgedPlugin protocol

jcesarmobile:
Make local plugins implement CAPBridgedPlugin protocol to declare their methods from swift so can get rid of `DefaultPlugins.m`


markemer (approved):
Love it!

theproducer (approved):

PR #7987, opened 2025-04-23
The app is crashed, if any low memory issue occured

merbin2012:
If any crash occurred due to a low memory issue, we just prevent the app from crashing and reload the webview.

To test this issue, please execute the following code
```
<script>
function crashWebView() {
    try {
        let a = [];
        while (true) {
            a.push(new Array(1000000).fill('*'));
        }
    } catch (e) {
        console.log('Crashed!', e);
    }
}
crashWebView();
</script>
```


eric-horodyski:
Thanks for opening this PR 🙏

PR #7990, opened 2025-04-25
chore: set skipLibCheck to true in tsconfig.json

jcesarmobile:
To prevent build failures when there are type problems on `node_modules`.


markemer (approved):
That's what I'm talking about. Nice work.

PR #7994, opened 2025-04-29
fix: add error code checking to `runPlatformHook`

markemer:


benmccarty91 (approved):
Is this really all it takes?

markemer:
> Is this really all it takes?

I think so? I'm still very suspicious of that `allSerial<T>` function - but that seems to be not what's failing here.

Chuckytuh (approved):

PR #7999, opened 2025-05-06
feat(cli): Initial Cordova plugin SPM support

jcesarmobile:
Initial Cordova plugin SPM support
It creates a `Package.swift` per Cordova plugin and links it in the `CapApp-SPM` project.

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -2,7 +2,7 @@ import { copy, remove, pathExists, readFile, realpath, writeFile } from 'fs-extr
 import { basename, dirname, join, relative } from 'path';
 
 import c from '../colors';
-import { checkPlatformVersions, runTask } from '../common';
+import { checkPlatformVersions, getCapacitorPackageVersion, runTask } from '../common';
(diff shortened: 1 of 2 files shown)


cli/src/ios/update.ts
```diff
@@ -56,13 +51,62 @@ async function updateIOSCocoaPods(config: Config, plugins: Plugin[], deployment:
...
+    headersText = `,
+            publicHeadersPath: "."`;
+  }
+
+  const content = `// swift-tools-version: 5.9
+
+import PackageDescription
+
+let package = Package(
+    name: "${p.name}",
```
markemer, line 88: If the Cordova plugin is mixed source, this will fail, right?

jcesarmobile: That's why the PR is called "Initial Cordova plugin SPM support", not all plugins will work and future improvements will need to be done, it worked with the ones I tested, but I couldn't find any with mixed sources as it's not as common as in Capacitor plugins where all of them had both Objective-C and Swift code, on Cordova most of them are Objective-C and a few of them use just Swift and mixed source is uncommon.

OS-kepatotorica (approved):

Issue #8003, opened 2025-05-11
[Bug]: `cap run android --list` fails with no error message when no devices are available


Pyroglyph:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.2.0
  @capacitor/core: 7.2.0
  @capacitor/android: 7.2.0
  @capacitor/ios: 7.2.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 7.2.0
  @capacitor/core: 7.2.0
  @capacitor/android: 7.2.0

[success] Android looking great! 👌

### Other API Details

```Shell
Bun 1.2.12
Node 23.11.0
```

### Platforms Affected

- [ ] iOS
- [ ] Android
- [ ] Web

### Current Behavior

`cap run android --list` fails with no error message when no devices are connected:
```
$ bunx cap run android --list
[fatal] native-run failed with error

  
  
        More details for this error may be available online: https://github.com/ionic-team/native-run/wiki/Android-Errors
```

Directly calling `native-run` works normally:
```
$ bunx native-run android --list

Connected Devices:

  No connected devices found

Virtual Devices:

  No virtual devices found
```

When connecting a device, the original command works:
```
$ bunx cap run android --list
Name           API      Target ID
----------------------------------
Sony XQ-BQ52   API 33   <redacted>

```

### Expected Behavior

I expect `cap run android --list` to not fail when no devices are present. Or at the very least, have an actual error message.

### Project Reproduction

You can just use the blank project created with @capacitor/app

PR #8015, opened 2025-05-23
fix(cli): unexpected error on target device list

Chuckytuh:
Fixes https://github.com/ionic-team/capacitor/issues/8003.

With these changes, if no devices/emulators/simulators are available are available and `--list` is run, an information log is written or an empty array is returned when using `--json` option:

```bash 
$ npx cap run android --list
[info] No devices found.

$ npx cap run ios --list
[info] No devices found.

$ npx cap run android --list --json
[info] No devices found. 
[]

$ npx cap run ios --list --json
[info] No devices found. 
[]
```

Note that, even though the output contains a log, that's written into stderr so it doesn't break with composition in the shell, e.g.

```bash
npx cap run android --list --json 2>&- | jq
```

If no devices/emulators/simulators are available and `run` is executed without any options, then a fatal error is logged and execution halted:

```
$ npx cap android run
✔ Copying web assets from dist to android/app/src/main/assets/public in 13.88ms
✔ Creating capacitor.config.json in android/app/src/main/assets in 398.71μs
[info] Inlining sourcemaps
✔ copy android in 29.71ms
✔ Updating Android plugins in 3.43ms
[info] Found 23 Capacitor plugins for android:
< redacted >
✔ update android in 51.43ms
[info] No devices found.
```

### DIFF
--- a/cli/src/util/native-run.ts
+++ b/cli/src/util/native-run.ts
@@ -23,7 +23,10 @@ export async function getPlatformTargets(platformName: string): Promise<Platform
   try {
     const output = await runNativeRun([platformName, '--list', '--json']);
     const parsedOutput = JSON.parse(output);
-    if (parsedOutput.devices.length >= 0 || parsedOutput.virtualDevices.length >= 0) {
+    if (
+      parsedOutput.errors.length === 0 &&
+      (parsedOutput.devices.length >= 0 || parsedOutput.virtualDevices.length >= 0)
+    ) {
       return [
         ...parsedOutput.devices.map((t: any) => ({ ...t, virtual: false })),
         ...parsedOutput.virtualDevices.map((t: any) => ({
--- a/cli/src/util/native-run.ts
+++ b/cli/src/util/native-run.ts
@@ -23,7 +23,7 @@ export async function getPlatformTargets(platformName: string): Promise<Platform
   try {
     const output = await runNativeRun([platformName, '--list', '--json']);
     const parsedOutput = JSON.parse(output);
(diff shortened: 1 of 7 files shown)


OS-kepatotorica (approved):

cli/src/util/native-run.ts
```diff
@@ -23,7 +23,7 @@ export async function getPlatformTargets(platformName: string): Promise<Platform
   try {
     const output = await runNativeRun([platformName, '--list', '--json']);
     const parsedOutput = JSON.parse(output);
-    if (parsedOutput.devices.length || parsedOutput.virtualDevices.length) {
+    if (parsedOutput.devices.length >= 0 || parsedOutput.virtualDevices.length >= 0) {
```
jcesarmobile, line 26: length can be 0 if native-run returned some error, with this change native-run errors won't be shown anymore as it will always be `>= 0`.

native-run doesn't throw errors, they are returned in `parsedOutput.errors` object, so better check if its length is 0 in the else and log the error there.

markemer: So instead of the native error we'd get `logger.info(`No target devices found for ${platformName}.\n`);` instead?

jcesarmobile: this is for when there are no native-run errors, which happens when there are no devices, when there is an error it would show the native-run error

markemer: Yeah, that's the best we can do for now - probably have to take a look at native-run one of these days. Knowing apple - we're gonna be forced to next month.

markemer (approved):

jcesarmobile (approved):

Issue #7972, opened 2025-04-11
[Bug]: Capacitorjs does not work with sveltekit compiled in hash routing mode on iOS - infinitely reloads webview.


jjgmckenzie:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.2.0
  @capacitor/core: 7.2.0
  @capacitor/android: 7.2.0
  @capacitor/ios: 7.2.0

Installed Dependencies:

  @capacitor/android: not installed
  @capacitor/cli: 7.2.0
  @capacitor/core: 7.2.0
  @capacitor/ios: 7.2.0

[success] iOS looking great! 👌

### Other API Details

```Shell
npm version 10.9.2
node version v23.7.0
pod version 1.16.2
```

### Platforms Affected

- [x] iOS
- [ ] Android (did not test - possibly?)
- [ ] Web (does not affect web)

### Current Behavior

When loading the build from a sveltekit application in hash routing mode, the application appears to continuously reload the webview - never initializing sveltekit:

(reproduction project provided)
```
npm install
npx vite build
npx cap sync
npx cap open ios
```

xcode output: (webview loads about every second)

```
⚡️  Loading app at capacitor://localhost...
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
⚡️  WebView loaded
...
```

This does not occur when in server mode from vite, nor in web.

### Expected Behavior

It should behave identically to when in server mode or web

### Project Reproduction

https://github.com/jjgmckenzie/sveltekit-hash-routing-capacitor-app

jcesarmobile:
Marking it as a bug because it works if we load `capacitor://localhost/` instead of `capacitor://localhost` (notice the `/` at the end), but looks like a sveltekit bug to me not handling well urls with non standard schemes such as `capacitor`

I don't think we would be changing the url to load to add the `/` as it could cause other issues to other users, but maybe add a configuration option to conditionally add it.

jjgmckenzie:
@jcesarmobile Thank you for the quick discovery of the root cause. What would be the idiomatic way to add the trailing slash  to capacitor currently, without a configuration option?

[...]

PR #8019, opened 2025-05-28
feat: expose `appStartPath` on cap config server configuration

benmccarty91:
Fixes [7972](https://github.com/ionic-team/capacitor/issues/7972).
closes [3912](https://github.com/ionic-team/capacitor/issues/3912)

These changes introduce a new `capacitor.config` flag, `appStartPath`.

On iOS and Android, the logic already existed, but there was no way to configure this value.  This PR simply exposes that configuration.

### DIFF
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -616,6 +616,7 @@ export interface CapacitorConfig {
      * Only affects iOS.
      *
      * @since 7.3.0
+     * @default null
      */
     appStartPath?: string;
   };
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -614,6 +614,7 @@ export interface CapacitorConfig {
     /**
      * Append a path to the app URL.
      *
+     * Allows loading from other paths than the default `/index.html`.
      * @since 7.3.0
      * @default null
      */
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -613,7 +613,6 @@ export interface CapacitorConfig {
 
     /**
      * Append a path to the app URL.
-     * Only affects iOS.
      *
      * @since 7.3.0
      * @default null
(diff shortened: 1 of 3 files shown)


cli/src/declarations.ts
```diff
@@ -490,6 +490,14 @@ export interface CapacitorConfig {
      */
     initialFocus?: boolean;
 
+    /**
+     * Whether to use hash-routing compatibility mode.
+     *
+     * @since 7.3.0
+     * @default false
+     */
+    hashRoutingCompatibility?: boolean;
```
jcesarmobile, line 499: Hash routing (vanilla) works fine, the problem here seems to be with sveltekit router when using hash routing, so I don't think we should name it like that as other people might enable it thinking they need it when they don't. The name should probably be related to what it does, like `appendSlashToHostname` or something like that. And should probably be put inside the `server` object since all other url configuration options are in there.
The description should mention that it's for iOS only.

jcesarmobile (changes requested):
Added a comment. 
Also this should be a `feat`, not a `fix` since it adds a new configuration option.

cli/src/declarations.ts
```diff
@@ -610,6 +610,14 @@ export interface CapacitorConfig {
      * @default null
      */
     errorPath?: string;
+
+    /**
+     * Append a path to the app URL.
+     * Only affects iOS.
+     *
+     * @since 7.3.0
```
markemer, line 618: is there a default here? `null` or `""`

cli/src/declarations.ts
```diff
@@ -610,6 +610,14 @@ export interface CapacitorConfig {
      * @default null
      */
     errorPath?: string;
+
+    /**
+     * Append a path to the app URL.
```
jcesarmobile, line 615: Mention that it allows to load from alternate files other than `index.html`

markemer (approved):

PR #8034, opened 2025-06-17
Readme and launch.json improvements

OS-kepatotorica:


.vscode/launch.json
benmccarty91, line 1: instead of running transpiled js, why not run `ts-node` or similar and debug that?

OS-kepatotorica: I tried to get that working but kept running into issues and wanted to get something out that was still useable.

benmccarty91 (approved):

Issue #7840, opened 2025-01-23
[Bug]: Capacitor HTTP POST application/x-www-form-urlencoded body not encoded on iOS


ryaa:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.0.1
  @capacitor/core: 7.0.1
  @capacitor/android: 7.0.1
  @capacitor/ios: 7.0.1

Installed Dependencies:

  @capacitor/cli: 7.0.1
  @capacitor/core: 7.0.1
  @capacitor/android: 7.0.1
  @capacitor/ios: 7.0.1

[success] iOS looking great! 👌
[success] Android looking great! 👌

### Other API Details

```Shell
alexryltsov@Alexs-MacBook-Pro-2 master % npm --version
10.5.0
alexryltsov@Alexs-MacBook-Pro-2 master % node --version
v20.11.1
alexryltsov@Alexs-MacBook-Pro-2 master % pod --version
1.16.2
```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

Capacitor HTTP POST with ContentType application/x-www-form-urlencoded body not encoded on iOS (please note that it is properly encoded on Android so the behavior differs between these platform) - see below

Here is the request on iOS (note that the special characters are not encoded)
<img width="1378" alt="Image" src="https://github.com/user-attachments/assets/cada06dd-b1af-456d-94b4-be4c13236b25" />

Here is the same request on Android (note that the special characters are properly encoded)
<img width="1379" alt="Image" src="https://github.com/user-attachments/assets/0c204280-bbd6-4220-8fcd-99d12ad17863" />

This causes the problem when the server does not recognize the content of the request as expected when sent from iOS platform (all works fine on Android)

It also seems that `shouldEncodeUrlParams` HttpOption is not supported on iOS (see https://capacitorjs.com/docs/apis/http#httpoptions)

### Expected Behavior

Request content/body encoding must be the same for the both iOS and Android platforms.

### Project Reproduction

https://github.com/ryaa/capacitor-http-post-application-x-www-form-urlencoded-body-not-encoded-on-ios

### Additional Information

See README in the Project Reproduction repository for more details.
[...]

PR #8037, opened 2025-06-25
fix(http): Properly URL-encode key and values during `x-www-form-urlencoded` POSTs

theproducer:
This PR fixes keys and values that were not url-encoded in CapacitorHttp on iOS during `application/x-www-form-urlencoded` HTTP POST requests.

closes: https://github.com/ionic-team/capacitor/issues/7840

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
+++ b/ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
@@ -44,7 +44,7 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
 
         obj.keys.forEach { (key: String) in
             let value = obj[key] as? String ?? ""
-            components.queryItems?.append(URLQueryItem(name: key.addingPercentEncoding(withAllowedCharacters: allowed) ?? key, value: value.addingPercentEncoding(withAllowedCharacters: allowed)))
+            components.queryItems?.append(URLQueryItem(name: key.addingPercentEncoding(withAllowedCharacters: allowed)?.replacingOccurrences(of: "%20", with: "+") ?? key, value: value.addingPercentEncoding(withAllowedCharacters: allowed)?.replacingOccurrences(of: "%20", with: "+")))
         }
 
         if components.query != nil {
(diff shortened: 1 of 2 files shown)


ios/Capacitor/Capacitor/Plugins/CapacitorUrlRequest.swift
```diff
@@ -40,8 +40,11 @@ open class CapacitorUrlRequest: NSObject, URLSessionTaskDelegate {
             throw CapacitorUrlRequestError.serializationError("[ data ] argument for request with content-type [ multipart/form-data ] may only be a plain javascript object")
         }
 
+        let allowed = CharacterSet(charactersIn: "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._")
```
jcesarmobile:
```suggestion
        let allowed = CharacterSet(charactersIn: "-._*").union(.alphanumerics)
```
This can be simplified with an union to `.alphanumerics` instead of typing them all
Also added `*` to match Android behavior, which uses `URLEncoder.encode` and according to the docs the `*` should remain the same:
> The special characters ".", "-", "*", and "_" remain the same.
https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html#encode-java.lang.String-java.lang.String-

I've noticed that this is also encoding the keys, tested and web also encodes the keys, but Android doesn't encode the keys.
Not sure which one should be the correct behavior, but I think all the 3 platforms should behave the same way, so all of them should encode the keys or none should.

jcesarmobile: Another thing, on web and Android whitespace (` `) is converted to `+`, while on iOS it's getting converted into `%20`.

theproducer: > Not sure which one should be the correct behavior, but I think all the 3 platforms should behave the same way, so all of them should encode the keys or none should.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/POST), keys being encoded is the correct behavior.  However, changing it on Android could be seen as breaking, unless we categorize non-encoded keys as a bug (which I do).

What do you think?

jcesarmobile: yeah, it should be considered a bug and should be fixed

PR #8039, opened 2025-06-26
Feature/rdmr 210 http content headers not sent when using axios

OS-kepatotorica:
The headers were not being set correctly when CapacitorHttp is set. The missing headers were Content-Type and Content-Length.
https://github.com/user-attachments/assets/482e3a3b-1e0e-43fc-8648-7ad65ad81b05

Changes not 100% related to the ticket are around the formatting of the logs 
1. making them a little more descriptive **native** -> **message to native** and **result** -> **message from native**
2. Changing the colors that were printing in the console to colors that work for both dark and light mode as reading the logs were quite difficult for me to read using dark mode.

### DIFF
--- a/core/native-bridge.ts
+++ b/core/native-bridge.ts
@@ -110,6 +110,9 @@ const convertBody = async (
     return {
       data: formData,
       type: 'formData',
+      headers: {
+        'Content-Type': `multipart/form-data; boundary=----WebKitFormBoundary${Math.random().toString(36).substring(2, 15)}`,
+      },
     };
   } else if (body instanceof File) {
     const fileData = await readFileAsBase64(body);
@@ -611,7 +614,6 @@ const initBridge = (w: any): void => {
 
         window.XMLHttpRequest = function () {
           const xhr = new win.CapacitorWebXMLHttpRequest.constructor();
-
           Object.defineProperties(xhr, {
             _headers: {
               value: {},
--- a/core/native-bridge.ts
+++ b/core/native-bridge.ts
@@ -321,13 +321,13 @@ const initBridge = (w: any): void => {
         const success = result.success === true;
 
         const tagStyles = success
-          ? 'font-style: italic; font-weight: lighter; color: gray'
+          ? 'font-style: italic; font-weight: lighter; color: green'
           : 'font-style: italic; font-weight: lighter; color: red';
 
         c.groupCollapsed(
-          '%cresult %c' + result.pluginId + '.' + result.methodName + ' (#' + result.callbackId + ')',
+          '%cmessage from native %c' + result.pluginId + '.' + result.methodName + ' (#' + result.callbackId + ')',
           tagStyles,
-          'font-style: italic; font-weight: bold; color: #444',
+          'font-style: italic; font-weight: bold; color: #4EA1E0',
         );
         if (result.success === false) {
           c.error(result.error);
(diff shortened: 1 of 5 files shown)


markemer:
We use conventional commits, so the title of this should be something like `fix: description of problem/fix`

markemer (changes requested):
Looks good, just need to fix the title.

OS-kepatotorica:
> We use conventional commits, so the title of this should be something like `fix: description of problem/fix`

Updated, thanks for the call out!

jcesarmobile:
I personally don't like this approach of setting the content-type ourselves.

First of all, this change affects both fetch and XHR, but fetch is working fine without this change.

Second, this change prevent's users from setting their own boundary since both headers get sent to the native code, the user provided one and the hardcoded one, and the native code sometimes picks the user provided one and other times it picks the hardcoded one.

Can at least be done only for XHR and check if the user provided a content-type header to not hardcode it in that case?

OS-kepatotorica:
> I personally don't like this approach of setting the content-type ourselves.
> 
> First of all, this change affects both fetch and XHR, but fetch is working fine without this change.
> 
> Second, this change prevent's users from setting their own boundary since both headers get sent to the native code, the user provided one and the hardcoded one, and the native code sometimes picks the user provided one and other times it picks the hardcoded one.
> 
> Can at least be done only for XHR and check if the user provided a content-type header to not hardcode it in that case?

Fixed

core/native-bridge.ts
```diff
@@ -726,9 +724,15 @@ const initBridge = (w: any): void => {
               });
 
               convertBody(body).then(({ data, type, headers }) => {
-                const otherHeaders =
+                let otherHeaders =
                   this._headers != null && Object.keys(this._headers).length > 0 ? this._headers : undefined;
 
+                if (body instanceof FormData) {
+                  otherHeaders = {
+                    ...otherHeaders,
```
jcesarmobile: In the previous commit you were checking the Content-Type to not override it if present, but this overrides it again.

Also when I said "content-type could be lower cased" I meant it can be both `Content-Type` or `content-type`, both are valid and both need to be checked, I wasn't asking you to change `Content-Type` to `content-type`.

OS-kepatotorica: So just to be clear, are you are saying that we want to compare against the headers included in the body, or the headers that are passed into convertBody (which is always left out for XHR?).
image

OS-kepatotorica: In the previous commit I was checking the contentType as it was passed in through the convertBody method, which for XHR is always null. I didn't think that would be relevant here since we don't have that variable, but I can do a check against the headers present on the body.

OS-kepatotorica: Updated

core/native-bridge.ts
```diff
@@ -726,9 +724,21 @@ const initBridge = (w: any): void => {
               });
 
               convertBody(body).then(({ data, type, headers }) => {
-                const otherHeaders =
+                let otherHeaders =
                   this._headers != null && Object.keys(this._headers).length > 0 ? this._headers : undefined;
 
+                if (body instanceof FormData) {
+                  const multipartHeaderAlreadyExists =
+                    this._headers['Content-Type']?.includes('multipart/form-data') ||
```
jcesarmobile: Just check if content-type exists, not if it's multipart or not.

While users shouldn't be doing that, they can, and when the http plugin is disabled, it gets sent with the provided content-type, so we shouldn't change that behavior.

OS-kepatotorica: Fixed

core/native-bridge.ts
```diff
@@ -2,6 +2,7 @@
  * Note: When making changes to this file, run `npm run build:nativebridge`
  * afterwards to build the nativebridge.js files to the android and iOS projects.
  */
+
```
jcesarmobile:
```suggestion
```
This new line here makes the comment to appear in the `native-bridge.js` files, which doesn't make much sense in them since it talks about making changes on the file and generating the .js files.

benmccarty91 (approved):

jcesarmobile:
You need to re run `npm run build:nativebridge` so the comment is removed from the `native-bridge.js` files

PR #8045, opened 2025-07-02
refactor(core): reference window.Capacitor

Chuckytuh:
Since the addition of the ability to inject misc JS modules into the WebView on https://github.com/ionic-team/capacitor/pull/7864 there's a possibility that some user provided JS module defines a "top-level" `const Capacitor` variable which ends up conflicting with the variable being referenced by cordova.js.

This change makes it so that cordova.js usage of global Capacitor object follows the other usages found throughout the framework where `window.` is prefixed.


benmccarty91 (approved):

theproducer (approved):

markemer:
Let's see if updating with Julio's fix makes lint pass.

markemer:
Booyah! Passes.

PR #8062, opened 2025-07-14
chore: update main branch for Capacitor 8

markemer:


theproducer (approved):

.github/workflows/publish-npm-latest.yml
```diff
@@ -15,7 +15,8 @@ permissions:
 
 jobs:
   deploy-npm-latest:
-    if: github.ref == 'refs/heads/main'
+    # Changed for safety until 8.0 ships
```
jcesarmobile: I don't think we need this, if I recall correctly, we have `ci:publish:latest-from-pre` because once the version is set to alpha/beta/pre, the latest publish fails and we need that script to publish the first release

markemer: Cool, I'll ditch it.

PR #8065, opened 2025-07-15
chore: Update release branches

jcesarmobile:
update the branch in lerna, android, ios and npm nightly workflows


markemer (approved):
Good catch, I wonder if we should write a script that can do this. Put that on my ever lengthening to do list.

PR #8077, opened 2025-07-23
feat(android-template)!: added density to configChanges to prevent reloading of the WebView

OS-kepatotorica:
Adding density to the default `android:configChanges` values prevents the webView from being reloaded on resize. Currently, there is a known issue with android that causes the repainting of the webView to fail on occasion. A bug has been created and can be tracked [here](https://issuetracker.google.com/issues/433671013). 

Alternative solutions:
- Disable hardware acceleration and resizing will work, but app functionality will be degraded.
- Tweak `android:configChanges` if you are okay with losing the state of your webView during resize events.

Example of behavior with the changes in this PR: 
https://github.com/user-attachments/assets/6cafbb1b-9ae9-4082-98ce-0bff36fd327b

### DIFF
--- a/android-template/app/src/main/AndroidManifest.xml
+++ b/android-template/app/src/main/AndroidManifest.xml
@@ -10,7 +10,7 @@
         android:theme="@style/AppTheme">
 
         <activity
-            android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode|navigation"
+            android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode|navigation|density"
             android:name=".MainActivity"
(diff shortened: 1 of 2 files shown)


jcesarmobile (changes requested):
Can you remove the migrate.ts changes?

It should be addressed in a future PR when we update the migrate command for Capacitor 8, and the changes are incorrect at the moment.

jcesarmobile:
Marking it as breaking despite is not technically breaking as will only affect new apps, but doing it to remember to document it and to make migrate do the change.

OS-kepatotorica:
> Can you remove the migrate.ts changes?
> 
> It should be addressed in a future PR when we update the migrate command for Capacitor 8, and the changes are incorrect at the moment.

Updated

benmccarty91 (approved):

PR #8087, opened 2025-07-28
Use JSONTokener for application/json reponse parsing

Crylion:
I'm changing the way the core HTTP plugin parses JSON responses, because I encountered a problem with a specific response from an API:
The request in question returns an application/json response that is just a 64bit number literal.

Screenshot 2025-07-28 at 12 21 10

Unfortunately, the way the parser in HttpRequestHandler.java works, is that it tries to parse number literals without a decimal point as a 32bit `Integer`, which fails for a number this size. 
Now, the easiest fix would have been to just change the type to `Long` here, but that seemed a bit short sighted, because there seem to have been a lot of changes to this parser for specific scenarios like this, so a new approach seemed the better choice. As you can see in my commits, I first tried to just copy&paste the parser logic that JSObject/JSArray use internally to literals, but that seemed silly to me, because at that point, why not just USE the internal parser instead of trying to recreate most of what it does.

So my change gets rid of all the specific logic for all the different types of literals, and just uses the parser of the JSONTokener instead. In it's [readLiteral](https://github.com/AndroidSDKSources/android-sdk-sources-for-api-level-35/blob/a1a930d749bc0d6b4cf96f4a029bd090e6c8851a/org/json/JSONTokener.java#L293) method, the Tokener basically does the exact same thing the long if/else statement here did before, but better.
Now I just try to parse the input as first an object, then an array, and if that does not work, I just use the Tokener directly to let it parse the input as a literal.

I tested it with all of the literals that were processed here before and the result for all of them is the same as before, but now Long values work too, and we hopefully don't have to add more special cases to this method anymore.


de-dan:
1. Is returning a Long value inside the JSON specification? I think by definition JSON number literals are treated as Double - if you need to transmit a 64-bit Long value, return it as String literal "17273747488494387373".
2. Did you check the iOS behavior for handling Long literals?

Crylion:
> 1. Is returning a Long value inside the JSON specification? I think by definition JSON number literals are treated as Double - if you need to transmit a 64-bit Long value, return it as String literal "17273747488494387373".

It is not that **I need** to transmit it that way, the API we are working with returns the value that way and I have no choice in the matter. And the problem is, that the current implementation in HttpRequestHandler does not pass that value along but fails while trying to parse it. So my actual application never actually receives the value. For sure, in my actual application code I would treat this as a string literal, but the reponse never makes it through the `parseJson` method in the Capacitor Android HTTP implementation.

But yes, in at least some versions of the standard, that is unfortunately a valid value.
Screenshot 2026-01-13 at 11 22 11


> 2. Did you check the iOS behavior for handling Long literals?

At the time of implementing this half a year ago, I did. In iOS the issue did not arise. But I can refresh my memory and double check how the iOS implementation handles this.

All in all, I just think it's generally a good idea to just use the JSONTokener implementation of the JSON parsing as I have proposed, instead of more or less reimplementing it in a very similar (yet incomplete) way. If there already is an available JSON parsing method in the system that correctly handles edge cases like this, why not use it instead of reinventing this particular wheel?

eric-horodyski:
Thanks for opening this PR 🙏

Issue #8085, opened 2025-07-27
[Bug]: Capacitor WebView treats URLs with dots as file requests, breaking SPA routing and page reload.


iamriajul:
### Capacitor Version

💊   Capacitor Doctor  💊 

Installed Dependencies:

  @capacitor/cli: 7.4.2
  @capacitor/core: 7.4.2
  @capacitor/android: 7.4.2
  @capacitor/ios: 7.4.2

[success] iOS looking great! 👌
[success] Android looking great! 👌

### Other API Details

```Shell

```

### Platforms Affected

- [x] iOS
- [x] Android
- [ ] Web

### Current Behavior

When navigating to URLs that contain dots (.) in the path, Capacitor's WebView treats them as file requests rather than SPA routes. This causes a 404 error instead of serving the `index.html` file for client-side routing.

For example:
- `http://localhost/@username` → Works (serves index.html)
- `http://localhost/@user.name` → Fails (404 - treats as file request)
- `http://localhost/page.section` → Fails (404 - treats as file request)

And shows console error in Xcode: 
⚡️  WebView failed provisional navigation
⚡️  Error: The file “@user.name” couldn’t be opened because there is no such file.

And in Android, An error page is shown with Page is not available.

<img width="576" height="1280" alt="Image" src="https://github.com/user-attachments/assets/d8c73c7e-3926-406a-9778-99779b8e1588" />

### Expected Behavior

All non-asset URLs should serve `index.html` to allow the SPA router to handle routing, regardless of whether the URL contains dots.

### Project Reproduction

https://github.com/iamriajul/capacitor-dot-in-urls-spa-navigation-issue-reproduction

iamriajul:
Requesting @jcesarmobile for taking a look at the PR: [fix: Capacitor WebView treats URLs with dots as file requests, breaking SPA routing and page reload. #8137](https://github.com/ionic-team/capacitor/pull/8137)

We're currently badly stuck with the issue in a production app. Would be very grateful if you can prioritize this PR to be merged and released a new version.

ingmar-crazygames:
We have encountered the same issue at CrazyGames, and are currently patching it manually in our codebase. A first-party fix would be nice.

PR #8137, opened 2025-09-02
fix: Capacitor WebView treats URLs with dots as file requests, breaking SPA routing and page reload. 

iamriajul:
Adds a new configuration option `routeWithFallback` to CapacitorConfig.server that enables
fallback to index.html for routes containing dots when the requested file doesn't exist.

This solves the issue where SPA routes like /@user.name or /api/data.json are incorrectly
treated as file requests, returning 404 instead of serving index.html for client-side routing.

The feature is disabled by default to maintain backward compatibility. When enabled:
- iOS: Checks if the file exists before serving, falls back to index.html if not found
- Android: Catches IOException when file not found, serves index.html as fallback

Fixes https://github.com/ionic-team/capacitor/issues/8085

### DIFF
--- a/ios/Capacitor/Capacitor/Router.swift
+++ b/ios/Capacitor/Capacitor/Router.swift
@@ -9,7 +9,7 @@
 import Foundation
 
 public protocol Router {
-    func route(for path: String, checkFileExists: Bool) -> String
+    func route(for path: String) -> String
     var basePath: String { get set }
 }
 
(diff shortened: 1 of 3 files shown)


iamriajul:
By setting `config.server.routeWithFallback` to `true`

I've verified the fix on Android:
Screenshot_1756794865

I've verified the fix on iOS:
simulator_screenshot_A2DA1918-F5BC-4927-8981-7E6BB38F4996

eric-horodyski:
Thanks for opening this PR 🙏

Issue #8101, opened 2025-08-11
[Bug]: REGRESSION: iOS bridge doesn't reset plugin listeners after WebContent process loss


gabeschine:
### Capacitor Version

7.4.2

### Other API Details

```Shell

```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

PR https://github.com/ionic-team/capacitor/pull/7962 seems to have regressed the issue described in https://github.com/ionic-team/capacitor/issues/7810 and that was fixed in https://github.com/ionic-team/capacitor/pull/7905.

### Expected Behavior

See https://github.com/ionic-team/capacitor/issues/7810

### Project Reproduction

See https://github.com/ionic-team/capacitor/issues/7810

### Additional Information

@markemer  bumping for your visibility.

I looked a little more at this. Is it possible that webView(_:didStartProvisionalNavigation:) races with the web content to clear out listeners before the web content starts processing? It's important that all the plugin listener state is cleared before _any_ javascript in the web content starts to run.

PR #8143, opened 2025-09-04
fix(ios): also call bridge.reset() on webViewWebContentProcessDidTerminate

jcesarmobile:
webViewWebContentProcessDidTerminate won't be called when the app is in the background but when you put it back in the foreground, so when it reloads the app and the didStartProvisionalNavigation gets called might be too late for some events

closes https://github.com/ionic-team/capacitor/issues/8101


markemer (approved):

gabeschine:
Thank you so much for taking care of this. Is there any chance we can get a 7.x minor release that includes this fix? We'd like to be able to upgrade but currently cannot due to this issue.

Issue #7944, opened 2025-03-26
[Bug]: Android Capacitor Returns UTF-8 Binary String Instead of Blob for Binary Responses


layola13:
### Capacitor Version

 Capacitor Doctor   

Latest Dependencies:

  @capacitor/cli: 7.1.0
  @capacitor/core: 7.1.0
  @capacitor/android: 
7.1.0
  @capacitor/ios: 7.1.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 7.1.0
  @capacitor/core: 7.1.0
  @capacitor/android: 7.1.0

[success] Android looking great! 👌

### Other API Details

```Shell
npm -v 10.9.2

 node -v v23.9.0
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

When making network requests with `responseType: 'blob'` on Android with Capacitor, instead of receiving a proper Blob object like on web platforms, the response is a UTF-8 encoded binary string. This inconsistency causes issues when trying to handle binary data like audio files across different platforms.

- **On web:** `response.data` is a proper Blob object

![Image](https://github.com/user-attachments/assets/586ce05f-c97d-42be-aabf-0d842b353f94)

- 
- **On Android:** `response.data` is a UTF-8 encoded binary string

![Image](https://github.com/user-attachments/assets/cc7e5495-5aaa-412b-8a0d-5337244f1b53)

### Expected Behavior

When making a request with `responseType: 'blob'`, the response should be a standard Blob object on all platforms (web and Android), allowing consistent code handling.

### Project Reproduction

https://gist.github.com/layola13/55bd89421c9e82d066d584b80f0da551

### Additional Information

This inconsistency makes it difficult to handle binary responses in a cross-platform manner. Ideally, Capacitor should normalize the response type to be consistent with web behavior, returning a proper Blob object when `responseType: 'blob'` is specified.

layola13:
https://github.com/layola13/capacitor-blob-bug-reproduction
@jcesarmobile

PR #8155, opened 2025-09-11
Response type correctly mapped on CapacitorHttp plugin call

cornflakes:
When using Capacitor Http Plugin then member data "responseType" is not mapped to the pluginCall.
This PR will fix that issue.


rsreimer:
fixes #7944

eric-horodyski:
Thanks for opening this PR 🙏

cornflakes:
Thanks for the ping. I’ve updated the branch with the latest from main. There were no merge conflicts, and the PR is up to date with the current releases. Ready for review whenever you have a moment!

PR #8160, opened 2025-09-18
chore(android)!: Remove bridge_layout_main.xml

jcesarmobile:
It was replaced by `capacitor_bridge_layout_main.xml` and kept in case somebody was using it


trevor-lambert (approved):

markemer (approved):
Good riddance!

PR #8180, opened 2025-10-03
feat: System Bars Plugin

theproducer:
This PR introduces a new core plugin, System Bars, designed to be a modern take on the Status Bar plugin, managing [Android System Bars](https://developer.android.com/design/ui/mobile/guides/foundations/system-bars), the iOS status bar, and fix issues related to broken safe area insets on Android.

### DIFF
--- a/core/systembars.md
+++ b/core/systembars.md
@@ -72,7 +72,7 @@ const setStatusBarAnimation = async () => {
 ## Configuration
 | Prop          | Type                 | Description                                                               | Default            |
 | ------------- | -------------------- | ------------------------------------------------------------------------- | ------------------ |
-| **`disableInsets`** | <code>boolean</code> | Enables the injection of device css insets into the webview.  This option is only supported on Android. | <code>false</code> |
+| **`disableInsets`** | <code>boolean</code> | Disables the injection of device css insets into the webview.  This option is only supported on Android. | <code>false</code> |
 | **`style`** | <code>string</code> | The style of the text and icons of the system bars. | <code>DEFAULT</code> |
 | **`hidden`** | <code>boolean</code> | Hide the system bars on start. | <code>false</code> |
 | **`animation`** | <code>string</code> | The type of status bar animation used when showing or hiding.  This option is only supported on iOS. | <code>FADE</code> |


pauldemarco:
This looks great, what more is needed before merge?

pauldemarco:
I am noticing that currently, without this plugin, the safe area and the status bar are working great on iOS -- it even changes the status bar text depending on light or dark mode when I toggle dark on the root: `document.documentElement.classList.toggle('dark')`

I am noticing this PR includes some styling on the status bar. Is there any way to have Android act like the iOS and have it "just work"?

I also want to make sure this won't break what we have working fine on iOS.

tafelnl:
> I am noticing this PR includes some styling on the status bar. Is there any way to have Android act like the iOS and have it "just work"?
> 
> I also want to make sure this won't break what we have working fine on iOS.

I am not sure if you have a correct understanding of this "inject css variables" thing. It doesn't inject `env` css variables, as that isn't possible. It detects the safe area insets in a best effort and injects them as custom `var` css variables. That means that you **will** have to change code to make it work on both iOS and Android. That's exactly one of the reasons I'm advocating for a native approach

tafelnl:
I'd like to add that as of November 1st (actually earlier, but Google granted an extension to developers that depend on older SDKs) no one will be able to publish a release to the play store if they're not supporting edge-to-edge. So I think this should be given the utmost priority by the Capacitor team

pauldemarco:
> > I am noticing this PR includes some styling on the status bar. Is there any way to have Android act like the iOS and have it "just work"?
> > I also want to make sure this won't break what we have working fine on iOS.
> 
> I am not sure if you have a correct understanding of this "inject css variables" thing. It doesn't inject `env` css variables, as that isn't possible. It detects the safe area insets in a best effort and injects them as custom `var` css variables. That means that you **will** have to change code to make it work on both iOS and Android. That's exactly one of the reasons I'm advocating for a native approach

I just discovered all of this was only a problem with an Android Medium emulator. I just tried the Pixel 9 emulator and safe area is working great without any plugins.

theproducer:
Hello everyone!  Stay tuned, I'm on the verge of getting this past the finish line, The plan is to make some final changes and wrap this up this week.

theproducer:
> I am noticing this PR includes some styling on the status bar. Is there any way to have Android act like the iOS and have it "just work"?

By "just work", you're expecting just the default status bar, without the fullscreen / transparent status bar style, correct?

tafelnl:
I'm curious which approach you've chosen to go with. I'm still advocating for going the native route. To demonstrate how that would work I made a simple demo proof of concept. Mind you that I've created this in like 5 minutes, so it's still not perfect, but you will get the main idea.

Demo: https://github.com/tafelnl/capacitor-demo/tree/edge-to-edge-poc
The magic happens in this file: https://github.com/tafelnl/capacitor-demo/blob/edge-to-edge-poc/android/app/src/main/java/com/example/demo/EdgeToEdgePlugin.kt

I also commented some todos which I haven't done because it's a proof of concept.

pauldemarco:
> > I am noticing this PR includes some styling on the status bar. Is there any way to have Android act like the iOS and have it "just work"?
> 
> By "just work", you're expecting just the default status bar, without the fullscreen / transparent status bar style, correct?

As of right now on iOS, without this plugin, the status bar is transparent and switches the text color automatically to the right color when I set the root html class to dark mode and light mode. Not sure how this is working.

core/systembars.md
```diff
@@ -0,0 +1,273 @@
...
+```css
+html {
+  padding-top: var(--safe-area-inset-top, env(safe-area-inset-top, 0px));
+  padding-bottom: var(--safe-area-inset-bottom, env(safe-area-inset-bottom, 0px));
+  padding-left: var(--safe-area-inset-left, env(safe-area-inset-left, 0px));
+  padding-right: var(--safe-area-inset-right, env(safe-area-inset-right, 0px));
+}
+```
+
+To disable the inset variable injections, set the configuration setting `enableInsets` to `false`.
```
tafelnl: The code doesn't do what this documentation suggests. IMO the documentation suggests that this config only toggles the injection of fallback variables. But in reality, it disables all of the logic. Meaning that on webview version < m140 the insets will be incorrect/broken.

That brings me to another issue; what will be done with the 'legacy' code related to insets. For example the `adjustMarginsForEdgeToEdge` logic

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,200 @@
...
+
+        return majorVersion >= 140;
+    }
+
+    private void initSystemBars() {
+        boolean enableInsets = getConfig().getBoolean("enableInsets", true);
+        String style = getConfig().getString("style", STYLE_DEFAULT).toUpperCase();
+        boolean hidden = getConfig().getBoolean("hidden", false);
+
+        if (enableInsets) {
```
tafelnl: It might be possible to look at the `viewport-fit=cover` meta tag here. I think that would be the cleanest solution, as that is the [official](https://www.w3.org/TR/css-round-display-1/#descdef-viewport-viewport-fit) way of a developer telling the webview/browser "I'll handle the safe area insets"

tafelnl: And if we decide not to do that, I think it should be very clearly stated in the docs that a developer should always handle the safe area insets when targeting the Android platform

riderx:
Thanks @theproducer for this; this is much needed !
Long ago I made this: https://github.com/Cap-go/capacitor-home-indicator
who was also made to help on that matter, I do think both are needed.
I do believe both should be core API :)
I'm up to implement it if needed

theproducer:
> Thanks @theproducer for this; this is much needed ! Long ago I made this: https://github.com/Cap-go/capacitor-home-indicator who was also made to help on that matter, I do think both are needed. I do believe both should be core API :) I'm up to implement it if needed

I think being able to control iOS's home indicator fits in to the use case here, thanks for the suggestion!

riderx:
Something i found out doing this plugin is hide and show doesn’t naturally update the css var for safe areas, so that something to take in account that would let us easy have reactive design. During my session i was not able to update the real value so i had to create new var. if you know how to handle this i would be glad ti be included too.
I know ionic framework has hack for that with ionic var, but it will be nice to have a generic solution who work for all, and documented in the website:)
Safe areas is a constant topic of beginners :)

tafelnl:
Not sure how this PR relates to it, but you might want to take a look at this PR: https://github.com/ionic-team/capacitor-plugins/pull/2388

tafelnl:
I think it makes sense to remove the workaround logic from the keyboard plugin and incorporate it in the capacitor core/this PR. Otherwise all users would have to install the keyboard plugin in order to make it work correctly

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,216 @@
...
+    @PluginMethod
+    public void setAnimation(final PluginCall call) {
+        call.resolve();
+    }
+
+    private void setupSafeAreaInsets(boolean hasFixedWebView, boolean hasMetaViewportCover) {
+        View decorView = getActivity().getWindow().getDecorView();
+
+        ViewCompat.setOnApplyWindowInsetsListener(decorView, (v, insets) -> {
+            if (hasFixedWebView || hasMetaViewportCover) {
```
tafelnl:
```suggestion
            if (hasFixedWebView && hasMetaViewportCover) {
```

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,216 @@
...
+                return insets;
+            }
+
+            Insets safeArea = insets.getInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout());
+            injectSafeAreaCSS(safeArea.top, safeArea.right, safeArea.bottom, safeArea.left);
```
tafelnl, line 126: I think in this case we should fallback to logic similar as to what `adjustMarginsForEdgeToEdge` does.

Also injecting the safe area insets as custom vars should be a toggleable option I think. And only respected if `hasMetaViewportCover==true` imo, so it stays consistent with the native inset behavior

theproducer: > I think in this case we should fallback to logic similar as to what adjustMarginsForEdgeToEdge does.

You are referring to adding margins directly to the native view?  That was meant to be a quick and dirty fix - we generally try to avoid manipulating the native views style-wise as much as possible to leave the power and final decision making in the hands of the web developer.  

Using CSS env vars (albeit a bit convoluted because of the Android bug: `var(--safe-area-inset-top, env(safe-area-inset-top, 0px));`) would be the normal way of handling edge to edge behavior, and we want to stick as close to that as possible.  Additionally, we have other requirements depending on these injected vars.

> Also injecting the safe area insets as custom vars should be a toggleable option I think.

This is something we can do, I'll work on adding a config option for disabling the inset injections altogether.

tafelnl: > You are referring to adding margins directly to the native view

Yes that's what I meant. It would align with iOS behavior. If a user does not enable edge-to-edge support through the `viewport-fit=cover` tag, the webview itself will already add margins for you. Otherwise every capacitor developer would need to add support for env variables. Which would be a major breaking change.

Ideally, of course, this is something that the Android team will just natively support out of the box. But iirc a few years/months ago one of the devs mentioned that it's not something they were looking into. I cannot find the comment anymore. I'll link it if I come across it

tafelnl:
> I think it makes sense to remove the workaround logic from the keyboard plugin and incorporate it in the capacitor core/this PR. Otherwise all users would have to install the keyboard plugin in order to make it work correctly

Actually, upon better looking into this, I think the workaround as is can be removed altogether. Instead we should do something like this inside the `setOnApplyWindowInsetsListener`

```java
Insets systemBarsInsets = windowInsets.getInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout());
Insets imeInsets = windowInsets.getInsets(WindowInsetsCompat.Type.ime());
boolean keyboardVisible = windowInsets.isVisible(WindowInsetsCompat.Type.ime());

ViewGroup.MarginLayoutParams mlp = (ViewGroup.MarginLayoutParams) parent.getLayoutParams();
if (keyboardVisible) {
    // https://issues.chromium.org/issues/457682720
    // this is a workaround to push the webview behind the keyboard for webview versions that have a bug
    // that causes the bottom inset to be incorrect if the IME is visible
    boolean hasWebViewWithKeyboardBug = true; // chromium version with this bugfix is yet to be released, so the version is unknown still
    if (hasWebViewWithKeyboardBug) {
        mlp.bottomMargin = imeInsets.bottom - systemBarsInsets.bottom;
    } else {
        mlp.bottomMargin = imeInsets.bottom;
    }
} else {
    mlp.bottomMargin = 0;
}
parent.setLayoutParams(mlp);

return windowInsets;
```

`parent` can be either `bridge.getWebView()` or `(ViewGroup) getBridge().getWebView().getParent()`. Not sure which is better

I think this is much cleaner than the current workaround. Additionally it makes the workaround also work across config changes such as rotating the screen (the current workaround would break, which is a bug currently)

core/src/core-plugins.ts
```diff
@@ -494,3 +494,155 @@ export const CapacitorHttp = registerPlugin<CapacitorHttpPlugin>('CapacitorHttp'
...
+   */
+  Light = 'LIGHT',
+
+  /**
+   * The style is based on the device appearance or the underlying content.
+   * If the device is using Dark mode, the statusbar text will be light.
+   * If the device is using Light mode, the statusbar text will be dark.
+   *
+   * @since 8.0.0
+   */
```
OS-pedrogustavobilro, line 524: Minor: technically it's not just the text right?

```suggestion
  /**
   * The style is based on the device appearance or the underlying content.
   * If the device is using Dark mode, the system bars content will be light.
   * If the device is using Light mode, the system bars content will be dark.
   *
   * @since 8.0.0
   */
```

core/src/core-plugins.ts
```diff
@@ -494,3 +494,155 @@ export const CapacitorHttp = registerPlugin<CapacitorHttpPlugin>('CapacitorHttp'
...
+ * Available inset edges.
+ */
+export type SystemBarsInsets = 'TOP' | 'BOTTOM' | 'LEFT' | 'RIGHT';
+
+export interface SystemBarsStyleOptions {
+  /**
+   * Style of the text and icons of the system bars.
+   *
+   * @since 8.0.0
+   * @default default
```
OS-pedrogustavobilro: Should be?

```suggestion
   * @default 'DEFAULT'
```

cli/src/declarations.ts
```diff
@@ -710,4 +710,31 @@ export interface PluginsConfig {
...
   };
+
+  /**
+   * System Bars plugin configuration
+   *
+   * @since 8.0.0
+   */
+  SystemBars?: {
+    /**
+     * The style of the text and icons of the system bars.
```
OS-pedrogustavobilro, line 727: Missing what would be the default value? 'DEFAULT'?

cli/src/declarations.ts
```diff
@@ -710,4 +710,31 @@ export interface PluginsConfig {
...
+     */
+    style?: string;
+
+    /**
+     * Hide the system bars on start.
```
OS-pedrogustavobilro, line 736: Missing what would be the default value? "false"?

core/systembars.md
```diff
@@ -0,0 +1,273 @@
+# SystemBars
+
+The SystemBars API provides methods for configuring the style and visibility of the device System Bars / Status Bar.  This API differs from the [Status Bar](https://capacitorjs.com/docs/apis/status-bar) plugin in that it is only intended to support modern edge to edge use cases moving forward.  For legacy functionality, use the [Status Bar](https://capacitorjs.com/docs/apis/status-bar) plugin.
```
OS-pedrogustavobilro, line 3: I think we could use more information on how this core plugin works with status-bar plugin, and when to use one or the other or both.

Unsure if this should reside here, in status bar documentation, or somewhere, just wanted to bring it up.

core/systembars.md
```diff
@@ -0,0 +1,273 @@
...
+const setStatusBarAnimation = async () => {
+  await SystemBars.setAnimation({ animation: "NONE" });
+}
+
+````
+
+## Configuration
+| Prop          | Type                 | Description                                                               | Default            |
+| ------------- | -------------------- | ------------------------------------------------------------------------- | ------------------ |
+| **`enableInsets`** | <code>boolean</code> | Enables the injection of device css insets into the webview.  This option is only supported on Android. | <code>true</code> |
```
OS-pedrogustavobilro: Question: We're not using `enableInsets` in the code anymore, in which case we should remove mentions of it in this here README right?

Or are we gonna still allow for configurability/control of inset injection, regardless of webview version?

core/src/core-plugins.ts
```diff
@@ -494,3 +494,155 @@ export const CapacitorHttp = registerPlugin<CapacitorHttpPlugin>('CapacitorHttp'
...
+   * Set the animation to use when showing / hiding the status bar.
+   *
+   * Only available on iOS.
+   *
+   * @since 8.0.0
+   */
+  setAnimation(options: SystemBarsAnimationOptions): Promise<void>;
+}
+
+export class SystemBarsPluginWeb extends WebPlugin implements SystemBarsPlugin {
```
OS-pedrogustavobilro, line 639: Correct me if I'm wrong but this core plugin will become the first (or at least the only one for Capacitor currently) that is only available in native mobile apps, not in web.

While for a lot of people this not being available in web would be obvious, maybe for some it wouldn't be. Wondering if we should make that explicit somewhere in the docs.

ios/Capacitor/Capacitor/Plugins/SystemBars.swift
```diff
@@ -0,0 +1,133 @@
...
+        case light = "LIGHT"
+        case defaultStyle = "DEFAULT"
+    }
+
+    @objc override public func load() {
+        let hidden = getConfig().getBoolean("hidden", false)
+
+        if let style = getConfig().getString("style") {
+            setStyle(style: style)
+        }
```
OS-pedrogustavobilro: Minor: don't think this matters much, but on Android we're using the 'DEFAULT' value to explicitly set the style. Unsure if it matters much, since it should load with a default style, just wanted to mention this since it seems like a minor inconsistency.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,216 @@
...
+                if (bridge != null && bridge.getWebView() != null) {
+                    String script = String.format(
+                        Locale.US,
+                        """
+                        try {
+                          document.documentElement.style.setProperty("--safe-area-inset-top", "%dpx");
+                          document.documentElement.style.setProperty("--safe-area-inset-right", "%dpx");
+                          document.documentElement.style.setProperty("--safe-area-inset-bottom", "%dpx");
+                          document.documentElement.style.setProperty("--safe-area-inset-left", "%dpx");
+                          window.dispatchEvent(new CustomEvent('safeAreaChanged'));
```
OS-pedrogustavobilro: What is this 'safeAreaChanged' event for? Something specific to Android WebView?

ios/Capacitor/Capacitor/Plugins/SystemBars.swift
```diff
@@ -0,0 +1,133 @@
...
+    public let identifier = "CAPSystemBarsPlugin"
+    public let jsName = "SystemBars"
+    public let pluginMethods: [CAPPluginMethod] = [
+        CAPPluginMethod(name: "setStyle", returnType: CAPPluginReturnNone),
+        CAPPluginMethod(name: "setAnimation", returnType: CAPPluginReturnNone),
+        CAPPluginMethod(name: "show", returnType: CAPPluginReturnNone),
+        CAPPluginMethod(name: "hide", returnType: CAPPluginReturnNone)
+    ]
+
+    public var hideHomeIndicator: Bool = false
```
OS-pedrogustavobilro: Don't need this to be set outside of this class right? Or do we? If not, then

```suggestion
    public private(set) var hideHomeIndicator: Bool = false
```

ios/Capacitor/Capacitor/Plugins/SystemBars.swift
```diff
@@ -0,0 +1,133 @@
...
+    @objc func setStyle(_ call: CAPPluginCall) {
+        setStyle(style: call.getString("style") ?? Style.defaultStyle.rawValue)
+        call.resolve()
+    }
+
+    @objc func show(_ call: CAPPluginCall) {
+        let inset = call.getString("inset")?.uppercased()
+
+        if let animation = call.getString("animation") {
+            setAnimation(animation: animation)
```
OS-pedrogustavobilro, line 45: Minor: if we're uppercasing the style, we could do it for animation as well? Here in `show`, in loading config, and `hide`?

OS-pedrogustavobilro (commented):
Haven't fully finished reviewing, but leaving some comments questions, for now mostly minor stuff / documentation

luisbytes:
> I think it makes sense to remove the workaround logic from the keyboard plugin and incorporate it in the capacitor core/this PR. Otherwise all users would have to install the keyboard plugin in order to make it work correctly

@tafelnl  I’ve noticed that with Capacitor v7 (targeting Android API 35), it’s necessary to enable resizeOnFullScreen for the webview resizing to work. However, even with it enabled, when changing between keyboard types while it’s already open (for example, from text to emoji or number), the WebView doesn’t resize correctly, leaving a blank space at the bottom.

In the following example there are 2 inputs, one of type text and another with the inputmode in numeric, as can be seen when switching between the two inputs the keyboard changes height but the height of the webview is not adjusted again.


https://github.com/user-attachments/assets/505ed8d0-9ce9-42a9-adc6-20c8da5751e5

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,216 @@
...
+            if (inset.isEmpty() || inset.equals(INSET_BOTTOM)) {
+                windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.navigationBars());
+            }
+            return;
+        }
+
+        if (inset.isEmpty() || inset.equals(INSET_TOP)) {
+            windowInsetsControllerCompat.show(WindowInsetsCompat.Type.systemBars());
+        }
+        if (inset.isEmpty() || inset.equals(INSET_BOTTOM)) {
```
OS-pedrogustavobilro: Wondering about this possible ambiguity: Unlike on iOS, depending on the phone's orientation, the inset to apply the change isn't necessarily the bottom one (if phone is in landspace, system navigation bar may be in left or right side, not bottom)

While I don't think we need to detect where the navigation bar is / orientation is, this perhaps could be made clearer in the documentation?

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,216 @@
...
+
+    @PluginMethod
+    public void setAnimation(final PluginCall call) {
+        call.resolve();
+    }
+
+    private void setupSafeAreaInsets(boolean hasFixedWebView, boolean hasMetaViewportCover) {
+        View decorView = getActivity().getWindow().getDecorView();
+
+        ViewCompat.setOnApplyWindowInsetsListener(decorView, (v, insets) -> {
```
OS-pedrogustavobilro: You're already aware of this @theproducer, but leaving here for context. It seems insets are being injected in Android < 15, even though the apps aren't edge-to-edge, meaning the actual UI gets left with additional empty space that otherwise wouldn't be there for an existing Capacitor app.

ios/Capacitor/Capacitor/Plugins/SystemBars.swift
OS-pedrogustavobilro: I am seeing app on iOS with light theme and phone with light theme, but status bar has dark setting (light text, assumed to be over dark background, which it isn't), so maybe this does matter?

tafelnl:
> > I think it makes sense to remove the workaround logic from the keyboard plugin and incorporate it in the capacitor core/this PR. Otherwise all users would have to install the keyboard plugin in order to make it work correctly
> 
> @tafelnl I’ve noticed that with Capacitor v7 (targeting Android API 35), it’s necessary to enable resizeOnFullScreen for the webview resizing to work. However, even with it enabled, when changing between keyboard types while it’s already open (for example, from text to emoji or number), the WebView doesn’t resize correctly, leaving a blank space at the bottom.
> 
> In the following example there are 2 inputs, one of type text and another with the inputmode in numeric, as can be seen when switching between the two inputs the keyboard changes height but the height of the webview is not adjusted again.
> 
>  Resize-Error.mp4

Thats exactly one of the reasons I proposed to do it like this instead: https://github.com/ionic-team/capacitor/pull/8180#issuecomment-3502850204. That approach solves your issue

luisbytes:
> Thats exactly one of the reasons I proposed to do it like this instead: [#8180 (comment)](https://github.com/ionic-team/capacitor/pull/8180#issuecomment-3502850204). That approach solves your issue

Okay, that should be the right approach. In my opinion, this behavior should also be in the core of Capacitor and not in the keyboard plugin.

theproducer:
> > I think it makes sense to remove the workaround logic from the keyboard plugin and incorporate it in the capacitor core/this PR. Otherwise all users would have to install the keyboard plugin in order to make it work correctly
> 
> @tafelnl I’ve noticed that with Capacitor v7 (targeting Android API 35), it’s necessary to enable resizeOnFullScreen for the webview resizing to work. However, even with it enabled, when changing between keyboard types while it’s already open (for example, from text to emoji or number), the WebView doesn’t resize correctly, leaving a blank space at the bottom.
> 
> In the following example there are 2 inputs, one of type text and another with the inputmode in numeric, as can be seen when switching between the two inputs the keyboard changes height but the height of the webview is not adjusted again.
> 
>  Resize-Error.mp4

Hello, @luisbytes !  Do you have a minimal reproduction of this example I can take a look at?

luisbytes:
> Hello, @luisbytes ! Do you have a minimal reproduction of this example I can take a look at?

Hi @theproducer
[This is the repo](https://github.com/luisbytes/capacitor-resize) with the example, please read the readme for more details and some screenshots.

eric-horodyski:
> > > I think it makes sense to remove the workaround logic from the keyboard plugin and incorporate it in the capacitor core/this PR. Otherwise all users would have to install the keyboard plugin in order to make it work correctly
> > 
> > 
> > @tafelnl I’ve noticed that with Capacitor v7 (targeting Android API 35), it’s necessary to enable resizeOnFullScreen for the webview resizing to work. However, even with it enabled, when changing between keyboard types while it’s already open (for example, from text to emoji or number), the WebView doesn’t resize correctly, leaving a blank space at the bottom.
> > In the following example there are 2 inputs, one of type text and another with the inputmode in numeric, as can be seen when switching between the two inputs the keyboard changes height but the height of the webview is not adjusted again.
> > Resize-Error.mp4
> 
> Thats exactly one of the reasons I proposed to do it like this instead: [#8180 (comment)](https://github.com/ionic-team/capacitor/pull/8180#issuecomment-3502850204). That approach solves your issue

Hey @tafelnl, thanks for all the discussion here! We're discussing your approach internally, and are considering making it an available option (in addition to the approach we're taking in this PR) in a future version of Capacitor.

ios/Capacitor/Capacitor/Plugins/SystemBars.swift
markemer: It probably should be default on iOS unless it's specifically changed. However, I don't know why "default" is acting funny in that case. It should match the content underneath it.

theproducer: I double checked this and it's working.  @OS-pedrogustavobilro if you are using the test app I sent you over slack, it had a sample plugin config that set the style to `DARK` on load for testing.  If you remove that, it should use the correct defaults.

markemer:
Kicked off an update to see if that would clear the fmt issue, since you clearly ran it as the last step.

tafelnl:
> Hey @tafelnl, thanks for all the discussion here! We're discussing your approach internally, and are considering making it an available option (in addition to the approach we're taking in this PR) in a future version of Capacitor.

Thanks, sounds promising! Which approach are you referring to exactly? I proposed several changes. Among which were the keyboard fix, `adjustMarginsForEdgeToEdge` and `viewport-fit=cover`

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
tafelnl: Found it: https://issues.chromium.org/issues/396827865#comment12

They do mention:

> Please file an [FR](https://b.corp.google.com/issues/new?component=76427&template=1564656) if these workarounds aren't sufficient for you.

So I guess if we properly file a FR, they might even be willing to support this out of the box. If we then also [file a FR](https://github.com/ionic-team/capacitor-plugins/pull/2388#issuecomment-3496623123) for the bug `Keyboard.resizeOnFullScreen` is trying to fix, we would have full safe area support out of the box (except for broken Chromium versions of course).

I would say this is definitely worth a try

tafelnl:
IMO the safe area workaround and styling of system bars are two different feature sets. To me it would make more sense to include the workaround in the core. Then deprecate/remove the `@capacitor/status-bar` plugin in favor of a new `@capacitor/system-bars` (as I think renaming would be justified) plugin and include the styling functionality there. Additionally the keyboard workaround can then be moved into the core and removed from the keyboard plugin

ios/Capacitor/Capacitor/Plugins/SystemBars.swift
OS-pedrogustavobilro: Oh my bad yeah, had not noticed 😅 After setting it to `"DEFAULT"` it's ok now. And also saw you added `"DEFAULT"` as default value to getting the config if nothing is provided, so all good in my book.

Resolving this comment.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
OS-pedrogustavobilro: Basically I'm wondering about the the need for `SystemBarsInsets`, because we're not controlling the style / visibility at specific insets, rather for types of system bars - status bar for android and iOS, navigation/gesture bar for android, and home indicator(?) for iOS.

Would a `SystemBarType` make more sense perhaps, draw less confusion? Let me know your thoughts.

core/src/core-plugins.ts
OS-pedrogustavobilro: @theproducer wondering if you had a chance to look at this, could be just a small doc update making explicit that this is not offered for web, if you think it makes sense.

core/systembars.md
OS-pedrogustavobilro: I see you renamed it to `disableInsets`, but I think the description still mentions "Enable the injection ...". Should be "Disable the injection ..." instead?

Also does this need to be added to `SystemBars` in `cli/src/declarations.ts`?

android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
```diff
@@ -288,7 +288,7 @@ private void deserializeConfig(@Nullable Context context) {
         webContentsDebuggingEnabled = JSONUtils.getBoolean(configJSON, "android.webContentsDebuggingEnabled", isDebug);
         zoomableWebView = JSONUtils.getBoolean(configJSON, "android.zoomEnabled", JSONUtils.getBoolean(configJSON, "zoomEnabled", false));
         resolveServiceWorkerRequests = JSONUtils.getBoolean(configJSON, "android.resolveServiceWorkerRequests", true);
-        adjustMarginsForEdgeToEdge = JSONUtils.getString(configJSON, "android.adjustMarginsForEdgeToEdge", "disable");
+        adjustMarginsForEdgeToEdge = JSONUtils.getString(configJSON, "android.adjustMarginsForEdgeToEdge", "auto");
```
tafelnl, line 291: I think `adjustMarginsForEdgeToEdge` can be removed?

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,209 @@
...
+        String majorVersionStr = matcher.group(0);
+        int majorVersion = Integer.parseInt(majorVersionStr);
+
+        return majorVersion >= 140;
+    }
+
+    private void initSystemBars() {
+        String style = getConfig().getString("style", STYLE_DEFAULT).toUpperCase(Locale.US);
+        boolean hidden = getConfig().getBoolean("hidden", false);
+        boolean disableCSSInsets = getConfig().getBoolean("disableInsets", false);
```
tafelnl, line 68: With a naming like `disableInsets` I would expect that either 1) insets are disabled entirely (similar to `adjustMarginsForEdgeToEdge: force`) or 2) that the injection of custom vars is disabled. Both isn't true. I think it makes sense to include a config var like `insetsCorrectionBehavior` that is an enum with the following values:

- `disabled`: this entirely disables the logic here
- `fallbackToInjection`: this enables the logic and also enables injection for older webviews
- `fallbackToMargins`: this enables the logic and for older webviews falls back to shrinking the webview using margins (IMO this should be the default as discussed before)

Right now this PR makes it quite hard to override this behavior if a developer is - for whatever reason - looking to change the core behavior (for example by installing a plugin that has a different philosophy about this).

On another note: the `disableCSSInsets` is checked inside the `evaluateJavascript` callback. But - with the current logic - it can be lifted to check it before `initSystemBars` is called

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -0,0 +1,209 @@
...
+            }
+        });
+
+        getBridge().executeOnMainThread(() -> {
+            setStyle(style, "");
+            setHidden(hidden, "");
+        });
+    }
+
+    @PluginMethod
```
tafelnl, line 83: The swift equivalents of the plugin methods here are annotated with a `returnType: none`. Would it make sense to change it to `@PluginMethod(returnType = PluginMethod.RETURN_NONE)`?

Issue #6232, opened 2023-01-19
bug: update ios - failed! followed by ERR_UNKNOWN: Path


mayerlench:
### Capacitor Version

```
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 4.6.2
  @capacitor/core: 4.6.2
  @capacitor/android: 4.6.2
  @capacitor/ios: 4.6.2

Installed Dependencies:

  @capacitor/cli: 4.6.2
  @capacitor/ios: 4.6.2
  @capacitor/core: 4.6.2
  @capacitor/android: 4.6.2

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)

IOS

### Current Behavior

1. Using ionic cli to create a new react tabs project
2. running `ionic cap add ios` i get this error
```
> capacitor add ios
[capacitor] ✔ Adding native Xcode project in ios in 14.25ms
[capacitor] ✔ add in 14.57ms
[capacitor] ✔ Copying web assets from build to ios/App/App/public in 20.54ms
[capacitor] ✔ Creating capacitor.config.json in ios/App/App in 232.33μs
[capacitor] ✔ copy ios in 29.47ms
[capacitor] ✔ Updating iOS plugins in 2.45ms
[capacitor] [info] Found 4 Capacitor plugins for ios:
[capacitor]        @capacitor/app@4.1.1
[capacitor]        @capacitor/haptics@4.1.0
[capacitor]        @capacitor/keyboard@4.1.1
[capacitor]        @capacitor/status-bar@4.1.1
[capacitor] ✖ Updating iOS native dependencies with pod install - failed!
[capacitor] ✖ update ios - failed!
[capacitor] [error] Command line invocation:
[capacitor]         /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -project App.xcodeproj clean
[capacitor]
[capacitor]         User defaults from command line:
[capacitor]         IDEPackageSupportUseBuiltinSCM = YES
[capacitor]
[capacitor]         error: Could not delete `/Users/mayerlench/Documents/VSCode/testing/ionic-capacitor-nativescript/ios/App/build` because it was not created by the build system.
[capacitor]         note: To mark this directory as deletable by the build system, run `xattr -w com.apple.xcode.CreatedByBuildSystem true /Users/mayerlench/Documents/VSCode/testing/ionic-capacitor-nativescript/ios/App/build` when it is created.
[capacitor]
[...]

PR #8188, opened 2025-10-21
Fix `update` failing because of "Could not delete [...]/ios/App/build" xcodebuild error

viteinfinite:
This PR fixes xcodebuild failing with `Could not delete [...]/ios/App/build because it was not created by the build system.`.

To solve this we append the scheme to the clean command.


viteinfinite:
This also fixes #6232

eric-horodyski:
Thanks for opening this PR 🙏

Issue #6021, opened 2022-10-24
bug: Serving local videos occasionally fails on Android


brospars:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 4.4.0
  @capacitor/core: 4.4.0
  @capacitor/android: 4.4.0
  @capacitor/ios: 4.4.0

Installed Dependencies:

  @capacitor/cli: 3.5.1
  @capacitor/core: 3.5.1
  @capacitor/android: 3.5.1
  @capacitor/ios: 3.5.1

[success] iOS looking great! 👌
[success] Android looking great! 👌
```

### Platform(s)
Android

### Current Behavior
On iOS : the video plays without a problem.
On Android : the video plays but we get a network error from time to time that stop the video and "crashes" the player (we have to change page and come back to replay it again). Sometimes the video won't play at all.

```
GET http://localhost/_capacitor_file_/data/user/0/com.myapp/files/videoXYZ.mp4 net::ERR_FAILED
```

### Expected Behavior
The video play thoroughly on android.

### Code Reproduction

We use a html5 video player in an Ionic/Angular app to play local video like such:

```html
<video *ngIf="src" #video  disablePictureInPicture playsinline preload="auto" [src]="src"></video>
```

Where `src`is set using an input : `Capacitor.convertFileSrc('video/videoXYZ.mp4')`

```js
@Component({
  selector: 'video-player',
  templateUrl: './video-player.component.html',
  styleUrls: ['./video-player.component.scss'],
})
export class VideoPlayerComponent implements OnInit, AfterViewInit {

    @ViewChild('video', {static: false}) videoRef: ElementRef;

    @Input() src: string;

    // ...
}
```

### Other Technical Details

We can have multiple videos on the same page. And the bug seems to happen more often when we seek through multiple videos quickly.

`npm --version` output: `8.7.0`

`node --version` output: `v14.19.0`

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
Ionitron 💙

brospars:
Just for the sake of it I upgraded to the latest versions of everything on the `v4` branch: https://github.com/brospars/bugvid/tree/v4

[...]

PR #8190, opened 2025-10-22
fix(Android): range requests

bolivir:
The responseHeaders map is shared between threads, causing simultaneous requests to return the same Content-Length to the browser in case of range requests. See the illustration in https://github.com/ionic-team/cordova-plugin-ionic-webview/issues/453 (the same issue is present Capacitor).

This PR proposes to return a copy of the map on each invocation. Based on https://github.com/ionic-team/capacitor/pull/5956

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -129,7 +129,7 @@ public class WebViewLocalServer {
         }
 
         public Map<String, String> getResponseHeaders() {
-            return responseHeaders;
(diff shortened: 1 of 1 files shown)


JeroenBoesten:
Would love to see this merged this should also fix https://github.com/ionic-team/capacitor/issues/6021.

eric-horodyski:
Thanks for opening this PR 🙏

OS-pedrogustavobilro:
Hey @bolivir, we've just merged this more recent PR https://github.com/ionic-team/capacitor/pull/8357 that does a similar fix to yours - So we'll actually be closing this one.

Feel free to ask questions if you have any, and we do appreciate your contribution nonetheless!

Issue #8111, opened 2025-08-15
[Feature]: cap run --target match name for team sharing


alexcroox:
### Description

Specifying a run target ID makes working in a team difficult because everyone's simulators have different IDs

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web

### Request or proposed solution

`npx cap run --target-name 'iPhone 6 (simulator)'`

This would allow it to run on any of my team mates machines as they all have different GUIDs for their simulators, but they all have the same name "iPhone 6 (simulator)"

Thanks!

PR #8199, opened 2025-10-28
feat(cli): Select a cap run target by target name

theproducer:
This PR adds a `--target-name` flag to `cap run ios|android` to select a run target by name instead of its ID.

example:
```
npx cap run ios --target-name "iPhone 17 Pro"
```
Its up to the developer to make sure their emulators / simulators have unique names.

To target emulators / simulators with similar names but different OS versions:
```
npx cap run ios --target-name "iPhone 16 Pro" --target-name-sdk-version "18.0"
```

closes: https://github.com/ionic-team/capacitor/issues/8111

### DIFF
--- a/android/capacitor/build.gradle
+++ b/android/capacitor/build.gradle
@@ -13,7 +13,6 @@ ext {
 
 
 buildscript {
-    ext.kotlin_version = project.hasProperty("kotlin_version") ? rootProject.ext.kotlin_version : '2.2.20'
     repositories {
         google()
         mavenCentral()
@@ -79,7 +78,6 @@ repositories {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation platform("org.jetbrains.kotlin:kotlin-bom:$kotlin_version")
     implementation "androidx.appcompat:appcompat:$androidxAppCompatVersion"
     implementation "androidx.core:core:$androidxCoreVersion"
     implementation "androidx.activity:activity:$androidxActivityVersion"
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -47,6 +47,7 @@ const coreVersion = 'next';
 const pluginVersion = 'next';
 const gradleVersion = '8.14.3';
 const iOSVersion = '15';
+const kotlinVersion = '2.2.20';
 let installFailed = false;
 
 export async function migrateCommand(config: Config, noprompt: boolean, packagemanager: string): Promise<void> {
@@ -489,7 +490,24 @@ async function updateBuildGradle(
       }
     }
   }
+
+  const beforeKotlinVersionUpdate = replaced;
+  replaced = replaceVersion(replaced, /(ext\.kotlin_version\s*=\s*['"])([^'"]+)(['"])/, kotlinVersion);
+  replaced = replaceVersion(replaced, /(org\.jetbrains\.kotlin:kotlin[^:]*:)([\d.]+)(['"])/, kotlinVersion);
+  if (beforeKotlinVersionUpdate !== replaced) {
+    logger.info(`Set Kotlin version to ${kotlinVersion}`);
+  }
   writeFileSync(filename, replaced, 'utf-8');
+}
+
+function replaceVersion(text: string, regex: RegExp, newVersion: string): string {
+  return text.replace(regex, (match, prefix, currentVersion, suffix) => {
+    const semVer = coerce(currentVersion)?.version;
+    if (gte(newVersion, semVer ? semVer : '0.0.0')) {
+      return `${prefix || ''}${newVersion}${suffix || ''}`;
+    }
+    return match;
+  });
 }
 
 async function updateAppBuildGradle(filename: string) {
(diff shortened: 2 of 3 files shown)


cli/src/common.ts
```diff
@@ -405,9 +406,20 @@ export async function promptForPlatformTarget(
   }
 
   const targetID = selectedTarget.trim();
-  const target = targets.find((t) => t.id === targetID);
+  const target = targets.find((t) => {
+    if (selectByName === true) {
+      return t.name === targetID;
+    }
+
+    t.id === targetID;
```
jcesarmobile:
```suggestion
    return t.id === targetID;
```
`t.id === targetID` works when it's a single expression, but since now it's a function it needs the return as you did for `return t.name === targetID;`

cli/src/common.ts
```diff
@@ -405,9 +406,20 @@ export async function promptForPlatformTarget(
   }
 
   const targetID = selectedTarget.trim();
-  const target = targets.find((t) => t.id === targetID);
+  const target = targets.find((t) => {
+    if (selectByName === true) {
```
OS-pedrogustavobilro, line 411: I see you specified in the PR that it is up to the dev to guarantee unique device names. I remember a time where I had simulators with iOS 18 and simulators with iOS 26, with the same name - also GitHub macos runners come with multiple simulators with the same name and different iOS versions - example for [macos-15](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#installed-simulators).

Currently we will just return the first one without checking if there are others. I'm wondering here if that's ok, or if it could have unexpected results for clients, and if  we should do something extra:

- Log to console that more than one was found, picking the one with id X
- Give a prompt to allow to select one of the targets? This becomes an issue if we'd want to run such a command in a CI/CD pipeline.
- Fail the command?
- Other..?

You probably thought about this already @theproducer, just putting the comment here to get your thoughts.

theproducer: > I remember a time where I had simulators with iOS 18 and simulators with iOS 26, with the same name - also GitHub macos runners come with multiple simulators with the same name and different iOS versions - example for [macos-15](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-arm64-Readme.md#installed-simulators).

Yeah, I can see that being a problem on CI.  I think passing a device name AND OS version should help make the selection specific enough.

cli/src/index.ts
```diff
@@ -235,6 +235,7 @@ export function runProgram(config: Config): void {
     .addOption(new Option('--json').hideHelp())
     .option('--target <id>', 'use a specific target')
     .option('--target-name <name>', 'use a specific target by name')
+    .option('--target-name-sdk-version <version>', 'use a specific sdk version when using --target-name')
```
OS-pedrogustavobilro: Wonder if it's relevant providing examples of the SDK version format to use (assume it's the SDK int for Android and regular iOS version for iOS, but maybe worth making explicit)?

cli/src/common.ts
```diff
@@ -405,9 +407,25 @@ export async function promptForPlatformTarget(
...
+      return t.name === targetID;
+    }
+
+    return t.id === targetID;
+  });
 
   if (!target) {
+    if (selectByName) {
+      fatal(
+        `Invalid target name: ${c.input(targetID)}.\n` +
```
jcesarmobile: If passing a SDK version this message should be different
I passed a valid name but and invalid SDK but the message says `Invalid target name: iPhone 16e.`, but the name is correct, what's incorrect is the SDK version, so should be like `Invalid target name: iPhone 16e. [26]`

cli/src/common.ts
```diff
@@ -405,9 +407,25 @@ export async function promptForPlatformTarget(
...
+          `Valid targets are: ${targets.map((t) => `${t.name} [${t.sdkVersion}]`).join(', ')}`,
```
jcesarmobile: It's hard to see the available targets, I think it would be better to separate with `\n` instead of `,`

cli/src/common.ts
```diff
@@ -405,10 +407,35 @@ export async function promptForPlatformTarget(
   }
 
   const targetID = selectedTarget.trim();
-  const target = targets.find((t) => t.id === targetID);
+  const target = targets.find((t) => {
+    if (selectByName === true) {
+      let name = t.name ?? t.model;
+      if (name) {
+        name = name.replace(/[\u2018\u2019]/g, "'").replace(/[\u201C\u201D]/g, '"');
```
OS-pedrogustavobilro, line 416: Minor: Maybe could leave a comment on what propped this line?

cli/src/common.ts
```diff
@@ -405,10 +407,35 @@ export async function promptForPlatformTarget(
...
   if (!target) {
-    fatal(`Invalid target ID: ${c.input(targetID)}.\n` + `Valid targets are: ${targets.map((t) => t.id).join(', ')}`);
+    if (selectByName) {
+      let invalidTargetName = targetID;
+      if (selectedTargetSdkVersion) {
+        invalidTargetName += ` [${selectedTargetSdkVersion}]`;
+      }
+      fatal(
+        `Invalid target name: ${c.input(invalidTargetName)}.\n` +
+          `Valid targets are:\n ${targets.map((t) => `${t.name} [${t.sdkVersion}]`).join('\n')}`,
```
OS-pedrogustavobilro: `t.name` can be undefined here, as it was for the filter, so we could instead do

```suggestion
          `Valid targets are:\n ${targets.map((t) => `${t.name ?? t.model} [${t.sdkVersion}]`).join('\n')}`,
```

alexcroox:
Thank you so much for this! <3

PR #8200, opened 2025-10-28
chore(ci): increase iOS tests timeout

jcesarmobile:
tests are failing a lot lately since we started using Xcode 26 because of timeouts, so I'm increasing the timeout from 30 minutes to 60 hoping it will finish in that time

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,7 +90,7 @@ jobs:
         working-directory: ./core
   test-ios:
     runs-on: macos-15
-    timeout-minutes: 30
+    timeout-minutes: 60
     needs:
       - setup
       - lint


OS-pedrogustavobilro:
How did the time increase by that much? Seeing a couple PRs ago it seemed to be 6 min? There must be something wrong with the the runner or something for it to be timing out in 30 minutes?

jcesarmobile:
They are investigating https://github.com/actions/runner-images/issues/13096

But even increasing to 60 failed, the tests seem to hang, there is a known issue on Xcode 26.1 about that, but we are using 26.0.1, not sure if it's also affected

OS-pedrogustavobilro:
Can temporarily go back down to Xcode 16 if this doesn't get fixed soon?

jcesarmobile:
we are dropping Xcode 16 support, so no

PR #8216, opened 2025-11-06
feat(iOS): Allow plugins to hook into handling WebView URL authentication challenges

theproducer:


cli/src/tasks/copy.ts
```diff
@@ -65,7 +65,11 @@ export async function copy(config: Config, platformName: string, inline = false)
     }
 
     let usesSSLPinning = false;
-    if (allPlugins.filter((plugin) => plugin.id === '@ionic-enterprise/ssl-pinning').length > 0) {
+    if (
+      allPlugins.filter(
+        (plugin) => plugin.id === '@ionic-enterprise/ssl-pinning' || plugin.id === '@capacitor/ssl-pinning',
```
OS-pedrogustavobilro, line 70: Just confirming if this should be in this PR?

ios/Capacitor/Capacitor/CAPPlugin.h
```diff
@@ -38,6 +39,11 @@
  * Returning nil will defer to the default Capacitor policy
  */
 - (NSNumber* _Nullable)shouldOverrideLoad:(WKNavigationAction* _Nonnull)navigationAction;
+/**
+ * Allows plugins to hook into and respond to the WebView's URL authentication challenge.
+ * Returning nil will defer to the default response of [.rejectProtectionSpace](https://developer.apple.com/documentation/Foundation/URLSession/AuthChallengeDisposition/rejectProtectionSpace).
```
OS-pedrogustavobilro: Since it's a default, wondering if [performDefaultHandling](https://developer.apple.com/documentation/foundation/urlsession/authchallengedisposition/performdefaulthandling) should be used instead?

ios/Capacitor/Capacitor/WebViewDelegationHandler.swift
```diff
@@ -161,6 +161,27 @@ open class WebViewDelegationHandler: NSObject, WKNavigationDelegate, WKUIDelegat
...
+        guard let bridge = bridge else {
+            completionHandler(.rejectProtectionSpace, nil)
+            return
+        }
+
+        for pluginObject in bridge.plugins {
+            let plugin = pluginObject.value
+            let selector = NSSelectorFromString("handleWKWebViewURLAuthenticationChallenge:")
+            if plugin.responds(to: selector) {
+                if let response = plugin.handleWKWebViewURLAuthenticationChallenge(challenge) {
```
OS-pedrogustavobilro: In my opinion, it may make more sense to do a different approach, more similar to what [Cordova has done recently](https://github.com/apache/cordova-ios/pull/1514/files#diff-76db62d77b6406dccf3dfcfa07b16cf4e68384fba993112df9c85a4b1bf63297R631).

Instead of expecting a plugin to return the response to the authentication challenge, pass the completion handler, and let the plugin return true or false, as in the challenge was handled or not.

This allows to support plugins that use third-party libraries that receive `completionHandler` (e.g. [Trustkit](https://github.com/datatheorem/TrustKit)) for SSL Pinning.

Let me know your thoughts.

cli/src/tasks/copy.ts
theproducer: I think it's fine for now.  Ideally we move this functionality out of core and into the plugin.

ios/Capacitor/Capacitor/CAPPlugin.h
OS-pedrogustavobilro: I see, it is [stated here](https://developer.apple.com/documentation/webkit/wknavigationdelegate/webview(_:didreceive:completionhandler:)#Discussion), and it seems to basically be the same as `performDefaultHandling`, so perhaps there ain't much difference between the two for authentication challenges in webviews.

I'll resolve this comment.

PR #8217, opened 2025-11-07
chore(cli): Update migrate command for Cap 8

OS-pedrogustavobilro:
## PR description

This PR updates the `migrate.ts` task to Capacitor 8, mostly updating dependency versions, but with a new addition to update app's `build.gradle` - see commits for list of changes.

## Testing

I tested this command on a few capacitor apps (as well as testing the apps after running the command), for instance:

- [Empty App created for testing](https://github.com/OS-pedrogustavobilro/test-migrate-capacitor-app/tree/test/05-cap8-migrate) - Refer to `test/` branches for a few different cases to make sure that the command works.
- [Capacitor test app](https://github.com/ionic-team/capacitor-testapp/tree/test/cap8-migrate-command-incremental) - Commit with changes of migration command, located [here](https://github.com/ionic-team/capacitor-testapp/commit/0091cdd16cb935824241d0591f54c1857171dfba)
- [Capacitor Google Maps](https://github.com/ionic-team/capacitor-google-maps/tree/test/cap8-migrate-app) - While not yet made capacitor compatible, this was more to test if everything still worked on android with [updated gradle dependencies](https://github.com/ionic-team/capacitor-google-maps/commit/6f713ecac466a8ff936fa81cf353a786cb46c8d2#diff-26c9c802fedb9ba02dda78ebbc672afeb5ff51cf6232499b6cc847dcee077c93)

You can test yourself by checking this branch, and mapping the cli version to the local source code directory (make sure you run `build` in cli directory beforehand), then run `npx cap migrate` on the app.

If you want to run a second time in the same app, you'll need to manually change the version for `cli` every time, because the `migrate` command will change the versions of capacitor next. If you're running the migrate command a second time, you can avoid running `npm install` (replying `n` to the second question), to not override the cli version afterwards (`core` and `android` and `ios` should already be using `next` from running migrate the first time, so nothing should change).

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -655,6 +655,7 @@ public class Bridge {
         this.registerPlugin(com.getcapacitor.plugin.CapacitorCookies.class);
         this.registerPlugin(com.getcapacitor.plugin.WebView.class);
         this.registerPlugin(com.getcapacitor.plugin.CapacitorHttp.class);
+        this.registerPlugin(com.getcapacitor.plugin.SystemBars.class);
 
         for (Class<? extends Plugin> pluginClass : this.initialPlugins) {
             this.registerPlugin(pluginClass);
@@ -1610,7 +1611,6 @@ public class Bridge {
 
             if (webView instanceof CapacitorWebView capacitorWebView) {
                 capacitorWebView.setBridge(bridge);
-                capacitorWebView.edgeToEdgeHandler(bridge);
             }
 
             bridge.setCordovaWebView(mockWebView);
--- a/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/CapConfig.java
@@ -54,7 +54,7 @@ public class CapConfig {
     private String errorPath;
     private boolean zoomableWebView = false;
     private boolean resolveServiceWorkerRequests = true;
-    private String adjustMarginsForEdgeToEdge = "disable";
+    private String adjustMarginsForEdgeToEdge = "auto";
 
     // Embedded
     private String startPath;
@@ -288,7 +288,7 @@ public class CapConfig {
         webContentsDebuggingEnabled = JSONUtils.getBoolean(configJSON, "android.webContentsDebuggingEnabled", isDebug);
         zoomableWebView = JSONUtils.getBoolean(configJSON, "android.zoomEnabled", JSONUtils.getBoolean(configJSON, "zoomEnabled", false));
         resolveServiceWorkerRequests = JSONUtils.getBoolean(configJSON, "android.resolveServiceWorkerRequests", true);
-        adjustMarginsForEdgeToEdge = JSONUtils.getString(configJSON, "android.adjustMarginsForEdgeToEdge", "disable");
+        adjustMarginsForEdgeToEdge = JSONUtils.getString(configJSON, "android.adjustMarginsForEdgeToEdge", "auto");
 
         String logBehavior = JSONUtils.getString(
             configJSON,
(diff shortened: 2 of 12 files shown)


alexgerardojacinto (approved):

OS-ruimoreiramendes (approved):

cli/src/tasks/migrate.ts
```diff
@@ -274,18 +278,18 @@ export async function migrateCommand(config: Config, noprompt: boolean, packagem
...
-              androidxCoreKTXVersion: '1.12.0',
-              googleMapsPlayServicesVersion: '18.2.0',
-              googleMapsUtilsVersion: '3.8.2',
-              googleMapsKtxVersion: '5.0.0',
-              googleMapsUtilsKtxVersion: '5.0.0',
-              kotlinxCoroutinesVersion: '1.7.3',
-              coreSplashScreenVersion: '1.0.1',
+              androidxBrowserVersion: '1.9.0',
+              androidxMaterialVersion: '1.13.0',
+              androidxExifInterfaceVersion: '1.4.0',
```
jcesarmobile:
```suggestion
              androidxExifInterfaceVersion: '1.4.1',
```
latest is 1.4.1

OS-pedrogustavobilro:
@jcesarmobile removed the method since all plugins should already be using `namespace` now, otherwise they would not work with Capacitor 7 (or older even).

cli/src/tasks/migrate.ts
```diff
@@ -621,48 +583,13 @@ async function updateAndroidManifest(filename: string) {
     return;
   }
 
-  if (txt.includes('navigation')) {
+  if (txt.includes('|density')) {
     return; // Probably already updated
   }
   const replaced = txt.replace(
-    'android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode"',
     'android:configChanges="orientation|keyboardHidden|keyboard|screenSize|locale|smallestScreenSize|screenLayout|uiMode|navigation"',
```
jcesarmobile: navigation is not assured to be there since people could have updated manually and it was marked as an optional update:
https://capacitorjs.com/docs/updating/7-0#add-navigation-to-configchanges

OS-pedrogustavobilro: Would you recommend then leaving the same replacement command as Cap 7, without `navigation`, or just add it to the beginning of `configChanges`, so that it could also work in cases where people manually added `navigation` to the beginning as opposed to the end (would need to change the if condition above as well but that's np).

jcesarmobile: Maybe just check the replacement string after the replacement and mark the task as failed if `density` is not there because it couldn't append it.

OS-pedrogustavobilro: I can also re-add the navigation append to the migrate command, if it's not guaranteed to be there?

jcesarmobile: yeah, but as you said, it could be at the beginning, it's not guaranteed that `density` will be appended or if some other part of the `configChanges` string will be different and will fail to append `density`, so we must warn if it wasn't appended. While on `navigation` we didn't warn since it was an optional change.

OS-pedrogustavobilro: Made a few changes to that step, and failed the task (which fails the migration command) if `density` was unable to be added -> 8ec0779b8ecf411401e215bbd7b803e7de7493e7

Let me know what tou think!

jcesarmobile: Sorry if I wasn't clear, but I didn't mean to throw an error but to log an error (with `logger.error`) as we do with other steps that fail, the migration should continue even if some steps didn't complete successfully.

jcesarmobile:
Can you remove the `, please check out the Ionic vscode extension for this functionality.` text from the monorepo warning since the Ionic vscode extension is no longer supported?

cli/src/tasks/migrate.ts
OS-pedrogustavobilro: My bad, I found it a bit strange to have the migration step log an error but be successful but yeah if we have other steps like that then wouldn't make sense to have this one different. I'll update the PR

OS-pedrogustavobilro:
> Can you remove the `, please check out the Ionic vscode extension for this functionality`. text from the monorepo warning since the Ionic vscode extension is no longer supported?

@jcesarmobile  So what should be used instead to allow to run this tool in a mono-repo type environment?

jcesarmobile:
I don't know, but not the Ionic extension.
Maybe just do the migration manually.

Issue #8063, opened 2025-07-14
[Bug]: PluginCall.resolve logs in Logcat inconsistently


croraf:
### Capacitor Version

@capacitor/core: 7.4.2
@capacitor/android: 7.4.2

### Other API Details

```Shell

```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

`PluginCall.resolve` logs in Logcat when called with no argument `call.resolve();`

> Capacitor/Console  I  File:  - Line 334 - Msg: undefined

But doesn't log anything if called with an argument:
```
JSObject test = new JSObject();
call.resolve(test);
```

---------------

In the **browser's console** I see log for both of these variants. Also it is confusing that file name in Logcat is empty while in the browser it is named like VM19 or so.

<img width="730" height="147" alt="Image" src="https://github.com/user-attachments/assets/8d459d5d-8e50-4cea-8d33-cd688ef9468a" />

### Expected Behavior

The log should either show or not show for both method overloads.

If shown it would be nice to show the file name also.

### Project Reproduction

[Reproduction repo](https://github.com/Istrability/capacitor-example-bug)  details in README.md

croraf:
Are you kidding me, why you need more information. It is such a plain issue and fully described. What is unclear?

croraf:
Added reproduction repo

jcesarmobile:
The name of the file not being shown is because the log comes from Capacitor's code, which gets injected into the WebView, so it's not treated as a regular file, but like a "VM", same as when you type code into Chrome's web debugger/console.
As you can see, a log in your screenshot is correctly shown for the checkLocationPermission message.

But we are hiding messages that are just `[object Object]`, which we probably shouldn't as it also hides user messages that try to log an object.

PR #8225, opened 2025-11-10
fix: make Plugin.resolve act consistently

markemer:
Closes: #8063

I've been using https://github.com/Istrability/capacitor-example-bug to test as it shows the bug on normal cap.

### DIFF
--- a/cli/src/util/native-run.ts
+++ b/cli/src/util/native-run.ts
@@ -16,7 +16,11 @@ export async function runNativeRun(args: readonly string[], options: RunCommandO
     fatal(`${c.input('native-run')} not found.`);
   }
 
-  return await runCommand(p, args, options);
+  if (process.versions.pnp) {
+    return await runCommand('yarn', ['node', p, ...args], options);
+  } else {
+    return await runCommand(p, args, options);
+  }
 }
 
 export async function getPlatformTargets(platformName: string): Promise<PlatformTarget[]> {


core/native-bridge.ts
```diff
@@ -330,7 +330,7 @@ const initBridge = (w: any): void => {
         if (result.success === false) {
           c.error(result.error);
         } else {
-          c.dir(result.data);
+          c.dir(JSON.stringify(result.data));
```
OS-pedrogustavobilro, line 333: Question: Can we expect the data to always be a json here?

markemer: Edit: I was thinking about this backwards. As long as that is a valid Javascript Value it should output a JSON string.

~~Yeah, probably not, and then it may just swallow that, or worse throw an error. I'll add some checking.~~

eric-horodyski: I believe `Plugin.resolve()` will fail if you don't pass it something that isn't JSON-encodable.

markemer: I think so too, now that I think about it. Because it needs to send it back as JSON.

jcesarmobile:
Good fix, I didn't think about this when I looked into the problem.
But the issue I mentioned on https://github.com/ionic-team/capacitor/issues/8063 about `console.log('[object Object]');` not being shown on logcat should also be addressed, should be as simple as removing the `msg.equalsIgnoreCase("[object Object]")` of `isValidMsg` method inside `BridgeWebChromeClient`.

markemer:
Cool, will do.

PR #8236, opened 2025-11-13
fix(android)!: Adding keyboard insets handling

theproducer:
This PR adds insets handling for open keyboards, and removes the no longer needed `adjustMarginsForEdgeToEdge` config option.

(Thanks to @tafelnl for the suggestions!)

### DIFF
--- a/android/capacitor/build.gradle
+++ b/android/capacitor/build.gradle
@@ -13,7 +13,6 @@ ext {
 
 
 buildscript {
-    ext.kotlin_version = project.hasProperty("kotlin_version") ? rootProject.ext.kotlin_version : '2.2.20'
     repositories {
         google()
         mavenCentral()
@@ -79,7 +78,6 @@ repositories {
 
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
-    implementation platform("org.jetbrains.kotlin:kotlin-bom:$kotlin_version")
     implementation "androidx.appcompat:appcompat:$androidxAppCompatVersion"
     implementation "androidx.core:core:$androidxCoreVersion"
     implementation "androidx.activity:activity:$androidxActivityVersion"
--- a/cli/src/android/run.ts
+++ b/cli/src/android/run.ts
@@ -12,9 +12,20 @@ const debug = Debug('capacitor:android:run');
 
 export async function runAndroid(
   config: Config,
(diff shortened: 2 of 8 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -123,7 +123,17 @@ private void setupSafeAreaInsets(boolean hasFixedWebView, boolean hasMetaViewpor
...
-            injectSafeAreaCSS(safeArea.top, safeArea.right, safeArea.bottom, safeArea.left);
+            Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
+            boolean keyboardVisible = insets.isVisible(WindowInsetsCompat.Type.ime());
+
+            int bottomInsets = safeArea.bottom;
+
+            if (keyboardVisible) {
+                // Remove when https://issues.chromium.org/issues/457682720 is fixed
+                bottomInsets = imeInsets.bottom - bottomInsets;
+            }
```
tafelnl, line 135: I'm not sure if this is something that should be removed when that issue is fixed, as that would effectively break it again for users with an older webview

OS-pedrogustavobilro: True, should instead be to review this logic once a fix gets shipped in a future webview version. @theproducer

cli/src/declarations.ts
```diff
@@ -297,17 +297,6 @@ export interface CapacitorConfig {
...
-    /**
-     * If set to "force", margins will be adjusted for edge to edge regardless of any other settings.
-     * If set to "auto", will check for Android 15 and the setting of [windowOptOutEdgeToEdgeEnforcement](https://developer.android.com/reference/android/R.attr#windowOptOutEdgeToEdgeEnforcement) and will adjust margins if on Android 15 and windowOptOutEdgeToEdgeEnforcement is false/missing.
-     * If set to "disable", or is missing, will not adjust margins at all.
-     * In Capacitor 8, this default will be changed to 'auto'
-     *
-     * @since 7.1.0
-     * @default disable
-     */
-    adjustMarginsForEdgeToEdge?: 'auto' | 'force' | 'disable';
```
OS-pedrogustavobilro, line 310: Just a note: This is a breaking change, that we'll have to document right?

PR #8238, opened 2025-11-14
chore(cli): Migrate task updates kotlin version in root build.gradle

OS-pedrogustavobilro:
## PR description

This PR updates the `migrate.ts` task to detect if the project's root `build.gradle` file has Kotlin defined (either via `kotlin_version` variable or set in the kotlin gradle plugin), and if so, update it to match the Kotlin version that Capacitor 8 uses (2.2.0).

## Testing

Here's a few examples of the command running (it was run before to make the other Cap 8 changes, so you'll only see the kotlin version change in diff):

- [Project with kotlin_version variable](https://github.com/OS-pedrogustavobilro/test-migrate-capacitor-app/commit/7bd3867ad6f837c68040ec989940dbd01924fb6f)
- [Project with version explicitly in gradle plugin](https://github.com/OS-pedrogustavobilro/test-migrate-capacitor-app/commit/02f17a0b75799b3392587a8e40724aaadff86007)

You can test yourself by checking this branch, and mapping the cli version to the local source code directory (make sure you run `build` in cli directory beforehand), then run `npx cap migrate` on the app.

If you want to run a second time in the same app, you'll need to manually change the version for `cli` every time, because the `migrate` command will change the versions of capacitor next. If you're running the migrate command a second time, you can avoid running `npm install` (replying `n` to the second question), to not override the cli version afterwards (`core` and `android` and `ios` should already be using `next` from running migrate the first time, so nothing should change).

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -491,12 +491,23 @@ async function updateBuildGradle(
     }
   }
 
-  if (replaced.includes('kotlin_version') || replaced.includes('org.jetbrains.kotlin:kotlin')) {
-    replaced = replaced.replace(/(ext\.kotlin_version\s*=\s*['"])([^'"]+)(['"])/, `$1${kotlinVersion}$3`);
-    replaced = replaced.replace(/(org\.jetbrains\.kotlin:kotlin-gradle-plugin:)([\d.]+)/, `$1${kotlinVersion}`);
+  const beforeKotlinVersionUpdate = replaced;
+  replaced = replaceVersion(replaced, /(ext\.kotlin_version\s*=\s*['"])([^'"]+)(['"])/, kotlinVersion);
+  replaced = replaceVersion(replaced, /(org\.jetbrains\.kotlin:kotlin[^:]*:)([\d.]+)(['"])/, kotlinVersion);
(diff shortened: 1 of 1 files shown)


cli/src/tasks/migrate.ts
```diff
@@ -489,6 +489,13 @@ async function updateBuildGradle(
       }
     }
   }
+
+  if (replaced.includes('kotlin_version') || replaced.includes('org.jetbrains.kotlin:kotlin')) {
+    const kotlinVersionString = '2.2.20';
```
jcesarmobile: put the version on top of the file where `gradleVersion`, `coreVersion`, etc. are so it's easier to spot for future updates

OS-pedrogustavobilro: Done in 4945f01

OS-pedrogustavobilro:
> Would be good to check the existing value so we don't downgrade it if set to 2.2.21 or greater

@jcesarmobile Done in [43860fc](https://github.com/ionic-team/capacitor/pull/8238/commits/43860fc7f9be5469a03b8363020a1ead78216bfb). I extracted some logic to a separate method, basically to replace `kotlin_version` and/or (most likely just or, but in theory could be both) gradle plugin.

Let me know what you think

PR #8249, opened 2025-11-19
fix(ios): remove percent-encoding of cookie values in CapacitorCookies

### LINKED ISSUE #8248: [Bug]: CapacitorCookies percent-encodes cookies on iOS (and not on Android)
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.4.4
  @capacitor/core: 7.4.4
  @capacitor/android: 7.4.4
  @capacitor/ios: 7.4.4

Installed Dependencies:

  @capacitor/cli: 7.4.2
  @capacitor/core: 7.4.2
  @capacitor/android: 7.4.2
  @capacitor/ios: 7.4.2

[info] Using Gemfile: RubyGems bundle installed

Doesn't really matter though because it's in current main branch too.

### Other API Details

```Shell

```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

When calling [`CapacitorCookies.setCookie()`](https://github.com/michaelwolz/capacitor/blob/d834600e639d0d3610b00586feb86c2549dd2494/ios/Capacitor/Capacitor/Plugins/CapacitorCookies.swift#L26-L36) the iOS implementation of this method percent-encodes the Cookie value via `encode` method [CapacitorCookieManager](https://github.com/michaelwolz/capacitor/blob/d834600e639d0d3610b00586feb86c2549dd2494/ios/Capacitor/Capacitor/Plugins/CapacitorCookieManager.swift#L49-L51). The problem is that requests using these cookies now include the encoded cookie value, which causes issues because servers typically do not decode cookie values. Although [RFC 6265 (see cookie-octet)](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1) permits certain special characters such as colons, these characters are being incorrectly encoded by the plugin.

In addition to that the [Android implementation](https://github.com/michaelwolz/capacitor/blob/d834600e639d0d3610b00586feb86c2549dd2494/android/capacitor/src/main/java/com/getcapacitor/plugin/CapacitorCookieManager.java#L152) of the CapacitorCookies plugin does not encode the values which causes an inconsistent behavior between iOS and Android.

**Example**: 

A cookie such as `key=chunks:2` becomes `key=chunks%3A2` in the http request and will not be accepted by the server.

### Expected Behavior

Cookie values should not be percent-encoded.

### Project Reproduction

Any usage of `CapacitorCookies.setCookie({ url: 'xxx', key: 'foo', value: 'bar:baz' })` along with CapacitorHttp being active to set the cookies in the request will reproduce the issue. However, network traffic needs to be intercepted or sever response to be checked to see it.

Minimal example can be found here: https://github.com/michaelwolz/capacitor-cookies (Replace [`WEBHOOK_URL`](https://github.com/michaelwolz/capacitor-cookies/blob/90c0b6889adbbdf2ea650b4736edbe0ae5888955/src/js/capacitor-welcome.js#L4) with your own for testing)

**Screenshot of a request sent on iOS:**

Image

**Same request on Android:**

Image

### Additional Information

CapacitorHttp must be active to make the request.

michaelwolz:
Remove the percent-encoding of cookie values in CapacitorCookies plugin implementation on iOS and therefore make it behave the same as on Android. 

Fixes: https://github.com/ionic-team/capacitor/issues/8248

Working example can be found here (build locally with non published packages): https://github.com/michaelwolz/capacitor-cookies/tree/working-version

**Screenshot**

ios-working


ios/Capacitor/Capacitor/Plugins/CapacitorCookieManager.swift
```diff
@@ -46,14 +46,6 @@ public class CapacitorCookieManager {
         return url
     }
 
-    public func encode(_ value: String) -> String {
-        return value.addingPercentEncoding(withAllowedCharacters: .urlPathAllowed)!
-    }
-
-    public func decode(_ value: String) -> String {
-        return value.removingPercentEncoding!
-    }
```
michaelwolz: This seems to not have been in use at all.

PR #8250, opened 2025-11-20
fix(cli): make migrate update to 8.0.0

jcesarmobile:


jcesarmobile:
We enable redirects from next to the path without next on capacitor-docs repository, so links don't need to get changed.
The links were already like that when you updated the migrate command.

OS-pedrogustavobilro:
> The links were already like that when you updated the migrate command.

Ah true, my bad I had forgotten 😅

Issue #8251, opened 2025-11-21
[Bug]: iOS Prerequisite Checks & Cordova Package Generation Race


MOHITKOURAV01:
### Capacitor Version

npx cap doctor
<paste your actual output here>

### Other API Details

```Shell
**npm --version output:**
<version here>

**node --version output:**
<version here>

**pod --version output (iOS CocoaPods issues only):**
<version here>
```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

Running the following commands:

npx cap add ios
npx cap update ios
npx cap doctor ios

causes inconsistent behavior on iOS.  
The CLI sometimes skips prerequisite checks and proceeds before Cordova package files are fully generated.  
This leads to intermittent build failures, missing Swift files, and incomplete iOS setup.

### Expected Behavior

Both Bundler and CocoaPods checks should always run during:

cap add ios

cap update ios

cap doctor ios

The CLI should wait for all Cordova Package `.swift` files to finish generating before continuing the update flow.

### Project Reproduction

A minimal reproduction repository:   <https://github.com/MOHITKOURAV01/capacitor.git>  Steps: 1. Clone the repo 2. Run `npm install` 3. Run `npx cap add ios` 4. Run `npx cap update ios` 5. Observe missing checks / race condition

### Additional Information

The issue appears more frequently on slow machines or when multiple plugins trigger Cordova Swift generation.  
Let me know if logs or screen recordings are required—happy to provide more details!

eric-horodyski:
Hi @MOHITKOURAV01 — thanks for the issue and fix submission! 🙌

When you get a moment, could you include the output from `npx cap doctor`? It gives us a helpful snapshot of your project setup before we dig into the reproduction.

We have planned work around the Cordova compatibility layer that we’ll be looking at after the Capacitor 8 GA launch. I’ve added this issue to the items we’ll be reviewing as part of that effort. Appreciate your patience in the meantime!

MOHITKOURAV01:
@eric-horodyski , It would look like this
 <img width="306" height="281" alt="Image" src="https://github.com/user-attachments/assets/2f5aff1c-9179-498c-8647-817229e7ac7d" />

MOHITKOURAV01:
@eric-horodyski  
As requested, I’ve attached a screenshot of my npx cap doctor output in the earlier comment. The iOS setup looks perfect and matches the latest dependencies.
Let me know if you need any other details or logs!
[...]

PR #8252, opened 2025-11-21
Fix iOS CLI checks and Cordova Package.swift generation race

MOHITKOURAV01:
### Related Issue
#8251

### Expected Behavior

1. **Ensure `checkBundler` and `checkCocoaPods` always run independently** during:
   - `cap add ios`
   - `cap update ios`
   - `cap doctor ios`  
   This guarantees that missing or misconfigured CocoaPods installations are detected reliably.

2. **Wait for all `Package.swift` files to finish writing** when generating Cordova SPM packages during:
   - `cap update ios`  
   This prevents intermittent issues where some Swift packages fail to register because files were not fully written before the next step executes.


MOHITKOURAV01:
hii @eric-horodyski !! Please review the latest changes and let me know if anything else is required. If the fixes look good, I’d appreciate your approval on this PR.

PR #8263, opened 2025-11-28
Add jsProfilingEnabled option for WebView responses

riderx:
Introduce an Android configuration option to enable JavaScript profiling in WebView responses by adding the `Document-Policy: js-profiling` HTTP header. This change allows users to leverage the JS Self-Profiling API within the WebView.


riderx:
This was reported in reddit here https://www.reddit.com/r/capacitor/comments/1p25gq5/comment/nqp0q5r/?context=1

PR #8264, opened 2025-11-28
fix(ios): move PrivacyInfo.xcprivacy to resource_bundles to fix build…

### LINKED ISSUE #8259: [Bug]: PrivacyInfo.xcprivacy should be in resources_bundle
### Capacitor Version

Capacitor comes as a transitive dependency of https://github.com/ionic-team/ionic-portals-react-native

```
COCOAPODS: 1.16.2
➜  ios git:(master) ✗ cat Podfile.lock | grep Capacitor
  - Capacitor (7.0.1):
    - CapacitorCordova
  - CapacitorCordova (7.4.4)
    - Capacitor (~> 7.0.0)
    - Capacitor
    - CapacitorCordova
  Capacitor: 7b6faa2a912c47313a231c76363a863cea0c6123
  CapacitorCordova: 1a2fefa999c426f71c4bb5cd7a787ffc6fd4ec3s 
```

### Other API Details

```Shell
cat package.json | grep ionic
    "@ionic/portals-react-native": "^0.8.3",


The app is using Expo framework and React Native's `privacy_file_aggregation_enabled` feature.
```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

After adding portals dependency (`npx expo install @ionic/portals-react-native`), the application build starts failing with following error.

```
❌  error: Multiple commands produce '/Users/<user>/Library/Developer/Xcode/DerivedData/<app-name>-bygnknwtcxmavsasyyxjtmjwoich/Build/Products/Debug-iphoneos/<app-name>.app/PrivacyInfo.xcprivacy'
```

The reason for this is that capacitor (3rd party dependency) is shipping its `PrivacyInfo.xcprivacy` manifest in `resources` instead of `resource_bundles`. Capacitor copies it's manifest into main bundle root and React Native can no longer create final aggregated manifest during the build.

These are two problematic places:
- https://github.com/ionic-team/capacitor/blob/main/ios/Capacitor.podspec#L20
- https://github.com/ionic-team/capacitor/blob/main/ios/CapacitorCordova.podspec#L23

## Some more context from Grok

`s.resources`: Files copied directly into the app's main bundle root. PrivacyInfo.xcprivacy lands at top level → conflicts with Expo's aggregation → error.
`s.resource_bundles`: Files placed inside a separate .bundle (here Capacitor.bundle). Privacy manifest goes into the bundle → CocoaPods/Expos privacy aggregation can find and merge it correctly → no error.

Use resource_bundles is required for privacy manifest aggregation in Xcode 15+/iOS 17+. resources no longer work

### Expected Behavior

I expect `@ionic/portals-react-native` to work with Expo framework.

### Project Reproduction

-

### Additional Information

_No response_

mrgear111:
This PR fixes a build issue where `PrivacyInfo.xcprivacy` conflicts with other libraries (like when using Expo or React Native) because it was being copied to the main bundle root.

## The Problem
When `PrivacyInfo.xcprivacy` is in `s.resources`, it gets copied to the top-level app bundle. If another library does the same thing, or if the build system (like Expo) tries to aggregate privacy manifests, it causes a "Multiple commands produce..." error.

## The Fix
I've moved `PrivacyInfo.xcprivacy` to `s.resource_bundles` in both [Capacitor.podspec](cci:7://file:///Users/dakshsaini/Desktop/OS/capacitor/ios/Capacitor.podspec:0:0-0:0) and [CapacitorCordova.podspec](cci:7://file:///Users/dakshsaini/Desktop/OS/capacitor/ios/CapacitorCordova.podspec:0:0-0:0). This places the manifest inside its own bundle (e.g., `Capacitor.bundle`), which namespaces it correctly and allows Xcode/CocoaPods to handle the aggregation without conflicts.

This is the recommended approach for shipping privacy manifests in libraries.

## Related Issue
Fixes #8259

## Verification
- Verified that `PrivacyInfo.xcprivacy` is now defined under `resource_bundles` in the podspecs.
- This matches the standard fix for this class of errors.

Thanks!

### DIFF
--- a/cli/src/common.ts
+++ b/cli/src/common.ts
@@ -537,6 +537,9 @@ export async function checkJDKMajorVersion(): Promise<number> {
 }
 
 export function parseApkNameFromFlavor(flavor: string): string {
-  const convertedName = flavor.replace(/([A-Z])/g, '$1').toLowerCase();
+  let convertedName = flavor.replace(/([A-Z])/g, '-$1').toLowerCase();
+
+  if (convertedName.startsWith('-')) convertedName = convertedName.replace('-', '');
+
   return `app-${convertedName ? `${convertedName}-` : ''}debug.apk`;
 }


mrgear111:
@andredestro @markemer Can we merge this now?

PR #7556, opened 2024-07-09
fix(ios): Match CapApp-SPM iOS version with project version

jcesarmobile:
Make the CapApp-SPM project use the configured iOS version (`IPHONEOS_DEPLOYMENT_TARGET`) on the App project instead of the hardcoded version 13.

closes https://github.com/ionic-team/capacitor/issues/7544

Issue #8276, opened 2025-12-05
[Bug]: V8 min version 15.5


riderx:
### Capacitor Version

I was preparing the update of our plugin to v8 in the beta branch and found out the minimum version for iOS is 15

In iOS 15, version 15.5 was released with a lot of critical security fixes: 
https://support.apple.com/en-us/120323

As an example, zlib in v15 has a path traversal security issue (https://nvd.nist.gov/vuln/detail/cve-2018-25032).

I do believe 15.5 would make a lot of sense to ensure the minimum version brings security on top of new features

### Other API Details

```Shell

```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

Min version is min is using major base

### Expected Behavior

Min version is defined as version with the most import security patches. 

### Project Reproduction

NONE

eric-horodyski:
Hey @riderx, thanks for the suggestion. 

The choice of iOS 15 (rather than 15.5) was intentional and driven primarily by customer and enterprise support requirements. Many Capacitor apps ship into managed or long-lived environments where OS upgrades lag behind Apple’s release cadence.

riderx:
Thanks a lot for the answer, how are we supposed to then select version ourselves in SPM ?
Right now if we put 15.5 in project file it get replace by 15 in spm file

riderx:
we decided to stop using the dependency that was creating the issue in our plugin.
but i would have love our Pr to be merged instead of PR be stale and not review for no reasons ...
I do know you are trying to improve on that and i hope you will but for now even if you closed a lot of old issues new one are not  addressed @eric-horodyski

PR #8284, opened 2025-12-17
feat: allow spm platforms to select os without being overwritten

WcaleNieWolny:
## Problem

When using Swift Package Manager (SPM), the Capacitor CLI currently defaults to iOS 15.0 as the minimum deployment target, regardless of what is configured in Xcode build settings. This causes issues when plugins require a higher minimum iOS version, particularly for security reasons.

### Related Work

[PR #7556](https://github.com/ionic-team/capacitor/pull/7556) attempted to address this issue by reading the iOS deployment target from the Xcode project configuration. However, it did not fully solve the problem because:

1. **It only reads the major version** - The implementation reads only 2 characters from the deployment target, which means it cannot handle minor versions like `15.5` (it would only read `15`)
2. **It uses the wrong format** - It generates `.iOS(.v15)` format, which doesn't support dot versions. According to Apple's documentation, versions with minor numbers (like 15.5) must use the `.iOS("15.5")` string format instead

This PR builds upon that work by properly reading the full version (including minor versions) and using the correct Swift Package Manager format.

### Security Impact

This is particularly critical for plugins that depend on libraries with security requirements. For example, [ZipArchive requires iOS 15.5+](https://github.com/ZipArchive/ZipArchive?tab=readme-ov-file#version-250-updates-minimum-os-versions) due to a security vulnerability (CVE-2018-25032) in zlib versions before 1.2.12. zlib 1.2.12 is only included in iOS 15.5+ and later versions.

When Capacitor CLI generates a `Package.swift` file with `.iOS(.v15)` instead of `.iOS("15.5")`, it prevents proper enforcement of the security requirement, potentially exposing apps to known vulnerabilities. Furthermore, the project will simply not build if the minimum iOS version is not met.

## Solution

This PR fixes the SPM platform version generation to:

1. **Read the full iOS deployment target** (including minor versions like 15.5) from the Xcode project configuration
2. **Use the correct Swift Package Manager format** based on Apple's documentation:
   - `.iOS("15.5")` when a dot version is detected (required for versions like 15.5)
   - `.iOS(.v15)` when no dot version is detected (for major versions only)

This ensures that the generated `Package.swift` accurately reflects the iOS deployment target configured in Xcode, allowing plugins with security requirements (like ZipArchive) to properly enforce their minimum iOS version requirements.

## Changes

- Added `getMajorMinoriOSVersion()` function to read the full iOS deployment target (including minor versions) from `project.pbxproj`
- Updated SPM `Package.swift` generation to use the correct platform version format based on whether a dot version is present
- Fixed iOS enum usage in platform version formatting

## Documentation

- [Apple's PackageDescription documentation](https://developer.apple.com/documentation/packagedescription/supportedplatform/ios(_:)-83bbf) - explains the `.iOS("15.5")` format requirement
- [ZipArchive security requirements](https://github.com/ZipArchive/ZipArchive?tab=readme-ov-file#version-250-updates-minimum-os-versions) - explains why iOS 15.5+ is required for security

## Availability

This fix is already available in our fork [Cap-go/capacitor-plus](https://github.com/Cap-go/capacitor-plus) and can be installed via npm:

```bash
npm install @capacitor-plus/cli
```

## Testing

Tested manually with a Capacitor app configured for iOS 15.5 deployment target. The generated `Package.swift` now correctly uses `.iOS("15.5")` format.

### DIFF
--- a/cli/src/ios/common.ts
+++ b/cli/src/ios/common.ts
@@ -125,13 +125,14 @@ export function getMajorMinoriOSVersion(config: Config): string {
   // Extract until semicolon or newline (typical end of value in pbxproj)
   const endIndex = pbx.indexOf(';', valueStart);
   const newlineIndex = pbx.indexOf('\n', valueStart);
-  const actualEnd = endIndex !== -1 && newlineIndex !== -1 
-    ? Math.min(endIndex, newlineIndex)
-    : endIndex !== -1 
-      ? endIndex 
-      : newlineIndex !== -1 
-        ? newlineIndex 
-        : pbx.length;
+  const actualEnd =
+    endIndex !== -1 && newlineIndex !== -1
+      ? Math.min(endIndex, newlineIndex)
+      : endIndex !== -1
+        ? endIndex
+        : newlineIndex !== -1
+          ? newlineIndex
+          : pbx.length;
   let iosVersion = pbx.substring(valueStart, actualEnd).trim();
   // Remove trailing .0 if present
   if (iosVersion.endsWith('.0')) {
--- a/cli/src/ios/common.ts
+++ b/cli/src/ios/common.ts
@@ -113,3 +113,29 @@ export function getMajoriOSVersion(config: Config): string {
   );
   return iosVersion;
 }
+
+export function getMajorMinoriOSVersion(config: Config): string {
+  const pbx = readFileSync(join(config.ios.nativeXcodeProjDirAbs, 'project.pbxproj'), 'utf-8');
+  const searchString = 'IPHONEOS_DEPLOYMENT_TARGET = ';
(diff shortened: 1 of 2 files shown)


eric-horodyski:
This Pull Request addresses the following Issue: https://github.com/ionic-team/capacitor/issues/8276.

eric-horodyski:
Hi @WcaleNieWolny - does this Pull Request address the scenario where there are multiple `IPHONEOS_DEPLOYMENT_TARGET` values?

riderx:

Patch ID: `upstream-pr-8284-cli`

```ts
plugins: {
  CapacitorPatch: {
    patches: ['upstream-pr-8284-cli'],
    strict: true,
  },
}
```

Issue #8294, opened 2026-01-06
[Bug]: SystemBars style gets reset when orientation changes


pomm0:
### Capacitor Version

```
💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 8.0.0
  @capacitor/core: 8.0.0
  @capacitor/android: 8.0.0
  @capacitor/ios: 8.0.0

Installed Dependencies:

  @capacitor/cli: not installed
  @capacitor/core: 8.0.0
  @capacitor/android: 8.0.0
  @capacitor/ios: 8.0.0

[error] Xcode is not installed
[success] Android looking great! 👌
```

### Other API Details

```Shell
npm: 10.9.3
node: v22.18.0
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

When setting the SystemBars style to LIGHT (probably anything else then DEFAULT)

```
// capacitor.config.json

"plugins": {
    "SystemBars": {
      "style": "DARK"
    }
  },
```

it boots with the proper style, but when changing the orientation to landscape for instace, the style is DEFAULT/LIGHT again.

https://github.com/user-attachments/assets/8cab1f9f-3d67-44f8-95a6-89932e7c22b9

### Expected Behavior

The style remains DARK as defined

### Project Reproduction

https://github.com/pomm0/capacitor-systembars-style-reset-issue

### Additional Information

Current workaround I'm using:

```
const styleStatusBar = async () => {
  await this.SystemBars.setStyle({
    style: SystemBarsStyle.Dark,
    bar: SystemBarType.StatusBar,
  });
};

void styleStatusBar();

await this.ScreenOrientation.addListener('screenOrientationChange', () => void styleStatusBar());
```

This is also an issue when setting the systemBars style via JS only, without the capacitor.config.json.

Not sure if iOS has this issue as well, cannot test it at the moment.

PR #8295, opened 2026-01-06
fix(SystemBars): get correct style on handleOnConfigurationChanged

theproducer:
fixes: https://github.com/ionic-team/capacitor/issues/8294

### DIFF
--- a/core/src/core-plugins.ts
+++ b/core/src/core-plugins.ts
@@ -617,14 +617,14 @@ export interface SystemBarsPlugin {
    *
    * @since 8.0.0
    */
-  show(options: SystemBarsVisibilityOptions): Promise<void>;
+  show(options?: SystemBarsVisibilityOptions): Promise<void>;
 
   /**
    * Hide the system bars.
    *
    * @since 8.0.0
    */
-  hide(options: SystemBarsVisibilityOptions): Promise<void>;
+  hide(options?: SystemBarsVisibilityOptions): Promise<void>;
 
   /**
    * Set the animation to use when showing / hiding the status bar.
--- a/core/system-bars.md
+++ b/core/system-bars.md
@@ -39,7 +39,7 @@ To control this behavior, use the `insetsHandling` configuration setting.
 ## Example
 
 ```typescript
-import { SystemBars, SystemBarsStyle } from '@capacitor/core';
+import { SystemBars, SystemBarsStyle, SystemBarType } from '@capacitor/core';
 
 const setSystemBarStyleDark = async () => {
   await SystemBars.setStyle({ style: SystemBarsStyle.Dark });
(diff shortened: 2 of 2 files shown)


andredestro (approved):

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -268,4 +268,22 @@ private String getStyleForTheme() {
...
     }
+
+    private String getStyleConfig() {
+        String style = getConfig().getString("style");
+
+        if (style == null) {
+            style = STYLE_DEFAULT;
+        }
+
+        return style;
```
jcesarmobile: Why are you changing the `getConfig().getString("style", STYLE_DEFAULT)` with all this?
This is how the `getConfig().getString("style", STYLE_DEFAULT)` is supposed to work internally, if `getConfig().getString("style")` is null it would return the `STYLE_DEFAULT` as it's the default value passed.

Is it not working as expected?

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -268,4 +268,22 @@ private String getStyleForTheme() {
...
+    }
+
+    private boolean getHiddenConfig() {
+        if (getConfig().getConfigJSON().has("hidden")) {
+            return getConfig().getBoolean("hidden", false);
+        }
+
+        return false;
```
jcesarmobile: Same here, why are you replacing the `getConfig().getBoolean("hidden", false)` with this? 

It's supposed to return `false` if the config doesn't include `hidden`, so you shouldn't need to check.
If it's not working like that it should be reported and fixed.
If it's working like that, not sure why all this extra code is being added.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
theproducer: No, sorry, this was just debug stuff that didn't get fully removed.

jcesarmobile:
You removed the `getStyleConfig()` method, so the code no longer compiles, maybe make it `getConfig().getString("style", STYLE_DEFAULT)`

jcesarmobile:
The only concern about this approach is, what if the user change the style programmatically? i.e. called `SystemBars.setStyle({ style: SystemBarsStyle.Dark });` but doesn't have it configured?
This will "reset" the value to default and change the appearance automatically.
Should `setStyle` plugin call save the latest style in a variable instead and use that value on the `handleOnConfigurationChanged` instead of reading the config value?
I think that's what status-bar plugin does.

Issue #8298, opened 2026-01-08
[Bug]: SystemBars no instructions how to install


croraf:
### Capacitor Version

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 8.0.0
  @capacitor/core: 8.0.0
  @capacitor/android: 8.0.0

### Other API Details

```Shell

```

### Platforms Affected

- [x] iOS
- [x] Android
- [x] Web

### Current Behavior

In https://capacitorjs.com/docs/apis/system-bars there are no instructions on how to install SystemBars plugin, contrary to all other API pages.

Does it require install, what is the plugin name on npm repository, etc... is unclear.

### Expected Behavior

Improve documentation for this API.

### Project Reproduction

a

OS-pedrogustavobilro:
Hey @croraf, SystemBars plugin is part of a set of Core Plugins bundled with `@capacitor/core`, so you don't have to install anything else to use this plugin.

I think the docs pages is missing a note that it is bundled with `@capacitor/core`, I'll open a quick PR to mention that.

Hope that clarifies it.

theproducer:
Resolved with https://github.com/ionic-team/capacitor/pull/8300

PR #8300, opened 2026-01-08
docs(SystemBars): Mention it's bundled with `@capacitor/core`

OS-pedrogustavobilro:
Addresses https://github.com/ionic-team/capacitor/issues/8298

### DIFF
--- a/cli/package.json
+++ b/cli/package.json
@@ -52,7 +52,7 @@
     "env-paths": "^2.2.0",
     "fs-extra": "^11.2.0",
     "kleur": "^4.1.5",
-    "native-run": "^2.0.2",
+    "native-run": "^2.0.3",
     "open": "^8.4.0",
     "plist": "^3.1.0",
     "prompts": "^2.4.2",


OS-pedrogustavobilro:
@jcesarmobile good point, I instead left the "This plugin is bundled with `@capacitor/core`" in the first paragraph, and moved mentions / distinctions from Status Bar plugin to a second paragraph, think it looks better like this.

jcesarmobile:
Yeah, that's what I meant by move "this" and then quoted the Status Bar text, move the Status Bar part to a separate paragraph, not the "This plugin is bundled with @capacitor/core".

OS-pedrogustavobilro:
Oh okay xD I misunderstood your suggestion 😅 but well in the end went with what you suggested

croraf:
I opened this issue. Please give me few hours to check this change and leave a comment.

croraf:
It is still confusing why this is a called a plugin at all.

Also https://capacitorjs.com/docs/plugins doesn't mention built in plugins conceptually, a plugin is something that can be plugged in, perhaps call this an option.

It seems this is the first such case in plugins list on the above plugins docs page, that is every other plugin is in a separate npm package.

--------------

So this plugin is enabled immediately, whenever `@capacitor/core` is installed and `npx cap sync` is run, no need to define it in plugins?

OS-pedrogustavobilro:
> a plugin is something that can be plugged in, perhaps call this an option.

I do suppose the naming of plugin usually doesn't mean bundled with Core Framework. But in that doc page you shared it says "Plugins in Capacitor enable JavaScript to interface directly with Native APIs.", and SystemBars is included in that category.

> It seems this is the first such case in plugins list on the above plugins docs page

It is not the first case though, [cookies](https://capacitorjs.com/docs/apis/cookies) and [http](https://capacitorjs.com/docs/apis/http) are also bundled with `@capacitor/core`, they always have been. Historically I believe all plugins as well were bundled with `@capacitor/core`, back in Capacitor 1 and 2. So for Capacitor this isn't a new concept that's being introduced.

> So this plugin is enabled immediately, whenever @capacitor/core is installed and npx cap sync is run, no need to define it in plugins?

Yes you don't need to install a separate package. You import it from `@capacitor/core`. See [the example](https://capacitorjs.com/docs/apis/system-bars). You won't see it listed in `cap sync` output, because it's part of Capacitor itself.

croraf:
Isn't the entire point of Capacitor to interface the native app and web app?

Or perhaps (as I'm reading now about Cordova and Capacitor differences) a major part (besides the interfacing  is also about configuring the native apps (ios and android) through the configuration file. Which then is not "plugins" part of Capacitor?

--------------

It would be cleaner to have all plugins as separate npm packages in my opinion, but as this would be a breaking change, perhaps then mention in https://capacitorjs.com/docs/plugins that some plugins come prebundled in `@capacitor/core` package and don't require separate installation.

-------------

> Yes you don't need to install a separate package. You import it from @capacitor/core. See [the example](https://capacitorjs.com/docs/apis/system-bars). You won't see it listed in cap sync output, because it's part of Capacitor itself.

Ok, I think I just need the css variables, not the javascript methods, and as per documentation they seem to be injected immediately without needing to do anything but to install `@capacitor/core@8.0' (and do the sync).

OS-pedrogustavobilro:
> perhaps then mention in https://capacitorjs.com/docs/plugins 

This we can do, but it will be in a separate PR, that will go on https://github.com/ionic-team/capacitor-docs

PR #8304, opened 2026-01-12
fix(http): Fix throwing data URI requests on iOS

frederikbosch:
Since Capacitor 8.0 data: requests throw because NSURLResponse is not a HTTPURLRepsonse

### DIFF
--- a/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
+++ b/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
@@ -238,7 +238,13 @@ open class HttpRequestHandler {
                 var output = [:] as [String: Any]
                 output["status"] = 200
                 output["headers"] = headers
-                output["data"] = String(data: data!, encoding: .utf8)
+
+                if let data = data {
+                    output["data"] = String(data: data, encoding: .utf8)
+                } else {
+                    output["data"] = ""
+                }
+
                 call.resolve(output)
             } else {
                 call.reject("Unknown response kind")


ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,21 @@ open class HttpRequestHandler {
...
+                call.resolve(self.buildResponse(data, response as! HTTPURLResponse, responseType: type))
+            } else if response != nil {
+                var headers = [:] as [String: Any]
+                headers["Content-Type"] = response?.mimeType
+                
+                var output = [:] as [String: Any]
+                output["status"] = 200
+                output["headers"] = headers
+                output["data"] = String(data: data!, encoding: .utf8)
+                call.resolve(output);
```
de-dan: Reuse self.buildResponse to keep unified return type

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,21 @@ open class HttpRequestHandler {
...
+                let type = ResponseType(rawValue: responseType) ?? .default
+                call.resolve(self.buildResponse(data, response as! HTTPURLResponse, responseType: type))
+            } else if response != nil {
+                var headers = [:] as [String: Any]
+                headers["Content-Type"] = response?.mimeType
+                
+                var output = [:] as [String: Any]
+                output["status"] = 200
+                output["headers"] = headers
+                output["data"] = String(data: data!, encoding: .utf8)
```
de-dan: Encoding format should be extracted from header and fallback to utf8, no?

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,21 @@ open class HttpRequestHandler {
...
-            call.resolve(self.buildResponse(data, response as! HTTPURLResponse, responseType: type))
+                                                          
+            if response is HTTPURLResponse {
+                setCookiesFromResponse(response as! HTTPURLResponse, config)
+    
+                let type = ResponseType(rawValue: responseType) ?? .default
+                call.resolve(self.buildResponse(data, response as! HTTPURLResponse, responseType: type))
+            } else if response != nil {
+                var headers = [:] as [String: Any]
+                headers["Content-Type"] = response?.mimeType
```
de-dan: Is the content-type the only extractable header from the response?

frederikbosch: There are no headers, look at

https://developer.apple.com/documentation/foundation/urlresponse?language=objc

When requesting `data:`  the response is `NSURLResponse` which does not have headers. We might be able to extract Content Length via [.expectedContentLength](https://developer.apple.com/documentation/foundation/urlresponse/expectedcontentlength?language=objc) but I think it is risky since there is also [NSURLResponseUnknownLength](https://developer.apple.com/documentation/foundation/nsurlresponseunknownlength?language=objc)

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
frederikbosch: There are no headers, look at

https://developer.apple.com/documentation/foundation/urlresponse?language=objc

When requesting `data:` the response is `NSURLResponse` which does not have headers

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
frederikbosch: This is not possible since the response is of the type `NSURLReponse`. I did however test for `NSURLResponse` explicitly now.

https://developer.apple.com/documentation/foundation/urlresponse?language=objc

theproducer:
Thanks @frederikbosch!  Do you have a reproduction for this issue, or can it be seen in basically any minimal project that uses CapacitorHttp?

frederikbosch:
@theproducer I don't at the moment. I have upgraded my app from 7 to 8  and I got crashes because of the forced casts. It might very well be that because of some other upgrades that I only now started with `data:` URI requests.

That data URIs fail can be proven quite easily by [this SwiftFiddle](https://swiftfiddle.com/?c=H4sIAAAAAAAAA21Sy27CMBD8la1PjhSlILVVlQpx6BMJqqqEnnIxZAGLxI5sA60Q%2F961IVCgucSandkZ73rDJixlsqq1cfCil6oQTmqVqwvoHd1am4VUs5zKJTpYmhI6MPrsc%2BsM4SnkjLgidfjtrutSSPUwFhbvbuLh65cdv953chZdNXKLlajn2iA1eZK2Fm4yHzYYX4lyiSm0Ik8njyFaSykSOxcGi8T7ZMIu%2BFq6eeqjRLABj8Zg0NZaWYwBjdEGJHUA%2BmoK6XjOPvd1cD81pjmLw4HraXqQRtGppGfhLcs%2BKEcjDrqGDvKiHnL7FnIK%2FrYHaud43OwYf40GvcHzMVfDTCpZYUYodLs0ZCVLGuSl%2BPm7xonDggzVjMZy0gL3xUetHCrXD5R9k%2B1ZWD9HChp%2B%2F4R8wokuyCYs25sMw%2F554fHwEHaLEDaFUfZyn1gsp9G512H%2FiZUzJUpOhG1CeZcV8jC%2BI2MtpOMRi9mKXutt0m4lbbb9Bee7gnq8AgAA).

Moreover, it is kind of documented behavior. The [`URLSession.dataTask` method documentation](https://developer.apple.com/documentation/foundation/urlsession/datatask(with:completionhandler:)-52wk8) says:

> If you are making an HTTP or HTTPS request, the returned object is actually an [HTTPURLResponse](https://developer.apple.com/documentation/foundation/httpurlresponse) object.

And a `data:` URI request is not a HTTP request. But documentation is not clear what the response is if the request is not a HTTP(S) request, but apparently it is a `NSURLResponse`.

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,26 @@ open class HttpRequestHandler {
                 call.reject(error.localizedDescription, (error as NSError).domain, error, nil)
                 return
             }
-
-            setCookiesFromResponse(response as! HTTPURLResponse, config)
-
-            let type = ResponseType(rawValue: responseType) ?? .default
-            call.resolve(self.buildResponse(data, response as! HTTPURLResponse, responseType: type))
+                                                          
+            if response is HTTPURLResponse {
```
markemer: You can make this `if let response = response as? HTTPURLResponse` and then all the `response as! HTTPURLResponse` is unneeded.

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,26 @@ open class HttpRequestHandler {
...
+                setCookiesFromResponse(response as! HTTPURLResponse, config)
+    
+                let type = ResponseType(rawValue: responseType) ?? .default
+                call.resolve(self.buildResponse(data, response as! HTTPURLResponse, responseType: type))
+            } else if response is NSURLResponse {
```
markemer: As above: You can make this `else if let response = response as? NSURLResponse` and then response will no longer be optional

frederikbosch:
@markemer Much better to unpack the response optional, I have updated the PR.

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,26 @@ open class HttpRequestHandler {
...
+                call.resolve(self.buildResponse(data, response, responseType: type))
+            } else if let response = response as? NSURLResponse {
+                // applicable to data: URI requests
+                var headers = [:] as [String: Any]
+                headers["Content-Type"] = response.mimeType
+                
+                var output = [:] as [String: Any]
+                output["status"] = 200
+                output["headers"] = headers
+                output["data"] = String(data: data!, encoding: .utf8)
```
theproducer: Just for safety purposes I would get rid of the `data!`:
```swift
if let data = data {
    output["data"] = String(data: data, encoding: .utf8)
} else {
    output["data"] = nil
}   
```

frederikbosch: Improved the PR. I did not assign `nil` to `output["data"]` but the empty string. In other places inside HttpRequestHandler.swift [it is also chosen to return the empty string](https://github.com/ionic-team/capacitor/blob/0969c5cd0b16cf23d2489a85a3b8fa1bee2ebf3b/ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift#L156).

ios/Capacitor/Capacitor/Plugins/HttpRequestHandler.swift
```diff
@@ -224,11 +224,32 @@ open class HttpRequestHandler {
                 call.reject(error.localizedDescription, (error as NSError).domain, error, nil)
                 return
             }
+                                                          
+            if let response = response as? HTTPURLResponse {
+                setCookiesFromResponse(response, config)
+    
+                let type = ResponseType(rawValue: responseType) ?? .default
+                call.resolve(self.buildResponse(data, response, responseType: type))
+            } else if let response = response as? NSURLResponse {
```
theproducer, line 233: You may need to rename `NSURLResponse` to `URLResponse` here as well.
Screenshot 2026-01-21 at 10 58 18 AM


Screenshot 2026-01-21 at 10 56 25 AM

theproducer (changes requested):

jcesarmobile:
Can you give a bit more information about what are you doing to get the crash?

Are you getting it while doing regular `fetch`/`XHR` calls with http plugin enabled?
Or are you calling the `CapacitorHttp` plugin API directly like `CapacitorHttp.get({url: 'data:text/plain;base64,SGVsbG8='})`?

If you are doing `fetch`/`XHR` calls and you get the crash, could you provide a sample app that reproduces the issue?
I've not been able to get the crash while doing that.

If you are doing `CapacitorHttp` plugin API calls to data urls, while it shouldn't crash, it shouldn't work neither. `CapacitorHttp` plugin is to avoid CORS problems, and all methods description include "make a Request **to a server**", so should only be used with `http`/`https` urls, not with `data` urls.
I've tested this and it makes the plugin crash on iOS, but on Android it returns an error saying `data` urls are not supported, iOS should also do this instead of crashing.

frederikbosch:
@jcesarmobile your comment has let me to investigate why this data uri call ends up in the native layer.

- in my app the following code `fetch(new URL('./file.css'))`
- is compiled by Vite as `fetch(new URL('data:text/css,body{background:red}'))`
- so, while I am not intentionally fetching a data URI, it is compiled as such, because the resource is so small that Vite modifies it to [an inline asset](https://vite.dev/guide/assets#importing-asset-as-url)

As you state that `CapacitorHttp` should not be used to for data URIs I have created PR #8347 to solve this issue differently. Once that one is merged, I will change this one to fail non HTTPURLResponse responses.

frederikbosch:
Fixed via #8386 and #8387

Issue #8289, opened 2025-12-28
[Bug]: App launches with persistent area equal to keyboard height


bcampru:
### Capacitor Version

Latest Dependencies:

  @capacitor/cli: 8.0.0
  @capacitor/core: 8.0.0
  @capacitor/android: 8.0.0
  @capacitor/ios: 8.0.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 8.0.0
  @capacitor/android: 8.0.0
  @capacitor/core: 8.0.0

[success] Android looking great! 👌

### Other API Details

```Shell
npm: 10.9.4
node: v22.21.1
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

- The app opens with a gray space at the bottom (roughly the keyboard height), and the content can scroll into that empty area.
- The gap persists even after opening/closing the keyboard; only fully closing and reopening the app clears it.
- Reproducible when launching the app from the app drawer while the keyboard is open (e.g., after typing in the app search).
- Observed on physical devices Pixel 8 and Pixel 8a running Android 16 (SDK 36).

https://github.com/user-attachments/assets/72d5c168-a6e1-4213-8541-bc92dbefd7a7

### Expected Behavior

- The WebView should render at full height when the keyboard is not visible.
- No persistent gray space or extra scroll area should appear after launch.

### Project Reproduction

https://github.com/bcampru/capacitor-scroll-space-bug/tree/main

### Additional Information

- Devices: Pixel 8 / Pixel 8a (physical), Android 16.
- Device props: `ro.build.version.sdk = 36`, `ro.product.cpu.abi = arm64-v8a`.
- Debug overlay snapshots when the issue happens show `visualViewport.height` reduced by ~310px while `innerHeight` and `document/body` heights remain full, and `keyboardState.visible = false` (suggesting stale IME insets despite no keyboard visible).
- Things we tried without success:
  - `windowSoftInputMode=adjustNothing` in AndroidManifest.
  - Forcing layout recalculation in JS (visualViewport/listeners).
  - CapacitorUpdater reload / JS reload.
  - Native fix in `MainActivity`: hide IME, toggle `SOFT_INPUT_ADJUST_NOTHING` → `ADJUST_RESIZE`, `requestApplyInsets()`, `requestLayout()`,

bcampru:
[...]

Issue #8214, opened 2025-11-06
[Bug]: Extra scrollable space below the fold exactly the height of the keyboard on Android 16 devices.


pattahgobhi:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.4.4
  @capacitor/core: 7.4.4
  @capacitor/android: 7.4.4
  @capacitor/ios: 7.4.4

Installed Dependencies:

  @capacitor/cli: 7.4.4
  @capacitor/core: 7.4.4
  @capacitor/android: 7.4.4
  @capacitor/ios: 7.4.4

### Other API Details

```Shell
npm --version : 10.9.2
node --version : v22.16.0
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

https://github.com/user-attachments/assets/2d3bc868-4d0c-4529-a62a-0ca9087edfcf

On Android 16 devices, there is an extra scroll space below the fold exactly the height of the keyboard when it opens.
I have tried: 
- adding android:windowSoftInputMode="adjustPan" to the main activity
- Installed "@capacitor/keyboard": "^7.0.3", and added this to capacitor config
 "Keyboard": {
      "resize": "none",
      "resizeOnFullScreen": false
    }

### Expected Behavior

I want no scrolling at all, even when the keyboard opens i do not want the webview to resize.

### Project Reproduction

//

ish1416:
@pattahgobhi I am interested in contributing and raising a pr!

ish1416:
pr #8219

bcampru:
Hello, any news on this issue? We are having the same problem with capacitor v8. Here is a project reproduction of the issue: https://github.com/bcampru/capacitor-scroll-space-bug/tree/main

We have only been able to replicate this bug with a physical phone on a google pixel 8 and a google pixel 8a, the bug arises when opening the app from the drawer and with the keyboard opened as shown in the video on this issue: https://github.com/ionic-team/capacitor-keyboard/issues/30.

The physical phone details are the following:

- ro.build.version.sdk: 36
- ro.product.cpu.abi: arm64-v8a

Let us know if you need more details.

Regards,
Biel

jules-gd-dev:
There is still the issue on Capacitor 8, and the mentionned pull request doesn't seems to fix the issue

mortenbo:
We have this issue too on physical Pixel 6 + 10. Using Capacitor 8.

[...]

PR #8309, opened 2026-01-19
fix(SystemBars): Skipping margin manipulation when on a fixed WebView

theproducer:
fixes: https://github.com/ionic-team/capacitor/issues/8289
fixes: https://github.com/ionic-team/capacitor/issues/8214

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -167,7 +167,7 @@ public class SystemBars extends Plugin {
     private void initWindowInsetsListener() {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM && insetHandlingEnabled) {
             ViewCompat.setOnApplyWindowInsetsListener((View) getBridge().getWebView().getParent(), (v, insets) -> {
-                boolean hasBrokenWebViewVersion = getWebViewMajorVersion() <= 130;
+                boolean hasBrokenWebViewVersion = getWebViewMajorVersion() <= 139;
 
                 if (hasViewportCover) {
                     Insets safeAreaInsets = calcSafeAreaInsets(insets);


praffn:
We've experienced the same issues as described in #8289, and this PR solves the problem

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -162,21 +167,25 @@ private void initSafeAreaInsets() {
     private void initWindowInsetsListener() {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM && insetHandlingEnabled) {
             ViewCompat.setOnApplyWindowInsetsListener((View) getBridge().getWebView().getParent(), (v, insets) -> {
+                boolean hasBrokenWebViewVersion = getWebViewMajorVersion() <= 130;
```
OS-pedrogustavobilro: What's broken on those lower webview versions? At first I thought it was [the same as this](https://github.com/ionic-team/capacitor/pull/8180/files#diff-272dad96b556082d3b777981c428581d8af440964cecbf1839934320904c3ea9R62), but it was set as 140 before, not 130.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
OS-pedrogustavobilro, line 1: Does this fix https://github.com/ionic-team/capacitor/issues/8214?

The issue says:

> ### Expected Behavior
> I want no scrolling at all, even when the keyboard opens i do not want the webview to resize.

jurienhamaker:
This PR also seems to solve our issue, tested on a pixel pro 10 & a nothing phone 2, both on android 16.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
theproducer: Yeah, that is supposed to be 139, not 130, thanks!

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
theproducer: Yes, if they happen to be on a fixed web view, we don't need to apply the margin manipulations we are applying to try to fix the insets issue.  At least for Capacitor 8.

OS-pedrogustavobilro: Got it, makes sense! I'll resolve this comment then.

markemer (approved):
Gave it a quick smoke test, looks good.

zarko-tg:
While this may have solved the original problem, 8.0.2 seems to be breaking our (@ionic/angular 6) app in multiple ways and places when testing on a Pixel 9 (Android 16). I cannot provide a full systematic description atm. due to the magnitude of the issue, but the VK popping up results in for ex.:
- A whole page overscrolling up including the header. The footer stays put / anchored.
- A page being covered on the bottom - the footer get covered by the VK.
- Other / some pages are unaffected.

OS-pedrogustavobilro:
Hey @zarko-tg, I can't say for certain, but would it be possible for you to try and update ionic dependencies to the latest version? I think that they include edge-to-edge fixes and features that might not be available in version 6.

If you find that you're still having those problems, please open new issue(s) with more information for us to take a look.

zarko-tg:
> Hey @zarko-tg, I can't say for certain, but would it be possible for you to try and update ionic dependencies to the latest version? I think that they include edge-to-edge fixes and features that might not be available in version 6.
> 
> If you find that you're still having those problems, please open new issue(s) with more information for us to take a look.

Hi, and thanks for the follow-up.
The latest v6 @ionic/angular seems to be last touched in 2023, so I doubt it has gotten any recent edge-to-edge fixes. Atm. I don't know if the problem originates in or it's a specific case where Capacitor 8 no longer "plays" well with it.
I guess I could run some independent tests soon of the latest @ionic/angular (8) with the latest Capacitor just to compare.

aeharding:
Is anyone else experiencing an issue after upgrading to capacitor v8.0.2 where the main page content always scrolls (doesn't resize) whenever keyboard is open? This is consistently reproducible for me on system webview v143.0.7499.146.0 and android v16. Latest version of Ionic & other dependencies.

Issue goes away downgrading to 8.0.0. I haven't tried 8.0.1 yet.

Happy to make an issue if needed.


https://github.com/user-attachments/assets/022cadf4-3012-4828-ad44-8da2ba939b32

zarko-tg:
> > Hey @zarko-tg, I can't say for certain, but would it be possible for you to try and update ionic dependencies to the latest version? I think that they include edge-to-edge fixes and features that might not be available in version 6.
> > If you find that you're still having those problems, please open new issue(s) with more information for us to take a look.
> 
> Hi, and thanks for the follow-up. The latest v6 @ionic/angular seems to be last touched in 2023, so I doubt it has gotten any recent edge-to-edge fixes. Atm. I don't know if the problem originates in or it's a specific case where Capacitor 8 no longer "plays" well with it. I guess I could run some independent tests soon of the latest @ionic/angular (8) with the latest Capacitor just to compare.

@OS-pedrogustavobilro Just for the record, we had to add
    Keyboard: {
      resizeOnFullScreen: true, // Android only (applies to versions >= 15)
    },
to capacitor.config.js, which did it for calculating the correct height of the app re. VK.

Issue #8310, opened 2026-01-19
[Bug]: Capacitor CLI 7.x is incompatible with tar@7.5.3


joaranda:
### Capacitor Version

Capacitor CLI 7.x is incompatible with tar@7.5.3, causing complete build failure with error "Cannot read properties of undefined (reading 'extract')" at template.js:9.

This blocks users from applying the security patch for CVE-2026-23745 (path traversal in tar ≤7.5.2).

Platforms Affected

 iOS
 Android

Reproduction:
1. Force tar@7.5.3 in package.json overrides
2. Delete node_modules and package-lock.json
3. npm install && npx cap sync ios
4. Build fails

### Other API Details

```Shell

```

### Platforms Affected

- [x] iOS
- [x] Android
- [ ] Web

### Current Behavior

Current Behavior
When tar@7.5.3 is installed (either directly or via transitive dependency update), running npx cap sync fails with:
✖ update ios - failed!
[error] TypeError: Cannot read properties of undefined (reading 'extract')
        at extractTemplate
        (/node_modules/@capacitor/cli/dist/util/template.js:9:25)

### Expected Behavior

Force tar@7.5.3 in package.json overrides

### Project Reproduction

Force tar@7.5.3 in package.json overrides

jcesarmobile:
fixed in #8313

mathieucaroff:
fixed in [#8312](https://github.com/ionic-team/capacitor/pull/8312) as well.

You target 7.x while I target main. I don't know what the Ionic teams prefers.

jcesarmobile:
mine is a cherry pick of another one already merged into main and which was opened before yours and also fixed the problem on the root package.json and some scripts files

main is for Capacitor 8, 7.x is for Capacitor 7 and this issue was about Capacitor 7

prajapatijay95:
The issue has been fixed in #8313 and merged into the 7.x branch.  Is there a new 7.x package with the fix?

andrui:
@jcesarmobile when are you going to release new build for 7.x branch with this fix?

PR #8313, opened 2026-01-20
fix(cli): Update tar package (#8311)

jcesarmobile:
closes https://github.com/ionic-team/capacitor/issues/8310


prajapatijay95:
The issue has been fixed in #8313  and merged into the 7.x branch. Is there a new 7.x package with the fix?

Issue #8314, opened 2026-01-20
[Bug]: New projects fail to build with Android Studio 2025.2.3


bdoms:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 8.0.1
  @capacitor/core: 8.0.1
  @capacitor/android: 8.0.1
  @capacitor/ios: 8.0.1

Installed Dependencies:

  @capacitor/android: 8.0.1
  @capacitor/core: 8.0.1
  @capacitor/ios: 8.0.1
  @capacitor/cli: 8.0.1

[success] Android looking great! 👌

### Other API Details

```Shell
npm --version: 11.6.2
node --version: v25.3.0
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

Assuming you have a SPA you want to convert to an Android app with a fresh install of Android Studio 2025.2.3 (released January 15th, 2026) on Ubuntu using Capacitor 8.0.1:

```bash
npx cap init
npx cap add android
npx cap sync
```

Open Android Studio.

Run suggested upgrade steps, namely "Upgrade AGP dependency from 8.13.0 to 9.0.0".

Attempt to build the project. It fails with this error:

```
Build file 'android/app/build.gradle' line: 22

A problem occurred evaluating project ':app'.
> `getDefaultProguardFile('proguard-android.txt')` is no longer supported since it includes `-dontoptimize`, which prevents R8 from performing many optimizations. Instead use `getDefaultProguardFile('proguard-android-optimize.txt)`, and if needed, temporarily use `-dontoptimize` in a custom keep rule file while fixing breakages.
```

### Expected Behavior

I expect it to build without failing.

### Project Reproduction

Existing projects may or may not have this issue. But taking any project that doesn't have Capacitor yet and adding it always fails for me.

### Additional Information

A workaround is to listen to the error and change the line in `android/app/build.gradle` to be:

```
proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
```

OS-pedrogustavobilro:
@bdoms thanks for opening this.

Any particular reason why you are doing the update "Upgrade AGP dependency from 8.13.0 to 9.0.0" - Do you need a particular feature from 9.0 for a plugin?

[...]

PR #8315, opened 2026-01-20
fix(android): AGP 9.0 no longer supports `proguard-android.txt`

terencehonles:
This change replaces the default proguard file `proguard-android.txt` with `proguard-android-optimize.txt` which allows proguard optimizations.

Alternatively, AGP 9 compatibility can be achieved by opting out of the change with `android.r8.globalOptionsInConsumerRules.disallowed=false`

fixes: https://github.com/ionic-team/capacitor/issues/8314

### DIFF
--- a/android/capacitor/build.gradle
+++ b/android/capacitor/build.gradle
@@ -53,7 +53,7 @@ android {
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
     lintOptions {


theproducer:
Thanks!  This should fix https://github.com/ionic-team/capacitor/issues/8314

terencehonles:
@theproducer I also opened https://github.com/ionic-team/capacitor-plugins/pull/2468 for the plugins in that repo if you have the ability to kick off the CI.

OS-pedrogustavobilro:
Do we want to merge this now? I would imagine we'd officially add support for a new AGP major version in a Capacitor Plugin major version. I don't know if there's precedent on this.

terencehonles:
> Do we want to merge this now? I would imagine we'd officially add support for a new AGP major version in a Capacitor Plugin major version. I don't know if there's precedent on this.

I would think it should be merged in a minor version since it doesn't _require_ an AGP upgrade, and without this change it is blocking AGP upgrades from downstream apps. The effect of this particular change is that r8 may perform optimizations it wasn't previously able to perform, but my thought would be if there are any optimizations that shouldn't be performed then the library's proguard rules should be updated so that it doesn't cause any breaking changes.

I'm not sure what those might be for `@capacitor/core` as it already has [a proguard file](https://github.com/ionic-team/capacitor/blob/main/android/capacitor/proguard-rules.pro), and I would hope those are up to date.

android-template/app/build.gradle
```diff
@@ -19,7 +19,7 @@ android {
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
```
theproducer, line 22: @terencehonles

What I would do for now is remove this change on the template from this PR and open another one just for android-template so that we can ship that in a separate release (possibly with other AGP 9.0 changes).

We are updating our guidance for now to tell people to upgrade to 8.13.0 in the meanwhile: ionic-team/capacitor-docs#515

terencehonles: That's true... I forgot I updated the template file, and that doesn't need to update (and *should* go in a major update)

terencehonles: done: https://github.com/ionic-team/capacitor/pull/8318

ShadeRF:
@capacitor/app still needs to be updated. Error is still present when using this package.

terencehonles:
@ShadeRF that's in https://github.com/ionic-team/capacitor-plugins/pull/2468

PR #8316, opened 2026-01-20
fix(ios): attach temp window to active scene

simplysparsh:
## Summary
- use an active `UIWindowScene` when creating the temp presentation window on iOS 13+
- fall back to `UIWindow(frame:)` when no scene is available

### Reason for patch: 

iOS 13+ requires modal presentation from a UIWindow attached to an active UIWindowScene. Capacitor’s presentVC used UIWindow(frame:), which isn’t scene‑attached, so SFSafariViewController (via @capacitor/browser) could fail to appear. We switched the temp window to use the foreground UIWindowScene when available, with a fallback to UIWindow(frame:) for older iOS or if no scene is active.

### Fix
We updated `CapacitorBridge.presentVC` to:

- Look up the **foreground active** `UIWindowScene` on iOS 13+
- Create the temp window with `UIWindow(windowScene:)`
- Fall back to `UIWindow(frame:)` when no scene is available (or on iOS < 13)

This keeps the presentation path correct for multi‑scene iOS and stabilizes Safari-based flows.

### Scope
The change is in **`@capacitor/ios`** core (not a plugin), but it benefits any plugin that presents view controllers via the bridge, especially `@capacitor/browser`.

### DIFF
--- a/ios/Capacitor/Capacitor/CapacitorBridge.swift
+++ b/ios/Capacitor/Capacitor/CapacitorBridge.swift
@@ -750,21 +750,13 @@ open class CapacitorBridge: NSObject, CAPBridgeProtocol {
         if viewControllerToPresent.modalPresentationStyle == .popover {
             self.viewController?.present(viewControllerToPresent, animated: flag, completion: completion)
         } else {
-            if #available(iOS 13.0, *) {
-                let windowScene = UIApplication.shared.connectedScenes
+            let windowScene = UIApplication.shared.connectedScenes
+                .compactMap { $0 as? UIWindowScene }
+                .first(where: { $0.activationState == .foregroundActive })
+                ?? UIApplication.shared.connectedScenes
                     .compactMap { $0 as? UIWindowScene }
-                    .first(where: { $0.activationState == .foregroundActive })
-                    ?? UIApplication.shared.connectedScenes
-                        .compactMap { $0 as? UIWindowScene }
-                        .first
-                if let windowScene = windowScene {
-                    self.tmpWindow = UIWindow(windowScene: windowScene)
-                } else {
-                    self.tmpWindow = UIWindow(frame: UIScreen.main.bounds)
-                }
-            } else {
-                self.tmpWindow = UIWindow(frame: UIScreen.main.bounds)
-            }
+                    .first
+            self.tmpWindow = windowScene.map { UIWindow(windowScene: $0) } ?? UIWindow(frame: UIScreen.main.bounds)
             self.tmpWindow?.rootViewController = TmpViewController.init()
             self.tmpWindow?.makeKeyAndVisible()
             self.tmpWindow?.rootViewController?.present(viewControllerToPresent, animated: flag, completion: completion)


simplysparsh:
> I don't think the fallback is necessary, since the minimum iOS version is v15 with Capacitor 8.

Capacitor 8’s minimum iOS is 15, but the fallback isn’t about older iOS versions—it’s about scene availability. Even on iOS 15, connectedScenes can be empty or not foregroundActive during app transitions, so creating a UIWindow(windowScene:) can fail. The fallback keeps presentation working in those edge states. It’s a defensive guard with no downside and aligns with Apple’s guidance to present from an active scene when available

@de-dan

simplysparsh:
@theproducer Any chance you can have a look?

simplysparsh:
@de-dan Thanks for the review! You're right, i didn't catch your first part initially. Since Capacitor 8 requires iOS 15+, the `#available(iOS 13.0, *)` check was unnecessary.

I've simplified the code to remove that check. I kept a minimal fallback (`?? UIWindow(frame:)`) for the edge case where no active `UIWindowScene` is found, but the nested conditionals are gone.

The change is now a single-line assignment:
```swift
self.tmpWindow = windowScene.map { UIWindow(windowScene: $0) } ?? UIWindow(frame: UIScreen.main.bounds)
```

Let me know if this looks good!

simplysparsh:
@jcesarmobile Would you mind taking a look at this PR when you have a chance? 

This fixes an issue where `SFSafariViewController` (used by `@capacitor/browser` for OAuth flows) wasn't appearing when presented via `presentVC` on scene-based apps. The temp `UIWindow` wasn't attached to an active `UIWindowScene`, so the Safari view would load but remain invisible.

theproducer (approved):

simplysparsh:
@eric-horodyski soft nudge

jcesarmobile:
The reason we are doing the `self.tmpWindow = UIWindow.init(frame: UIScreen.main.bounds)` is because there was a bug where presenting a view controller from the same window would "freeze" the javascript execution of the WKWebView that was presenting a view controller, i.e. you couldn't call `Browser.close()` programmatically to close the browser as the browser on top of the WebView would freeze the execution, so that's why we use a separate new window that is not attached, that was the only way of not freeze the javascript execution.

I've tested and on my real iOS 15.6.1 device and was also tested on iOS 14 simulators, that's not really needed anymore, just using `self.viewController?.present(viewControllerToPresent, animated: flag, completion: completion)` (as we do for popover presentations) seems to be enough, so I think that would be preferred to change the whole `presentVC` method to just wrap a `self.viewController?.present(viewControllerToPresent, animated: flag, completion: completion)` call and get rid of the tpmWindow/tmpViewController code that causes other set of issues.

Can you address those changes?
If not we can do it in a separate PR.
If you do, can you also mark the `presentVC` method as deprecated?
since it would be a single line now there is no need for it anymore, so we would delete it in Capacitor 9 or another future major release.

jcesarmobile:
Since there was no response and the requested changes were no addressed, I've created a new PR
https://github.com/ionic-team/capacitor/pull/8338

PR #8318, opened 2026-01-27
feat(android-template)!: update Android template to support AGP 9.0

terencehonles:
This change updates the Android template and replaces the default proguard file `proguard-android.txt` with `proguard-android-optimize.txt` which allows proguard optimizations.

This PR is a follow up to https://github.com/ionic-team/capacitor/pull/8315, and can be merged outside of a fix release.

### DIFF
--- a/android-template/app/build.gradle
+++ b/android-template/app/build.gradle
@@ -19,7 +19,7 @@ android {
     buildTypes {
         release {
             minifyEnabled false
-            proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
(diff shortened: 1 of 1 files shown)


ximex:
should the android-template also get changed?
see: https://github.com/ionic-team/capacitor/blob/main/android-template/app/build.gradle#L22

terencehonles:
> should the android-template also get changed? see: https://github.com/ionic-team/capacitor/blob/main/android-template/app/build.gradle#L22

This PR is for the file you referenced (I updated the description to make it clearer it's updating the template).

ximex:
:facepalm: i checked before the changes from https://github.com/ionic-team/capacitor/pull/8315 and overlooked that this is another PR that is already for that i mentioned. sry

PR #8319, opened 2026-01-27
chore: update latest workflow to only allow pointing at 8.x

markemer:


OS-pedrogustavobilro:
The [publish-latest.yml](https://github.com/ionic-team/capacitor/blob/main/.github/workflows/publish-latest.yml), which can be called from any branch, and also publishes to Maven Central / CocoaPods, should that be changed here as well?

theproducer:
> The [publish-latest.yml](https://github.com/ionic-team/capacitor/blob/main/.github/workflows/publish-latest.yml), which can be called from any branch, and also publishes to Maven Central / CocoaPods, should that be changed here as well?

`publish-android` and `publish-ios` depends on the changed action:
```
needs: publish-npm-latest
```

So, in theory, if its ran on a non 8.x branch, the  `deploy-npm-latest` job won't run, and others will that depend on the outcome of `deploy-npm-latest` will be skipped.

Issue #8127, opened 2025-08-28
[Bug]: Capacitor finds unrelated Gemfile in an SPM project


william-stacken:
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.4.3
  @capacitor/core: 7.4.3
  @capacitor/android: 7.4.3
  @capacitor/ios: 7.4.3

Installed Dependencies:

  @capacitor/cli: 7.4.3
  @capacitor/core: 7.4.3
  @capacitor/android: 7.4.3
  @capacitor/ios: 7.4.3

[success] iOS looking great! 👌
[success] Android looking great! 👌

### Other API Details

```Shell
npm 10.9.2
node v22.15.1
```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

I am running `npx cap update` in my iOS project that I have migrated to SPM. It results in the following:

```
[info] Using Gemfile: Bundler update needed...
[error] ERROR:  While executing gem ... (Gem::FilePermissionError)
        You don't have write permissions for the /Library/Ruby/Gems/2.6.0 directory.
```

It seems capacitor thinks I am still using cocoapods because it uses the exit code of `bundle &> /dev/null ; echo $?` to determine if it should update bundler.

https://github.com/ionic-team/capacitor/blob/7.4.3/cli/src/ios/common.ts#L21

I am running capacitor from a subdirectory of a directory that happens to contain a `Gemfile`, and bundler thus finds it. When capacitor tries to update bundler, it fails because it is my system bundler. But I don't suppose capacitor should update bundler in an SPM project, let alone touch unrelated Gemfiles?

### Expected Behavior

Capacitor does not try to update bundler when I am using an SPM project, and also does not write to that unrelated Gemfile. Perhaps it can try to locate the Gemfile itself first, so that bundler won't find any unrelated Gemfiles?

### Project Reproduction

https://github.com/william-stacken/capacitor-bug

PR #8321, opened 2026-01-29
feat(cli): Add packageManager to iOS config

jcesarmobile:
When bundler support [was added](https://github.com/ionic-team/capacitor/pull/5205), it broke a lot of things.

First of all, Capacitor already had bundler support, but by using an executable that would run the `bundle exec pod` command instead of `pod` command.
The PR shouldn't have repurposed `podPath` as it was for pointing to the executable path, but in that PR it was changed to contain the command to run, which breaks some checks like `isInstalled(podPath)`, and also doesn't really work if passed to `runCommand` as it only expects one command and that would pass `bundle exec pod` if using bundler. (no bug here since we were reading the value and manually calling just `bundle` and then `exec` and `pod` as params.

Also added a bunch of `checkBundler()` calls which cause issues and race conditions when not using bundler.

So, this PR adds a new `packageManager` property to the iOS config, which gets the value from a new `determinePackageManager` function with a similar logic to what `determineGemfileOrCocoapodPath` used to have (including a SPM check and only sets the value to bundler if installed.
The values can be `bundler`, `Cocoapods` or `SPM`
Then uses that property to only run `checkBundler()` if using bundler and only run `checkCocoaPods()` if using CocoaPods, and no checks at all if using SPM.
So it closes https://github.com/ionic-team/capacitor/issues/8127

`determineGemfileOrCocoapodPath` has been deleted since it wasn't exported, so I don't think it's a breaking change.
Added a new `determineCocoapodPath` with the original logic that was removed on the mentioned bundler PR, so it only checks if the `CAPACITOR_COCOAPODS_PATH` env variable is defined or not.

Deprecated `checkPackageManager` and `updateChecks` since those were exported and removing them would be a breaking change.
We can delete them on main, but this should be cherry picked into 7.x and 8.x, so can't be deleted yet.

Changed some checks from `== 'Cocoapods'` to `!== 'SPM'` since `bundler` and `Cocoapods` are basically the same, except on a few places where I explicitly check for `bundler`.

Removes some `isInstalled('bundler')` and `isInstalled('pod')` since those are tested beforehand, and for `pod` it shouldn't check that but `isInstalled(podPath)` in case `podPath` was set to an executable path.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,8 @@ jobs:
           - /Applications/Xcode_26.0.app
     steps:
       - run: sudo xcode-select --switch ${{ matrix.xcode }}
+      - run: xcrun simctl list > /dev/null
+      - run: xcodebuild -downloadPlatform iOS
       - uses: actions/setup-node@v6
         with:
           node-version: 22.x
--- a/.github/workflows/publish-ios.yml
+++ b/.github/workflows/publish-ios.yml
@@ -13,6 +13,8 @@ jobs:
     timeout-minutes: 30
     steps:
       - run: sudo xcode-select --switch /Applications/Xcode_26.0.app
+      - run: xcrun simctl list > /dev/null
+      - run: xcodebuild -downloadPlatform iOS
       - uses: actions/setup-node@v6
         with:
           node-version: 22.x
--- a/.github/workflows/publish-npm-latest.yml
+++ b/.github/workflows/publish-npm-latest.yml
@@ -15,7 +15,7 @@ permissions:
 
 jobs:
   deploy-npm-latest:
-    if: github.ref == 'refs/heads/main'
+    if: github.ref == 'refs/heads/8.x'
     runs-on: macos-15
     timeout-minutes: 30
     steps:
--- a/android/package.json
+++ b/android/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@capacitor/android",
-  "version": "8.0.2",
+  "version": "9.0.0-dev.0",
   "description": "Capacitor: Cross-platform apps with JavaScript and the web",
   "homepage": "https://capacitorjs.com",
   "author": "Ionic Team <hi@ionic.io> (https://ionic.io)",
@@ -23,7 +23,7 @@
     "verify": "./gradlew clean lint build test -b capacitor/build.gradle"
   },
   "peerDependencies": {
-    "@capacitor/core": "^8.0.0"
+    "@capacitor/core": "^9.0.0-dev.0"
   },
   "publishConfig": {
     "access": "public"
--- a/cli/package.json
+++ b/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@capacitor/cli",
-  "version": "8.0.2",
+  "version": "9.0.0-dev.0",
   "description": "Capacitor: Cross-platform apps with JavaScript and the web",
   "homepage": "https://capacitorjs.com",
   "author": "Ionic Team <hi@ionic.io> (https://ionic.io)",
--- a/core/package.json
+++ b/core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@capacitor/core",
-  "version": "8.0.2",
+  "version": "9.0.0-dev.0",
   "description": "Capacitor: Cross-platform apps with JavaScript and the web",
   "homepage": "https://capacitorjs.com",
   "author": "Ionic Team <hi@ionic.io> (https://ionic.io)",
--- a/ios/package.json
+++ b/ios/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@capacitor/ios",
-  "version": "8.0.2",
+  "version": "9.0.0-dev.0",
   "description": "Capacitor: Cross-platform apps with JavaScript and the web",
   "homepage": "https://capacitorjs.com",
   "author": "Ionic Team <hi@ionic.io> (https://ionic.io)",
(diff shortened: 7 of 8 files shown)


cli/src/index.ts
```diff
@@ -29,6 +27,16 @@ export async function run(): Promise<void> {
   }
 }
 
+async function getPackageManager(config: Config, packageManager: any): Promise<string> {
```
de-dan: Return type could be narrowed down instead of Promise<string>.

cli/src/config.ts
```diff
@@ -415,13 +417,20 @@ async function determineAndroidStudioPath(os: OS): Promise<string> {
...
+}
+
+async function determinePackageManager(
   rootDir: string,
   platformDir: any,
   nativeProjectDirAbs: string,
-): Promise<string> {
-  if (process.env.CAPACITOR_COCOAPODS_PATH) {
-    return process.env.CAPACITOR_COCOAPODS_PATH;
+): Promise<'Cocoapods' | 'bundler' | 'SPM'> {
```
de-dan: This return type is used several times, maybe it makes sense to create a determined type for this.

cli/src/ios/doctor.ts
```diff
@@ -19,7 +19,7 @@ export async function doctorIOS(config: Config): Promise<void> {
   // check online datebase of common errors
   // check if www folder is empty (index.html does not exist)
   try {
-    await check([() => checkBundler(config) || checkCocoaPods(config), () => checkWebDir(config), checkXcode]);
+    await check([() => checkWebDir(config), checkXcode, ...(await getCommonChecks(config))]);
```
OS-pedrogustavobilro, line 22: The order of methods was changed here, don't know if on purpose or accidental?

cli/src/ios/common.ts
```diff
@@ -25,6 +26,16 @@ function execBundler() {
   }
 }
 
+export async function getCommonChecks(config: Config): Promise<CheckFunction[]> {
```
OS-pedrogustavobilro, line 29: Should this be something like

```suggestion
export async function getPackageManagerChecks(config: Config): Promise<CheckFunction[]> {
```

or is there plans to add different kind of checks in the future?

cli/src/tasks/update.ts
```diff
@@ -46,6 +45,28 @@ export async function updateCommand(config: Config, selectedPlatformName: string
...
+      continue;
+    } else {
+      throw `Platform ${platformName} is not valid.`;
+    }
+  }
+  return checks;
+}
+
+/**
+ * @deprecated use addUpdateChecks
```
OS-pedrogustavobilro, line 65: Should we provide more context here on why it's deprecated, or you feel it's not necessary?

Idk what's usually done on CLI method deprecations.

cli/src/util/spm.ts
```diff
@@ -12,13 +12,18 @@ import { getMajoriOSVersion } from '../ios/common';
...
-import { runCommand, isInstalled } from '../util/subprocess';
+import { runCommand } from '../util/subprocess';
 
 export interface SwiftPlugin {
   name: string;
   path: string;
 }
 
+/**
+ * @deprecated use config.ios.packageManager
```
OS-pedrogustavobilro, line 23: Same comment about reason for deprecation here.

cli/src/config.ts
```diff
@@ -415,13 +417,20 @@ async function determineAndroidStudioPath(os: OS): Promise<string> {
   return '';
 }
 
-async function determineGemfileOrCocoapodPath(
+async function determineCocoapodPath(): Promise<string> {
```
OS-pedrogustavobilro, line 421: Minor: Why does this need to be async?

cli/src/ios/doctor.ts
jcesarmobile: on purpose, but the order doesn't matter, they are all run

cli/src/config.ts
jcesarmobile: It doesn't but since `podPath` is `Promise<string>` and this is the method setting that value, so it's easier to keep it `Promise<string>` too.
Changing `podPath` to `string` would be a breaking change.

cli/src/ios/doctor.ts
OS-pedrogustavobilro: Got it, my suspicion was that the order didn't matter, but thanks for confirming!

cli/src/ios/common.ts
jcesarmobile: The method is basically a simplified version of `updateChecks`, but since it's now used in most of the commands,  I changed the name to "common" instead of "update" and put it inside `common.ts` file.
While it's only doing package manager checks at the moment, it could do other checks in the future, or not, can't know for sure.
If we need to add new checks in the future, then `getPackageManagerChecks` will have to be renamed (breaking) or add a new method for the new checks, if we don't, then might be a slightly more suitable name.

cli/src/tasks/update.ts
jcesarmobile: The method was not documented, so I don't think more explanations are needed, it's more for us to not use them as we should remove them in Capacitor 9.

While it's technically possible to use any exported method, I don't think people is using them, but we can't remove them just in case as it will be cherry picked into 8.x and 7.x and would be breaking.

cli/src/util/spm.ts
jcesarmobile: The method was not documented, so I don't think more explanations are needed, it's more for us to not use them as we should remove them in Capacitor 9.

While it's technically possible to use any exported method, I don't think people is using them, but we can't remove them just in case as it will be cherry picked into 8.x and 7.x and would be breaking.

OS-pedrogustavobilro: Makes sense! Thanks for the explanation

cli/src/tasks/update.ts
OS-pedrogustavobilro: Makes sense! Thanks for the explanation

cli/src/ios/common.ts
OS-pedrogustavobilro: Fair enough then, I'm fine with being as-is.

PR #8330, opened 2026-02-04
fix(cookies): only send expires param on web if valid date

ItsChaceD:
This PR fixes an issue with `CapacitorCookies.setCookie()` on web / PWA where cookies were not being set properly if the expires date was empty or improperly formatted

### DIFF
--- a/core/src/core-plugins.ts
+++ b/core/src/core-plugins.ts
@@ -110,7 +110,7 @@ export class CapacitorCookiesPluginWeb extends WebPlugin implements CapacitorCoo
       const encodedValue = encode(options.value);
 
       // Clean & sanitize options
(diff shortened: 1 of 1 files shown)


core/src/core-plugins.ts
OS-pedrogustavobilro, line 1: I'm still seeing this warning when running the PWA, don't know if you're getting it as well, and don't think it's related to setCookies, because I get this when opening the cookies page of the test app in PWA, maybe it's the date time component of the ionic framework or how we're using it? 

```
[Ionic Warning]: Unable to parse date string: . Please provide a valid ISO 8601 datetime string.
```

ItsChaceD: Yes it's an issue with the date picker component in the test app

Issue #5951, opened 2022-09-23
CapacitorBridge.presentVC does not work in Scenes based application


joeflateau:
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `@capacitor/ios@3.0.1` for the project I'm working on.

In a Scenes-based application (ie: any app that uses CarPlay) `@capacitor/browser`'s `Browser.open({ url })` does not work because they underlying `CapacitorBridge.presentVC` does not work.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/@capacitor/ios/Capacitor/Capacitor/CapacitorBridge.swift b/node_modules/@capacitor/ios/Capacitor/Capacitor/CapacitorBridge.swift
index 749ab1f..199dc83 100644
--- a/node_modules/@capacitor/ios/Capacitor/Capacitor/CapacitorBridge.swift
+++ b/node_modules/@capacitor/ios/Capacitor/Capacitor/CapacitorBridge.swift
@@ -658,7 +658,14 @@ internal class CapacitorBridge: NSObject, CAPBridgeProtocol {
         if viewControllerToPresent.modalPresentationStyle == .popover {
             self.viewController?.present(viewControllerToPresent, animated: flag, completion: completion)
         } else {
-            self.tmpWindow = UIWindow.init(frame: UIScreen.main.bounds)
+            if #available(iOS 13.0, *) {
+                if let windowScene = self.viewController?.view.window?.windowScene {
+                    self.tmpWindow = UIWindow.init(windowScene: windowScene)
+                }
+            }
+            if (self.tmpWindow == nil) {
+                self.tmpWindow = UIWindow.init(frame: UIScreen.main.bounds)
+            }
             self.tmpWindow?.rootViewController = TmpViewController.init()
             self.tmpWindow?.makeKeyAndVisible()
             self.tmpWindow?.rootViewController?.present(viewControllerToPresent, animated: flag, completion: completion)
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

Ionitron:
This issue may need more information before it can be addressed. In particular, it will need a reliable Code Reproduction that demonstrates the issue.

Please see the Contributing Guide for [how to create a Code Reproduction](https://github.com/ionic-team/capacitor/blob/HEAD/CONTRIBUTING.md#creating-a-code-reproduction).

Thanks!
[...]

PR #8338, opened 2026-02-11
fix(ios): remove tmpWindow usages on presentVC/dismissVC

jcesarmobile:
closes https://github.com/ionic-team/capacitor/issues/5951


markemer:
When we deprecate these days, maybe we should create a PR to next that removes them

PR #8346, opened 2026-02-13
fix!: replace deprecated `@UIApplicationMain` with `@main`

jworreth:
This merge request concerns only the iOS platform.

With Swift 6, the @UIApplicationMain has been deprecated and make the build to fail. This MR will use the @main instead on the AppDelegate on the iOS template used when capacitor adds ios as a platform


markemer:
As this is a breaking change it will be put against the `next` branch. More details on branching to come, but it will be released with Capacitor 9.

PR #8347, opened 2026-02-15
Do not pass data: fetch requests to the native http request handler

frederikbosch:
For all sorts of reasons, a fetch request of data URI might be executed.

```js
fetch(new URL('data:text/css,body{background:red}'))
```

At the moment these requests are forwarded to the native request, causing crashes like the one that PR #8304 is trying to solve. With this PR, `data:` requests will not be forwarded anymore to native request layer.

Just to explain why, in my app, I am fetching a DATA CSS Uri.
- in my app the following code `fetch(new URL('./file.css'))`
- is compiled by Vite as `fetch(new URL('data:text/css,body{background:red}'))`
- so, while I am not intentionally fetching a data URI, it is compiled as such, because the resource is so small that Vite modifies it to [an inline asset](https://vite.dev/guide/assets#importing-asset-as-url)

### DIFF
--- a/core/native-bridge.ts
+++ b/core/native-bridge.ts
@@ -508,7 +508,7 @@ const initBridge = (w: any): void => {
             options.headers = headers;
           }
           const request = new Request(resource, options);
-          if (request.url.startsWith(`${cap.getServerUrl()}/`)) {
+          if (request.url.startsWith(`${cap.getServerUrl()}/`) || request.url.startsWith('data:')) {
             return win.CapacitorWebFetch(resource, options);
           }
           const { method } = request;


jcesarmobile:
Thanks for the PR, but while this change solves your issue, it doesn't address the real problem.
The real problem is that we are not handling `URL` objects on fetch calls.
I've addressed it here https://github.com/ionic-team/capacitor/pull/8386

As I commented on https://github.com/ionic-team/capacitor/pull/8304, I had already tried a `fetch` call with a `data://` url and it was working fine, because I passed it as string, since now you have shared that you were using an `URL` object, I tried and verified that it causes the crash.

frederikbosch:
Thanks @jcesarmobile

PR #8350, opened 2026-02-18
chore(cli)!: Remove unused livereload method and parameter

OS-pedrogustavobilro:
Saw some TODOs set to remove code in `livereload.ts`, addressing them in this PR.

Since it's possibly users are consuming `CapLiveReload` directly, this is a breaking change.

Now that we have a next major planned, this can be done in `next` branch.


OS-pedrogustavobilro:
@jcesarmobile believe you were the one who added the TODOs originally, from what I saw the changes needed were just those, but do let me know if you think I missed something.

Issue #8335, opened 2026-02-09
[Feature]: Support SPM package traits for plugin dependencies


robingenz:
### Description

Allow Capacitor plugin consumers to enable/disable SPM package traits on plugin dependencies

### Platforms

- [x] iOS
- [ ] Android
- [ ] Web

### Request or proposed solution

Swift 6.1 introduced [Package Traits (SE-0450)](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md) which allow package authors to define conditional dependencies that consumers can toggle via a `traits` parameter:

```swift
// Plugin's Package.swift
let package = Package(
    name: "CapacitorFirebaseAnalytics",
    traits: [
        .trait(name: "AdIdSupport", description: "Include IDFA collection support"),
    ],
    targets: [
        .target(
            name: "FirebaseAnalyticsPlugin",
            dependencies: [
                .product(name: "FirebaseAnalyticsCore", package: "firebase-ios-sdk"),
                .product(name: "FirebaseAnalyticsIdentitySupport", package: "firebase-ios-sdk",
                         condition: .when(traits: ["AdIdSupport"])),
            ]),
    ]
)
```

```swift
// Consumer's Package.swift
dependencies: [
    .package(url: "https://github.com/example/some-plugin.git", from: "1.0.0", traits: ["AdIdSupport"])
]
```

Since Capacitor manages SPM dependency resolution on behalf of the app developer, there is currently no way for a Capacitor app to pass traits to plugin dependencies. It would be great if Capacitor supported this, e.g. via `capacitor.config.ts`:

```typescript
const config: CapacitorConfig = {
  ios: {
    packageTraits: {
      '@capawesome-team/capacitor-firebase-analytics': ['AdIdSupport'],
    },
  },
};
```

A concrete use case is [capawesome-team/capacitor-firebase#942](https://github.com/capawesome-team/capacitor-firebase/issues/942): the Firebase Analytics plugin needs to offer a choice between `FirebaseAnalytics` (with IDFA) and `FirebaseAnalyticsCore` (without IDFA). CocoaPods handles this via subspecs, but SPM has no equivalent mechanism — [package traits](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/) are the intended solution.

### Additional Information

- [SE-0450 proposal](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0450-swiftpm-package-traits.md)
[...]

PR #8351, opened 2026-02-19
feat(cli): support SPM package traits in generated Package.swift

robingenz:
## Summary
- Add `ios.spm.packageTraits` config option to pass package traits to SPM plugin dependencies
- Use `swift-tools-version: 6.1` when traits are configured, otherwise keep `5.9`
- Append `traits: [...]` to `.package()` dependency lines for plugins with configured traits

Close https://github.com/ionic-team/capacitor/issues/8335


OS-pedrogustavobilro:
Hey @robingenz thanks for the PR, tested here and it's working pretty well. I did a few minor changes, most notably to allow to set the `.defaults` attribute (was being written in quotes).

I'll ask for an additional review of the team as well. For reference - Tested with https://github.com/OS-pedrogustavobilro/test-capacitor-pr8351-spm-traits

robingenz:
@OS-pedrogustavobilro Thank you so much! We are really looking forward to the release as we have 2-3 plugins that are just waiting for it. Let me know if I can help in any way.

jcesarmobile:
Since the feature requires swift-tools-version: 6.1, even if that version will only be used if traits are present, I would prefer not merging this until Capacitor uses Swift 6.

My understanding is that having swift-tools-version: 6.1 in the Package.swift has implications for the dependencies being fetched and could cause issues if some of the dependencies have xcframeworks built with previous versions

robingenz:
@jcesarmobile Good point—I totally understand the concern. Is there a timeline for Swift 6 support in Capacitor?

The reason this feature is somewhat time-sensitive is that while it was supported in CocoaPods, SPM is now the default package manager. This has created restrictions for our plugin users who have transitioned to SPM.

To mitigate the risk, I have two suggestions:

1. Log a warning when traits are configured and the tools version is bumped to 6.1, so users are aware of potential compatibility issues with certain dependencies.
2. Add a separate ios.spm.swiftToolsVersion config option instead of auto-bumping, so users have to explicitly opt in. Traits would only kick in when this is set to 6.1 or higher.

Let me know what you think!

OS-pedrogustavobilro:
Hey @robingenz, getting back to you on your points. They're both reasonable, but we feel they might not be enough so we went a step further.

We're planning on going with essentially an "extension" of the second option you provided: adding an **experimental** setting for `swift-tools-version` [in this PR](https://github.com/ionic-team/capacitor/pull/8372).

Users will need to set that config explicitly to be able to use the package traits feature from this PR, which we will be also place under the `experimental` object.

Unfortunately we can't yet provide a timeline for when Capacitor will support Swift 6, but I think I can say it won't be Capacitor 8, only in a future major. As such, we think placing these settings in an `experimental` block is a fair compromise between having this functionality available in the near future, while also making sure users are aware of the change their making in their apps, that can cause issues.

We hope this doesn't entirely dissuade users of your plugin to not use the feature, as I think as long they properly test their apps after making this change and find everything is still working, they should be alright.

The other PR is still going to be reviewed, but if/when it's merged, we will also be doing a few updates in this PR, particularly to mark the config option as experimental.

Feel free to share your feedback on this option, thank you!

robingenz:
Hey @OS-pedrogustavobilro, thank you! That looks like a good approach. I do think there will be some developers who will be put off by a setting labeled `experimental`, but as long as the possibility to use traits exists at all, it would be a good step forward. I really hope Swift 6 makes it into Capacitor 9.

OS-pedrogustavobilro:
Temporarily placing this PR in draft until #8372 is merged.

PR #8364, opened 2026-02-25
Add method to disable autosuggestion

tyilo:
Re-creation of #6876 

We have a use-case where our app is run on a tablet in a public setting.

Thus we would like to disable autosuggestion to prevent the Android keyboard from remembering and suggesting emails of previous users.


riderx:

Patch ID: `upstream-pr-8364-android`

```ts
plugins: {
  CapacitorPatch: {
    patches: ['upstream-pr-8364-android'],
    strict: true,
  },
}
```

PR #8368, opened 2026-03-01
fix(android): handle lowercase range header

bwees:
This fixes issues with HTTP range requests when the range header is lowercase on Android. 

iOS handles the case insensitivity via `urlSchemeTask.request.value(forHTTPHeaderField: "Range")`. However on Android, headers are placed in a case sensitive Map that requires manual handling.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -352,13 +352,6 @@ public class WebViewLocalServer {
                 String[] parts = rangeString.split("=");
                 String[] streamParts = parts[1].split("-");
                 String fromRange = streamParts[0];
-                int bytesToSkip;
-                try {
-                    bytesToSkip = Integer.parseInt(fromRange);
-                    if (bytesToSkip > 0) {
-                        responseStream.skip(bytesToSkip);
-                    }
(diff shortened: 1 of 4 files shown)


OS-pedrogustavobilro:
Hey @bwees, thanks for the PR.

I'm not getting any issues with the current code in `main`. I'm sending a lowercase "range" header, and it's still arriving "Range" to native Android, I assume Chrome does some capitalization of headers names.

Were you getting an issue with a particular request or situation? And if so can you share a reproduction app for us to check?

bwees:
I am getting this issue when working with MapLibre GL JS and using PMTiles as a source. 

An example app is here: https://github.com/bwees/range-request-capacitor-repro

If you look at the PMTiles responses received by the client (viewed from Chrome Devtools debugging the running android app), you will see that the proper range response headers are only sent once these changes are added. 

On `main`, the "server" responds with a 200 response instead of a 206 response. Once this PR is applied, the responses come back as 206 with the Range specific headers in the response.

Note: Even with this fix, the map will not load as HTTP Range requests are quite broken on Capacitor. They do not currently follow spec which causes issues with the PMTiles decoder. I narrowed down the issue late last night, #8371 .

bwees:
Hey sorry about that! It appears I must have deleted the wrong map components while stripping my app to a minimum example. I have updated it and pushed to the branch so #8371 should be easy to reproduce now.

Thanks for your help!

Issue #8303, opened 2026-01-10
[Bug]: `cap run` ignores server config


aolsenjazz:
### Capacitor Version: 8.0.0

Thanks Ionic team for the work you do! The root issue that I'm calling out with this issue is that `cap run -l` ignores server config, but there are a few related issues imo worth solving here. I'm hoping that this issue can be a suitable place to discuss them at a high level before creating small-scope issues.

_If preferred, I'm happy to break these out into small-scope tickets now, just lmk._

## Current Behavior
`cap run -l` ignores `capacitor.config.{ts|json|etc}` server config. Instead, what happens is as follows:

- If `--host` cli arg is used, use that value
  - If not, resolve to a reasonable external network interface || localhost
- If `--port` cli arg is used, use that value
  - Otherwise, use port 3000
- Concatenate values to form connection string `http://{host}:{port}`

### Hardcoded `http` protocol
**Paint point 1**: My company's auth provider refuses to work on non-localhost && non-encrypted (`http://`) connections. I.e. `http://localhost` is fine, but `http://192.168.x.x` fails. This of course creates problems for running locally on hardware devices.

**Pain point 2**: I would like to add tunneling to my app via a `--tunnel` flag. A standard tunneling setup might use ngrok, and would therefore use a connection string like `https://some-endpoint.ngrok.com`. In order to do this, I would need to pass that endpoint to `cap run -l` at runtime. Using just `--host` and `--port` this is impossible; I need to be able to control the protocol.

**Solutions**: Add support for a `--https`/`--protocol`/`ssl` flag to `cap run -l`.

### Bad values imputed when no `host`/`port`
**Pain point 1**: `cap run -l` auto selects a reasonable network interface to try to load the app from. Oftentimes, a non-peer-accessible network is selected. I regularly experience this behavior:
- I'm at a coffee shop/on a mobile hotspot which blocks peer traffic on e.g. the 10.x.x.x address space
  - 169.254.x.x address space is peer-accessible

**Pain point 2**: `cap run -l` uses port `3000` when a `--port` cli arg isn't passed. We cannot reasonably make the assumption that we should use port `3000`, especially when one of the dominant bundlers uses `5173` by default.

**Solutions**: Three distinct behaviors here would constitute a solution:
- If `--host` and `--port` are unset, use `server.url`
  - If `server.url` is unset, use reasonable default according to existing strategy
- If `--host` is unset, use `server.hostname` for host
- Add `port` to server config
[...]

Issue #6388, opened 2023-03-13
feat: better localhost development


kristojorg:
### Description
I often want to run my app with live reloading for development on both iOS and Android platforms. This is currently very cumbersome. I have tried using the Ionic CLI as referenced in the docs, but it has never worked for me and I have been told by Ionic employees that the Capacitor CLI is more up to date and the Ionic one is not maintained. 

I would like to propose an easier way to run an app with live reloading.

#### The Problem(s)

1. The iOS simulator and Android emulator have different means of accessing a server on the host machine's `localhost`. In the iOS simulator this can be done by simply connecting to `localhost` or `127.0.0.1` just like a normal web browser on the host machine. In the Android emulator, the `localhost` refers to the emulator's loopback interface, not the host machine's. Thus there is no one setting I have found for the `server.url` field in `capacitor.config.ts` that will tell iOS and Android where to find the web bundle. 
2. Changing the `server.url` field every time you want to develop your app is problematic. **This field should not ever be set in production**, yet it is very easy to mistakenly commit the change. It's not a great development experience. 

### Platform(s)

- iOS
- Android
- Capacitor CLI
- **Not** the Ionic CLI

### Preferred Solution
One solution I have thought of is:

1. A way to set a `server.url` when running capacitor commands. This should override the value in `capacitor.config.ts`, allowing us to set `"sync: live": "npx cap sync --server-url http://0.0.0.0:8100"` in our package.json and then just run `npm run sync:live`. No more setting temporary values in the `capacitor.config.ts` file.
2. If we set `server.url` to `localhost`, capacitor should interpret that as the host machine's localhost and set it to `10.0.2.2` in the Android emulator.

### Alternatives

There are many other options - and maybe something better? Another option is to allow multiple `capacitor.config.ts` files. One for development and another for production. There is probably another way to solve the ip address issue as well, and honestly it would be awesome if there was one command (`npx cap run --live`) that runs the dev server, syncs, and runs both the iOS simulator and Android emulator. 

### Additional Context
This is related to many tickets and posts in the Ionic Forum referencing the `ERR_CONNECTION_REFUSED` message on Android.

jcesarmobile:
`npx cap run` now has a `-l, --live-reload` parameter, that in conjunction with ` --host` and `--port` allow you to run your app directly on the device or emulator on the desired live reload url, only when you are running the app, and removes the values once you stop the command.
https://github.com/ionic-team/capacitor/pull/6831

[...]

PR #8370, opened 2026-03-02
fix(cli): Preserve Cap Config `server.url` for live reload

OS-pedrogustavobilro:
Closes https://github.com/ionic-team/capacitor/issues/8303. 

## Description

Prior to this PR, the `server.url` config property would always get overridden when running `cap run --live-reload`, regardless of whether `--host` and `--port` were specified or not, even though the docs state the `url` is meant to be used for live reload.

This PR fixes that issue addressing different scenarios as follows:

1. Neither `-host` nor`--port` provided and config `server.url` exists - `server.url` is used - **Fixed by this PR**. 
2. Either `-host` and/or `--port` provided and config `server.url` exists - `server.url` not used, `--host` and/or `--port` used - Same as before.
3. Either `-host` and/or `--port` provided and config `server.url` does not exist - `server.url` not used - Same as before.
4. Neither `-host` nor`--port` provided and config `server.url` does not exist - `server.url` not used, will use existing defaults for host and port - Same as before.

## Testing

You can pull this branch, and point the `@capacitor/cli` to the local folder containing the code from this PR (make sure you do `npm run build` before), and test `npx cap run --live-reload` with different combinations of `--host`, `--port`, `--https`, and capacitor config `server.url`

For reference: I've done most of my tests in https://github.com/OS-pedrogustavobilro/capacitor-test-livereload

### DIFF
--- a/cli/src/util/livereload.ts
+++ b/cli/src/util/livereload.ts
@@ -169,7 +169,7 @@ class CapLiveReload {
       if (!options.https && !options.port) {
         options.port = '3000';
       }
-      url = `http://${options.host}:${options.port ? `:${options.port}` : ''}`;
+      url = `http://${options.host}${options.port ? `:${options.port}` : ''}`;
     } else {
       // use existing url in capacitor config json
       url = configJson.server.url;


jcesarmobile:
`--live-reload` should only be used when you want to set the live reload url from the `--host`/`port` values.
It was created for that, for being able to configure the `url` from command line.

If the `url` is already provided in the capacitor config file, then `cap run` should be used without `--live-reload`/`-l`.

If `--live-reload` is passed, it's expected to override the existing value of the `url` if it exists.

The `--live-reload` naming is confusing because it gives the impression that it will do something else to "enable" live reload, but it doesn't, it just configures the `url` value. The server used as params is the one in charge or doing all live reload "work".

OS-pedrogustavobilro:
@jcesarmobile thanks for bringing this up. I am less confused now, but still slightly confused, what's the point of `server.url`? When you want to use live reload but without the `--live-reload` flag in cap run / run it directly from Xcode / Android studio?  So that you can do it at runtime and without the risk of accidentally going to production with `server.url` option? I guess that's what the original issue states https://github.com/ionic-team/capacitor/issues/6388

But if that's the case, if the purpose of `--live-reload` is solely to allow setting a live-reload url in command line (rather than in capacitor config), then we should instead update the [description for the option in CLI](https://github.com/ionic-team/capacitor/blob/5db81e68c67652e9d2b29d7ad30629b423d2ad30/cli/src/index.ts#L252), as it states

```
     option('-l, --live-reload', 'Enable Live Reload')
```

which is at the very least ambiguous/misleading (it misled me and other people 😅)

OS-pedrogustavobilro:
Since https://github.com/ionic-team/capacitor/pull/8373 was merged, I am closing this one.

PR #8373, opened 2026-03-03
docs(cli): Better descriptions for live-reload parameters

OS-pedrogustavobilro:
Alternative PR to https://github.com/ionic-team/capacitor/pull/8370

I had not realised that the `--live-reload` option overrides the `server.url` CapacitorConfig on purpose - In fact, the purpose of this property is really to allow to point to a certain live-reload server in the CLI, without having to set it in Capacitor Config (e.g. to avoid accidentally deploying that URL to production).

The description of the CLI option however was not the best, as it indicated "Enable Live Reload", when it fact it doesn't really do that. You can set `server.url` in Capacitor Config, and run `cap run` without `--live-reload`, and live reload should still work.

Therefore, in this PR I update the descriptions of the live reload CLI options to better reflect what they actually do.

FYI @aolsenjazz

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -249,11 +249,11 @@ export function runProgram(config: Config): void {
     )
     .option('--no-sync', `do not run ${c.input('sync')}`)
     .option('--forwardPorts <port:port>', 'Automatically run "adb reverse" for better live-reloading support')
-    .option('-l, --live-reload', 'Set live-reload server URL via CLI (uses defaults, overrides server.url config)')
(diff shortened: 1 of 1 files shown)


OS-pedrogustavobilro:
> rather than "live-reload server", I would say "live-reload url", since this doesn't really have any effect in the server

Yeah that's better, I was unsure if you meant it for `host` and `port` or `live-reload`, but I updated for all of them.

> Also, instead of "Change default" I think it should say "Configure"

Yeah that's also better, updated it.

> we should not really have a default port value

I was thinking of opening a PR to `next` soon to make that (breaking) change.

PR #8375, opened 2026-03-03
docs(cli): Better descriptions for live-reload parameters

OS-pedrogustavobilro:
## Description

Equivalent to https://github.com/ionic-team/capacitor/pull/8373, but for 7.x maintenance branch.

## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [x] Documentation

## Rationale / Problems Fixed

The reason why we're including this PR in 7.x is to provide better clarity on live-reload functionality (or lack there-of) in all maintained versions of Capacitor, since it has been a cause for confusion.

Didn't do a straight cherry-pick because 7.x does not have the `--https` feature in the CLI.

## Tests or Reproductions

N/A

## Screenshots / Media

N/A

## Platforms Affected
- [ ] Android
- [ ] iOS
- [ ] Web

## Notes / Comments

N/A

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -244,9 +244,9 @@ export function runProgram(config: Config): void {
     .option('--target <id>', 'use a specific target')
     .option('--no-sync', `do not run ${c.input('sync')}`)
     .option('--forwardPorts <port:port>', 'Automatically run "adb reverse" for better live-reloading support')
-    .option('-l, --live-reload', 'Enable Live Reload')
-    .option('--host <host>', 'Host used for live reload')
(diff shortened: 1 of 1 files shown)


jcesarmobile:
and why don't we have `--https` in 7.x?
7.x is still in maintenance, so we should cherry pick features too (unless they are breaking)

OS-pedrogustavobilro:
@jcesarmobile true, there's no technical reason why the `--https` flag can't be in 7.x

I'll go ahead and do that then, and I'll close this PR and instead just cherry-pick it directly from main

Thanks!

PR #8194, opened 2025-10-27
feat(cli): Add --https option for --live-reload

Shellishack:
Added a --https flag to allow `cap run` to use https instead of http. 

**[Edit by @OS-pedrogustavobilro]** Fixes  https://github.com/ionic-team/capacitor/issues/8049. Partially addresses https://github.com/ionic-team/capacitor/issues/8303

For example, now running with --https flag, cap cli will have:
```
> cap run android -l --host=192.168.2.101 --https
...
[info] App running with live reload listing for: https://192.168.2.101:3000. Press Ctrl+C to quit.
```

## Live reload remote
It works out of the box.
e.g. `cap run android -l --host=capacitorjs.com --https --port=443`

## Live reload local https server 
⚠️**Self-signed SSL issue:**
Using --https may cause capacitorjs not to load the content in WebView if the web view has a self-signed SSL certificate. 
### My local workaround (fully local, but make your own CA):
For local development, I ended up working around this by making a local CA certificate, and sign server certificates with this CA. Then I would need to install the CA file on my android device. Now live reloading a https local dev server is possible. 

For nextjs:
If you have run `next dev --experimental-https` before, delete ./certificates folder. Then continue.
```
mkdir certificates
cd certificates
```

**Do the below in `./certificates` folder**

1.  Generate a local CA.
    Use devCA.key to install on device later.

```bash
openssl genrsa -out devCA.key 2048
openssl req -x509 -new -nodes -key devCA.key -sha256 -days 3650 -out devCA.crt -subj "/CN=Local Development CA"
```

2. Create a .cnf
   e.g.

```bash
[req]
default_bits       = 2048
prompt             = no
default_md         = sha256
distinguished_name = dn
req_extensions     = req_ext

[dn]
CN = 192.168.1.100 (or your local dev server's ip)

[req_ext]
subjectAltName = @alt_names

[alt_names]
IP.1 = 192.168.1.100
DNS.1 = mypc
DNS.2 = mypc.local
DNS.3 = localhost
DNS.4 = (your custom domain for local testing)
```

3. Generate private key

```bash
openssl genrsa -out localhost-key.pem 2048
openssl req -new -key localhost-key.pem -out localhost.csr -config localhost.cnf
```

4. Sign the certificate with CA
```bash
openssl x509 -req -in localhost.csr -CA devCA.crt -CAkey devCA.key -CAcreateserial -out localhost.pem -days 365 -sha256 -extensions req_ext -extfile localhost.cnf
```

[...]

PR #8376, opened 2026-03-03
fix(cli)!: Remove default port for --live-reload

OS-pedrogustavobilro:
## Description

This PR removes the default `--port` value for `--live-reload`. When adding the [--https](), we already opted out of having a default, and with this (breaking) change there's consistency.

With this PR, if no `--port` nor `--http` is specified, the url will be like so `http://<host>`; in other words, will run in the default port for http (80).

An alternative to this is to merge `--host`, `--port`, and `--https` CLI args into a single `--url` or `--live-reload-url` arg, that would directly override `server.url` config. I am inclined to open a separate PR for that and the team can decide which one is preferable.

## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [x] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Port '3000' isn't widely used for local servers, maybe it made sense once upon a time, but not anymore.

There has been confusion regarding the use of a default port for `--live-reload` (there's probably older links, but these should suffice to justify the change):

- https://github.com/ionic-team/capacitor/issues/8303
- https://github.com/ionic-team/capacitor/pull/8194#pullrequestreview-3862531009 

Since the port may change depending on what's the underlying stack used to setup live-reload, the framework shouldn't make assumptions on the port to use.

## Tests or Reproductions

You can use any existing capacitor app to test, you just need to point to the local capacitor cli version from this PR's branch.

For reference, I used https://github.com/OS-pedrogustavobilro/capacitor-test-livereload

## Screenshots / Media

Not relevant

## Platforms Affected
- [ ] Android
- [ ] iOS
- [ ] Web

## Notes / Comments

N/A

### DIFF
--- a/cli/src/tasks/run.ts
+++ b/cli/src/tasks/run.ts
@@ -43,9 +43,6 @@ export async function runCommand(
   options: RunCommandOptions,
 ): Promise<void> {
   options.host = options.host ?? CapLiveReloadHelper.getIpAddress() ?? 'localhost';
-  if (!options.https && !options.port) {
-    options.port = '3000';
-  }
   if (selectedPlatformName && !(await isValidPlatform(selectedPlatformName))) {
     const platformDir = resolvePlatform(config, selectedPlatformName);
     if (platformDir) {


jcesarmobile:
I personally prefer a single `--url` param rather than having to pass 4 different params: `--live-reload`, `--host` `--port` and `--https`, with undocumented defaults if you don't pass some of them.

When you run a dev server and/or ngrok they usually tell you, the app is running on `http://192.168.1.53:3713`/`https://4f8a-83-53-7-217.ngrok-free.app` or similar message, so with `--url` you could simply copy the whole url and paste as the `--url` param, while with current implementation you have to pass `--live-reload --host 192.168.1.53 --port 3713`, or if using ngrok not pass the `--port`, but pass `--https`.

Also I find `--live-reload` param very confusing, we just changed the description recently to try to make it cleared, but still the param name is confusing as it doesn't really allow live reload, it just configures the `server.url`, whether it points to a server that supports live reload or not, and doesn't do any "live reload magic" by itself.

PR #6526, opened 2023-04-20
feat(cli): inline CSS sourcemaps in addition to JS sourcemaps

terencehonles:
This change updates the cli copy's `--inline` handling to also inline the CSS sourcemaps and not just the JS sourcemaps.

PR #8377, opened 2026-03-04
fix(cli): inline CSS sourcemaps in addition to JS sourcemaps

terencehonles:
## Description

This change updates the cli copy's `--inline` handling to also inline the CSS sourcemaps and not just the JS sourcemaps.

reopening: https://github.com/ionic-team/capacitor/pull/6526

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Allow CSS sourcemaps to be included properly

## Platforms Affected
- [x] Android
- [x] iOS
- [x] Web

### DIFF
--- a/.github/ionic-issue-bot.yml
+++ b/.github/ionic-issue-bot.yml
@@ -20,7 +20,7 @@ comment:
       message: >
         This issue needs more information before it can be addressed.
         In particular, the reporter needs to provide a minimal sample app that demonstrates the issue.
-        If no sample app is provided within 15 days, the issue will be closed.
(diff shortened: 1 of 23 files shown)


jcesarmobile:
I'm considering this a fix instead of feat since the `--inline` description says
> if true, all source maps will be inlined for easier debugging on mobile devices
since it says "all source maps", but it's not doing it for css files, I think it should and doing it is a fix

terencehonles:
Works for me, thanks!

PR #8379, opened 2026-03-06
chore(ci): release as latest-7

jcesarmobile:


jcesarmobile:
yeah, but I've checked and we have always released a 8.x right after, so it's not been a big problem, other than `latest-7` tag was not up to date, has been pointing to 7.4.4 for 3 months

PR #8383, opened 2026-03-09
fix(android): allow synchronous replies on the modern bridge

### LINKED ISSUE #8382: [Bug]: first Android modern-bridge call can miss the reply proxy if synchronous
### Capacitor Version

8.2.0

### Other API Details

```Shell

```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

On Android's modern `WebMessageListener` bridge, `MessageHandler` dispatches the incoming message before it stores the `JavaScriptReplyProxy` for that message.

That means a plugin method that resolves synchronously during the first modern-bridge call can reach `sendResponseMessage()` while `javaScriptReplyProxy` is still `null`. When that happens, Capacitor falls back to the legacy WebView JavaScript response path instead of replying through `replyProxy.postMessage(...)`.

This can make the first synchronous plugin call flaky: the native plugin resolves successfully, but the JavaScript caller may never receive the result if that legacy fallback path is ineffective at that moment.

### Expected Behavior

The reply proxy for a modern bridge message should be available before plugin dispatch begins, so a synchronous response from the first call is always delivered through `replyProxy.postMessage(...)` and does not depend on the legacy fallback path.

### Project Reproduction

.

### Additional Information

  Minimal failing conditions:

  1. Android modern bridge is enabled
  2. The first plugin call arrives through `WebMessageListener`
  3. The plugin resolves _synchronously_
  4. The JavaScript caller expects the normal modern reply path

  The current listener ordering is effectively:

  ```java
  postMessage(message.getData());
  javaScriptReplyProxy = replyProxy;
  ```

A focused Android unit test demonstrates the problem by asserting that the first synchronous modern-bridge call should use the reply proxy instead of `WebView.post(...)`. Reordering those two lines makes that test pass.

richavos:
## Rationale
On Android's modern bridge, `MessageHandler` stored `javaScriptReplyProxy` _after_ dispatching the message to `postMessage(...)`.

That meant a plugin that resolved synchronously during the first modern-bridge call could reach `sendResponseMessage()` before the reply proxy was available, causing Capacitor to fall back to the legacy `WebView` JavaScript response path.

This is unreliable as a fallback, so can result in promises never being settled for plugin calls.

## Changes
This change stores the reply proxy before dispatching the message, so that _all_ plugin calls should be settlable, and adds a regression test for this scenario. 

## Testing
To provide a minimal test, I've added a small `TextUtils` host-test stub so the real `postMessage` path can be run in the test.

I've also added two reflection-based methods that attempt to surgically insert only what is required to allow the "normal" bridge call codepath to be run.

The tradeoff here is concision against brittleness - let me know/feel free to adjust the test to fit with the project as a whole.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

Closes https://github.com/ionic-team/capacitor/issues/8382

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/MessageHandler.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/MessageHandler.java
@@ -26,8 +26,7 @@ public class MessageHandler {
         if (WebViewFeature.isFeatureSupported(WebViewFeature.WEB_MESSAGE_LISTENER) && !bridge.getConfig().isUsingLegacyBridge()) {
             WebViewCompat.WebMessageListener capListener = (view, message, sourceOrigin, isMainFrame, replyProxy) -> {
                 if (isMainFrame) {
-                    postMessage(message.getData());
-                    javaScriptReplyProxy = replyProxy;
+                    onModernBridgeMessage(message.getData(), replyProxy);
                 } else {
                     Logger.warn("Plugin execution is allowed in Main Frame only");
                 }
@@ -40,6 +39,11 @@ public class MessageHandler {
         } else {
             webView.addJavascriptInterface(this, "androidBridge");
         }
+    }
+
+    void onModernBridgeMessage(String data, JavaScriptReplyProxy replyProxy) {
+        javaScriptReplyProxy = replyProxy;
+        postMessage(data);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


riderx:

Patch ID: `upstream-pr-8383-android`

```ts
plugins: {
  CapacitorPatch: {
    patches: ['upstream-pr-8383-android'],
    strict: true,
  },
}
```

PR #8384, opened 2026-03-10
fix(SystemBars): use native safe area insets on Android

### LINKED ISSUE #8287: [Bug]: System Bar safe-area-inset-x ignores Keyboard state
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 8.0.0
  @capacitor/core: 8.0.0
  @capacitor/android: 8.0.0
  @capacitor/ios: 8.0.0

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 8.0.0
  @capacitor/core: 8.0.0
  @capacitor/android: 8.0.0

### Other API Details

```Shell
npm --version: 11.6.2
node --version: v24.11.1
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

With `plugin.SystemBars.insetsHandling` = `css` and no keyboard my app works fine using:
```
body {
    box-sizing: border-box;
    height: 100dvh;
    width: 100dvw;
    padding-top: var(--safe-area-inset-top, env(safe-area-inset-top, 0px));
    padding-bottom: var(--safe-area-inset-bottom, env(safe-area-inset-bottom, 0px));
    padding-left: var(--safe-area-inset-left, env(safe-area-inset-left, 0px));
    padding-right: var(--safe-area-inset-right, env(safe-area-inset-right, 0px));
    display: flex;
}
```

With the keyboard displayed, the bottom inset remains, creating superfluous spacing between the upper border of the on screen keyboard and the webview content.

As a workaround I added as css style ...
```
body.no-bottom-inset {
    --safe-area-inset-bottom: 0px;
}
```
... and these handlers in my code:
```    
    Keyboard.addListener('keyboardWillShow', info => {
        console.log('keyboard did show with height:', info.keyboardHeight);
        document.body.classList.add('no-bottom-inset');
    });

    Keyboard.addListener('keyboardWillHide', () => {
        console.log('keyboard did hide');
        document.body.classList.remove('no-bottom-inset');
    });
```

### Expected Behavior

I feel making `safe-area-inset-x` variables keyboard aware would be more natural.

### Project Reproduction

https://github.com/schwebke/capacitor-system-bar-keyboard-interaction

### Additional Information

Screenshots showing the issue and the effect of the workaround:
https://github.com/schwebke/capacitor-system-bar-keyboard-interaction/tree/main/screenshots

### LINKED ISSUE #8329: [Bug]: Too much bottom safe area when keyboard is visible
### Capacitor Version

```
➜  test-cap git:(master) ✗ npx cap doctor
💊   Capacitor Doctor  💊

Latest Dependencies:

  @capacitor/cli: 8.0.2
  @capacitor/core: 8.0.2
  @capacitor/android: 8.0.2
  @capacitor/ios: 8.0.2

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/android: 8.0.2
  @capacitor/core: 8.0.2
  @capacitor/cli: 8.0.2

```

### Other API Details

```Shell
➜  test-cap git:(master) ✗ npm --version
11.6.1
➜  test-cap git:(master) ✗ node --version
v24.11.0
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

With the latest capacitor v8.0.2, I noticed a strange issue where the bottom padding safe area will be ~ the size of the open keyboard, when the keyboard is open. This makes too much scrollable space appear in the webview.

This was originally spotted here: https://github.com/ionic-team/capacitor/pull/8309#issuecomment-3812996139

#### ❌ Incorrect behavior

https://github.com/user-attachments/assets/bd75db9f-8569-41a8-9df1-54371695bc42

### Expected Behavior

The bottom safe area should be 0 when the keyboard is open.

#### ✅ Correct behavior

This issue does NOT happen with capacitor v8.0.1 or v8.0.0. In those versions it works as expected. Example:

https://github.com/user-attachments/assets/b90b82fa-aa29-453e-a4c2-92fb5c6f2385

### Project Reproduction

https://github.com/aeharding/cap-kb-blank-space

### Additional Information

Device info: Google Pixel 7 with GrapheneOS. Android 16, webview v143.0.7499.146.0

### LINKED ISSUE #8343: [Feature]: use native safe area insets
### Description

Deprecate the old way of using safe area insets (custom vars) and use the native approach instead

### Platforms

- [ ] iOS
- [ ] Android
- [ ] Web

Would you be willing to do this? We've discussed this before. For example [here](https://github.com/ionic-team/capacitor/pull/8268#issuecomment-3613142117) and [here](https://github.com/ionic-team/capacitor/pull/8180#issuecomment-3520458761).

The current approach taken by Capacitor is quite different than the native approach and to me it seems inferior.  I've outlined the differences [here](https://github.com/capacitor-community/safe-area?tab=readme-ov-file#differences-between-this-plugin-system-bars-and-status-bar).

The main reason you took this approach, was

> For us, the issue isn’t spec compliance — it’s whether this is reliable enough to serve as a runtime contract across OEMs, OS versions, etc.

However, as far as I can see, it seems, the current approach achieves quite the opposite. There were a few bugs that needed to be fixed (while it already was marked as production-ready) and there are still other open bugs ([ex. 1](https://github.com/ionic-team/capacitor/issues/8329), [ex. 2](https://github.com/ionic-team/capacitor/issues/8287)).

What's also bothering me, is that the code has a few quirks.

- The code doesn't bother for Android versions prior v15, which makes older Android versions just break without a developer directly noticing ([source](https://github.com/ionic-team/capacitor/pull/8268#discussion_r2588782642))
  - <sup>This seems to contradict the reason why you chose to go this route btw</sup>
- When installing Capacitor v8 fresh, a developer would never know (without reading the docs thoroughly) that his app is broking on Chromium versions < M140 ([source](https://github.com/ionic-team/capacitor/pull/8268#discussion_r2605888892))

There are a few other code changes, which I do not quite understand ([ex. 1](https://github.com/ionic-team/capacitor/pull/8268#discussion_r2605905909), [ex. 2](https://github.com/ionic-team/capacitor/pull/8268#discussion_r2605911761), [ex. 3](https://github.com/ionic-team/capacitor/pull/8268#discussion_r2638910919))

Meanwhile, the https://github.com/capacitor-community/safe-area seems to work without any problems on a few enterprise-sized production apps. With one of its advantages being that it doesn't need any configuration changes when integrated in the Capacitor core.

I, myself, do not see any reason to not switch strategies. But if you're not up for it, you can close this issue and I will stop pursuing this going forward. Instead I will then just keep maintaining the community plugin

### Alternatives

_No response_

### Additional Information

_No response_

theproducer:
## Description
This PR implements native safe area inset handling as demonstrated in https://github.com/capacitor-community/safe-area

(CSS Safe Area vars are still intact for OS upstream)

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed
closes https://github.com/ionic-team/capacitor/issues/8343
closes https://github.com/ionic-team/capacitor/issues/8329
closes https://github.com/ionic-team/capacitor/issues/8287

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -82,12 +82,6 @@ public class SystemBars extends Plugin {
     }
 
     @Override
-    protected void handleOnResume() {
-        super.handleOnResume();
-        getBridge().executeOnMainThread(() -> getBridge().getWebView().requestApplyInsets());
-    }
-
-    @Override
     protected void handleOnConfigurationChanged(Configuration newConfig) {
         super.handleOnConfigurationChanged(newConfig);
 
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -73,8 +73,8 @@ public class SystemBars extends Plugin {
             new WebViewListener() {
                 @Override
                 public void onPageCommitVisible(WebView view, String url) {
-                super.onPageCommitVisible(view, url);
-                getBridge().getWebView().requestApplyInsets();
+                    super.onPageCommitVisible(view, url);
+                    getBridge().getWebView().requestApplyInsets();
                 }
             }
         );
@@ -194,24 +194,30 @@ public class SystemBars extends Plugin {
                 // We need to correct for a possible shown IME
                 v.setPadding(0, 0, 0, keyboardVisible ? imeInsets.bottom : 0);
 
-                return new WindowInsetsCompat.Builder(insets).setInsets(
-                    WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout(),
-                    Insets.of(
-                        systemBarsInsets.left,
-                        systemBarsInsets.top,
-                        systemBarsInsets.right,
-                        getBottomInset(systemBarsInsets, keyboardVisible)
+                return new WindowInsetsCompat.Builder(insets)
+                    .setInsets(
+                        WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout(),
+                        Insets.of(
+                            systemBarsInsets.left,
+                            systemBarsInsets.top,
+                            systemBarsInsets.right,
+                            getBottomInset(systemBarsInsets, keyboardVisible)
+                        )
                     )
-                ).build();
+                    .build();
             }
 
             // We need to correct for a possible shown IME
-            v.setPadding(systemBarsInsets.left, systemBarsInsets.top, systemBarsInsets.right, keyboardVisible ? imeInsets.bottom : systemBarsInsets.bottom);
+            v.setPadding(
+                systemBarsInsets.left,
+                systemBarsInsets.top,
+                systemBarsInsets.right,
+                keyboardVisible ? imeInsets.bottom : systemBarsInsets.bottom
+            );
 
-            return new WindowInsetsCompat.Builder(insets).setInsets(
-                WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout(),
-                Insets.of(0, 0, 0, 0)
-            ).build();
+            return new WindowInsetsCompat.Builder(insets)
+                .setInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout(), Insets.of(0, 0, 0, 0))
+                .build();
         });
     }
 
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -103,7 +103,7 @@ public class SystemBars extends Plugin {
         }
 
         initWindowInsetsListener();
-        initSafeAreaInsets();
+        initSafeAreaCSSVariables();
 
         getBridge().executeOnMainThread(() -> {
             setStyle(style, "");
@@ -166,7 +166,7 @@ public class SystemBars extends Plugin {
         return Insets.of(safeArea.left, safeArea.top, safeArea.right, safeArea.bottom);
     }
 
-    private void initSafeAreaInsets() {
+    private void initSafeAreaCSSVariables() {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM && insetHandlingEnabled) {
             View v = (View) this.getBridge().getWebView().getParent();
             WindowInsetsCompat insets = ViewCompat.getRootWindowInsets(v);
@@ -185,7 +185,7 @@ public class SystemBars extends Plugin {
             Insets imeInsets = insets.getInsets(WindowInsetsCompat.Type.ime());
             boolean keyboardVisible = insets.isVisible(WindowInsetsCompat.Type.ime());
 
-            if (hasViewportCover) {
+            if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM && hasViewportCover && insetHandlingEnabled) {
                 Insets safeAreaInsets = calcSafeAreaInsets(insets);
                 injectSafeAreaCSS(safeAreaInsets.top, safeAreaInsets.right, safeAreaInsets.bottom, safeAreaInsets.left);
             }
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -36,6 +36,9 @@ public class SystemBars extends Plugin {
     static final String INSETS_HANDLING_CSS = "css";
     static final String INSETS_HANDLING_DISABLE = "disable";
 
+    private static final int WEBVIEW_VERSION_WITH_SAFE_AREA_FIX = 140;
+    private static final int WEBVIEW_VERSION_WITH_SAFE_AREA_KEYBOARD_FIX = 144;
+
     static final String viewportMetaJSFunction = """
         function capacitorSystemBarsCheckMetaViewport() {
             const meta = document.querySelectorAll("meta[name=viewport]");
@@ -70,11 +73,17 @@ public class SystemBars extends Plugin {
             new WebViewListener() {
                 @Override
                 public void onPageCommitVisible(WebView view, String url) {
-                    super.onPageCommitVisible(view, url);
-                    getBridge().getWebView().requestApplyInsets();
+                super.onPageCommitVisible(view, url);
+                getBridge().getWebView().requestApplyInsets();
                 }
             }
         );
+    }
+
+    @Override
+    protected void handleOnResume() {
+        super.handleOnResume();
+        getBridge().executeOnMainThread(() -> getBridge().getWebView().requestApplyInsets());
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -169,41 +178,47 @@ private void initSafeAreaInsets() {
     }
 
     private void initWindowInsetsListener() {
```
OS-pedrogustavobilro, line 175: I could be mistaken, but just want to confirm: Can't this create a situation where native insets get applied + css variables are injected.

If css variables are consumed in the app, wouldn't this possibly create a double padding?

Maybe this type of situation doesn't actually happen, but from looking at the code it seems possible.

theproducer: Yeah, it could happen on a web view less than 140 (I verified on device).  Should be fixed now.

OS-pedrogustavobilro: > a situation where native insets get applied + css variables are injected.

I'm getting an issue here that seems related to this, although from my tests it's both happening with the latest version of Capacitor and this PR, so it's not something introduced in this PR, but would think it could be fixed in this PR (but maybe not). I've checked Android 11 Samsung and Android 12 Nokia, both with System WebView 145, both have this issue. I didn't reproduce in an emulator in these conditions. Not happening on Android 16 from what I tested.

The issue is basically: the first time you open the app, everything's fine. If you kill the app via app switcher, then re-open, you will see some additional spacing below the status bar (like it's applying insets twice). I also tested trying to rotate the device while in the splash screen, and that causes additional padding on the side as well.

I used https://github.com/aeharding/cap-kb-blank-space (from #8329), but maybe this can be reproduced with other apps.

Video from Nokia Android 12 (you can see the issue I'm saying right at the beginning): https://github.com/user-attachments/assets/e96633fa-096f-4360-ad79-2623eba2171e

Screenshot from Samsung Android 11:  
samsung_android11_second_app_open

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -36,6 +36,9 @@ public class SystemBars extends Plugin {
     static final String INSETS_HANDLING_CSS = "css";
```
OS-pedrogustavobilro, line 35: This field isn't being used anymore in this PR, is it by design? I would assume that injecting CSS variables would still be conditional to the capacitor config property.

Also there's an unused `import android.view.ViewGroup;`

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
OS-pedrogustavobilro: Update: It doesn't even seem to be just running the second time, looks like I'm getting a difference behavior when running from Android Studio / Adb and opening the app normally.

I cleared app data, and launched it regularly, and the issue happened, didn't need to close and re-open. Strange 🤔

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
OS-pedrogustavobilro: > This field isn't being used anymore in this PR

Nevermind, going back to 8.0.0 this was always like this, we just look at the disable option, otherwise we assume css variables are being injected

OS-pedrogustavobilro: But also, does this change the meaning of `disabled` property? Because it mentions [in docs](https://capacitorjs.com/docs/config#schema): 

> `disable` = Disable all inset handling.

Correct me if I'm wrong but now native insets are always applied, so `disable` means more disable css injection?

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -77,6 +80,12 @@ public void onPageCommitVisible(WebView view, String url) {
         );
     }
 
+    @Override
+    protected void handleOnResume() {
+        super.handleOnResume();
+        getBridge().executeOnMainThread(() -> getBridge().getWebView().requestApplyInsets());
+    }
```
OS-pedrogustavobilro: Wondering the rationale behind this. From what I see an equivalent doesn't seem to exist in the community safe-areas plugin? Was it to fix a particular open issue?

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -36,6 +36,9 @@ public class SystemBars extends Plugin {
     static final String INSETS_HANDLING_CSS = "css";
     static final String INSETS_HANDLING_DISABLE = "disable";
 
+    private static final int WEBVIEW_VERSION_WITH_SAFE_AREA_FIX = 140;
+    private static final int WEBVIEW_VERSION_WITH_SAFE_AREA_KEYBOARD_FIX = 144;
```
OS-pedrogustavobilro, line 41: Minor: For historical purposes / ease of consulation, wouldn't mind having the links to the chromium issues like the safe-area community plugin has. Unless we don't really do that for Capacitor codebase.

https://issues.chromium.org/issues/40699457 and https://issues.chromium.org/issues/457682720 for 140 and 144 respectively.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -169,41 +178,54 @@ private void initSafeAreaInsets() {
...
-    private void setViewMargins(View v, Insets insets) {
-        ViewGroup.MarginLayoutParams mlp = (ViewGroup.MarginLayoutParams) v.getLayoutParams();
-        mlp.leftMargin = insets.left;
-        mlp.bottomMargin = insets.bottom;
-        mlp.rightMargin = insets.right;
-        mlp.topMargin = insets.top;
-        v.setLayoutParams(mlp);
+            WindowInsetsCompat newInsets = new WindowInsetsCompat.Builder(insets)
+                .setInsets(WindowInsetsCompat.Type.systemBars() | WindowInsetsCompat.Type.displayCutout(), Insets.of(0, 0, 0, 0))
+                .build();
```
OS-pedrogustavobilro, line 218: Minor: Wondering if worth stating the comment as in safe-areas community plugin. 

```java
            // Returning `WindowInsetsCompat.CONSUMED` breaks recalculation of safe area insets
            // So we have to explicitly set insets to `0`
            // See: https://issues.chromium.org/issues/461332423
```

OS-pedrogustavobilro (commented):
A few comments on stuff I noticed while looking at the existing community safe-areas plugin.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
OS-pedrogustavobilro: Update 2: this happens when setting `SystemBars#insetsHandling` to both `css` (default) and `disable`, so this might really be an unrelated bug.

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
theproducer: No, this was part of some tinkering I was doing earlier, thanks for calling it out!

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
theproducer: Yeah...I was able to reproduce this as well.  Lets consider this a separate issue.

OS-pedrogustavobilro: Note: We aligned internally that this is a separate issue, and we will create some backlog to look into this later.

Resolving this thread.

OS-pedrogustavobilro (approved):
Tested on several apps shared across different GitHub issues, as well as a few internal apps, all seems to be working well. Very nice!

tafelnl:
Awesome! Thank you so much for working on this and merging it! 🚀

A few points I'd like to discuss:

- what is the main reason for using `DOMContentLoaded` over just using `onPageCommitVisible`?
   <sup>see also: https://github.com/ionic-team/capacitor/pull/8268#discussion_r2605905909</sup>
   - Does this approach also work when navigating to content inside the webview that hasn't Capacitor loaded in?
- If you prefer using `DOMContentLoaded` and making JS responsible for this, couldn't we then skip the overhead of using `evaluateJavascript` and just let the native code know the value directly
   <sup>see also: https://github.com/ionic-team/capacitor/pull/8268#discussion_r2605911761</sup>
- The initial value of `hasViewportCover` is set to `false`. In my testing this could lead to slightly jumping of content when the value switches from `false` to `true`. It might be good to set it to `true` (if the majority of users has that as the default). Or even better: allow to configure this initial value. For the community plugin I chose `initialViewportFitCover`
- I think it's wise to remove all of the `setOnApplyWindowInsetsListener` related logic inside the `@capacitor/keyboard` plugin.
   <sup>see also: https://github.com/ionic-team/capacitor-keyboard/pull/57 and https://github.com/ionic-team/capacitor-keyboard/pull/57#issuecomment-4080337674</sup>
- We might want to update the documentation. Stating something along the lines of:
   > "Capacitor v8 takes care of patching the safe area related bugs in Chromium out of the box. Using `viewport-fit="cover"` should work as expected. Alternatively you could use CSS variable injection if you don't want to rely on the natively reported inset envs"

Thanks again for your work on this. Cheers!

optimix:
I tested today's version of `SystemBars.java` and noticed that the webView is now always becoming smaller when the keyboard is being shown (compared to the version 8.2.0).
Is there any setting to display the keyboard on top of the webView and leave the webView size unchanged ?
I thought that setting `android:windowSoftInputMode="adjustNothing"` would help achieving this but this is not the case.

optimix:
I had an issue with the latest code: the UI was jumping at startup because `shouldPassthroughInsets` is `false` at the beginning and then set to `true` once the `onDOMReady()` and then `runOnUiThread()` finish setting `hasViewportCover`.
I did a small change on my end to bypass this: https://github.com/ionic-team/capacitor/commit/745560ca4676e436701eda46d66277e3ed799ed9

In order to programatically control the behaviour of resizing the web view or not when the keyboard changes, here is an example with a new `setResizeWebViewMode()` function allowing to control this:
https://github.com/ionic-team/capacitor/commit/0b4283d50c83b2d397060daa958efcd19e533413

piotrblasiak:
@optimix same problem here with `adjustNothing`. Opened #8611 with a reproduction and #8612 adding an opt-in `keyboardInsetsHandling: "none"`. Note that removing the IME padding alone is not enough on WebView >= 139: the IME insets also have to be zeroed before they reach the WebView, otherwise it resizes its visual viewport and the page becomes pannable.

tafelnl:
Keep in mind that the code at the time of optimix' commenting was fundamentally different to how the code behaves on the current version of main

PR #8403, opened 2026-03-26
fix(http): correctly serialize Blob objects in request bodies and FormData

vamsi2246:
## Bug Report / Context
When using `CapacitorHttp` to intercept requests, dynamically generated `Blob` objects appended to `FormData` or passed as the request body are silently corrupted. Because the native bridge only checks `if (value instanceof File)`, generic `Blob`s fall through the type check and are improperly serialized as the literal string `"[object Blob]"`. 

## What This PR Does
Changes the strict `File` instance checks to `Blob` in `native-bridge.ts`. Since all `File` objects are instances of `Blob`, and `FileReader.readAsBinaryString()` natively accepts a `Blob`, this safely fixes the data corruption for both pure `Blob`s and `File`s without introducing any breaking changes.

- Replaced `instanceof File` with `instanceof Blob` in `convertFormData` and `convertBody`.
- Updated `readFileAsBase64` signature to accept `Blob`.
- Added a fallback `fileName` for generic Blobs that do not inherently have a `.name` property.

### DIFF
--- a/android/capacitor/src/main/assets/native-bridge.js
+++ b/android/capacitor/src/main/assets/native-bridge.js
@@ -48,14 +48,14 @@ var nativeBridge = (function (exports) {
         const newFormData = [];
         for (const pair of formData.entries()) {
             const [key, value] = pair;
-            if (value instanceof File) {
+            if (value instanceof Blob) {
                 const base64File = await readFileAsBase64(value);
                 newFormData.push({
                     key,
                     value: base64File,
                     type: 'base64File',
                     contentType: value.type,
-                    fileName: value.name,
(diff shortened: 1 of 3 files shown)


jcesarmobile:
can you first create an issue with a sample app that reproduces this problem?

riderx:

Patch IDs: `upstream-pr-8403-android`, `upstream-pr-8403-ios`

```ts
plugins: {
  CapacitorPatch: {
    patches: [
      'upstream-pr-8403-android',
      'upstream-pr-8403-ios',
    ],
    strict: true,
  },
}
```

PR #8404, opened 2026-03-26
fix(cli): properly copy all valid SSL certs and fix log typo

vamsi2246:
Currently, `copySSLCert` returns early if any given certificate path is invalid, preventing other valid certificates from being copied. Also, there is a typo in the log message. This PR fixes both.


jcesarmobile:
please, before you continue creating pull requests, create issues and provide reproductions

PR #8405, opened 2026-03-26
fix(core): properly propagate rejections in addListenerNative

vamsi2246:
Currently, `addListenerNative` creates a new promise that only handles the success case of the underlying `addListener` call, causing it to hang on failure. This PR refactors it to use `.then()` on the original call, ensuring that rejections are correctly propagated. This also fixes an unhandled rejection.


jcesarmobile:
please, before you continue creating pull requests, create issues and provide reproductions

PR #8417, opened 2026-03-30
chore: fix various typos discovered by codespell

ankitkumar572005:
Fixed various typos across the codebase found using codespell (e.g., datebase -> database, inital -> initial, compatable -> compatible).


OS-ruimoreiramendes:
Hey @ankitkumar572005, thanks for the contribution! While reviewing this PR, we found that the changes to the test files incorrectly added a space to `afterAll()`, turning it into `after all()` which breaks the syntax. Could you fix that in the following files?

- cli/test/add.android.spec.ts
- cli/test/add.ios.spec.ts
- cli/test/init.spec.ts
- cli/test/update.android.spec.ts
- cli/test/update.ios.spec.ts

Everything else looks good! Thanks again.

core/cordova.js
```diff
@@ -621,7 +621,7 @@
     };
     var channel = {
       /**
-       * Calls the provided function only after all of the channels specified
+       * Calls the provided function only afterAll of the channels specified
```
OS-ruimoreiramendes, line 624: This change doesn’t make sense, since it’s not referring to any testing hooks. It should remain as “after all”.

PR #8369, opened 2026-03-01
fix(android): invalid http range seeking

bwees:
Fixes a regression introduced in #8357.

Closes #8367.

Issue #8367, opened 2026-03-01
[Bug]: PR #8357 Regression


bwees:
### Capacitor Version

8.1.1-nightly-20260227T151709.0

### Other API Details

```Shell

```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

The addition of 

```
int bytesToSkip;
try {
   bytesToSkip = Integer.parseInt(fromRange);
   if (bytesToSkip > 0) {
       responseStream.skip(bytesToSkip);
   }
} catch (NumberFormatException ignored) {}
```

in PR #8357 improperly seeks the responseStream an additional bytesToSkip. 

I have checked the binary output via chrome devtools on 8.1.0 and then on 8.1.1 nightly and compared the outputs to a curl request against a vite dev server (which is my "source of truth" here since it properly serves the range requests when using it for hot reload).

#### Vite Dev Server (first 16 bytes)

This is the "expected" output from Capacitor server

```
curl 'http://localhost:5173/map/cstat.pmtiles' \
  -H 'sec-ch-ua-platform: "Android"' \
  -H 'Referer: https://localhost/' \
  -H 'User-Agent: Mozilla/5.0 (Linux; Android 16; SM-S911U1 Build/BP2A.250605.031.A3; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/142.0.7444.102 Mobile Safari/537.36' \
  -H 'sec-ch-ua: "Chromium";v="142", "Android WebView";v="142", "Not_A Brand";v="99"' \
  -H 'range: bytes=389189-420083' \
  -H 'sec-ch-ua-mobile: ?1' --output test.pmtiles
```

Output (first 16 bytes): `1F 8B 08 00 00 00 00 00 00 FF 7C BC 09 98 54 D5`

#### Android on Capacitor 8.1.0 (same byte range as vite server)

Output (first 16 bytes): `1F 8B 08 00 00 00 00 00 00 FF 7C BC 09 98 54 D5`

#### Android on Capacitor 8.1.1-nightly-20260227T151709.0 (same byte range as vite server)

Output (first 16 bytes): `77 B0 9C B5 45 DE 28 50 43 EA E8 48 F3 54 B1 9D`

### Expected Behavior

The range request should seek to the correct position in the file as it did in v8.1.0. 

### Project Reproduction

See the current behavior.

### Additional Information

[...]

PR #8357, opened 2026-02-24
fix(android): Concurrent Range Requests for assets

biguphpc:
Rebased against the latest main Capacitor branch and updated to resolve merge conflicts for PR #5956.
Checking quickly for requirement in https://github.com/ionic-team/capacitor/pull/5956#issuecomment-3954571388

PR #8418, opened 2026-04-01
fix(android): range request truncation

### LINKED ISSUE #8371: [Bug]: HTTP Range Responses not within spec
### Capacitor Version

8.1.0

### Other API Details

```Shell

```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

This issue has been reported: https://github.com/ionic-team/capacitor/issues/7151#issuecomment-1897113728

When constructing the response for an HTTP Range request, the InputStream is not properly truncated at the end of the requested range. This results in the response including data from the specified start index, all the way to the end of the file.

The results of this failed truncation are as follows:
- The `Content-Length` response header gives a byte length that would be correct if the InputStream was truncated
- The `Content-Range` response header gives a byte range that would be correct if the InputStream was truncated

From my understanding of the HTTP Range spec, if the server plans to send the remainder of the file in the response (which is the current behavior of the Capacitor implementation), the Content-Length and Content-Range must be updated to reflect the extra returned length and ending byte range (length - 1).

This appears to cause issues specifically with the PMTiles decoder since it appears to rely on the response headers to understand how much data it received.

#### Reproduction Steps

1. Install dependencies and build app with pnpm (crucial for patches to apply)

 ```bash
pnpm i
pnpm build
pnpm cap sync android
pnpm cap run android
```

3. Using the demo app, observe a `cstat.pmtiles` request from within Chrome Devtools. 
4. You will see a Content-Length header in the response.
5. Convert the Content-Length value to hexadecimal
6. View the Response data tab in Devtools
7. Go to the Hexadecimal index you calculated in Step 3.
8. See that there is lots of data returned that exceeds the Content-Length. 
9. You will see that the data continues to the end of the file (`index = totalFileLength - startRange - 1` => Convert to Hex)
#### Example

Request Range: `bytes=850902-889367`
Response `Content-Length` Header: `38466` (0x00009642)
Response `Content-Range` Header: `bytes 850902-889367/3949417`

**Response Data @ 0x00009642**

Image

**Response Data @ 0x002F4792 (See step 7: 3949417-850902-1)**

Image

### Expected Behavior

The stream is properly truncated at the specified byte index or the response headers match the fact that the remainder of the file is sent. Ideally the stream is properly truncated to minimize large data transfer.

### Project Reproduction

https://github.com/bwees/range-request-capacitor-repro/tree/fix/lowercase-range

UPDATE: March 3 17:00 UTC, the example project was broken and has since been fixed

### Additional Information

The provided project branch includes a pnpm patch that resolves issues that are fixed in #8368. This is needed to properly debug this problem as otherwise the range response headers are not sent.

I also took multiple attempts at creating a PR to solve this issue but could not get the stream to properly truncate without a ERR:FAILED error being thrown or the Content-Length header being overridden before response to the client. Maybe someone more versed in the Web Server implementation has some insight but I was not able to solve it

bwees:
## Description

Range requests on Android did not properly truncate their streams leading to the remainder of the file being returned instead of only the specified range. The stream was properly seeked (seemingly automatically as discovered in #8369) but continued to respond with the rest of the file.

This PR creates a wrapper around InputStream that properly sends EOF once the specified number of bytes has been read.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Closes https://github.com/ionic-team/capacitor/issues/8371

## Tests or Reproductions

The minimum reproduction app now loads the map with the patch applied in https://github.com/bwees/range-request-capacitor-repro/tree/range-fix-repro.

I also added additional tests for the BoundedInputStream class

## Screenshots / Media

N/A
## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

## Notes / Comments
Claude was used to build the BoundedInputStream class and corresponding tests were written to ensure accuracy.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/WebViewLocalServer.java
@@ -348,16 +348,22 @@ public class WebViewLocalServer {
             Map<String, String> tempResponseHeaders = handler.buildDefaultResponseHeaders();
             int statusCode = 206;
             try {
-                int totalRange = responseStream.available();
+                int totalSize = responseStream.available();
                 String[] parts = rangeString.split("=");
                 String[] streamParts = parts[1].split("-");
-                String fromRange = streamParts[0];
-                int range = totalRange - 1;
-                if (streamParts.length > 1) {
-                    range = Integer.parseInt(streamParts[1]);
+                int fromRange = Integer.parseInt(streamParts[0]);
+                int endRange = totalSize - 1;
+                if (streamParts.length > 1 && !streamParts[1].isEmpty()) {
+                    endRange = Integer.parseInt(streamParts[1]);
                 }
+
+                // Truncate the stream at the end of the requested range.
+                // Somewhere in the request pipeline, the stream gets automatically
+                // seeked to the correct start position, so we don't need to skip to the fromRange position here.
+                responseStream = new BoundedInputStream(responseStream, endRange + 1);
+
                 tempResponseHeaders.put("Accept-Ranges", "bytes");
-                tempResponseHeaders.put("Content-Range", "bytes " + fromRange + "-" + range + "/" + totalRange);
+                tempResponseHeaders.put("Content-Range", "bytes " + fromRange + "-" + endRange + "/" + totalSize);
             } catch (IOException e) {
                 statusCode = 404;
             }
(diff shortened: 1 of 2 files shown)


bviksoe:
This fixed a problem I had embedding a .pmtiles tile-map file for offline viewing of map data.

bwees:
Hey there is there any updates on if this PR can be merged? I am happy to make any additional changes that the team requests.

Jin1c-3:
Third app in this thread, for whatever the pile-on is worth: I maintain an Obsidian plugin that draws offline basemaps, and it lands on the same `_capacitor_file_` path with the same PMTiles-shaped problem.

Measured against a 100 MB archive on Android 16 / WebView 133, with byte `i` written as `i % 251` so the offsets are actually checkable:

| read | body actually delivered | time |
| --- | --- | --- |
| 127-byte header at offset 0 | **104,857,600 bytes** | 286 ms |
| 10 bytes at offset 90,000,000 | 14,857,600 bytes | 64 ms |
| 50 scattered 1 KB tile reads | **2.73 GB** | 5,619 ms |

The first row is the one that gets you. A PMTiles header sits at offset 0, so the very first read of the archive drags the whole thing through the renderer — 127 bytes asked for, 100 MB delivered, before you've drawn a single tile.

I arrived at the comment in your diff from the other end: the `skip()` in #8357 double-offsetting (#8367) only makes sense if something below the Java layer is already seeking off `Content-Range`. Good to see the patch say that out loud instead of working by accident.

We ended up with a renderer-side workaround — read the stream, stop once you have the requested length, `cancel()` — which pulls one 64–256 KB chunk instead of the remainder of the file. Those 50 tile reads go from 2.73 GB / 5.6 s to about 3 MB / 67 ms. It only works because the *start* offset is honoured, so the bytes are correct, just absurdly over-delivered. It isn't a fix: anything that trusts `Content-Length` still gets a malformed response, and everyone downstream pays for a workaround they shouldn't have to write.

@bwees I wouldn't write it off. #8271 sat open since December and landed last month, and a couple of other outside PRs merged within a day or two recently. One commit, tests, no conflicts — as easy as a review gets.

PR #7571, opened 2024-07-19
fix(android): respect all allowed origins on modern WebViewCompat JS injection

martin-braun:
Fixes #7454

I applied the fix from @rfe-css at https://github.com/ionic-team/capacitor/issues/7454, but I figured my device would skip this, so please someone can verify this before merge.

PR #8429, opened 2026-04-10
fix: Fix missing `window.Capacitor` on Android after redirection

### LINKED ISSUE #7454: [Bug]: Capacitor on Android detected as web platform after redirecting with `allowNavigation`
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 6.0.0
  @capacitor/core: 6.0.0
  @capacitor/android: 6.0.0
  @capacitor/ios: 6.0.0

Installed Dependencies:

  @capacitor/cli: 6.0.0
  @capacitor/core: 6.0.0
  @capacitor/ios: 6.0.0
  @capacitor/android: 6.0.0

### Other API Details

```Shell
`npm --version`
> 10.7.0

`node --version`
> v20.11.0
```

### Platforms Affected

- [ ] iOS
- [X] Android
- [ ] Web

### Current Behavior

After redirecting to an external URL that is allowed in the `server.allowNavigation` config setting, an android app reports being on a web platform instead of android.
This can be seen by calling `Capacitor.getPlatform()` (which returns `android` before redirecting and `web` after redirecting) or by trying to use a native API, which will either fallback to the web version of the API, or fail to run at all

All this works on iOS, where after redirecting, the app can still use native iOS APIs

### Expected Behavior

After redirecting, the android app should still have access to all the native Android APIs (and report itself as being an Android platform)

### Project Reproduction

https://github.com/Dylancyclone/capacitor-android-redirect/tree/main

### Additional Information

Reproduction details are in the repo, but the basic reproduction steps are as follows:

1. Serve a valid capacitor app on your computer (i.e. with `npx serve .`)
2. Add your computer's IP to the `server.allowNavigation` capacitor config setting
3. Load the application on your (real or simulated) device
4. Notice that `Capacitor.getPlatform()` returns `android` on android and `ios` on iOS
5. Redirect to the hosted server (any method works, can be `window.location.assign()`, `window.location.href = ""`, `window.location.replace()`, or redirecting through a form
6. Notice that `Capacitor.getPlatform()` now returns `web` on android but still `ios` on iOS

Possibly related to https://github.com/ionic-team/capacitor/issues/5455

Here are videos showing the issue. Even though these were recorded on a simulator, the results are identical to a physical device

https://github.com/ionic-team/capacitor/assets/19830705/076fc893-1732-4bab-9aea-b7a90b4ec142

https://github.com/ionic-team/capacitor/assets/19830705/4522dcef-bbdb-42e4-8085-7009f55176bf

Our use case is to have one app that connects to different backends on different subdomains, very similar to Slack. So far everything is working except the android app loses access to its native APIs

caffeinated10xprogrammer:
## Description
This fixes #7454 with rebased code from PR #7571

## Change Type
- [x ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed
Fix missing `window.Capacitor` on Android after redirection

## Tests or Reproductions
Refer to #7454 

## Platforms Affected
- [ x] Android
- [ ] iOS
- [ ] Web

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/Bridge.java
@@ -265,8 +265,13 @@ public class Bridge {
         JSInjector injector = getJSInjector();
         if (WebViewFeature.isFeatureSupported(WebViewFeature.DOCUMENT_START_SCRIPT)) {
             String allowedOrigin = Uri.parse(appUrl).buildUpon().path(null).fragment(null).clearQuery().build().toString();
+            final String finalAllowedOrigin = allowedOrigin;
+            Set<String> allowedOrigins = new HashSet<String>() {{
+                add(finalAllowedOrigin);
+                addAll(allowedOriginRules);
+            }};
             try {
-                WebViewCompat.addDocumentStartJavaScript(webView, injector.getScriptString(), Collections.singleton(allowedOrigin));
+                WebViewCompat.addDocumentStartJavaScript(webView, injector.getScriptString(), allowedOrigins);
                 injector = null;
             } catch (IllegalArgumentException ex) {
                 Logger.warn("Invalid url, using fallback");


damrbaby:
We hit this same issue after upgrading to Capacitor 8 on Android. Our app uses allowNavigation to navigate to a remote URL from a local index.html, and window.Capacitor was no longer available on the remote page. We are using the same patch and it resolves the issue. Would love to see this merged so we can drop the patch.

PR #8438, opened 2026-04-23
fix(ios): support Cordova plugins with Package.swift

jcesarmobile:
If the Cordova plugin already has a Package.swift, point to that file instead of generating a new one, but patching the repository url and version to point to the Capacitor one, also, in that case, don't copy the plugin files to `capacitor-cordova-ios-plugins` folder since it will be using the ones in `node_modules`

Also fixes an issue where the plugin files were copied to `capacitor-cordova-ios-plugins/sourcesstatic/pluginName` but the `Package.swift` was being generated, and pointed to `capacitor-cordova-ios-plugins/sources/pluginName` and failed to link the plugin since it had no files. The `sourcesstatic` folder is only needed for some CocoaPods plugins, not when using SPM.

### DIFF
--- a/cli/src/util/spm.ts
+++ b/cli/src/util/spm.ts
@@ -11,7 +11,7 @@ import { fatal } from '../errors';
 import { getMajoriOSVersion } from '../ios/common';
 import { logger } from '../log';
 import type { Plugin } from '../plugin';
-import { getPluginType, PluginType } from '../plugin';
+import { getPluginPlatform, getPluginType, PluginType } from '../plugin';
 import { runCommand } from '../util/subprocess';
 
 export interface SwiftPlugin {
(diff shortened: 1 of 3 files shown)


cli/src/util/spm.ts
OS-pedrogustavobilro, line 65: Wondering if this needs to be changed as well, right now testing on an app that contains a Cordova Plugin with `Package.swift`, we get `⠙ update ios [warn] Some installed packages are not compatable with SPM` (also there's a typo on "compatable")

OS-pedrogustavobilro (commented):
I'm still running more tests with different apps and combinations of plugins, but had noticed this here and just asking to see if it's relevant.

cli/src/util/spm.ts
jcesarmobile: The `checkPluginsForPackageSwift` only checks Capacitor plugins, so it's unrelated to the PR changes, but yeah, there is a bug there that should be fixed and the typo too (in a separate PR). It should check for `iOSCapacitorPlugins.length` instead of `plugins.length` as the later includes Cordova plugins, and the messages should be more clear about being Capacitor plugins.

Or alternatively, we might want to check Cordova plugins with `Package.swift` too, but since we patch the ones without a `Package.swift`, the message could still be confusing as most would still work even if they don't have a `Package.swift`. The `Package.swift` is not a requirement for Cordova plugins, while it is for Capacitor plugins.

OS-pedrogustavobilro: > there is a bug there that should be fixed and the typo too (in a separate PR)

Right, fair enough, I was under the impression it could include Cordova Plugins, but then yeah I'd say for now the simpler thing would be to fix it in a separate PR to apply only to Capacitor, and if we need to apply it for Cordova Plugins as well we can re-evaluate. Anywho, I'll resolve this comment since for this PR it won't require changes.

OS-pedrogustavobilro (approved):
Tested on a few different apps with different plugins (Capacitor, Cordova with Package.swift, Cordova without Package.swift and without any dependencies), it's working good!

OS-ruimoreiramendes (approved):

PR #8443, opened 2026-04-27
fix(cli): handle Cordova plugins without iOS source files

OS-ruimoreiramendes:
- Fix `getPluginPlatform` to return `undefined` instead of `[]` when a plugin has no `<platform>` tags
- Skip `Package.swift` generation for Cordova plugins with `<platform name="ios">` but no source files
- Skip adding those plugins to the generated `CapApp-SPM/Package.swift`

Reference: https://outsystemsrd.atlassian.net/browse/RMET-5159

**Tests**
Validated with three dummy plugins:
`test-plugin-no-platform`: no `<platform>` tags -> correctly skipped as incompatible
`test-plugin-no-sources`: `<platform name="ios">` with only <config-file> entries -> sync completes, config still processed
`test-plugin-with-sources`: `<platform name="ios">` with source files -> continues to work normally

### DIFF
--- a/cli/src/util/spm.ts
+++ b/cli/src/util/spm.ts
@@ -156,21 +156,20 @@ let package = Package(
                 .product(name: "Cordova", package: "capacitor-swift-pm")`;
 
   for (const plugin of plugins) {
+    let pluginText = `,\n                .product(name: "${plugin.ios?.name}", package: "${plugin.ios?.name}")`;
     if (getPluginType(plugin, config.ios.name) === PluginType.Cordova) {
       const platformTag = getPluginPlatform(plugin, config.ios.name);
       if (platformTag.$?.package) {
-        packageSwiftText += `,\n                .product(name: "${plugin.id}", package: "${plugin.id}")`;
+        pluginText = `,\n                .product(name: "${plugin.id}", package: "${plugin.id}")`;
       } else {
         const sourceFiles = getPlatformElement(plugin, config.ios.name, 'source-file');
         const headerFiles = getPlatformElement(plugin, config.ios.name, 'header-file');
         if (sourceFiles.length === 0 && headerFiles.length === 0) {
-          continue;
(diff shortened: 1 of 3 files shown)


jcesarmobile:
if the plugin gets skipped as incompatible, then the copying of the plugin's `.js` files also gets skipped, that's why `getPluginPlatform` is returning `[]` instead of `undefined`, so even if there is no native files, the `.js` files still get copied.
So `getPluginPlatform` shouldn't be changed to return `undefined`.

OS-ruimoreiramendes:
> if the plugin gets skipped as incompatible, then the copying of the plugin's `.js` files also gets skipped, that's why `getPluginPlatform` is returning `[]` instead of `undefined`, so even if there is no native files, the `.js` files still get copied. So `getPluginPlatform` shouldn't be changed to return `undefined`.

confirmed your point. With `undefined`, plugins without a `<platform>` tag were marked as Incompatible and their `.js` files were not copied. With `[]` it works correctly. 
Reverted.

alexgerardojacinto:

cli/src/util/spm.ts
```diff
@@ -151,14 +156,21 @@ let package = Package(
                 .product(name: "Cordova", package: "capacitor-swift-pm")`;
 
   for (const plugin of plugins) {
-    let pluginText = `,\n                .product(name: "${plugin.ios?.name}", package: "${plugin.ios?.name}")`;
```
jcesarmobile, line 154: I added this variable to avoid duplicating the same text twice, since it's a complex text and in case we need to update it in the future, we don't need to update it twice, so please, don't remove it.
You can set the value to `''` when there are no source/header files

OS-ruimoreiramendes: ok ok, will revert this

jcesarmobile (approved):

PR #8445, opened 2026-04-28
fix(cli): generate binaryTarget entries for custom xcframeworks in Package.swift

OS-ruimoreiramendes:
When Capacitor CLI generates a `Package.swift` for a Cordova plugin in SPM mode, custom `.xcframework` files declared via `<framework custom="true" src="..." />` in `plugin.xml` were not being translated into `binaryTarget` entries. The xcframework was already being copied to the correct location by `copyPluginsNativeFiles`, but the generated `Package.swift` had no reference to it, causing build failures.

This PR adds `binaryTarget` entries and the corresponding target dependencies for each custom `.xcframework` found in the plugin's `plugin.xml`.

Reference https://outsystemsrd.atlassian.net/browse/RMET-5155

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -98,6 +98,25 @@ async function generateCordovaPackageFile(p: Plugin, config: Config) {
     );
     await writeFile(packageSwiftPath, content);
   } else {
+    const frameworks = getPlatformElement(p, platform, 'framework');
+    const customXcframeworks = frameworks.filter((f: any) => f.$.custom === 'true' && f.$.src.endsWith('.xcframework'));
+    const binaryTargetsText = customXcframeworks
+      .map((f: any) => {
+        const name = f.$.src.split('/').pop().replace('.xcframework', '');
+        return `,
+        .binaryTarget(
+            name: "${name}",
+            path: "${f.$.src}"
+        )`;
+      })
+      .join('');
+    const binaryDepsText = customXcframeworks
+      .map((f: any) => {
+        const name = f.$.src.split('/').pop().replace('.xcframework', '');
+        return `,\n                .target(name: "${name}")`;
+      })
+      .join('');
+
     const content = `// swift-tools-version: 5.9
 
 import PackageDescription
@@ -118,10 +137,10 @@ let package = Package(
         .target(
             name: "${p.name}",
             dependencies: [
-                .product(name: "Cordova", package: "capacitor-swift-pm")
+                .product(name: "Cordova", package: "capacitor-swift-pm")${binaryDepsText}
             ],
             path: "."${headersText}
-        )
+        )${binaryTargetsText}
     ]
 )`;
 
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -98,17 +98,7 @@ async function generateCordovaPackageFile(p: Plugin, config: Config) {
     );
     await writeFile(packageSwiftPath, content);
   } else {
-    const frameworks = getPlatformElement(p, platform, 'framework');
-    const { binaryTargetsText, binaryDepsText } = buildBinaryTargetEntries(p, frameworks);
-    await writeGeneratedPackageSwift(
-      p,
-      config,
-      iosVersion,
-      iosPlatformVersion,
-      headersText,
-      binaryTargetsText,
-      binaryDepsText,
-    );
+    await writeGeneratedPackageSwift(p, config, iosVersion, iosPlatformVersion, headersText);
   }
 }
 
@@ -151,9 +141,9 @@ async function writeGeneratedPackageSwift(
   iosVersion: string,
   iosPlatformVersion: string,
   headersText: string,
-  binaryTargetsText: string,
-  binaryDepsText: string,
 ) {
+  const frameworks = getPlatformElement(p, platform, 'framework');
+  const { binaryTargetsText, binaryDepsText } = buildBinaryTargetEntries(p, frameworks);
   const content = `// swift-tools-version: 5.9
 
 import PackageDescription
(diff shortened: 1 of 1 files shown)


jcesarmobile (commented):
RMET-5155 is about both .xcframework and .framework, this seems to only handle .xcframework.
Not sure if .framework aren't compatible with binaryTarget or if you missed that.
If they are not compatible we should warn some message if .framework files are present as the plugin won't work.

OS-ruimoreiramendes:
> RMET-5155 is about both .xcframework and .framework, this seems to only handle .xcframework. Not sure if .framework aren't compatible with binaryTarget or if you missed that. If they are not compatible we should warn some message if .framework files are present as the plugin won't work.

Regular `.framework` bundles with `custom="true"` are not supported by SPM's `binaryTarget`, only `.xcframework` is. I'll add a `logger.warn` to inform the developer when a custom `.framework` is found, since they'd need to convert it to `.xcframework` for SPM compatibility.
Also, I'm planning to push the RMET-5154 changes (system frameworks - `linkerSettings`) to this same branch, since both tasks are related to framework handling. Let me know if you'd prefer a separate PR for that.

jcesarmobile:
I prefer to keep them separate

cli/src/ios/update.ts
```diff
@@ -98,6 +98,32 @@ async function generateCordovaPackageFile(p: Plugin, config: Config) {
     );
     await writeFile(packageSwiftPath, content);
   } else {
+    const frameworks = getPlatformElement(p, platform, 'framework');
```
OS-pedrogustavobilro: I reckon probably as this method may grow, we may want to split it in a few ones, e.g. one to get the xcframework(s) and binary target, another to write `Package.swift`, and stuff like that.

But doesn't necessarily need to happen in this PR.

cli/src/ios/update.ts
```diff
@@ -98,6 +98,32 @@ async function generateCordovaPackageFile(p: Plugin, config: Config) {
...
+    if (customFrameworks.length > 0) {
+      customFrameworks.forEach((f: any) => {
+        logger.warn(
+          `${p.id}: custom .framework files are not supported as binaryTarget in SPM (${f.$.src}). Convert to .xcframework for SPM compatibility.`,
+        );
+      });
+    }
+    const binaryTargetsText = customXcframeworks
+      .map((f: any) => {
+        const name = f.$.src.split('/').pop().replace('.xcframework', '');
```
OS-pedrogustavobilro: Could this create a situation where the target name becomes duplicate of an existing one? Causing build to fail.

As a silly example I did was making an `.xcframework` file name as the exact same as a Cordova Plugin identifier, which indeed does cause the build to fail.

Just thinking if we should prevent an edge case like this or if it's kind of irrelevant.

OS-pedrogustavobilro (commented):
I've been testing different apps and everything seems to be working, leaving a couple of comments though, but nothing major I think.

cli/src/ios/update.ts
OS-ruimoreiramendes: makes sense, I can do it in this PR.

cli/src/ios/update.ts
OS-ruimoreiramendes: Good point. One approach I'm thinking would be to add a suffix to the binary target name (e.g. `"${name}Binary"` or `"${name}Framework"`) to always avoid any potential collision. However, this would apply even when there's no actual conflict, which may not be ideal. An alternative would be to only add the suffix when a collision is detected. 
What do you think is the best approach here? Or any other suggestion

jcesarmobile: I don't think there are plugins using a .xcframework with the same name as the actual plugin identifier, so let's not worry about it for now

jcesarmobile: Also, I think the name of the target/binaryTarget is the name used in the import, so if we change the binaryTarget name, the import in the code would fail.

cli/src/ios/update.ts
OS-ruimoreiramendes: done, take another look when you can

cli/src/ios/update.ts
OS-ruimoreiramendes: yeah, I initially thought renaming the binaryTarget would be fine since the module name is defined inside the xcframework itself, but you're right that the binaryTarget name is what's actually used for the import. Let's leave it as is.

OS-pedrogustavobilro: Makes sense! Yeah it was kind of far-fetched anyways xD I'll resolve this comment, thanks for the insight guys!

OS-pedrogustavobilro (approved):
Tested again, working great here!

alexgerardojacinto:

cli/src/ios/update.ts
jcesarmobile: Since we will need to add more entries to the `Package.swift`(csettings, resource files, etc.), I would rather not have the `writeGeneratedPackageSwift` method as we will have to keep adding more and more params.
Or if we have it, it should have the methods for getting the texts inside instead of as params, i.e. for this PR, `buildBinaryTargetEntries` should be inside `writeGeneratedPackageSwift` to get the texts, rather than being called before to pass the texts as params.

OS-ruimoreiramendes: Makes sense, moved `buildBinaryTargetEntries` inside `writeGeneratedPackageSwift` so that as more entries are added in future PRs, they can be handled there directly without growing the parameter list.
I'll apply the same for  #8448  as well. Depending on which PR gets merged first there may be a conflict, but it should be straightforward to resolve.

OS-pedrogustavobilro (approved):

PR #8446, opened 2026-04-28
refactor(android-template)!: remove capacitor-cordova-android-plugins module

jcesarmobile:
## Description
This just removes the capacitor-cordova-android-plugins module from the android-template.
Will follow up with a PR that modifies the CLI to conditionally add the module to the project if Cordova plugins are installed.

It's a breaking change since we are modifying the template, migrate command will have to remove it or users will have to manually remove it.

## Change Type
- [ ] Fix
- [ ] Feature
- [x] Refactor
- [ ] Breaking Change
- [ ] Documentation

### DIFF
--- a/android-template/app/build.gradle
+++ b/android-template/app/build.gradle
@@ -24,12 +24,6 @@ android {
     }
 }
 
-repositories {
-    flatDir{
-        dirs '../capacitor-cordova-android-plugins/src/main/libs', 'libs'
-    }
-}
-
 dependencies {
     implementation fileTree(include: ['*.jar'], dir: 'libs')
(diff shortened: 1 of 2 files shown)


OS-pedrogustavobilro (approved):
LGTM. Just a note on "migrate command will have to remove it or users will have to manually remove it", should we create some backlog to make that decision of being either migrate command or docs or both and implement it? Just to make sure we don't forget. Assuming you haven't done that already

jcesarmobile:
Eric created a ticket for the migrate command, while it doesn't explicitly mentions this change or any changes, the work usually consists in checking the changes in android-template and both ios templates and make the script do the same.
That's why we try to do template changes separately, so it's easier to spot what changed between versions.

PR #8447, opened 2026-04-29
fix(cli): add system framework and weak framework support in SPM Package.swift

OS-ruimoreiramendes:
Adds support for translating `<framework src="..." />` declarations from `plugin.xml` into `linkerSettings` entries in the generated `Package.swift`. Frameworks marked as `weak="true"` are handled by setting `type: .dynamic` on the library product, following the same approach used in [ion-ios-camera](https://github.com/ionic-team/ion-ios-camera/commit/123ba380ba06720bb9885c5706f225c6602cf4bf).
Tested with a dummy plugin declaring `SwiftUICore.framework`with `weak="true"`, no crash on iOS 15 simulator, successful build on a physical device.

Reference: https://outsystemsrd.atlassian.net/browse/RMET-5154

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -98,6 +98,20 @@ async function generateCordovaPackageFile(p: Plugin, config: Config) {
     );
     await writeFile(packageSwiftPath, content);
   } else {
+    const frameworks = getPlatformElement(p, platform, 'framework');
+    const systemFrameworks = frameworks.filter((f: any) => !f.$.custom && f.$.src.endsWith('.framework'));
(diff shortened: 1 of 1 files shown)


alexgerardojacinto:

jcesarmobile:
it still shows conflicts

OS-ruimoreiramendes:
> Member

the conflicts are now resolved. They appeared after merging one of the related PRs, but everything is up to date now.

OS-pedrogustavobilro (approved):
LGTM. From my tests this code doesn't seem to add difference for cdv plugins with regular system frameworks (without weak), it seems to work for me without this PR, but from what we discussed internal, that is expected.

PR #8449, opened 2026-04-29
fix(cli): conditionally add capacitor-cordova-android-plugins module

jcesarmobile:
Only link the `capacitor-cordova-android-plugins` module if Cordova plugins are present.
Also skip some Cordova related tasks and some Cordova related gradle lines if there are no Cordova plugins.


cli/src/android/update.ts
```diff
@@ -202,6 +212,18 @@ project(':${getGradlePackageName(p.id)}').projectDir = new File('${relativePlugi
     });
     prefsArray = prefsArray.concat(getAllElements(p, platform, 'preference'));
   });
+  const cdvRepoApplyText = enableCordova
+    ? `
+repositories {
+    flatDir{
+        dirs '../capacitor-cordova-android-plugins/src/main/libs', 'libs'
```
OS-pedrogustavobilro, line 219: Just for my context, this is where stuff like `.aar` declared by cordova plugins in `plugin.xml` go to? Or is it for something else?

cli/src/android/update.ts
OS-pedrogustavobilro, line 1: Just to make sure I understand, Capacitor always comes with `capacitor-cordova-android-plugins` when you run npm install, but with these changes, it will be only be included and used in the app if there's Cordova Plugins?

I was thinking if there was any concerns of dead code in the generated binary, but because the gradle module is not actually there, it should be fine. I mean the code is still dead in the sense that for a Capacitor app with no Cordova Plugins it's not doing anything there, but it's not doing any harm either.

So it seems good. Unless something in my understanding is incorrect xD

OS-pedrogustavobilro (approved):
Overall looks good! I left a couple questions just to improve my understanding 😅

cli/src/android/update.ts
jcesarmobile: I think it's for .jar files and I think it's deprecated and we might end up deleting it for gradle 9 support, but for now just moving here what was removed on https://github.com/ionic-team/capacitor/pull/8446/changes and add it only if cordova plugins are present

cli/src/android/update.ts
jcesarmobile: yeah, `npx cap add android` will untar the `capacitor-cordova-android-plugins` template, but will be basically empty if there are no Cordova plugins, and after https://github.com/ionic-team/capacitor/pull/8446/changes, it no longer appears as a module on the user app, if the module is not there, the files will not be part of the app despite they are in the user's filesystem.

With this PR, it will link the module back whenever there are Cordova plugins installed.

I could also delete the whole folder if there are no plugins, but would require to untar the folder again whenever a Cordova plugin gets installed, and also change the add command to delete the folder as add doesn't run update and I don't think the untar can be prevented.

alexgerardojacinto (approved):

cli/src/android/update.ts
OS-pedrogustavobilro: Yeah that makes sense, thanks for the explanation!

PR #8453, opened 2026-05-01
fix(android): conditionally add Cordova

alexgerardojacinto:
# ⚠️🤖  AI-Assisted development with Claude Code 🤖⚠️

## Description

- Introduces `android/capacitor-cordova/` as a dedicated Gradle module (`:capacitor-cordova-android`) containing the Cordova bridge adapter classes (e.g. CordovaPlugin). This mirrors the iOS architecture where `CapacitorCordova` is a separate pod/SPM product, distinct from the main Capacitor module.

- The `org.apache.cordova:framework` Gradle dependency and the Cordova-specific ProGuard rule are removed from `@capacitor/android`. The new module owns both. `android/settings.gradle` is updated to declare both `:capacitor-android` and :`capacitor-cordova-android` for standalone builds.

- The CLI is updated to conditionally include `:capacitor-cordova-android` in the generated `capacitor.settings.gradle` and `capacitor.build.gradle` only when Cordova plugins are present, alongside the existing conditional inclusion of `:capacitor-cordova-android-plugins`. With no Cordova plugins installed, neither module is included — the built APK contains no `org.apache.cordova` classes and `cordova-android` does not appear in the Gradle dependency graph.

- **Extra:** Fixes a bug in `Bridge.java` where `isDeployDisabled()` and `shouldKeepRunning()` looked up the Cordova plugin by the name "__CordovaHandle", which was never registered anywhere. The correct name is "__CordovaPlugin", matching the @CapacitorPlugin annotation in `CordovaPlugin.java`. Before this fix, both methods always fell through to their default return values regardless of the DisableDeploy and KeepRunning preferences in config.xml.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

- Cordova classes and dependencies don't need to be included if there are no Cordova plugins present.

## Tests or Reproductions

- Tested with simple Capacitor app:
    - Started with a simple Android app without any Cordova plugins. Built the app and confirmed that the `cordova-android` Gradle dependency isn't present, and also that the `org.apache.cordova` package with the Cordova classes isn't present either.
    - Then installed a Cordova plugin (e.g. cordova-plugin-device) and confirmed that the Gradle dependency and Cordova classes are now present, and that the app worked properly (calling the plugin works).

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web


ItsChaceD (changes requested):
I tried installing in a test app and then adding a cordova plugin, and it failed with the error:

```
 > Could not resolve project :capacitor-cordova-android.
     Required by:
         project :app
      > No matching variant of project :capacitor-cordova-android was found.
```

I believe this can be resolved by adding the new src dependencies to the `android/package.json`. To verify I ran `npm pack --dry-run` and the tarball contains none of the new `capacitor-cordova/*` files.

So when a user installs `@capacitor/android` from npm and adds a Cordova plugin, the CLI writes:
```
include ':capacitor-cordova-android'
project(':capacitor-cordova-android').projectDir = new File('../node_modules/@capacitor/android/capacitor-cordova')
include ':capacitor-cordova-android-plugins'
project(':capacitor-cordova-android-plugins').projectDir = new File('./capacitor-cordova-android-plugins/')
```

but that directory will not exist in `node_modules`, so Gradle sync will fail with the missing project error. 

I was able to resolve this locally by adding this to the files entry in the `android/package.json`:
```
"capacitor-cordova/build.gradle",
"capacitor-cordova/proguard-rules.pro",
"capacitor-cordova/src/main/",
```

(although not sure if we need all of these)

alexgerardojacinto:
@ItsChaceD you're right, me and Claude missed that 😅 I didn't test with `npm pack` and the tarball and I should have to make sure

I did the changes now, can you check again?

ItsChaceD (approved):

PR #8455, opened 2026-05-04
fix(cli): generate resource entries in Package.swift

jcesarmobile:
If using SPM:
- copy the resources to the plugin folder instead of a separate resources folder.
- generate resources entry in Package.swift
- patch code to use `SWIFTPM_MODULE_BUNDLE` instead of `[NSBundle bundleForClass`
- patch code to use `SWIFTPM_MODULE_BUNDLE URLForResource` instead of `[NSBundle mainBundle] URLForResource` (we can't just replace all `[NSBundle mainBundle]` occurrences with `SWIFTPM_MODULE_BUNDLE` as some plugins use `[NSBundle mainBundle]` to get actual main bundle files such as the Info.plist information.

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -136,7 +136,7 @@ function buildResourcesText(resources: any[]) {
   return resources.length > 0
     ? `,
             resources: [
-                ${resourceEntry.join(',')}
+                ${resourceEntry.join(',\n                ')}
             ]`
     : '';
 }
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -128,6 +128,19 @@ function buildBinaryTargetEntries(p: Plugin, frameworks: any[]): { binaryTargets
   return { binaryTargetsText, binaryDepsText };
 }
 
+function buildResourcesText(resources: any[]) {
+  const resourceEntry = [];
+  for (const resource of resources) {
+    resourceEntry.push(`.copy("resources/${resource.$.src.split('/').pop()}")`);
+  }
+  return resources.length > 0
+    ? `,
+            resources: [
+                ${resourceEntry.join(',')}
+            ]`
+    : '';
+}
+
 function buildCSettingsText(p: Plugin, sourceFiles: any[]): string {
   const pluginId = p.id;
   const allFlags = new Set<string>();
@@ -190,8 +203,9 @@ async function writeGeneratedPackageSwift(p: Plugin, config: Config, iosPlatform
       ? `,
             publicHeadersPath: "."`
       : '';
+  const resources = getPlatformElement(p, platform, 'resource-file');
   const sourceFiles = getPlatformElement(p, platform, 'source-file');
-  if (sourceFiles.length === 0 && headerFiles.length === 0) {
+  if (sourceFiles.length === 0 && headerFiles.length === 0 && resources.length === 0) {
     return;
   }
   const frameworks = getPlatformElement(p, platform, 'framework');
@@ -200,6 +214,7 @@ async function writeGeneratedPackageSwift(p: Plugin, config: Config, iosPlatform
   const systemFrameworks = frameworks.filter((f: any) => !f.$.custom && f.$.src.endsWith('.framework'));
   const hasWeakFrameworks = systemFrameworks.some((f: any) => f.$.weak === 'true');
   const requiredSystemFrameworks = systemFrameworks.filter((f: any) => f.$.weak !== 'true');
+  const resourcesText = buildResourcesText(resources);
 
   const libraryTypeText = hasWeakFrameworks ? `\n            type: .dynamic,` : '';
   const linkerSettingsText =
@@ -232,7 +247,7 @@ let package = Package(
             dependencies: [
                 .product(name: "Cordova", package: "capacitor-swift-pm")${binaryDepsText}
             ],
-            path: "."${headersText}${cSettingsText}${linkerSettingsText}
+            path: "."${resourcesText}${headersText}${cSettingsText}${linkerSettingsText}
         )${binaryTargetsText}
     ]
 )`;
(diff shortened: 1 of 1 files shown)


cli/src/ios/update.ts
```diff
@@ -557,8 +572,16 @@ async function copyPluginsNativeFiles(config: Config, cordovaPlugins: Plugin[])
...
+            const bundleName = isSPM ? 'SWIFTPM_MODULE_BUNDLE' : '[NSBundle mainBundle]';
+            fileContent = fileContent.replace('[NSBundle bundleForClass:[self class]]', bundleName);
+            fileContent = fileContent.replace('[NSBundle bundleForClass:[CDVCapture class]]', bundleName);
+            await writeFile(fileDest, fileContent, { encoding: 'utf-8' });
+          }
+          if (isSPM && fileContent.includes('[NSBundle mainBundle] URLForResource')) {
+            fileContent = fileContent.replace(
+              '[NSBundle mainBundle] URLForResource',
+              'SWIFTPM_MODULE_BUNDLE URLForResource',
+            );
```
ItsChaceD, line 584: Should we use `replaceAll` for this, or is it not possible to have multiple `NSBundle`?

```suggestion
            const bundleName = isSPM ? 'SWIFTPM_MODULE_BUNDLE' : '[NSBundle mainBundle]';
            fileContent = fileContent.replaceAll('[NSBundle bundleForClass:[self class]]', bundleName);
            fileContent = fileContent.replaceAll('[NSBundle bundleForClass:[CDVCapture class]]', bundleName);
            await writeFile(fileDest, fileContent, { encoding: 'utf-8' });
          }
          if (isSPM && fileContent.includes('[NSBundle mainBundle] URLForResource')) {
            fileContent = fileContent.replaceAll(
              '[NSBundle mainBundle] URLForResource',
              'SWIFTPM_MODULE_BUNDLE URLForResource',
            );
```

cli/src/ios/update.ts
```diff
@@ -128,6 +128,19 @@ function buildBinaryTargetEntries(p: Plugin, frameworks: any[]): { binaryTargets
   return { binaryTargetsText, binaryDepsText };
 }
 
+function buildResourcesText(resources: any[]) {
+  const resourceEntry = [];
+  for (const resource of resources) {
+    resourceEntry.push(`.copy("resources/${resource.$.src.split('/').pop()}")`);
+  }
+  return resources.length > 0
```
markemer, line 136: A little confusing what's going on here. Could we split this up without the ternary? I thought it was returning bool, and then thought there was dead code, before I noticed what was up.

cli/src/ios/update.ts
markemer: If I understand this correctly, `replaceAll` might be safer. Although, I think replace should work in like nearly all cases.

jcesarmobile: that code already exists for CocaPods, so I prefer to not change it in this PR.
I just changed the replacement from `[NSBundle mainBundle]` if CocoaPods to `SWIFTPM_MODULE_BUNDLE` if SPM

cli/src/ios/update.ts
jcesarmobile: This is in line with other string generations in the codebase:
https://github.com/ionic-team/capacitor/blob/1f7e33fca43d183332ec19d22b0d75ef81d8cc6d/cli/src/ios/update.ts#L204
https://github.com/ionic-team/capacitor/blob/1f7e33fca43d183332ec19d22b0d75ef81d8cc6d/cli/src/ios/update.ts#L189-L192
https://github.com/ionic-team/capacitor/blob/1f7e33fca43d183332ec19d22b0d75ef81d8cc6d/cli/src/ios/update.ts#L206-L211
https://github.com/ionic-team/capacitor/blob/1f7e33fca43d183332ec19d22b0d75ef81d8cc6d/cli/src/util/spm.ts#L137-L144

cli/src/ios/update.ts
```diff
@@ -128,6 +128,19 @@ function buildBinaryTargetEntries(p: Plugin, frameworks: any[]): { binaryTargets
...
+    ? `,
+            resources: [
+                ${resourceEntry.join(',')}
```
OS-ruimoreiramendes: Not sure if this was intentional, but the resources entries are currently generated inline on a single line. It might be more readable to have each entry on its own line.

```suggestion
                ${resourceEntry.join(',\n                ')}
```

jcesarmobile: yeah, good catch, the plugins I tested only had one resource, so I didn't notice

OS-ruimoreiramendes: I tested it using the same test setup I've been using for the other CLI tasks, I created a dummy plugin with multiple resources and that's when I noticed it.

cli/src/ios/update.ts
```diff
@@ -128,6 +128,19 @@ function buildBinaryTargetEntries(p: Plugin, frameworks: any[]): { binaryTargets
   return { binaryTargetsText, binaryDepsText };
 }
 
+function buildResourcesText(resources: any[]) {
```
OS-ruimoreiramendes, line 131: Also, not sure if this is a real concern, but `buildResourcesText` uses `.copy()` for all resource types. For resources like `.xcassets`, `.storyboard`, and `.xib`, SPM typically requires `.process()` to compile them correctly, otherwise they might not be accessible at runtime. That said, the best way to validate this would probably be with a real plugin that actually uses these file types, since the dummy plugin I created for testing had XIB/xcassets files that weren't properly generated by Xcode. If the plugin you tested already included these file types and everything worked fine, then this is probably not a concern!

ItsChaceD (approved):

OS-ruimoreiramendes (approved):

cli/src/ios/update.ts
jcesarmobile: I've only found plugins using `.bundle` file extension and they required `.copy`.

For those file extensions we shouldn't even need to add a resources entry as it's supposed to auto detect them according to the docs, but having them with the copy might conflict, so we will have to revisit if some user reports that they don't work and provide some Cordova plugin to reproduce. We will merge as it is now.


> When you add a resource to your Swift package, Xcode detects common resource types for Apple platforms and treats them as a resource automatically. For example, you don’t need to make changes to your package manifest for the following resources:

> Interface Builder files; for example, XIB files and storyboards

> Core Data files; for example, xcdatamodeld files

> Asset catalogs

> .lproj folders you use to provide localized resources

https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package

PR #8463, opened 2026-05-07
fix(cli): mark Cordova CocoaPods plugins as incompatible in SPM

### LINKED ISSUE #8456: [Bug]: Migrating to Swift Package Manager fails
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 8.3.1
  @capacitor/core: 8.3.1
  @capacitor/android: 8.3.1
  @capacitor/ios: 8.3.1

Installed Dependencies:

  @capacitor/cli: 8.3.1
  @capacitor/android: 8.3.1
  @capacitor/core: 8.3.1
  @capacitor/ios: 8.3.1

### Other API Details

```Shell
npm --version
11.12.1
node --version
v24.14.1
pod --version
1.16.2
```

### Platforms Affected

- [x] iOS
- [ ] Android
- [ ] Web

### Current Behavior

Hi there. I am trying to migrate our app from pods to swift package manager as described here: https://ionic.io/blog/swift-package-manager-and-capacitor

It seems that our cordova plugin from salesforce makes problems. Generating a fresh app with angular / capacitor and swift package manager works without problems. 

The problematic plugin is `cordova-plugin-marketingcloudsdk`

When adding CapApp-SPM i get error:

`the package at '/Users/olivierschmid/AppDev/map-core-b2c-app/ios/capacitor-cordova-ios-plugins/sources/CordovaPluginMarketingcloudsdk' cannot be accessed (Error Domain=NSCocoaErrorDomain Code=260 "The folder “CordovaPluginMarketingcloudsdk” doesn’t exist." UserInfo={NSUserStringVariant=(
- unexpectedly did not find the new dependency in the package graph: fileSystem(identity: capapp-spm, absolutePath: SwiftPM.SPMAbsolutePath(absolutePath: <AbsolutePath:"/Users/olivierschmid/AppDev/map-core-b2c-app/ios/App/CapApp-SPM">))`

Then after selecting "add anyway" i cannot build the app.

Image

Image

Image

### Expected Behavior

App should be migrated to Swift Package Manager and build works again.

### Project Reproduction

I can't provide the whole app. I desired i can try to set up a fresh app including this plugin.

### Additional Information

Thanks!

jcesarmobile:
If the plugin has a podspec entry and doesn't have a package=SPM on the ios platform entry in plugin.xml, mark it as incompatible as it won't work.

This is a bit hacky, the proper place to do it would be in `resolvePlugin` as that's where we mark plugins as incompatible, but we don't have the `config` object available to check if it's SPM, so doing it inside update.

closes https://github.com/ionic-team/capacitor/issues/8456

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -44,8 +44,14 @@ export async function updateIOS(config: Config, deployment: boolean): Promise<vo
 }
 
 async function updatePluginFiles(config: Config, plugins: Plugin[], deployment: boolean) {
+  const isSPM = (await config.ios.packageManager) === 'SPM';
   await removePluginsNativeFiles(config);
-  const cordovaPlugins = plugins.filter((p) => getPluginType(p, platform) === PluginType.Cordova);
+  let cordovaPlugins = plugins.filter((p) => getPluginType(p, platform) === PluginType.Cordova);
+  let incompatibleCordovaPlugins = plugins.filter((p) => getPluginType(p, platform) === PluginType.Incompatible);
+  if (isSPM) {
+    incompatibleCordovaPlugins = incompatibleCordovaPlugins.concat(cordovaPlugins.filter((p) => needsSPMDep(p)));
+    cordovaPlugins = cordovaPlugins.filter((p) => !needsSPMDep(p));
+  }
   if (cordovaPlugins.length > 0) {
     await copyPluginsNativeFiles(config, cordovaPlugins);
   }
@@ -54,7 +60,7 @@ async function updatePluginFiles(config: Config, plugins: Plugin[], deployment: 
   }
   await handleCordovaPluginsJS(cordovaPlugins, config, platform);
   await checkPluginDependencies(plugins, platform, config.app.extConfig.cordova?.failOnUninstalledPlugins);
-  if ((await config.ios.packageManager) === 'SPM') {
+  if (isSPM) {
     await generateCordovaPackageFiles(cordovaPlugins, config);
 
     const validSPMPackages = await checkPluginsForPackageSwift(config, plugins);
@@ -66,7 +72,6 @@ async function updatePluginFiles(config: Config, plugins: Plugin[], deployment: 
   }
   await logCordovaManualSteps(cordovaPlugins, config, platform);
 
-  const incompatibleCordovaPlugins = plugins.filter((p) => getPluginType(p, platform) === PluginType.Incompatible);
   printPlugins(incompatibleCordovaPlugins, platform, 'incompatible');
 }
 
@@ -676,4 +681,10 @@ async function replaceFrameworkVariables(config: Config, prefsArray: any[], fram
     );
   });
   return frameworkString;
+}
+
+function needsSPMDep(plugin: Plugin): boolean {
+  const podspecs = getPlatformElement(plugin, 'ios', 'podspec');
+  const platformTag = getPluginPlatform(plugin, platform);
+  return podspecs.length > 0 && !platformTag.$?.package;
 }


cli/src/ios/update.ts
```diff
@@ -44,8 +44,14 @@ export async function updateIOS(config: Config, deployment: boolean): Promise<vo
 }
 
 async function updatePluginFiles(config: Config, plugins: Plugin[], deployment: boolean) {
+  const isSPM = (await config.ios.packageManager) === 'SPM';
```
OS-pedrogustavobilro, line 47: Just to confirm, we wouldn't be including this in `resolvePlugin` because we'd need to change the signature of an exported method, which is a breaking change. Is that it?

Would we want to change this in `next` branch to move this logic to `resolvePlugin`, or do you think it's not worth it?

OS-ruimoreiramendes (approved):
Tested this by creating a dummy plugin with a `<podspec>` entry in `plugin.xml`. The plugin correctly shows up as incompatible in the output instead of failing silently. LGTM!

cli/src/ios/update.ts
OS-pedrogustavobilro, line 1: [Sorry for the long comment 😅]
There's a few things I'm thinking here that make me wonder if this PR makes sense or not:

1. Correct me if I'm wrong, prior to this PR (and maybe depending on the plugin), users would be hit with an Error on Xcode / xcodebuild when having a Cordova Plugin with pod dependencies and not SPM compatible. Now, the users would see an info text in CLI about the plugin being incompatible and not being installed. The app should compile, but if the app is run on a device, and tries to access a method from the plugin, it will get a runtime error because the cordova plugin is unavailable. In such a situation, we're replacing a possible build-time error with a runtime error. Which is ok for scenarios where users are manually running cap sync and builds (if users ignore the message, it's their risk, although one could argue that for a big app with a large cap sync output, it can be easy to miss), but for automated build systems (I'm thinking OutSystems), this would become a problem I think.
2. The plugin is marked as incompatible, but we don't say why in CLI output, even though we know the reason we mark them as incompatible is because they need to support SPM.
3. This is maybe an edge case but I think could happen: A plugin whose only native dependencies are podspec(s) that aren't actually used in native iOS code (thinking like dead code that could be removed but hasn't). So the plugin would technically work with SPM, but this CLI is marking it as incompatible.

In my mind it would be nicer to log a warning related specifically to SPM incompatibility (possibly like we do for Capacitor Plugins), but do not mark them as incompatible explicitly. Maybe that doesn't fix the issue stated in the PR, but at the end of the day the three options are either to make the Cordova Plugin support SPM, or use CocoaPods until then, or remove the problematic plugin; neither of them are ideal (the first one the better option but if the developer doesn't own the plugin it's a third-party blocker).

Some of these points you may have considered already, so would like to hear your thoughts on the matter.

OS-pedrogustavobilro (commented):
I tested the PR itself and it works fine, but I have some doubts regarding the advantage of this PR over how it is today, leaving a comment that expands into that.

cli/src/ios/update.ts
jcesarmobile: yeah, it would be a breaking change

cli/src/ios/update.ts
jcesarmobile: Historically, incompatible plugins are plugins that don't compile, so we exclude them, and plugins with CocoaPods dependencies don't compile on SPM projects because of the missing dependencies.

But since we now have a different messaging for Capacitor plugins without Package.swift I could also do a similar thing.

Nevertheless, I find the Package.swift messaging a bit confusing at the moment, in example this is my output at the moment

⠼ update ios [info] Found 6 Cordova plugins for ios:
       cordova-plugin-device@3.0.0
       cordova-plugin-dialogs@2.0.2
       cordova-plugin-file@8.1.3
       cordova-plugin-firebase-analytics@5.0.0-OS21
       cordova-plugin-secure-storage@3.0.2
**⠇ update ios [warn] `@capacitor-community/firebase-analytics` does not have a Package.swift**
**[warn] Some installed Capacitor plugins are not compatible with SPM**
[info] Writing Package.swift
[info] Found 2 incompatible Cordova plugins for ios, skipped install:
       cordova-plugin-add-swift-support@2.0.3-OS2
       cordova-plugin-marketingcloudsdk@8.1.1
**[info] Found 1 Capacitor plugin for ios:
       @capacitor-community/firebase-analytics@8.0.0**
       
The warning for `@capacitor-community/firebase-analytics` not having Package.swift is on the same line as `⠇ update ios [warn]`, so it's easy to miss

Then it says 
  [info] Found 1 Capacitor plugin for ios:
  `@capacitor-community/firebase-analytics@8.0.0`
  
  Despite it's also being skipped on the linking, so I find the "Incompatible" way better and I think the SPM incompatibility should have been handled in a similar way, but it wasn't.
  
  But if you think it should work as Capacitor plugins I can change it.

PR #8468, opened 2026-05-11
feat(cli)!: merge live-reload options into `--url`

OS-pedrogustavobilro:
# ⚠️🤖  AI-Assisted development with Claude Code 🤖⚠️

## Description

This PR unifies the live reload CLI arguments in cap run`--live-reload` (or `-l`) `--host` `--port` `--https` -> into a single `--url` argument. if we wanted to point to a local network vite server, we had to separate host from the port like so:

```
npx cap run -l --host 192.168.1.181 --port 5173
```

Now it's simply copy the url from the command line that vite outputs and paste it into can run command:

```
npx cap run --url http://192.168.1.181:5173
```

There was a bit of logic related to getting a local IP in the CLI that now becomes unused, so I removed it.

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [x] Breaking Change - to go in Capacitor 9. Since the CLI options change it can break existing apps that are using the current live reload parameters.
- [ ] Documentation

## Rationale / Problems Fixed

Alternative solution to a recent PR that removed a default port in live reload, as suggested in https://github.com/ionic-team/capacitor/pull/8376#issuecomment-4067299292, that some time ago the team internally agreed was better, but never reserved the time to do it.

## Tests or Reproductions

You can test this updated functionality in any capacitor app. For convenience I have https://github.com/OS-pedrogustavobilro/capacitor-test-livereload.

There you need to run `npm ci`, `npm run build` + `npx cap sync`. So start a vite server you do `npm run dev`, to use HTTPS you can use a tool like ngrok `ngrok http 5173`; in another terminal tab you paste the url from vite / ngrok into the cap run command.

### DIFF
--- a/cli/src/tasks/run.ts
+++ b/cli/src/tasks/run.ts
@@ -30,11 +30,8 @@ export interface RunCommandOptions {
   targetNameSdkVersion?: string;
   sync?: boolean;
   forwardPorts?: string;
-  liveReload?: boolean;
-  host?: string;
-  port?: string;
+  url?: string;
   configuration?: string;
-  https?: boolean;
 }
 
 export async function runCommand(
@@ -42,7 +39,16 @@ export async function runCommand(
   selectedPlatformName: string,
   options: RunCommandOptions,
 ): Promise<void> {
-  options.host = options.host ?? CapLiveReloadHelper.getIpAddress() ?? 'localhost';
+  if (options.url) {
+    try {
+      const parsed = new URL(options.url);
+      if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+        fatal(`Invalid --url: must use http:// or https:// (got "${parsed.protocol}//")`);
+      }
+    } catch {
+      fatal(`Invalid --url: "${options.url}" is not a valid URL`);
+    }
+  }
   if (selectedPlatformName && !(await isValidPlatform(selectedPlatformName))) {
     const platformDir = resolvePlatform(config, selectedPlatformName);
     if (platformDir) {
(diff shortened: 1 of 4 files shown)


cli/src/tasks/run.ts
```diff
@@ -30,19 +30,25 @@ export interface RunCommandOptions {
...
   config: Config,
   selectedPlatformName: string,
   options: RunCommandOptions,
 ): Promise<void> {
-  options.host = options.host ?? CapLiveReloadHelper.getIpAddress() ?? 'localhost';
+  if (options.url) {
+    try {
+      const parsed = new URL(options.url);
+      if (parsed.protocol !== 'http:' && parsed.protocol !== 'https:') {
+        fatal(`Invalid --url: must use http:// or https:// (got "${parsed.protocol}//")`);
```
jcesarmobile: calling fatal inside the try-catch will trigger the catch, so this message is never displayed, if you want to get this message you should put only the `const parsed = new URL(options.url);` inside the try-catch and then if it wasn't triggered, throw the fatal if the protocol is not valid

jcesarmobile (changes requested):

cli/src/tasks/run.ts
OS-pedrogustavobilro: The message doesn't really matter that much I guess, but I did the update as you suggested to keep both messages.

Note to test: Currently it's crashing in iOS with SPM app, but it's unrelated to this change (getting the crash with `9.0.0-alpha.2` atm), I'm looking into it. If you'd like to test, recommend either Android (e.g. for https://github.com/OS-pedrogustavobilro/capacitor-test-livereload) or iOS CocoaPods

OS-pedrogustavobilro: Also I think lint is failing because it's missing syncing `next` with changes from `main`.

cli/src/tasks/run.ts
jcesarmobile, line 124:
```suggestion
      if (options.url) {
        await CapLiveReloadHelper.revertCapConfigForLiveReload();
      }
      if (!isFatal(e)) {
```
While it's not strictly related to your changes, while testing it, I've noticed that live reload configuration is not being reverted on run/build failure, I've fixed it for main here https://github.com/ionic-team/capacitor/pull/8485
but since your PR removes `options.liveReload` that fix won't be valid for next branch, so, can you do something like this here?

OS-pedrogustavobilro: Yeah I'll do it here! Good shout

OS-pedrogustavobilro: Done in https://github.com/ionic-team/capacitor/pull/8468/commits/f2cb5d304f8fbc091a046b835349a1299db79776

jcesarmobile (approved):

PR #8471, opened 2026-05-15
feat(cli): add experimental packageOptions

### LINKED ISSUE #8451: [Bug]: SwiftPM identity collision between plugin and transitive dependency

jcesarmobile:
In addition of the folder name conflict, I found another conflict with target names, so for the folder name problem I added `symlink` option, which will create a symbolic link to the node_modules folder inside a new `symlinks` folder (so we can remove it on update to remove old symbolic links before doing the copy and configuration of existing plugins), since we don't keep track of existing plugins, it's better to just remove everything and start over to not have dangling symlinks after plugin removals.

For target name conflicts I added `moduleAliases`, where you put the existing target name, and the name to be used as alias

can be tested on https://github.com/jcesarmobile/devicecollision which has both `@capacitor/device` and `@capacitor-community/device` and when you fix the folder name problem with `symlink: true`, the target name problem appears. Adding this to solve both problems:

```
experimental: {
    ios: {
      spm: {
        packageOptions: {
          '@capacitor-community/device': {
            symlink: true, 
            moduleAliases: {
              'DevicePlugin': 'CommunityDevicePlugin',
            }
        }
        }
      }
    }
  }
  ```
  
closes https://github.com/ionic-team/capacitor/issues/8451
  
We should probably put the packageTraits inside packageOptions, but didn't do it since it would be breaking, we could do it for Capacitor 9, unless we update to Swift 6 and we could make them non experimental.

### DIFF
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -559,6 +559,13 @@ export interface CapacitorConfig {
          * @since 8.3.0
          */
         packageTraits?: { [pluginId: string]: string[] };
+        /**
+         * Define options to apply to the package.
+         * The key is the plugin ID (e.g. `@capacitor-community/device`)
+         *
+         * @since 8.4.0
+         */
+        packageOptions?: { [pluginId: string]: PackageOptions };
       };
     };
   };
@@ -798,4 +805,17 @@ export interface PluginsConfig {
      */
     animation?: 'FADE' | 'NONE';
   };
+}
+
+export interface PackageOptions {
+  /**
+   * Create a symlink to the plugin folder instead of pointing to the plugin path.
+   * Useful when plugin names conflict.
+   */
+  symlink: boolean;
+  /**
+   * Useful to avoid target name conflicts in dependencies
+   * [see](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/modulealiasing/)
+   */
+  moduleAliases: { [target: string]: string };
 }
--- a/cli/src/declarations.ts
+++ b/cli/src/declarations.ts
@@ -812,10 +812,10 @@ export interface PackageOptions {
    * Create a symlink to the plugin folder instead of pointing to the plugin path.
    * Useful when plugin names conflict.
    */
-  symlink: boolean;
+  symlink?: boolean;
   /**
    * Useful to avoid target name conflicts in dependencies
    * [see](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/modulealiasing/)
    */
-  moduleAliases: { [target: string]: string };
+  moduleAliases?: { [target: string]: string };
 }
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -639,6 +639,9 @@ async function copyPluginsNativeFiles(config: Config, cordovaPlugins: Plugin[]) 
 
 async function removePluginsNativeFiles(config: Config) {
   await remove(config.ios.cordovaPluginsDirAbs);
+  if ((await config.ios.packageManager) === 'SPM') {
+    await remove(join(config.ios.nativeProjectDirAbs, 'CapApp-SPM', 'symlinks'));
+  }
   await extractTemplate(config.cli.assets.ios.cordovaPluginsTemplateArchiveAbs, config.ios.cordovaPluginsDirAbs);
 }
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: cli/src/declarations.ts [hunk 1]
```diff
@@ -799,3 +806,16 @@ export interface PluginsConfig {
...
+  /**
+   * Create a symlink to the plugin folder instead of pointing to the plugin path.
+   * Useful when plugin names conflict.
+   */
+  symlink: boolean;
+  /**
+   * Useful to avoid target name conflicts in dependencies
+   * [see](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/modulealiasing/)
+   */
+  moduleAliases: { [target: string]: string };
```
OS-pedrogustavobilro: Wondering if these values should be mandatory or not? Maybe right now it doesn't matter as much, but if we plan on placing `packageTraits` in this object for Cap 9 as you hint in the PR, maybe it matters for them to be optional? 

Still, leaving the question here to check with you.

OS-pedrogustavobilro: I guess to phrase my question differently is do we always need both values to fix the naming conflicts in SPM?

jcesarmobile: yeah, probably don't need both, and probably could need just moduleAliases and not symlink in some cases, so made them both optional

OS-pedrogustavobilro (approved):
Tested with your app and a few others, everything's working well, very nice!

PR #8476, opened 2026-05-20
chore: Decrease timeout for CI jobs from 60 to 30 minutes

markemer:
## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Platforms Affected
- [ ] Android
- [ ] iOS
- [ ] Web

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ concurrency:
 jobs:
   setup:
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 120
     steps:
       - name: Get Latest
         uses: actions/setup-node@v6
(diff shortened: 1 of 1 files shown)


jcesarmobile (changes requested):
the problem is that the runner hangs, so increasing the timeout just makes things worse since the more time we put the more time it will take to fail, but will fail anyway, we should in fact change it back to 30 min, if it hangs it's fasters and we can manually restart

We could try updating to macos 26 and see if that solves the hang, but since it's only occasional we won't be sure if it worked or not

test-ios has been unstable since we started using Xcode 26, we have a workaround to prevent the hang, but doesn't seem to work every time.

jcesarmobile (approved):

Issue #6177, opened 2022-12-19
Consolidated issue cataloguing all Fetch/HTTP issues with test harness


silviogutierrez:
### Capacitor Version
`4.6.1`

### Platform(s)
Android / iOS

### Code Reproduction
I took the time to put together a test harness that reproduces all the issues HTTP/Fetch has right now on iOS and Android. There may be others, but this is what I've seen so far.

Go to https://github.com/silviogutierrez/capacitor-http-tests and clone the repository.

Run `npm install` and `npm cap sync` to get both emulators working.

Then run `npm run build && node server.js` to start the local server.

Now open `http://localhost:3000` and see how everything works on the browser.

After that run `npx open ios` and run the simulator. And `npx open android` and run the emulator. You'll see what's failing and what's not.

I've tried to associate the bugs with the issue number.

Hopefully this is helpful.

Related issues: #6174, #5945, #6165, #6142, #6170, #6126,

silviogutierrez:
Update: I was linking to the wrong repo. Apologies! The updated repo has the right tests.

Screenshot below:

Web:

<img width="435" alt="image" src="https://user-images.githubusercontent.com/92824/208961908-01086d90-c49f-4f59-bd05-9793a973f81d.png">


iOS:

<img width="565" alt="image" src="https://user-images.githubusercontent.com/92824/208962033-eb7e77fa-bdac-4b6b-bded-da35b70a78a4.png">


Android:

<img width="412" alt="image" src="https://user-images.githubusercontent.com/92824/208962080-2427a579-31d3-4723-aa54-8833e30cbc68.png">

lubomirblazekcz:
We were having an issue that set-cookie was doubled in fetch post or xhr request only on Android. Might this be also a know issue?

silviogutierrez:
@lubomirblazekcz, perhaps: https://github.com/ionic-team/capacitor/issues/6280 or https://github.com/ionic-team/capacitor/issues/6308 ?

marcelklehr:
Is there an ETA on when this will be fixed?

eric-horodyski:
We’re reviewing older issues as part of our backlog cleanup effort:
[https://ionic.io/blog/keeping-capacitors-backlog-healthy](https://ionic.io/blog/keeping-capacitors-backlog-healthy)

To keep this open, please provide an updated minimal reproduction and environment details on the latest version of Capacitor within 7 days.

[...]

PR #8478, opened 2026-05-23
fix(http): support binary request bodies

iamanaws:
## Description

Fix patched `fetch` support for binary request bodies.

This updates the native HTTP bridge path so `fetch()` requests with `Blob`/binary bodies preserve the original bytes instead of serializing the body as `{}`.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Related issue: https://github.com/ionic-team/capacitor/issues/6177

Specific repro comment: https://github.com/ionic-team/capacitor/issues/6177#issuecomment-4526122970

I updated the original HTTP/fetch repro to the latest Capacitor version here:

https://github.com/iamanaws/capacitor-http-tester

The original HTTP/fetch cases from the issue now pass for me on web, Android, and iOS, but one binary request-body issue remained: a plain `fetch()` POST with a `Blob` body works on web, but native Capacitor patched `fetch` did not preserve the Blob bytes correctly.

Observed Android result before this fix:

```json
{
  "contentType": "application/octet-stream",
  "length": 2,
  "text": "{}"
}
```
This looks similar to #6126, but for sending a Blob instead of receiving one.

This also breaks floccus (https://github.com/floccusaddon/floccus/issues/1608) because isomorphic-git/http/web uses fetch, and Git smart HTTP needs binary request bodies.

## Tests or Reproductions

Failing repro on current Capacitor version:
https://github.com/iamanaws/capacitor-http-tester

Proof branch using this Capacitor patch (from local source):
https://github.com/iamanaws/capacitor-http-tester/tree/repro/binary-request-body-support

Tested with the repro app on Android, iOS, and Web
With this patch, the Blob request-body test passes on Android and iOS.

## Screenshots / Media

N/A

## Platforms Affected
- [x] Android
- [x] iOS
- [ ] Web

## Notes / Comments

Web already preserves the Blob body correctly. This PR fixes native patched fetch request body handling.


marcelklehr:
@iamanaws Woop woop, thank you so much for this contribution!

marcelklehr:
@jcesarmobile Would be great if this could get looked at :pray:

riderx:

Patch IDs: `upstream-pr-8478-android`, `upstream-pr-8478-ios`

```ts
plugins: {
  CapacitorPatch: {
    patches: [
      'upstream-pr-8478-android',
      'upstream-pr-8478-ios',
    ],
    strict: true,
  },
}
```

PR #8480, opened 2026-05-25
fix(android): show only the requested system bar

### LINKED ISSUE #8479: [Bug]: Android SystemBars.show({ bar: StatusBar }) also shows the navigation bar
### Capacitor Version

Current `main` (`e456de08`). I checked history as well: this behavior appears to have been present since the SystemBars plugin was introduced in `a32216ac` on November 12, 2025.

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

On Android, `SystemBars.show({ bar: SystemBarType.StatusBar })` also shows the navigation bar.

The `hide()` path treats the two bars separately, but the `show()` path does not:

- `hide(StatusBar)` -> `WindowInsetsCompat.Type.statusBars()`
- `hide(NavigationBar)` -> `WindowInsetsCompat.Type.navigationBars()`
- `show(StatusBar)` -> `WindowInsetsCompat.Type.systemBars()`
- `show(NavigationBar)` -> `WindowInsetsCompat.Type.navigationBars()`

Since `systemBars()` includes both status and navigation bars, a status-only show request reveals both bars.

### Expected Behavior

`SystemBars.show({ bar: SystemBarType.StatusBar })` should only show the status bar, leaving the navigation bar unchanged.

### Reproduction

```ts
import { SystemBars, SystemBarType } from '@capacitor/core';

await SystemBars.hide({ bar: SystemBarType.NavigationBar });
await SystemBars.show({ bar: SystemBarType.StatusBar });
```

Expected:
The status bar appears and the navigation bar stays hidden.

Actual:
The navigation bar appears too.

### Why this looks like a bug

The public API/docs describe `bar` as "The system bar to hide or show", which implies `StatusBar` and `NavigationBar` should be handled independently.

The Android implementation already does that correctly in the `hide()` path, and iOS also behaves symmetrically. The Android `show()` path is the odd one out.

### Relevant code

`android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java`

Current `show()` branch:

```java
if (bar.isEmpty() || bar.equals(BAR_STATUS_BAR)) {
    windowInsetsControllerCompat.show(WindowInsetsCompat.Type.systemBars());
}
if (bar.isEmpty() || bar.equals(BAR_GESTURE_BAR)) {
    windowInsetsControllerCompat.show(WindowInsetsCompat.Type.navigationBars());
}
```

### Suggested fix

```java
if (bar.isEmpty()) {
    windowInsetsControllerCompat.show(WindowInsetsCompat.Type.systemBars());
} else if (bar.equals(BAR_STATUS_BAR)) {
    windowInsetsControllerCompat.show(WindowInsetsCompat.Type.statusBars());
} else if (bar.equals(BAR_GESTURE_BAR)) {
    windowInsetsControllerCompat.show(WindowInsetsCompat.Type.navigationBars());
}
```

### Additional Information

I did not find a recent commit that introduced this. `git blame` points back to the original SystemBars plugin implementation, so this looks like an original Android mapping bug rather than a regression from the recent inset work.

fallintoplace:
## Summary
- show only the requested Android system bar when `SystemBars.show({ bar })` is called
- add a focused Android unit test covering the `show()` mapping for the empty, status-bar, and navigation-bar cases

## Why
The Android `hide()` path already treats `StatusBar` and `NavigationBar` separately, but `show(StatusBar)` was calling `WindowInsetsCompat.Type.systemBars()`, which also reveals the navigation bar.

Fixes #8479.

## Testing
- `JAVA_HOME=/opt/homebrew/opt/openjdk@21 ANDROID_HOME=/Users/hoangvu/Library/Android/sdk ANDROID_SDK_ROOT=/Users/hoangvu/Library/Android/sdk PATH=/opt/homebrew/opt/openjdk@21/bin:$PATH ./gradlew -p capacitor testDebugUnitTest --tests com.getcapacitor.plugin.SystemBarsTest`

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -288,11 +288,12 @@ public class SystemBars extends Plugin {
         WindowInsetsControllerCompat windowInsetsControllerCompat = WindowCompat.getInsetsController(window, window.getDecorView());
 
         if (hide) {
-            if (bar.isEmpty() || bar.equals(BAR_STATUS_BAR)) {
-                windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.statusBars());
-            }
-            if (bar.isEmpty() || bar.equals(BAR_GESTURE_BAR)) {
-                windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.navigationBars());
+            if (bar.isEmpty()) {
+              windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.systemBars());
+            } else if (bar.equals(BAR_STATUS_BAR)) {
+              windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.statusBars());
+            } else if (bar.equals(BAR_GESTURE_BAR)) {
+              windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.navigationBars());
             }
             return;
         }
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -305,11 +305,11 @@ public class SystemBars extends Plugin {
 
         if (hide) {
             if (bar.isEmpty()) {
-              windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.systemBars());
+                windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.systemBars());
             } else if (bar.equals(BAR_STATUS_BAR)) {
-              windowInsetsControllerCompat.hide(WindowInsetsCompat.Type.statusBars());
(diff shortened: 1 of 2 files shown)


theproducer (approved):
@fallintoplace Thanks for the unit tests as well!

android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
```diff
@@ -297,10 +297,11 @@ private void setHidden(boolean hide, String bar) {
             return;
         }
 
-        if (bar.isEmpty() || bar.equals(BAR_STATUS_BAR)) {
+        if (bar.isEmpty()) {
             windowInsetsControllerCompat.show(WindowInsetsCompat.Type.systemBars());
```
OS-pedrogustavobilro, line 318: @theproducer should we propagate this equivalent logic to `hide` (above in this same `setHidden` method)? In general systemBars should just be statusBars+navigationBars, but maybe this could change in the future? But I suppose if it does then that probably means additional changes in this core plugin.

OS-pedrogustavobilro (approved):

PR #8328, opened 2026-02-03
feat!: make Cordova support optional

markemer:
### Test instructions
The best way to test this is with 4 apps. (iOS and Android)

1. An app with just Capacitor Plugins that uses Cocoapods
2. An app with Capacitor and at least 1 Cordova plugin that uses Cocoapods
3. An app with just Capacitor Plugins that uses SPM
4. An app with Capacitor and at least 1 Cordova plugin that uses SPM

In all cases running `npx cap sync` should work as expected, not including Cordova when there are no plugins, and including it when there are.

### Some other things to try

- Remove all Cordova plugins and re-sync
- Add a Cordova plugin and re-sync
- Any other weird case you can come up with

PR #8483, opened 2026-05-27
chore: merge main into next

alexgerardojacinto:
# ℹ️ This PR was created using Claude

## Summary

Merges the latest from `main` (up to `b4b297a5`, telemetry endpoint update and `feat(cli): capture ios_package_manager`) into `next`.

## Conflict resolutions

- **`lerna.json` + all `package.json`** — kept `next`'s `9.0.0-alpha.2` (main is on the `8.3.x` line).
- **All `CHANGELOG.md`** — combined: `9.0.0-alpha.x` entries on top, then `8.3.4` / `8.3.3` from main, then existing history below.
- **`cli/src/ios/update.ts`** — kept `next`. The Cordova-related helpers (`generateCordovaPodspecs`, `copyPluginsNativeFiles`, `removePluginsNativeFiles`, `removeNoSystem`, etc.) were moved to `cli/src/util/cordova-ios.ts` in #8328 (`feat!: make Cordova support optional`), so the entire `main`-side block at this conflict no longer belongs in `update.ts`.
- **`cli/src/util/cordova-ios.ts`** — ported the SPM `symlinks` cleanup that main added to `removePluginsNativeFiles` in #8471 (`feat(cli): add experimental packageOptions`). On `next` that function lives in `cordova-ios.ts`, so the fix was applied there to keep behavior consistent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## [8.3.3](https://github.com/ionic-team/capacitor/compare/8.3.2...8.3.3) (2026-05-08)
+
+### Bug Fixes
+
+- **cli:** copy plugin files in CocoaPods projects ([#8467](https://github.com/ionic-team/capacitor/issues/8467)) ([b2d7719](https://github.com/ionic-team/capacitor/commit/b2d771926a180e60deea31992d7d4abcd5ca3bc7))
+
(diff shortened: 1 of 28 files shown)


OS-pedrogustavobilro:
@alexgerardojacinto We shouldn't squash-merge this into `next`, as we will lose the commit hashes that come from main. Unless this is to double-check in this PR and then manual merge (fast-forward?) `chore/merge-main-into-next` into `next`

OS-pedrogustavobilro (approved):
The merged changes looks good. As we aligned internally this approval serves more as a ok to merge manually locally then push rather that squash-merge the PR, which is not what we want here. Therefore I'm closing the PR.

PR #8485, opened 2026-05-28
fix(cli): revert live reload config on failure

jcesarmobile:
if run/build fails the live reload configuration is kept, should be reverted


OS-pedrogustavobilro (approved):
Ah nice, I had noticed this before but didn't realised it was tied to failures.

PR #8497, opened 2026-06-03
chore(cli)!: bump compile/targetSdkVersion to 37

jcesarmobile:
bump compile/targetSdkVersion to 37 in cli related packages

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -90,7 +90,7 @@ export async function migrateCommand(config: Config, noprompt: boolean, packagem
 
   const { migrateconfirm } = noprompt
     ? { migrateconfirm: 'y' }
-    : await logPrompt(`Capacitor 8 sets a deployment target of iOS ${iOSVersion} and Android 16 (SDK 36). \n`, {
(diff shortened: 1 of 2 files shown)


alexgerardojacinto (approved):

cli/src/tasks/migrate.ts
```diff
@@ -90,7 +90,7 @@ export async function migrateCommand(config: Config, noprompt: boolean, packagem
 
   const { migrateconfirm } = noprompt
     ? { migrateconfirm: 'y' }
-    : await logPrompt(`Capacitor 8 sets a deployment target of iOS ${iOSVersion} and Android 16 (SDK 36). \n`, {
+    : await logPrompt(`Capacitor 8 sets a deployment target of iOS ${iOSVersion} and Android 17 (SDK 37). \n`, {
```
OS-pedrogustavobilro, line 93: Should be Cap 9 instead?


```suggestion
    : await logPrompt(`Capacitor 9 sets a deployment target of iOS ${iOSVersion} and Android 17 (SDK 37). \n`, {
```

jcesarmobile: yeah, but it's unrelated to my change

we should really update all the migrate command at once, but since rui has been updating other values in here I also did it for the SDK version

PR #8498, opened 2026-06-03
chore(android)!: bump AGP to 9.2.1 and Gradle to 9.5.1

OS-ruimoreiramendes:
## Description

Bump the Android Gradle Plugin from `8.13.0` to `9.2.1` and the Gradle wrapper from `8.14.3` to `9.5.1`

## Change Type
- [ ] Fix
- [ ] Feature
- [ ] Refactor
- [x] Breaking Change
- [ ] Documentation
- [x] Other (CI, chores, etc.)

## Rationale / Problems Fixed

 Keep Capacitor 9 on the latest stable toolchain.

 Refs: [RMET-4709](https://outsystemsrd.atlassian.net/browse/RMET-4709)

## Tests or Reproductions

  - `npm run build` in `cli/`: TypeScript compiles
  - `npm run verify` in `android/`: **BUILD SUCCESSFUL**
  - Android library `assembleRelease` and unit tests: **BUILD SUCCESSFUL**
  - CLI unit test suite passes
  - Generated a fresh app via local CLI (`npx cap add android`) and built it: **BUILD SUCCESSFUL** with AGP `9.2.1` / Gradle `9.5.1`

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

### `-b` flag replacement
Gradle 9 removed `-b/--build-file` (see [upgrade guide][gradle9]). The `verify` script in `android/package.json` and `scripts/publish-android.sh` now use qualified task paths (`:capacitor-android:<task>`) to keep the same scope.

[gradle9]: https://docs.gradle.org/9.0.0/userguide/upgrading_major_version_9.html#removal_of_custom_build_layout_options

[RMET-4709]: https://outsystemsrd.atlassian.net/browse/RMET-4709?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jcesarmobile (commented):
latest plugin is 9.2.1

the gradle-wrapper.jar, gradlew and gradle.bat files also need to be updated to gradle 9.5.1, not just the properties file

OS-ruimoreiramendes:
> latest plugin is 9.2.1
> 
> the gradle-wrapper.jar, gradlew and gradle.bat files also need to be updated to gradle 9.5.1, not just the properties file

@jcesarmobile bumped AGP to 9.2.1 and regenerated the Gradle wrapper files for both android/ and android-template/.

jcesarmobile (approved):
Changes look good.

But I've noticed that doing the upgrade using AGP, it adds a bunch of settings on `android/gradle.properties`, are we sure we don't need any of those?

Not asking to add them to the PR, but we should have a task for evaluating them, and if we don't need them, since in the docs we usually recommend using AGP Upgrade Assistant, we should add a note about removing those changes.

Also gradle 9 enables kotlin plugin by default, which breaks kotlin plugin, you should create a ticket about that or if we have one about updating plugins to gradle 9, update it to add a note about the kotlin plugin being enabled.

Also in gradle 9 targetSdkVersion default to compileSdk, so we could remove it from capacitor and templates (again, not asking to do it in this PR, but create ticket for evaluating it)

PR #8508, opened 2026-06-12
fix(cli): make SPM dependency patch work on prereleases

jcesarmobile:
if the capacitor/ios version is 9.0.0-alpha the CLI is setting the dependency to 9.0.0, which doesn't exist yet, so check if it's a prerelease and use the prerelease version as dependency instead of x.0.0

### DIFF
--- a/cli/src/ios/update.ts
+++ b/cli/src/ios/update.ts
@@ -1,6 +1,6 @@
 import { copy, remove, pathExists, readFile, realpath, writeFile } from 'fs-extra';
 import { basename, dirname, join, relative } from 'path';
-import { major } from 'semver';
+import { major, prerelease } from 'semver';
 
 import c from '../colors';
 import { checkPlatformVersions, getCapacitorPackageVersion, runTask } from '../common';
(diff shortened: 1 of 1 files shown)


cli/src/ios/update.ts
```diff
@@ -70,11 +70,13 @@ async function updatePluginFiles(config: Config, plugins: Plugin[], deployment:
         const version = content.match(regex)?.[1];
         const majorCapVersion = major(iosPlatformVersion);
         if (version && major(version) != majorCapVersion) {
+          const preCapVersion = prerelease(iosPlatformVersion);
+          const forceVersion = preCapVersion ? iosPlatformVersion : `${majorCapVersion}.0.0`;
```
OS-pedrogustavobilro, line 74: This also makes the change for dev and nightly builds of capacitor, which I'm unsure if we want that since those aren't available as `capacitor-swift-pm` versions. We could check the output of `prerelease` and check if it contains alpha, beta, or rc, but would like to get your thoughts if such a thing is useful to have or not.

jcesarmobile: we should start publishing those then, since we already set the exact version in `ios/App/CapApp-SPM/Package.swift`, setting the Capacitor dependency value on the installed `@capacitor/ios` dependency, so that functionality is already broken.

OS-pedrogustavobilro: Good point then, it doesn't impact solely this PR; I tested it with a capacitor app and seems to work fine! Approving it ✅

OS-pedrogustavobilro (approved):

PR #8511, opened 2026-06-20
fix(http): make HttpResponse data type generic

medbenmakhlouf:
## Description

This change makes HttpResponse generic so callers can type the response payload with HttpResponse<T>. It remains backward-compatible because the default type parameter is any, so existing usages of HttpResponse continue to work unchanged. The result is a small TypeScript API improvement with no runtime impact.     

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Platforms Affected
- [ ] Android
- [ ] iOS
- [ ] Web

### DIFF
--- a/core/http.md
+++ b/core/http.md
@@ -44,7 +44,11 @@ export default config;
 ## Example
 
 ```typescript
-import { CapacitorHttp } from '@capacitor/core';
+import { CapacitorHttp, type HttpResponse } from '@capacitor/core';
+
+interface MyResponse {
+  foo: string;
+}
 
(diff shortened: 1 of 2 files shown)


medbenmakhlouf:
Hey @jcesarmobile @markemer   👋 Small change, big DX win :)`HttpResponse<T = any>` makes the type generic so callers can type their payload without casting. Fully backward-compatible, zero runtime impact. Would love your feedback!

PR #8518, opened 2026-06-24
fix(cli): Don't add kotlin plugin for Cordova plugins

jcesarmobile:
With Gradle 9, kotlin plugin is enabled, and causes issues to apply it, so remove all kotlin plugin related code from the CLI


cli/src/android/update.ts
```diff
@@ -263,7 +263,6 @@ if (hasProperty('postBuildExtras')) {
 
 export async function handleCordovaPluginsGradle(config: Config, cordovaPlugins: Plugin[]): Promise<void> {
   const pluginsGradlePath = join(config.android.cordovaPluginsDirAbs, 'build.gradle');
-  const kotlinNeeded = await kotlinNeededCheck(config, cordovaPlugins);
   const kotlinVersionString = config.app.extConfig.cordova?.preferences?.GradlePluginKotlinVersion ?? '2.4.0';
```
markemer: This lint warning appears to be correct, [Line 324](https://github.com/ionic-team/capacitor/pull/8518/changes#diff-ba4645a42326cd344120764d6f71384511748b2d3efca587facf244d2950c455L324) is the only place we use this.

markemer (changes requested):
Just that one fix and I think we're good.

markemer (approved):

PR #8521, opened 2026-07-01
chore(cli): migrate changes for Capacitor 9

jcesarmobile:
The links to the docs don't exist yet, we should create them before releasing this, but the docs tasks are not on the sprint.

Also it will probably need some more changes in the future depending on what breaking changes we make in the plugins or if we make more changes on the templates.

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -357,14 +357,23 @@ async function installLatestLibs(dependencyManager: string, runInstall: boolean,
   for (const devDepKey of Object.keys(pkgJson['devDependencies'] || {})) {
     if (libs.includes(devDepKey)) {
       pkgJson['devDependencies'][devDepKey] = coreVersion;
-    } else if (plugins.includes(devDepKey) && !pkgJson['devDependencies'][devDepKey].startsWith('file:')) {
+    } else if (
+      plugins.includes(devDepKey) &&
+      !(
+        pkgJson['devDependencies'][devDepKey].startsWith('file:') ||
+        pkgJson['devDependencies'][devDepKey].startsWith('workspace:')
+      )
+    ) {
       pkgJson['devDependencies'][devDepKey] = pluginVersion;
     }
   }
   for (const depKey of Object.keys(pkgJson['dependencies'] || {})) {
     if (libs.includes(depKey)) {
       pkgJson['dependencies'][depKey] = coreVersion;
-    } else if (plugins.includes(depKey) && !pkgJson['dependencies'][depKey].startsWith('file:')) {
+    } else if (
+      plugins.includes(depKey) &&
+      !(pkgJson['dependencies'][depKey].startsWith('file:') || pkgJson['dependencies'][depKey].startsWith('workspace:'))
+    ) {
       pkgJson['dependencies'][depKey] = pluginVersion;
     }
   }
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -47,7 +47,6 @@ const coreVersion = 'next';
 const pluginVersion = 'next';
 const gradleVersion = '9.5.1';
 const iOSVersion = '16';
-const kotlinVersion = '2.4.0';
 let installFailed = false;
 
 export async function migrateCommand(config: Config, noprompt: boolean, packagemanager: string): Promise<void> {
@@ -511,23 +510,7 @@ async function updateBuildGradle(
     }
   }
 
-  const beforeKotlinVersionUpdate = replaced;
-  replaced = replaceVersion(replaced, /(ext\.kotlin_version\s*=\s*['"])([^'"]+)(['"])/, kotlinVersion);
-  replaced = replaceVersion(replaced, /(org\.jetbrains\.kotlin:kotlin[^:]*:)([\d.]+)(['"])/, kotlinVersion);
-  if (beforeKotlinVersionUpdate !== replaced) {
-    logger.info(`Set Kotlin version to ${kotlinVersion}`);
-  }
   writeFileSync(filename, replaced, 'utf-8');
-}
-
-function replaceVersion(text: string, regex: RegExp, newVersion: string): string {
-  return text.replace(regex, (match, prefix, currentVersion, suffix) => {
-    const semVer = coerce(currentVersion)?.version;
-    if (gte(newVersion, semVer ? semVer : '0.0.0')) {
-      return `${prefix || ''}${newVersion}${suffix || ''}`;
-    }
-    return match;
-  });
 }
 
 async function updateAppBuildGradle(filename: string) {
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -177,9 +177,7 @@ export async function migrateCommand(config: Config, noprompt: boolean, packagem
         }
 
         await runTask(`Migrating AppDelegate.swift`, () => {
-          return updateAppDelegate(
-            join(config.ios.nativeTargetDirAbs, 'AppDelegate.swift')
-          );
+          return updateAppDelegate(join(config.ios.nativeTargetDirAbs, 'AppDelegate.swift'));
         });
       }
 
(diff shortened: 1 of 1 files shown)


cli/src/tasks/migrate.ts
```diff
@@ -622,3 +637,21 @@ async function updateAndroidManifest(filename: string) {
...
+async function updateAppDelegate(filename: string) {
+  const txt = readFile(filename);
+  if (!txt) {
+    return;
+  }
+
+  if (txt.includes('@main')) {
+    return; // Probably already updated
+  }
+  // Since navigation was an optional change in Capacitor 7, attempting to add density and/or navigation
```
OS-pedrogustavobilro: This comment seems out of place? It sounds it's referring to changes in Android Manifest but this is AppDelegate.

cli/src/tasks/migrate.ts
OS-pedrogustavobilro, line 19: If this year we'll be launching `next` versions for all official plugins (I think we are aiming to do that?), should we update this list perhaps?

OS-pedrogustavobilro: While I suppose for this PR it would mean pointing everything to next, when Cap 9 rolls out we'd need to adapt this script to map other plugin versions that aren't 9.x (e.g. barcode scanner)
That being said, since they are official plugins, they should be here in this script?

cli/src/tasks/migrate.ts
```diff
@@ -43,8 +43,8 @@ const plugins = [
   '@capacitor/text-zoom',
   '@capacitor/toast',
 ];
-const coreVersion = '^8.0.0';
-const pluginVersion = '^8.0.0';
+const coreVersion = 'next';
+const pluginVersion = 'next';
 const gradleVersion = '9.5.1';
 const iOSVersion = '16';
 const kotlinVersion = '2.4.0';
```
OS-pedrogustavobilro: Should we no longer be replacing kotlin version for `classpath`, and instead remove it? Because it's now bundled in AGP 9? 

[We added it in Capacitor 8](https://github.com/ionic-team/capacitor/pull/8238) to fix an issue related to an app declaring the kotlin version in root `build.gradle` and it differing from the one Capacitor / Plugins use. [Internal Slack Thread](https://outsystems.slack.com/archives/C07CWA9D0R4/p1763055431231829). But AGP bundles kotlin gradle plugin it's probably better to not have it in the app, and instead remove it altogether perhaps?

But also, we are using Kotlin 2.4.0 here and [AGP 9.2 uses 2.3.10](https://developer.android.com/build/releases/agp-9-2-0-release-notes#fixed-issues-agp-9.2.0). But since it's a minor version change it should be ok (where last year it was a major version).

If you think this is better left to a separate task / PR let me know, but just wanted to highlight it.

cli/src/tasks/migrate.ts
```diff
@@ -518,14 +528,19 @@ async function updateAppBuildGradle(filename: string) {
...
-  const gradlePproperties = ['compileSdk', 'namespace', 'ignoreAssetsPattern'];
-  for (const prop of gradlePproperties) {
-    // Use updated Groovy DSL syntax with " = " assignment
-    const regex = new RegExp(`(^\\s*${prop})\\s+(?!=)(.+)$`, 'gm');
-    replaced = replaced.replace(regex, (_match, key, value) => {
-      return `${key} = ${value.trim()}`;
-    });
-  }
+  replaced = replaced.replace(`proguard-android.txt`, `proguard-android-optimize.txt`);
+  replaced = replaced.replace('        targetSdkVersion rootProject.ext.targetSdkVersion\n', '');
```
OS-pedrogustavobilro, line 524: In AGP 9, users can change the behavior of targetSDK by setting `android.sdk.defaultTargetSdkToCompileSdkIfUnset` to false in `gradle.properties` - in which case it would default to the minSdk, which would break the app (don't know if it would compile, but you wouldn't be able to submit to play store). This will no longer be possible in AGP 10 (there's a warning in gradle related to this).

I don't know why anyone would do this voluntarily, but they could if they want to. What seems more likely to happen is users that migrated to AGP 9 while in Capacitor 8 - `android.sdk.defaultTargetSdkToCompileSdkIfUnset=false`; I tested Android Studio's AGP migration assistant and it adds that property.

So the way to overcome this issue would be one of or a combination of:

1. Keep `targetSdkVersion` for now. Only remove it in a future version whenever we move to AGP 10. This would maybe require re-changing the template (although the template is new Capacitor apps, this point I'm making would be more for existing ones). For plugins it doesn't really matter if `targetSdkVersion` is there or not, because we don't run plugins on-device, only apps.
2. Have the migrate command remove the properties added - Probably in a separate PR because I'm seeing there may be other properties that we may need to update for AGP 9. While I think we should do that, it has the nefarious effect of potentially removing flags that the particular Capacitor app needed, but there's no way for us to know if they were added automatically by Android Studio and have no actual impact, or if they have specific flags to fix specific issues in their native setup - will post more info on that internally in the next couple of days.

Let me know if this makes sense, and if you have a preference on the approach (only 1., only 2., both, other...)

jcesarmobile: this has been already removed from the template, migrate command is to make the changes we made in the template, so if we want to keep the targetSdk, we should revert the template change

OS-pedrogustavobilro: I'm thinking now, option 2 may suffice (such changes will happen in a future task / PR, if the team agrees with the proposed approach for those flags). As for this comment, I'll resolve this comment.

cli/src/tasks/migrate.ts
jcesarmobile: I've only added the new changes, haven't removed anything from last year, we usually don't remove previous changes for a while since some apps might have missed some changes, if updated during beta and didn't rerun migrate once it was final, I've found this problem specially in our test apps 
https://github.com/ionic-team/capacitor-geolocation/pull/89/changes#diff-e525140166db601268205422ea95ecdeaac1306e2b96807222cdfcde1cb0c951
https://github.com/ionic-team/capacitor-geolocation/pull/89/changes#diff-e56dd649cbbb6fce23fa286ee7e1e8df7abeecb68bd1a86f623eea74c98a4746

But yeah, we should remove this, we shouldn't have added it to begin with since the official template doesn't have kotlin, and using 2.4.0 causes issues as it's not yet compatible with latest AGP, the value was updated by Rui in a previous PR, I haven't changed it.

cli/src/tasks/migrate.ts
jcesarmobile: we have a ticket for updating versions to 9.0.0 once it's final, but yeah, if we add all plugins that don't have 9.0.0 we should specify the latest version for them, or use `latest` tag (which I don't like)

OS-pedrogustavobilro: I guess the next aren't out yet for all our plugins, so making the change now could mean some plugins have outdated versions (not Cap 9 ones). So I'll add it to that task that we should also add missing official plugins.

cli/src/tasks/migrate.ts
OS-pedrogustavobilro: Just to make sure we're on the same page, when I said to remove it, I mean for the command to remove the kotlin versions / gradle plugin from app's root `build.gradle`, not remove the logic that changes the version and have the command not touch anything in regards to kotlin.

This could mean changing the apps that explicitly declare the kotlin version, but I'd say it's worse to leave it there if AGP version comes with kotlin gradle plugin (especially if the version declared is lower than the one AGP uses it would fail the build)

jcesarmobile: I don't plan to do that, migrate command is to keep apps in sync with our templates and our android template doesn't use kotlin.
None our the test apps I've updated so far has kotlin enabled since it's not needed to have kotlin enabled in the app to use kotlin plugins.

The OutSystems template used to have kotlin enabled and that might be why we had that into consideration in previous years, but it doesn't have it enabled anymore and also OS users will most likely not use migrate command anyway since they get latest template on app build and if they did, it might not work since the templates are different.

But I've removed the logic to update the kotlin version since 2.4.0 is not compatible with current AGP.

OS-pedrogustavobilro: Alrighty, fair enough, thank you! Resolving comment and approving PR.

OS-pedrogustavobilro (approved):

andredestro (approved):

PR #8524, opened 2026-07-06
fix(ios): remove Cordova.framework from Capacitor project

jcesarmobile:


ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
```diff
@@ -56,7 +56,6 @@
 		62959B452524DA7800A3D7F1 /* CAPPlugin.h in Headers */ = {isa = PBXBuildFile; fileRef = 62959B132524DA7700A3D7F1 /* CAPPlugin.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		62959B462524DA7800A3D7F1 /* CAPBridge.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62959B142524DA7700A3D7F1 /* CAPBridge.swift */; };
 		62959B472524DA7800A3D7F1 /* CAPNotifications.swift in Sources */ = {isa = PBXBuildFile; fileRef = 62959B152524DA7700A3D7F1 /* CAPNotifications.swift */; };
-		62959BA52526475A00A3D7F1 /* Cordova.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 62959BA02526474300A3D7F1 /* Cordova.framework */; };
```
OS-pedrogustavobilro, line 59: Because of this change, the `capacitor-swift-pm` xcframework build, as it stands now, will fail for Cap 9.

The [build-cap](https://github.com/ionic-team/capacitor-swift-pm/blob/main/build-cap) command Capacitor workspace and then generates Capacitor and Cordova xcframeworks from it, but the Cordova build directory will not be found anymore. I tested it on a fork using this branch and it [failed for that reason](https://github.com/OS-pedrogustavobilro/capacitor-swift-pm/actions/runs/28861386150/job/85600557386) - `error: the path does not point to a valid framework: /Users/runner/work/capacitor-swift-pm/capacitor-swift-pm/capacitor-checkout/ios/Capacitor/Build/iOS-Simulator.xcarchive/Products/Library/Frameworks/Cordova.framework`.

With some elbow grease from Claude I managed to get [a working version in my fork](https://github.com/OS-pedrogustavobilro/capacitor-swift-pm/blob/main/build-cap) with which I could at least test, but essentially this means either:

1. We fix the capacitor-swift-pm build to differ for 9.X versions .
2. Or we remove the xcframework build logic altogether for Cap 9, keeping in mind it still needs to exist for Cap 8.

Which probably warrants re-discussion internally with the team again, just wanted to bring it up.

OS-pedrogustavobilro (approved):
I tested it in a Cap 9 app and it works, but there's some additional considerations that I'll leave in comment (that don't impact this PR necessarily, it should still be merged imo).

ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
jcesarmobile: I've referenced Capacitor.framework in the CapacitorCordova project, since now its needed, not sure if that could fix the issue or break it more.

Also linked the `Plugin.swift` in the CapacitorCordova project since that wasn't done, so it wasn't part of the SPM project.

OS-pedrogustavobilro (approved):

ios/Capacitor/Capacitor.xcodeproj/project.pbxproj
OS-pedrogustavobilro: Looks good! Re-approved 

> I've referenced Capacitor.framework in the CapacitorCordova project, since now its needed, not sure if that could fix the issue or break it more.

Uhh I don't think it would fix the issue I referred to in `capacitor-swift-pm` no, but still makes sense.

jcesarmobile: ah, yeah, didn't notice we are only building Capacitor scheme, and Cordova scheme was being built since it was a dependency of Capacitor scheme.

We can probably change `capacitor-swift-pm` to build Cordova scheme now instead of Capacitor scheme since I made Capacitor a dependency, would be simpler than your changes, but would still need to be conditional for using Cordova in 9.x and Capacitor in previous versions.

OS-pedrogustavobilro: Sounds good to me!

markemer (approved):

PR #8527, opened 2026-07-10
fix(ios): adopt UIScene life cycle in default iOS app templates

nilspep:
## Problem

Apple's [TN3187 – Migrating to the UIKit scene-based life cycle](https://developer.apple.com/documentation/technotes/tn3187-migrating-to-the-uikit-scene-based-life-cycle) states:

> In the next major release following iOS 26, UIScene life cycle will be required when building with the latest SDK; otherwise, your app won't launch.

Both default iOS app templates used by the CLI (`ios-spm-template` and `ios-pods-template`) currently only implement the legacy `AppDelegate`-only life cycle, with no `SceneDelegate` and no `UIApplicationSceneManifest` in `Info.plist`.

**Impact:** any app scaffolded today via `npx cap add ios` (or `--packagemanager CocoaPods`) will fail to launch when built against that newer SDK. There is no crash and no crash log — the process starts, the screen goes black, and the OS silently returns to the Home Screen. This is very hard to diagnose without knowing to look for this specific runtime issue in the Xcode console (`UIApplicationEvaluateRuntimeIssueForNoSceneLifecycleAdoption`).

I hit exactly this while testing a Capacitor app on a real device on a newer iOS SDK, and confirmed the fix against Apple's official (and recently revised) documentation before proposing it here.

## Fix

Minimal changes to both templates, following Apple's documented migration path 1:1:

- **`SceneDelegate.swift`** (new): implements `UIWindowSceneDelegate` with the standard lifecycle stubs. Since the app's root view controller (`CAPBridgeViewController`) is loaded from `Main.storyboard`, the system auto-configures the scene's window from `UISceneStoryboardFile` — no manual `UIWindow` setup is required.
- **`AppDelegate.swift`**: adds `application(_:configurationForConnecting:options:)` and `application(_:didDiscardSceneSessions:)`.
- **`Info.plist`**: adds `UIApplicationSceneManifest` with `UIApplicationSupportsMultipleScenes = false` and a single `UIWindowSceneSessionRoleApplication` configuration pointing at `SceneDelegate` and the existing `Main` storyboard — this matches Apple's current official example verbatim (same keys, same `UISceneConfigurationName`/`UISceneDelegateClassName` pattern).
- **`project.pbxproj`**: registers `SceneDelegate.swift` in the `App` target (file reference, build file, group, Sources build phase).

`UIMainStoryboardFile`/`UILaunchStoryboardName` are left untouched — they remain valid and are simply superseded for windowing purposes by the scene manifest, matching Apple's guidance for storyboard-based apps.

## Verification

- `Info.plist` and `project.pbxproj` validated with `plutil -lint` (both templates).
- Applied the identical change pattern to a real Capacitor app project, rebuilt with `xcodebuild`, and verified on a physical device: without the fix the app silently fails to launch (black screen, no crash log); with the fix it launches and runs normally.
- Cross-checked the `Info.plist` snippet and `SceneDelegate` pattern against Apple's official TN3187 example to ensure an exact match (not just "a" fix, but *the* documented fix).

## Notes for reviewers

- This is a non-breaking, additive change (existing storyboard-based launch flow is preserved), so I've targeted `main` per `CONTRIBUTING.md`.
- I don't have `npm install` run against the full monorepo tooling in this environment, so I wasn't able to run the repo's own SwiftLint config end-to-end; I did install SwiftLint locally and manually reviewed the new/changed Swift files against the existing file style (4-space indentation, matching comment style, no trailing whitespace, no force-unwraps). Happy to address any CI lint findings.
- Let me know if you'd like the `AppDelegate.configurationForConnecting` method dropped since it's technically redundant once `Info.plist` fully specifies the scene manifest — I kept it since Apple's own template ships it too and it's harmless, but I'm happy to simplify if preferred.

### DIFF
--- a/ios-pods-template/App/App.xcodeproj/project.pbxproj
+++ b/ios-pods-template/App/App.xcodeproj/project.pbxproj
@@ -10,6 +10,7 @@
 		2FAD9763203C412B000D30F8 /* config.xml in Resources */ = {isa = PBXBuildFile; fileRef = 2FAD9762203C412B000D30F8 /* config.xml */; };
 		50379B232058CBB4000EE86E /* capacitor.config.json in Resources */ = {isa = PBXBuildFile; fileRef = 50379B222058CBB4000EE86E /* capacitor.config.json */; };
 		504EC3081FED79650016851F /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 504EC3071FED79650016851F /* AppDelegate.swift */; };
+		2C6EFDE1674F365EA0601FC0 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6197F4D7F73A0E1131CB2963 /* SceneDelegate.swift */; };
 		504EC30D1FED79650016851F /* Main.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 504EC30B1FED79650016851F /* Main.storyboard */; };
 		504EC30F1FED79650016851F /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 504EC30E1FED79650016851F /* Assets.xcassets */; };
 		504EC3121FED79650016851F /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 504EC3101FED79650016851F /* LaunchScreen.storyboard */; };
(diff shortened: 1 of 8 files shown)


eric-horodyski:
Hi @nilspep, thank you for the thorough Pull Request. The team and I appreciate the time and effort put into this!

However, we have been working on a solution to adopt UIScene internally that we will be releasing imminently with Capacitor 8.5. Adopting UIScene required quite the amount of regression testing, the creation of migration documentation and tooling, and we're also releasing a Capacitor skill to help those with very custom AppDelegate files migrate. It is something we felt needed to be tackled internally. 

That said, I'll be closing this Pull Request.

As Engineering Manager for Capacitor, I'd like to find a better way to share what the team is currently working on with the community so we don't duplicate effort. I take complete blame on that, and sincerely apologize.

PR #8424, opened 2026-04-07
fix(SystemBars): make `safe-area-inset-x` available on API <= 34

theproducer:
## Description
This PR makes the custom injected `safe-area-inset-x` CSS variables available for Android devices on API <= 34, to make dealing with UI safe area logic more consistent across Android versions.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed
fixes: https://github.com/ionic-team/capacitor/issues/8394

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

PR #8528, opened 2026-07-12
fix(SystemBars): avoid extra IME padding on API <= 34 in insets passthrough

### LINKED ISSUE #8525: [Bug]: Capacitor v8.3.0 extra gray space when keyboard is visible
### Capacitor Version

Latest Dependencies:

  @capacitor/cli: 8.4.1
  @capacitor/core: 8.4.1
  @capacitor/android: 8.4.1
  @capacitor/ios: 8.4.1

Installed Dependencies:

  @capacitor/ios: not installed
  @capacitor/cli: 8.4.1
  @capacitor/core: 8.4.1
  @capacitor/android: 8.4.1

### Other API Details

```Shell
npm 11.6.2
node v24.12.0

Android OS version 10
```

### Platforms Affected

- [ ] iOS
- [x] Android
- [ ] Web

### Current Behavior

When the keyboard is visible (e.g. from opening an input field), there is a grey area that appears above the keyboard, taking up a lot of space. The height seems to be the same as the height of the keyboard.

I ran `npm init @capacitor/app@latest`, added an input field, and compiled a debug apk to install on my phone.

Only seems to affect my phone (android version 10). Tested on one other android phone (OS 13) with no issue. Couldn't seem to run the apk in android home emulator (just got a white screen).

https://github.com/user-attachments/assets/53688c75-7ae3-4c4c-a725-0b1055370296

### Expected Behavior

Do not expect to see the gray area

### Project Reproduction

Changed basically nothing, except that I removed `@capacitor/camera` and set `"launchAutoHide": true` in capacitor config file (otherwise I would be stuck at splash screen)

### Additional Information

The capacitor version I installed here is v8.4.1, but from my test the version that caused this issue was v8.3.0. The previous version (8.2.0) did not have this issue.

And from the changelog, it seems that the only notable change was the SystemBars using native safe area insets (https://github.com/ionic-team/capacitor/pull/8384)?

Probably the same issue as #8459 and #8412 but upgrading to latest capacitor version (8.4.1) did not help fix this issue.

Setting `insetsHandling` to `disable` fixes the issue but not sure how that would affect other android versions.

MedOussemaNjimi:
Fixes #8525

## What

Gates the IME padding in the insets **passthrough** branch of `SystemBars` (WebView >= 140 + `viewport-fit=cover`) behind `Build.VERSION.SDK_INT >= VANILLA_ICE_CREAM`, the same gate that #8439 added to the non-passthrough branch.

## Why

Since 8.3.0 (#8384), when an input is focused on Android 10, a gray area the height of the keyboard appears above the keyboard — reproducible with a fresh `npm init @capacitor/app@latest` app (whose template uses `viewport-fit=cover`, so the passthrough branch is taken on devices with WebView >= 140).

On API <= 34 Capacitor does not run edge-to-edge, so the system already resizes the window when the IME is shown. On API 29 the compat `WindowInsetsCompat.Type.ime()` insets are still reported as the keyboard height after that resize, so `v.setPadding(0, 0, 0, imeInsets.bottom)` compensated for the keyboard a second time, producing the extra gray space. (On API 30–34 the dispatched IME insets are already consumed after the resize, which is why those versions were unaffected.)

#8439 fixed the exact same double-compensation in the non-passthrough branch; the passthrough branch added later in #8424 reintroduced it without the SDK gate. This PR applies the same gate there.

On API 35+ (enforced edge-to-edge) the behavior is unchanged.

## Testing

- `./gradlew compileReleaseJavaWithJavac -b capacitor/build.gradle` passes.
- Repro per the issue: `npm init @capacitor/app@latest`, add an input field, run on Android 10 — the gray space above the keyboard no longer appears, while API 35+ keyboard handling is untouched.

### DIFF
--- a/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
+++ b/android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
@@ -204,8 +204,12 @@ public class SystemBars extends Plugin {
             boolean keyboardVisible = insets.isVisible(WindowInsetsCompat.Type.ime());
 
             if (shouldPassthroughInsets) {
-                // We need to correct for a possible shown IME
-                v.setPadding(0, 0, 0, keyboardVisible ? imeInsets.bottom : 0);
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.VANILLA_ICE_CREAM) {
+                    // We need to correct for a possible shown IME
+                    // On API <= 34 the window is not edge-to-edge, so the system already
+                    // resizes it for the IME and extra padding would double the space
+                    v.setPadding(0, 0, 0, keyboardVisible ? imeInsets.bottom : 0);
+                }
 
                 Insets safeAreaInsets = calcSafeAreaInsets(insets);
                 injectSafeAreaCSS(safeAreaInsets.top, safeAreaInsets.right, safeAreaInsets.bottom, safeAreaInsets.left);


android/capacitor/src/main/java/com/getcapacitor/plugin/SystemBars.java
tafelnl, line 230: The actual fix would be removing this weird if-statement. See my reference implementation here: https://github.com/capacitor-community/safe-area/blob/3042a26e278c7babf83c72c37fa0e1e9c0a32d35/android/src/main/java/com/getcapacitor/community/safearea/SafeAreaPlugin.java#L142

PR #8544, opened 2026-07-28
feat(cli): add migrator functionality for adopting UIScene

theproducer:
## Description
This PR updates the CLI migrator to assist with migrating pre-UIScene Capacitor projects to adopting UIScene: https://github.com/ionic-team/capacitor/pull/8536

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation
- [ ] Other (CI, chores, etc.)

## Platforms Affected
- [ ] Android
- [x] iOS
- [ ] Web

PR #8536, opened 2026-07-22
feat(ios): UIScene Support

theproducer:
## Description
Adds UIScene support to Capacitor, which is required in iOS 27

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [x] Breaking Change
- [ ] Documentation
- [ ] Other (CI, chores, etc.)

## Rationale / Problems Fixed
fixes: https://github.com/ionic-team/capacitor/issues/7961

## Platforms Affected
- [ ] Android
- [x] iOS
- [ ] Web

PR #8534, opened 2026-07-20
fix(cli): support TypeScript 7 when loading capacitor.config.ts

conbrad:
TypeScript 7 dropped the classic compiler API (`transpileModule`, `ModuleKind`, etc.) from its default export, so `requireTS`'s use of `ts.transpileModule` throws on projects with `typescript@7` installed and every command that reads `capacitor.config.ts` fails immediately.

This change detects when that API isn't available and falls back to Node's own built-in TypeScript syntax stripping via a real dynamic `import()` (stable since Node 23.6, available behind `--experimental-strip-types` since Node 22.6) instead of transpiling the file ourselves. Falls back to a clear error if neither is available. Projects on older TypeScript versions are unaffected: the existing `transpileModule` path is unchanged.

Fixes #8531

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation
- [ ] Other (CI, chores, etc.)

## Rationale / Problems Fixed
TypeScript 7 dropped the classic compiler API (`transpileModule`, `ModuleKind`, etc.) from its default export: `require('typescript')` now only exposes `version`/`versionMajorMinor`, with the replacement living behind non-call-compatible `typescript/unstable/*` subpaths. `requireTS` in `cli/src/util/node.ts` calls `ts.transpileModule(...)` with `module: ts.ModuleKind.CommonJS` to transpile `capacitor.config.ts` on the fly, so on any project with `typescript@7` installed, every command that loads the config (`sync`, `run`, `copy`, `config`, etc.) fails immediately with `TypeError: Cannot read properties of undefined (reading 'CommonJS')`.

This PR detects when the classic compiler API isn't available and falls back to loading the config via a real dynamic `import()`, letting Node's own built-in TypeScript type-stripping do the work instead (stable since Node 23.6, available behind `--experimental-strip-types` since 22.6, Capacitor CLI already requires Node >=22). Projects on TypeScript <7 are unaffected; the existing `transpileModule` path is unchanged.

Fixes #8531

## Tests or Reproductions
Verified against the minimal repro project linked in #8531 ([bradenbiz/cap-ts7-repro](https://github.com/bradenbiz/cap-ts7-repro)), which pins `typescript@7.0.2` and `@capacitor/cli@8.3.3`. `npx cap config` reproduces the exact failure from the issue before this fix, and correctly resolves the config (`appId: 'com.example.repro'`, `appName: 'repro'`, `webDir: 'dist'`) after it.

[...]

PR #8554, opened 2026-08-06
chore: merge 8.5.0 into next

andredestro:
## What

Ports the 8.5 line into the Capacitor 9 branch so the two lines don't diverge. This brings `next` up to date with `main` as of the 8.5.0 release:

- feat(ios): UIScene Support (#8536)
- feat(cli): add migrator functionality for adopting UIScene (#8544)
- fix(cli): support TypeScript 7 when loading capacitor.config.ts (#8534)
- fix(cli): use POSIX paths in CapApp-SPM Package.swift (#8549)

## Conflict resolutions

Most conflicts were version bookkeeping. The interesting ones are where 8.5's UIScene work overlapped with Capacitor 9 refactors:

- **Version files** (`lerna.json`, `package.json` x4, CHANGELOGs): kept `9.0.0-alpha.6`, with the 8.5.0 changelog entries interleaved below the alpha.6 sections.
- **`cli/src/tasks/migrate.ts`**: kept the v9 `updateAppDelegate` task and added the 8.5 `migrateToUIScene` step after it. The UIScene migrator self-detects already-migrated or partial projects and skips them, so running both is safe for apps coming from any 8.x version.
- **`CapacitorBridge.swift`**: kept the v9 `setupListeners` structure (Cordova decoupled, `cordovaIsPresent` guard) and adopted the 8.5 scene-filtered `UIScene` resume/pause observers in place of the `UIApplication` ones.
- **`project.pbxproj`**: kept the v9 removals of `CAPFile.swift` and `CAPBridge.swift` (#8551) and added `CAPSceneDelegateProxy.swift` to the Sources build phase.
- **`CAPSceneDelegateProxy.swift`** (semantic conflict, not textual): dropped the `CDVPluginHandleOpenURL` post. Cordova is decoupled from the Capacitor framework on the v9 line, and `CAPApplicationDelegateProxy` on `next` already dropped the same post, so the scene proxy now matches that behavior. Everything else from 8.5 is carried over unchanged, including the `.capacitorOpenURL` back-compat payload and the new scene-aware notifications.

## Verification

- CLI: `npm run build` passes, all 84 tests pass (including the new `migrate-uiscene` and `xcode` specs)
- iOS: `Capacitor` scheme builds and all 81 unit tests pass (iPhone 17 simulator, iOS 26.5)
- iOS: `Cordova` scheme builds

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,17 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+# [8.5.0](https://github.com/ionic-team/capacitor/compare/8.4.2...8.5.0) (2026-07-31)
+
+### Bug Fixes
+
+- **cli:** support TypeScript 7 when loading capacitor.config.ts ([#8534](https://github.com/ionic-team/capacitor/issues/8534)) ([4c1c870](https://github.com/ionic-team/capacitor/commit/4c1c8709413b9c19b008c99122ca330cc3c90e6f))
+
+### Features
+
+- **cli:** add migrator functionality for adopting UIScene ([#8544](https://github.com/ionic-team/capacitor/issues/8544)) ([984fa85](https://github.com/ionic-team/capacitor/commit/984fa85ba0adab0aacf895aed6323bf4b503dccb))
+- **ios:** UIScene Support ([#8536](https://github.com/ionic-team/capacitor/issues/8536)) ([3fa04a3](https://github.com/ionic-team/capacitor/commit/3fa04a357c92af34cd6fccb8124791963804a9dc))
+
 ## [8.4.2](https://github.com/ionic-team/capacitor/compare/8.4.1...8.4.2) (2026-07-14)
 
 ### Bug Fixes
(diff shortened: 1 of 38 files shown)


OS-pedrogustavobilro:
@andredestro the problem with this being a PR is that it will squash all previous commits. Ideally we'd do a straight merge to `next` locally and push. Is the PR because the merge from `main` had conflicts and you want some validation that everything's okay?

markemer:
> @andredestro the problem with this being a PR is that it will squash all previous commits. Ideally we'd do a straight merge to `next` locally and push. Is the PR because the merge from `main` had conflicts and you want some validation that everything's okay?

I just turned on merge commits for a hot second here.

andredestro:
> @andredestro the problem with this being a PR is that it will squash all previous commits. Ideally we'd do a straight merge to `next` locally and push. Is the PR because the merge from `main` had conflicts and you want some validation that everything's okay?

Yes, exactly. The merge had conflicts in a few areas where 8.5 overlapped with Capacitor 9 refactors, and one of them was semantic rather than textual, so I wanted a second pair of eyes before this lands on next.

theproducer (approved):

PR #8563, opened 2026-08-11
refactor(android)!: drop legacy @NativePlugin annotation support

OS-ruimoreiramendes:
## Description

Removes the legacy `@NativePlugin` annotation and its associated permission/activity result machinery, as part of the Capacitor 9 cleanup. `@NativePlugin` has been `@Deprecated` since Capacitor 3.0.0 (May 2021) in favor of `@CapacitorPlugin`.

ref: [RMET-5305](https://outsystemsrd.atlassian.net/browse/RMET-5305) (decision spike + implementation)

## Change Type
- [ ] Fix
- [ ] Feature
- [x] Refactor
- [x] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Capacitor 9 is a major release, the natural moment to drop APIs that have been deprecated for several majors.
`@NativePlugin` has been deprecated for 5+ years, and every published plugin still using it is frozen at Capacitor 3–4 (2020–2022) with many other breaking API dependencies that already prevent it from running on modern Capacitor.

## Tests or Reproductions

- `./gradlew :capacitor-android:compileDebugJavaWithJavac` and `:capacitor-cordova-android:compileDebugJavaWithJavac` pass with no warnings.
- `./gradlew :capacitor-android:testDebugUnitTest` passes.
- Full `./gradlew build` passes.
- CLI regex change verified via `npx cap sync android` on `capacitor-testapp` (20 plugins discovered correctly).
- Runtime validation on a Samsung Galaxy S24 FE (Android 16) with `@capacitor/android` linked locally against this branch:
    - `@capacitor/camera`:  permission dialog + camera activity + photo delivered back.
    - `@capacitor/share`:  4 scenarios (success, cancel, empty payload validation, share of photo path).
    - `@capacitor/geolocation` (via [capacitor-geolocation example-app](https://github.com/ionic-team/capacitor-geolocation)): permission dialog, native execution, JS callback with error propagation.
    - `cordova-plugin-inappbrowser`: `__CordovaPlugin` registers, Cordova bridge interceptor routes JS to Native, InAppBrowser lifecycle completes cleanly. Confirms Cordova compatibility is not broken.

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

## Notes / Comments

- A migration guide entry for these removals is in ionic-team/capacitor-docs#586.

[RMET-5305]: https://outsystemsrd.atlassian.net/browse/RMET-5305?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


android/capacitor/src/main/java/com/getcapacitor/Plugin.java
```diff
@@ -323,19 +323,15 @@ public boolean hasDefinedPermissions(String[] permissions) {
...
-            // Check for legacy plugin annotation, @NativePlugin
-            NativePlugin legacyAnnotation = handle.getLegacyPluginAnnotation();
-            return hasDefinedPermissions(legacyAnnotation.permissions());
-        } else {
-            for (Permission perm : annotation.permissions()) {
-                for (String permString : perm.strings()) {
-                    if (!PermissionHelper.hasDefinedPermission(getContext(), permString)) {
-                        return false;
-                    }
+            return false;
```
OS-pedrogustavobilro: I'm seeing that before, if there was no `@CapacitorPlugin` nor `@NativePlugin` annotation, `hasDefinedRequiredPermissions` and `hasRequiredPermissions`, it would previously `return true` where now it does `return false`. I think this is of no consequence because not having that annotation means the plugin wouldn't really be registered I think, so we should be okay?

Just wanted to bring it up.

OS-ruimoreiramendes: Good point, after this PR the `PluginHandle` constructor throws when `@CapacitorPlugin` is missing, so `getPluginAnnotation()` can't return null on a registered plugin. The `if (annotation == null)` branch is effectively unreachable. Went with `return false` as the defensive default, but on reflection keeping `return true` (matching the previous behavior in the unreachable path) is probably safer. I'll flip both methods to `true` if you agree.

OS-pedrogustavobilro: It could be that there isn't much consequence to returning false or true, because the plugin not having an annotation will be a bigger issue than just permissions.

But, for sake of consistency, doesn't hurt to leave `true` as it was before

OS-ruimoreiramendes: Done

OS-pedrogustavobilro (approved):

PR #8567, opened 2026-08-11
fix(android-template): declare app/build.gradle dependency versions explicitly

OS-ruimoreiramendes:
## Description

Fixes the "Implicit lookup of properties in parent projects" Gradle 9.6 deprecation in the Android app template.

ref: [RMET-5369](https://outsystemsrd.atlassian.net/browse/RMET-5369)

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Gradle 9.6 deprecated implicit parent property lookup. This will fail with an error in Gradle 10. Declaring the versions locally removes the implicit lookup and makes the template forward-compatible.

## Tests or Reproductions

Verified on a fresh Capacitor 9 app (Gradle 9.6.0, `--warning-mode=all`) that the 6 warnings on `app/build.gradle` are gone after the fix, and that running `updateAppBuildGradle` against a Cap 8-style sample inserts the `def` block.

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

[RMET-5369]: https://outsystemsrd.atlassian.net/browse/RMET-5369?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -553,6 +553,23 @@ async function updateAppBuildGradle(filename: string) {
 `,
     '',
   );
+
+  if (!replaced.includes('def androidxAppCompatVersion = rootProject.ext.androidxAppCompatVersion')) {
+    replaced = replaced.replace(
+      `apply plugin: 'com.android.application'\n\n`,
+      `apply plugin: 'com.android.application'
+
+def androidxAppCompatVersion = rootProject.ext.androidxAppCompatVersion
+def androidxCoordinatorLayoutVersion = rootProject.ext.androidxCoordinatorLayoutVersion
+def coreSplashScreenVersion = rootProject.ext.coreSplashScreenVersion
+def junitVersion = rootProject.ext.junitVersion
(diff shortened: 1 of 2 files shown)


cli/src/tasks/migrate.ts
OS-pedrogustavobilro, line 1: We actually had a [separate task to update the migrator command](https://outsystemsrd.atlassian.net/browse/RMET-5370) (and other Gradle / AGP changes). At the risk of being "that guy" and nitpicky 😅, this would be better to go in a separate PR, if you don't mind 🙏

OS-ruimoreiramendes: Ah, missed that ticket. I'll move the `migrate.ts` change out of this PR and open a separate ONE

OS-ruimoreiramendes: reverted

OS-pedrogustavobilro (approved):

PR #8572, opened 2026-08-13
fix(cli): replace usesCleartextTraffic with network-security-config

OS-ruimoreiramendes:
## Description

Migrates `usesCleartextTraffic opt-in in the generated `capacitor-cordova-android-plugins` module from the deprecated `android:usesCleartextTraffic` attribute to `network-security-config.xml` with `cleartextTrafficPermitted="true"`. Any legacy `usesCleartextTraffic` attribute injected by Cordova plugins via `edit-config` is filtered out, so the final manifest is free of the deprecated attribute. The `android:networkSecurityConfig` reference is set on the library module's manifest and propagated to the app's final manifest at build time.

The three existing triggers still work: `server.cleartext=true` in the Capacitor config, live-reload, and Cordova plugins declaring `usesCleartextTraffic="true"` via `edit-config`. When none apply, the `network-security-config.xml` file is removed.

ref: [RMET-5243](https://outsystemsrd.atlassian.net/browse/RMET-5243)

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

`android:usesCleartextTraffic` is [deprecated in a future Android version](https://developer.android.com/about/versions/17/behavior-changes-all#uses-clear-traffic-deprecation) and will eventually be removed.
The recommended replacement is `network-security-config` with `cleartextTrafficPermitted`.

## Tests or Reproductions

Verified (inspecting the generated manifest + `network-security-config.xml` file) and in runtime on a Samsung SM-S721B using the `@capacitor/inappbrowser` example app pointed at `http://httpforever.com` and `https://outsystems.com`:

- Default (no cleartext trigger): HTTP blocked, HTTPS loads.
- `server.cleartext=true`: HTTP loads, HTTPS loads.
- Cordova plugin `edit-config` with `usesCleartextTraffic="true"` (and `server.cleartext=false`): HTTP loads, HTTPS loads. Legacy attribute filtered out of the final manifest.
- live-reload: HTTP loads, HTTPS loads.

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

[RMET-5243]: https://outsystemsrd.atlassian.net/browse/RMET-5243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/cli/src/cordova.ts
+++ b/cli/src/cordova.ts
@@ -867,14 +867,30 @@ export async function writeCordovaAndroidManifest(
     });
   });
   const cleartextString = 'android:usesCleartextTraffic="true"';
-  const cleartextValue =
-    (cleartext || config.app.extConfig.server?.cleartext) && !applicationXMLAttributes.includes(cleartextString)
-      ? cleartextString
+  const networkSecurityConfigValue =
+    cleartext || config.app.extConfig.server?.cleartext || applicationXMLAttributes.includes(cleartextString)
+      ? 'android:networkSecurityConfig="@xml/network_security_config"'
       : '';
+  const networkSecurityConfigDir = join(config.android.cordovaPluginsDirAbs, 'src', 'main', 'res', 'xml');
+  const networkSecurityConfigPath = join(networkSecurityConfigDir, 'network_security_config.xml');
+  if (networkSecurityConfigValue) {
+    await ensureDir(networkSecurityConfigDir);
+    await writeFile(
+      networkSecurityConfigPath,
+      `<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="true" />
+</network-security-config>
+`,
+    );
+  } else if (await pathExists(networkSecurityConfigPath)) {
+    await remove(networkSecurityConfigPath);
+  }
+
   let content = `<?xml version='1.0' encoding='utf-8'?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
 xmlns:amazon="http://schemas.amazon.com/apk/res/android">
-<application ${applicationXMLAttributes.join('\n')} ${cleartextValue}>
+<application ${applicationXMLAttributes.filter((attr) => !attr.includes('android:usesCleartextTraffic')).join('\n')} ${networkSecurityConfigValue}>
 ${applicationXMLEntries.join('\n')}
 </application>
 ${rootXMLEntries.join('\n')}


cli/src/cordova.ts
```diff
@@ -867,14 +867,30 @@ export async function writeCordovaAndroidManifest(
...
+      networkSecurityConfigPath,
+      `<?xml version="1.0" encoding="utf-8"?>
+<network-security-config>
+    <base-config cleartextTrafficPermitted="true" />
+</network-security-config>
+`,
+    );
+  } else if (await pathExists(networkSecurityConfigPath)) {
+    await remove(networkSecurityConfigPath);
+  }
```
OS-pedrogustavobilro, line 888: Because se are not writing to the app's manifest directly (but maybe we might need to, in relation to my other comment), I think we're fine. However, I wonder if you tested this in an app that has their own `network-security-config` (maybe with clearTestTraffic=false, or with pinning configuration like SSL Pinning Plugin).

Wondering if those could clash and fail the build (which may not fail before with the deprecated way), and if we need to place additional guards to protect those cases.

Let me know if that makes sense.

cli/src/cordova.ts
```diff
@@ -867,14 +867,30 @@ export async function writeCordovaAndroidManifest(
...
   });
   const cleartextString = 'android:usesCleartextTraffic="true"';
-  const cleartextValue =
-    (cleartext || config.app.extConfig.server?.cleartext) && !applicationXMLAttributes.includes(cleartextString)
-      ? cleartextString
+  const networkSecurityConfigValue =
+    cleartext || config.app.extConfig.server?.cleartext || applicationXMLAttributes.includes(cleartextString)
+      ? 'android:networkSecurityConfig="@xml/network_security_config"'
       : '';
+  const networkSecurityConfigDir = join(config.android.cordovaPluginsDirAbs, 'src', 'main', 'res', 'xml');
```
OS-pedrogustavobilro, line 874: Now that I see it, I'm skeptical of this working in general for Capacitor 9.

Since capacitor 9 only comes with the `cordovaPluginsDirAbs` tied to the app project if the app has cordova plugins, we could run into situations where setting `config.app.extConfig.server?.cleartext` would have no effect because the app has no Cordova Plugins - which would lead me to indicate that perhaps this needs to be written somewhere else?

This is kind of separate from your PR, but I guess is highlighted by it.

Did you run into this when testing, or did you always test in apps with a Cordova Plugin?

cli/src/cordova.ts
OS-ruimoreiramendes: Tested with `server.cleartext=true` + app declaring its own `network-security-config`, in order to understand what could happen, and here are the findings:

- **Same filename** (`network_security_config.xml`): build passes, but the app's file wins, our `server.cleartext=true` is silently ignored. Also tested with: config `false` + app `network_security_config` `true` still allows HTTP.

- **Different filename**: build fails with manifest merger conflict (different values for `android:networkSecurityConfig` between library and app). The error itself suggests the standard Android fix: add `tools:replace="android:networkSecurityConfig"` to the app's `<application>`, once the user does that, build passes and the app's file wins.

`Manifest merger failed : Attribute application@networkSecurityConfig value=(@xml/my_security_config) from AndroidManifest.xml:12:9-64
	is also present at [:capacitor-cordova-android-plugins] AndroidManifest.xml:8:18-78 value=(@xml/network_security_config).
	Suggestion: add 'tools:replace="android:networkSecurityConfig"' to <application> element at AndroidManifest.xml:5:5-39:19 to override.
`

 So when the app has its own NSC, our injection is either silently overridden or breaks the build.

OS-pedrogustavobilro: Hmm, that's why we may need to write the config directly to the app, but in a way that doesn't break the existing config (I'm fine with overriding the cleartext to true if the config requests but the app has it at false, which is technically what the CLI is meant to be doing, but not so much overwriting the entire `networkSecurityConfig `file).

But also that has implications in relation to the other PR comment.

cli/src/cordova.ts
OS-ruimoreiramendes: Yes, ran into it during testing. Started in an app without any Cordova plugins with `server.cleartext=true` set, and on the first sync noticed the mechanism wasn't triggered, `writeCordovaAndroidManifest` only runs with effect when there are Cordova plugins (via the `if(enableCordova)` path in `update.ts`). Without one, the library manifest stayed empty (no network-security-config, no attribute), and `server.cleartext=true` was a no-op. Ended up installing `cordova-plugin-device` so the mechanism would trigger and I could test in runtime.

From what I understand this is inherited from the pre-existing behavior, the old `usesCleartextTraffic="true"` injection had the same limitation on Cap 9, but this PR keeps it that way.

OS-pedrogustavobilro: Yes you are correct that this is not a fault with your PR specifically, but it should be fixed for Cap 9 imo.

Now, that doesn't mean it needs to be fixed in your PR necessarily, but maybe because of the other comment it could need further changes? We can align offline as well.

OS-pedrogustavobilro (approved):
Tested here and seems to work fine with a cordova plugin installed. As we aligned privately, the other changes in PR comments will be scoped to a separate task / PR.

OS-ruimoreiramendes:
> Tested here and seems to work fine with a cordova plugin installed. As we aligned privately, the other changes in PR comments will be scoped to a separate task / PR.

Here is the separate [task](https://outsystemsrd.atlassian.net/browse/RMET-5420) to address the other comments.

PR #8576, opened 2026-08-18
feat: include option to force cordova inclusion

markemer:
## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Platforms Affected
- [x] Android
- [x] iOS
- [x] Web

### DIFF
--- a/cli/src/android/update.ts
+++ b/cli/src/android/update.ts
@@ -43,6 +43,9 @@ export async function updateAndroid(config: Config): Promise<void> {
   await removePluginsNativeFiles(config);
   if (enableCordova) {
     logger.info('Found Cordova Plugins: Including Android Cordova Support');
+    if (config.app.forceCordova) {
+      logger.info('Cordova support installation has been forced')
+    }
     await copyPluginsNativeFiles(config, cordovaPlugins);
   }
   if (!(await pathExists(config.android.webDirAbs))) {
--- a/cli/src/android/update.ts
+++ b/cli/src/android/update.ts
@@ -44,7 +44,7 @@ export async function updateAndroid(config: Config): Promise<void> {
   if (enableCordova) {
     logger.info('Found Cordova Plugins: Including Android Cordova Support');
     if (config.app.forceCordova) {
-      logger.info('Cordova support installation has been forced')
+      logger.info('Cordova support installation has been forced');
     }
     await copyPluginsNativeFiles(config, cordovaPlugins);
   }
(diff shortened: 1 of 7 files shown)


cli/src/ios/update.ts
```diff
@@ -43,7 +43,7 @@ export async function updateIOS(config: Config, deployment: boolean): Promise<vo
 async function updatePluginFiles(config: Config, plugins: Plugin[], deployment: boolean) {
   await removePluginsNativeFiles(config);
   const cordovaPlugins = plugins.filter((p) => getPluginType(p, platform) === PluginType.Cordova);
-  const enableCordova = cordovaPlugins.length > 0;
+  const enableCordova = cordovaPlugins.length > 0 || config.app.forceCordova;
 
   if (enableCordova) {
     logger.info('Found Cordova Plugins: Including iOS Cordova Support');
```
OS-pedrogustavobilro, line 49: Minor: should this log message be tweaked since now cordova support may exist even if there are no Cordova Plugins?

cli/src/android/update.ts
OS-pedrogustavobilro, line 45: Equivalent (minor) comment: should this log message be tweaked since now cordova support may exist even if there are no Cordova Plugins?

OS-pedrogustavobilro: @markemer Missing a check for `config.app.forceCordova` here I think?

markemer: Good catch. How'd that one sneak by?

cli/src/android/update.ts
```diff
@@ -37,12 +37,15 @@ export async function updateAndroid(config: Config): Promise<void> {
...
-  const enableCordova = cordovaPlugins.length > 0;
+  const enableCordova = cordovaPlugins.length > 0 || config.app.forceCordova;
 
   await writePluginsJson(config, capacitorPlugins, enableCordova);
   await removePluginsNativeFiles(config);
   if (enableCordova) {
     logger.info('Found Cordova Plugins: Including Android Cordova Support');
+    if (config.app.forceCordova) {
+      logger.info('Cordova support installation has been forced');
+    }
```
OS-pedrogustavobilro, line 48: Minor: This if block logs "Found Cordova Plugins" even if there are 0 cordova plugins with `config.app.forceCordova` set to true, which might be misleading? Unless that's what you were going with.

(also applies to the iOS version of `update.ts`)

OS-pedrogustavobilro (approved):
I tested with a [few combination of Capacitor apps](https://github.com/OS-pedrogustavobilro/capacitor-test-apps-for-cordova-optional), everything seems to work fine!

Left an additional non-blocking comment about the logging, but approving the PR.

cli/src/android/update.ts
markemer: Yeah, I really have to think about this language/logic a little, but I thought just "warning" (I did wan't to use warn, but I thought people might complain) when it's enabled and forced would be enough for now.

PR #8578, opened 2026-08-19
feat!(iOS): support for source-based Swift package

theproducer:
## Description
Converts Capacitor's iOS library from an Xcode-project-based, dual CocoaPods/prebuilt-XCFramework distribution model to a single source-of-truth Swift Package Manager package, while keeping the CocoaPods path fully working and unchanged for existing apps.

## Change Type
- [ ] Fix
- [x] Feature
- [ ] Refactor
- [x] Breaking Change
- [ ] Documentation

## Platforms Affected
- [ ] Android
- [x] iOS
- [ ] Web


.github/workflows/ci.yml
```diff
@@ -89,15 +89,15 @@ jobs:
       - run: npm test
         working-directory: ./core
   test-ios:
-    runs-on: macos-15
+    runs-on: macos-26-intel
```
markemer: Why intel?

theproducer: No reason - other than just that `macos-15` was Intel.  I can try with the ARM variant, it'll probably work fine.

ios/Sources/Capacitor/CAPPlugin.swift
```diff
@@ -0,0 +1,208 @@
...
+
+        listenersForEvent.removeObject(at: listenerIndex)
+    }
+
+    @objc public func notifyListeners(_ eventName: String, data: [String: Any]?) {
+        notifyListeners(eventName, data: data, retainUntilConsumed: false)
+    }
+
+    @objc public func notifyListeners(_ eventName: String, data: [String: Any]?, retainUntilConsumed: Bool) {
+        guard let listenersForEvent = eventListeners.object(forKey: eventName) as? [CAPPluginCall] else {
```
andredestro:
```suggestion
        let listenersForEvent = eventListeners.object(forKey: eventName) as? [CAPPluginCall] ?? []
        guard !listenersForEvent.isEmpty else {
```

The Obj-C version this replaces checked two conditions:

```objc
NSArray<CAPPluginCall *> *listenersForEvent = [self.eventListeners objectForKey:eventName];
if (listenersForEvent == nil || [listenersForEvent count] == 0) {
```

The Obj-C original checked `listenersForEvent == nil || [listenersForEvent count] == 0`. The Swift port expresses "no listeners" as a failed cast, which only covers the `nil` half: an empty `NSMutableArray` bridges fine, so the retain branch is skipped and the event is dropped.

The array ends up empty whenever a listener is added and removed (`removeEventListener` never removes the key), so `retainUntilConsumed: true` breaks exactly in the re-attach case it exists for. Confirmed with a throwaway test on this branch.


Worth adding a regression test for the full cycle, since there is currently no test file covering `CAPPlugin`:

```swift
func testRetainedEventIsDeliveredAfterListenerIsRemovedAndReAdded() {
    let plugin = CAPPlugin()
    plugin.addEventListener("myEvent", listener: firstListener)
    plugin.removeEventListener("myEvent", listener: firstListener)

    plugin.notifyListeners("myEvent", data: ["foo": "bar"], retainUntilConsumed: true)
    plugin.addEventListener("myEvent", listener: secondListener)

    XCTAssertEqual(receivedEvents.count, 1)
}
```

cli/src/util/cordova-ios.ts
```diff
@@ -353,18 +354,26 @@ export async function generateCordovaPackageFile(p: Plugin, config: Config): Pro
             publicHeadersPath: "."`;
   }
 
+  const useSourceSPM = major(iosPlatformVersion) >= 9;
+  const capacitorPackageName = useSourceSPM ? 'capacitor' : 'capacitor-swift-pm';
+  const capacitorPackageUrl = useSourceSPM
+    ? 'https://github.com/ionic-team/capacitor'
+    : 'https://github.com/ionic-team/capacitor-swift-pm.git';
+  const capacitorPackageVersionSuffix = useSourceSPM
+    ? ` branch: "feature/source-spm"`
```
andredestro:
```suggestion
    ? ` branch: "next"`
```

ios/Sources/Capacitor/InstanceConfiguration.swift
```diff
@@ -0,0 +1,115 @@
...
+            self.loggingEnabled = true
+        case .debug:
+            self.loggingEnabled = isDebug
+        case .none:
+            self.loggingEnabled = false
+        @unknown default:
+            self.loggingEnabled = false
+        }
+
+        self.localURL = URL(string: "\(descriptor.urlScheme)://\(descriptor.urlHostname)")!
```
andredestro, line 74: The force-unwrap turns a nil URL into a launch-time crash.
But I'm not 100% sure whether it's a possible scenario.

ios/Sources/Capacitor/CAPPluginCall.swift
```diff
@@ -0,0 +1,186 @@
...
+        options as? JSObject ?? [:]
+    }
+}
+
+@objc extension CAPPluginCall: BridgedJSValueContainer {
+    public var dictionaryRepresentation: NSDictionary {
+        options as NSDictionary
+    }
+
+    public static var jsDateFormatter: ISO8601DateFormatter {
```
andredestro: Was this change intentional? It was previously stored as a static variable, so this makes `jsDateFormatter` return a new instance on every access.

ios/Sources/Capacitor/CAPPlugin.swift
theproducer: Ah, thanks for spotting this!

theproducer:
Closing in favor of an updated approach better compatible with objective-c plugins due to https://forums.swift.org/t/cannot-subclass-a-class-with-objc-subclassing-restricted-attribute/5674/

PR #8579, opened 2026-08-25
fix(cli): run gradle wrapper before root build.gradle migration

jcesarmobile:
In https://github.com/ionic-team/capacitor/pull/8543 the order of updating files was changed to edit settings.gradle and app/build.gradle before updating gradle files as it could cause issues, but the root build.gradle was also moved to run before updating the wrapper files.
The wrapper files need to be updated before updating the root build.gradle as if the AGP version is not compatible with current gradle it will fail to update the wrapper files, it doesn't happen with current versions, but could break in the future, in example if we update to AGP 9.4.0.

Added a comment as it's not the first time the wrapper files update was moved after the root build.gradle update and broke in a following update.

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -202,10 +202,6 @@ export async function migrateCommand(config: Config, noprompt: boolean, packagem
           join(config.android.platformDirAbs, 'gradle', 'wrapper', 'gradle-wrapper.properties'),
         );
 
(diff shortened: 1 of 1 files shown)


alexgerardojacinto (approved):

OS-pedrogustavobilro (approved):
Well caught.

I had looked through the commit history before to try and understand if there was a reasoning for that particular order, but could not find it.

The comment is a good add too, should prevent further blunders 😅

PR #8580, opened 2026-08-26
fix(cli): restore Cordova Package.swift generator for Cap 9

OS-pedrogustavobilro:
## Description

Restores `buildBinaryTargetEntries`, `buildResourcesText`, `buildDependencyTexts`, `buildCSettingsText`, and the weak/required system-framework linker-settings logic in `generateCordovaPackageFile()`'s generated `Package.swift` branch.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

These methods were originally added by #8445, #8447, #8448, #8455, and #8457 on `main`, but then were collectively lost on a merge from `main` to `next` branch: `next` had already relocated this generator into `cli/src/util/cordova-ios.ts` (via #8328's Cordova-optional refactor), so `main`'s still-monolithic update.ts was probably incorrectly treated as a stale deletion and none of the five PRs' new content was ported into Capacitor 9.

It's likely this has been like this for some time, and simply it missed us

I analysed other commits and merges (along with Claude 🤖), did not find any other kind of these issues

## Tests or Reproductions

Test current Capacitor 9 alpha with a Cordova Plugin that requires updates to the auto-generated `Package.swift` file (I chose https://github.com/OutSystems/cordova-outsystems-healthfitness that contains an xcframework). You'll find that it fails to build on Xcode, but with this PR it builds.

## Platforms Affected
- [ ] Android
- [x] iOS
- [ ] Web

## Notes / Comments

For PR reviewers:

- Internal Jira Reference: https://outsystemsrd.atlassian.net/browse/RMET-5435
- [Internal App used to test](https://github.com/ionic-team/cordova-plugins-ios-spm-test-apps/tree/main/healthfitness_capacitor9_spm)

### DIFF
--- a/cli/src/util/cordova-ios.ts
+++ b/cli/src/util/cordova-ios.ts
@@ -5,6 +5,7 @@ import { getCapacitorPackageVersion } from '../common';
 import { needsStaticPod } from '../cordova';
 import type { Config } from '../definitions';
 import { getMajoriOSVersion } from '../ios/common';
+import { logger } from '../log';
 import {
   PluginType,
   getAllElements,
@@ -12,6 +13,7 @@ import {
   getPlatformElement,
   getPluginPlatform,
   getPluginType,
+  resolvePlugin,
 } from '../plugin';
 import type { Plugin } from '../plugin';
 import { setAllStringIn } from '../tasks/migrate';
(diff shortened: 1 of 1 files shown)


alexgerardojacinto:
@OS-pedrogustavobilro I tested this myself first and it worked fine on my test. Then I asked Claude to verify and this is what it told me:

> One real regression I found by running it, not just reading it: the original main code had a guard — skip writing Package.swift entirely if the plugin has no source files, header files, or resources (i.e., a JS-only Cordova plugin). That guard didn't make it into this restore. I proved it by running the function against a JS-only plugin fixture: it now writes an empty, sourceless Package.swift where main would write nothing.
> 
> I also checked whether this actually breaks builds today: cli/src/util/spm.ts (untouched by this PR) independently guards against wiring such plugins into the app's SPM dependency graph, so I don't believe this breaks swift build right now — it's a stray/dead file, not a build-breaker. Still worth a fix since it's a behavioral divergence from main and a latent trap for future code that assumes an empty target never exists.
> 
> 
> Recommendation: minor fix needed before merge — restore the early-return guard in generateCordovaPackageFile's else branch. Everything else looks solid and the core fix is verified working.

OS-pedrogustavobilro:
Thanks for the review!

I believe that is referring to https://github.com/ionic-team/capacitor/pull/8443/changes - I've added it in this commit https://github.com/ionic-team/capacitor/pull/8580/commits/24a0912ac45009bb92a07fac2d9eef6661c0465c

andredestro (approved):

PR #8584, opened 2026-08-28
fix(cli): remove AGP 9 property shims and jcenter() in cap migrate

andredestro:
## Description

Extends `cap migrate` with two Android cleanup steps that Gradle 9 / AGP 9 make necessary:

1. **`gradle.properties`**: removes the 10 compatibility flags that Android Studio's AGP Upgrade Assistant writes to preserve the AGP 8 defaults. `android.useAndroidX=true` is kept.
2. **`jcenter()`**: removes it from `android/build.gradle`.

Both steps run **before** the gradle wrapper upgrade. The wrapper task runs on Gradle 9.5.1 the second time around, so a leftover `jcenter()` would fail it at the configuration phase and the wrapper upgrade would be lost.

`mavenCentral()` is not added back, and `android/app/build.gradle` is not scanned, per review feedback: the 3 -> 4 migrator always added `mavenCentral()` regardless of the opt-in `jcenter()` removal, it has been in the template since Capacitor 4, and `jcenter()` has only ever been in the root `build.gradle` of the templates.

Adds `cli/test/migrate-android-gradle.spec.ts`: 15 tests covering the transformations and a block that runs them against a project on disk.

## Change Type
- [x] Fix
- [ ] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed

Users have been running the AGP 9 migration in Android Studio while still on Capacitor 8. The Assistant injects flags that a Capacitor app does not need: most are deprecated and will be removed in AGP 10, and a couple break a Capacitor 9 build (`android.builtInKotlin=false` disables the Kotlin support AGP 9 now bundles, `android.sdk.defaultTargetSdkToCompileSdkIfUnset=false` stops AGP from inferring `targetSdkVersion`).

`RepositoryHandler.jcenter()` was removed in Gradle 9.0. Any `build.gradle` that still calls it fails immediately at the configuration phase with `Could not find method jcenter()`.

## Platforms Affected
- [x] Android
- [ ] iOS
- [ ] Web

### DIFF
--- a/cli/src/tasks/migrate.ts
+++ b/cli/src/tasks/migrate.ts
@@ -69,8 +69,6 @@ const agpMigrationAssistantProperties = [
 ];
 
 const jcenterLine = /^jcenter\(\)\s*(\/\/.*)?$/;
-const mavenCentralLine = /^mavenCentral\(\)\s*(\/\/.*)?$/;
-const repositoriesBlockStart = /(^|\s)repositories\s*\{/;
 
 export async function migrateCommand(config: Config, noprompt: boolean, packagemanager: string): Promise<void> {
   if (config === null) {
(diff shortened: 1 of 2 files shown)


jcesarmobile:
I don't think migrate command should add `mavenCentral()`, it was already added in the 3 to 4 migrator, the deletion of `jcenter()` was optional, but the addition of `mavenCentral()` was always done, so we could assume all apps have it.

`jcenter()` has never been in `android/app/build.gradle` in our templates, only in `android/build.gradle`, so I don't think we should worry about that neither.

jcesarmobile (approved):

PR #8585, opened 2026-09-02
feat(cli): migrate apps to ESLint 10 in cap migrate

ItsChaceD:
## Description
`npx cap migrate` can now optionally update apps that use `@ionic/eslint-config`: bumps ESLint to v10 and the config to v1, replaces the `eslintConfig` block and `.eslintignore` with an `eslint.config.cjs`, and drops `--ext ts` from scripts. Opt-in only: an "Update ESLint setup? (y/N)" prompt (default no), or the `--update-eslint` flag for non-interactive runs. When skipped or absent, nothing changes. A block with custom rules is kept so they can be ported manually.

## Change Type
- [ ] Fix
- [X] Feature
- [ ] Refactor
- [ ] Breaking Change
- [ ] Documentation

## Rationale / Problems Fixed
https://outsystemsrd.atlassian.net/browse/RMET-4734

## Platforms Affected
- [ ] Android
- [ ] iOS
- [X] Web

## Notes
Should not ship in a release before `@ionic/eslint-config@1.0.0` is published.

### DIFF
--- a/cli/src/index.ts
+++ b/cli/src/index.ts
@@ -384,11 +384,12 @@ export function runProgram(config: Config): void {
     .command('migrate')
     .option('--noprompt', 'do not prompt for confirmation')
     .option('--packagemanager <packageManager>', 'The package manager to use for dependency installs (npm, pnpm, yarn)')
+    .option('--update-eslint', 'also update ESLint to v10 and @ionic/eslint-config to v1 (breaking change to the lint setup)')
     .description('Migrate your current Capacitor app to the latest major version of Capacitor.')
     .action(
-      wrapAction(async ({ noprompt, packagemanager }) => {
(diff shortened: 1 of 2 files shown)


jcesarmobile:
I don't think we should do this for user apps, and I doubt any apps is using `@ionic/eslint-config`, that package is meant for npm tooling, I don't think it plays well with web code, we don't include it in the apps created by `npm init @capacitor/app`, nor by apps created with `@ionic/cli` (they use eslint rules for react, vue or angular based on the chosen framework), and on plugins created with `npm init @capacitor/plugin`, since it also creates an app with `npm init @capacitor/app` in the process, we exclude the generated `example-app` from `eslint` and `prettier` as they cause issues on them.

PR #8593, opened 2026-09-08
fix(cli): load default exports from JavaScript config

### LINKED ISSUE #8267: [Bug]: capacitor.config.js not working
### Capacitor Version

💊   Capacitor Doctor  💊 

Latest Dependencies:

  @capacitor/cli: 7.4.4
  @capacitor/core: 7.4.4
  @capacitor/android: 7.4.4
  @capacitor/ios: 7.4.4

Installed Dependencies:

  @capacitor/cli: 7.4.4
  @capacitor/core: 7.4.4
  @capacitor/ios: 7.4.4
  @capacitor/android: 7.4.4

### Other API Details

```Shell
npm --version
10.9.2

node -v
v22.15.1
```

### Platforms Affected

- [x] iOS
- [x] Android
- [x] Web

### Current Behavior

capacitor.config.js has export default config object, but cli processed it wrong way:

`capacitor.config.js`

```js
const isProduction = process.env.CAP_ENV == 'production' ? true : false;

const config = {
  appId: 'com.svelte.app',
  appName: 'SvelteCap',
  webDir: 'build/frontend',
  // server добавляем только если включен режим разработки
  server: !isProduction ? {
    url: 'http://192.168.1.102:5173',
    cleartext: true
  } : undefined,
  plugins: {
    Geolocation: {
      enabled: true
    }
  }
};

export default config;
```

Try to run `cap sync` cmd:

```sh
npx cap sync
[error] Could not find the web assets directory: ./www.
        Please create it and make sure it has an index.html file. You can change the path of this directory in
        capacitor.config.js (webDir option). You may need to compile the web assets for your app (typically npm run
        build). More info: https://capacitorjs.com/docs/basics/workflow#sync-your-project
```

The issue lies here https://github.com/ionic-team/capacitor/blob/5b27fdf44a924f6aaf9b15555999bc839e7bfe81/cli/src/config.ts#L143.

await require() returns object like this: {default: {/* config object */}.

To fix this we need to get same config require as TS version do, this line: https://github.com/ionic-team/capacitor/blob/5b27fdf44a924f6aaf9b15555999bc839e7bfe81/cli/src/config.ts#L116.

### Expected Behavior

capacitor.config.js should work as expected, i change for example webDir config and cli start to sync new folder.

### Project Reproduction

https://github.com/boomfly/capacitor-config-js-issue

### Additional Information

_No response_

jabrailkhalil:
When `capacitor.config.js` uses `export default`, Node returns a module namespace object. The CLI reads settings from that wrapper, so `appId` and `appName` become empty and `webDir` falls back to `www`.

Unwrap the default export using the same logic as the TypeScript config loader. Add regression tests that run the built config loader in a separate Node process, covering real ES modules and existing CommonJS configurations without Jest transforming the module format.

Fixes #8267.

Validation on Windows, Node 22.21.1:
- Original source: CommonJS case passes; ES module case fails with empty app identity and `webDir: "www"`.
- CLI build passes.
- Config and framework-detection suites: 19 tests pass.
- ESLint, Prettier, and `git diff --check` pass.

Native Android/iOS builds and the full platform integration suite were not run.


jabrailkhalil:
The CI workflows for the current head `08c356893b` are waiting for contributor workflow approval (`action_required`): [CI](https://github.com/ionic-team/capacitor/actions/runs/34279347690). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

PR #8594, opened 2026-09-09
fix(cli): update gradle-wrapper.properties version to avoid JDK 25 problems

jcesarmobile:
gradlew command with gradle 8.x doesn't play well with JDK 25, making migrate command to fail to update the gradle files.

By updating the wrapper version in gradle-wrapper.properties, when calling the gradlew command it will download gradle 9.x first and gradle 9.x works fine with JDK 25


OS-pedrogustavobilro (approved):

jcesarmobile:
re-requesting review since I applied the change proposed on https://github.com/ionic-team/migrate-capacitor-plugin/pull/33#discussion_r3978080130

OS-pedrogustavobilro (approved):
LGTM

PR #8595, opened 2026-09-09
fix(ios): do not forward scene lifecycle events to the page before it has loaded

tgabi333:
## What

With UIScene adopted (8.5.0), `CapacitorBridge.setupCordovaCompatibility` observes `UIScene.willEnterForegroundNotification` / `didEnterBackgroundNotification` and forwards them to the page as `resume` / `pause` document events via `window.Capacitor.triggerEvent(...)`.

On a **cold start** the `willEnterForeground` notification is posted while the initial page load is still in flight — before `window.Capacitor` has been defined — so the evaluation throws inside the web view and every launch of every UIScene-migrated app logs:

```
⚡️  Loading app at capacitor://localhost...
⚡️  JS Eval error A JavaScript exception occurred
```

The same happens while the page is being reloaded after the web content process was terminated (the scenario described in ionic-team/capacitor-plugins#2357 for the App plugin).

This PR forwards the event only once the initial load has finished (`webViewLoadingState == .subsequentLoad`) and the web view is not currently loading.

## Why drop instead of defer

In both cases the page is about to load from scratch, so there is nothing for it to resume or pause. Deferring the event until after the load would deliver a `resume` right after a fresh load and make the page react to a transition it never went through. Real background/foreground transitions of a loaded page are unaffected.

## Reproduction

1. `npm init @capacitor/app` with Capacitor 8.5.x, run `npx cap migrate` so the project adopts UIScene (or use any app migrated per the 8.5 guide).
2. Run on a device or simulator and watch the Xcode console on a cold start.
3. `⚡️ JS Eval error A JavaScript exception occurred` appears right after `Loading app at capacitor://localhost...`, before the page's first bridge call. It never appears in the same app before the UIScene migration.

## Verification

- `xcodebuild clean test` on the `Capacitor` scheme (iPhone 16 simulator): 9 tests, 0 failures.
- A/B on a physical iPhone 15 Pro Max with a UIScene-migrated 8.5.1 app, same build, only `CapacitorBridge.swift` swapped: the cold-start line is gone; `SplashScreen.hide` / `App.getInfo` still cross the bridge; `pause` / `resume` still fire on real background/foreground transitions once the page is loaded.

Related: ionic-team/capacitor-plugins#2357 (same symptom in `@capacitor/app` after a web-process termination, closed for lack of a reproduction — the cold-start case above reproduces deterministically).


theproducer (approved):
Thanks!

